PR #47, opened 2022-05-29
Delegate to `datafusion-proto` for logical plan serde

andygrove:

Part of https://github.com/apache/arrow-ballista/issues/2

 # Rationale for this change

We should delegate to `datafusion-proto` for logical plan serde so that Ballista automatically picks up support for new operators and expressions.

- Delegate to `datafusion-proto` for serde, removing around 1,000 lines of code here
- Remove logical plan types from `ballista.proto`
- Update `datafusion.proto` to match latest from `datafusion-proto`  (note that we will eventually remove this file from this repo)

No


andygrove:
@thinkharderdev @yahoNanJing fyi

alamb (approved):
Makes sense to me

thinkharderdev:
@andygrove Why remove the `AsLogicalPlan` type param from the codec? In our project we use a custom logical plan representation (based on a different query AST) and it is very helpful to be able to plugin our own `LogicalPlan` type.

andygrove:
> @andygrove Why remove the `AsLogicalPlan` type param from the codec? In our project we use a custom logical plan representation (based on a different query AST) and it is very helpful to be able to plugin our own `LogicalPlan` type.

@thinkharderdev the main goal was to delegate to `datafusion-proto` for logical plan serde so that we don't have to duplicate all of this code in Ballista. A secondary goal was to use the `datafusion-proto` API for converting plans to bytes and bytes to plans without having dependencies on internals of that crate (so that we can easily change the format in the future). If we create a trait for this API does that make it easy for you to plug in your own implementation?

andygrove:
Would be good to get input from @alamb as well since he is a consumer of the `datafusion-proto` API.

andygrove:
Moving this back to draft for now

andygrove:
Having re-read the comments, I can see that this approach is not compatible with having a custom logical plan representation and this is not a feature I had considered. Also, I don't think we have any tests for this feature?

@thinkharderdev Perhaps you could add a test for your use case? I will rethink the approach here.

thinkharderdev:
> > @andygrove Why remove the `AsLogicalPlan` type param from the codec? In our project we use a custom logical plan representation (based on a different query AST) and it is very helpful to be able to plugin our own `LogicalPlan` type.
> 
> @thinkharderdev the main goal was to delegate to `datafusion-proto` for logical plan serde so that we don't have to duplicate all of this code in Ballista. A secondary goal was to use the `datafusion-proto` API for converting plans to bytes and bytes to plans without having dependencies on internals of that crate (so that we can easily change the format in the future). If we create a trait for this API does that make it easy for you to plug in your own implementation?

Yeah, my understanding is that `AsLogicalPlan` is that trait. So `datafusion-proto` can provide a set of protobuf definitions for serializing a `LogicalPlan` in a "canonical" way and implement the `AsLogicalPlan` trait for those definitions. I think it makes sense to continue using that implementation in Ballista as the default representation and rely on the implementation in `datafusion-proto`.

alamb:
Sorry I am catching up with PRs -- sounds like there is a good plan forward here. Thanks @andygrove and @thinkharderdev

andygrove:
Closed in favor of https://github.com/apache/arrow-ballista/pull/57

PR #66, opened 2022-06-13
Update to datafusion 9.0.0

nl5887:
This commit will fix compilation errors related to upcoming datafusion 9.0.0.


andygrove:
Unfortunately we cannot use DF 9.0.0 due to a breaking API change. I created https://github.com/apache/arrow-ballista/pull/68 to upgrade to the latest version which has the fix the that.

nl5887:
> Unfortunately we cannot use DF 9.0.0 due to a breaking API change. I created #68 to upgrade to the latest version which has the fix the that.

This was a work in progress commit, to be merged when all breaking changes have been fixed. It has a lot of duplicates with #68, so we can close this commit.

andygrove:
Closing this since another PR went in to update to a recent version. Thanks @nl5887

PR #67, opened 2022-06-13
Feature return task error on panic

nl5887:
Currently in case of a panic in the task executor, no task status will be returned. This commit will catch panics and return the error as task result.


thinkharderdev (approved):
Looks like a good improvement to me!

andygrove:
@nl5887 Do you have examples of panics that this fixes? Can we fix the root cause by using `Result` instead of calling `panic!` or `unimplemented!`?

PR #68, opened 2022-06-14
Use latest DataFusion

andygrove:

N/A

 # Rationale for this change

Stay up to date

Bump DF version

### DIFF
--- a/ballista/rust/core/src/serde/physical_plan/mod.rs
+++ b/ballista/rust/core/src/serde/physical_plan/mod.rs
@@ -836,7 +836,7 @@ impl AsExecutionPlan for PhysicalPlanNode {
                 .groups()
                 .iter()
                 .flatten()
-                .map(|b| *b)
+                .copied()
                 .collect();
 
             let group_names = exec
--- a/ballista/rust/core/src/serde/physical_plan/mod.rs
+++ b/ballista/rust/core/src/serde/physical_plan/mod.rs
@@ -235,16 +235,22 @@ impl AsExecutionPlan for PhysicalPlanNode {
             PhysicalPlanType::GlobalLimit(limit) => {
                 let input: Arc<dyn ExecutionPlan> =
                     into_physical_plan!(limit.input, registry, runtime, extension_codec)?;
-                Ok(Arc::new(GlobalLimitExec::new(
-                    input,
-                    None,
-                    Some(limit.limit as usize),
-                )))
+                let skip = if limit.skip > 0 {
+                    Some(limit.skip as usize)
+                } else {
+                    None
+                };
+                let fetch = if limit.fetch > 0 {
+                    Some(limit.fetch as usize)
+                } else {
+                    None
+                };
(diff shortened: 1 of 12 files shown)


andygrove:
Hi @thinkharderdev could I get a review when you have time

ballista/rust/core/src/serde/physical_plan/mod.rs
```diff
@@ -799,11 +803,13 @@ impl AsExecutionPlan for PhysicalPlanNode {
...
+                .expr()
                 .iter()
                 .map(|expr| expr.0.to_owned().try_into())
                 .collect::<Result<Vec<_>, BallistaError>>()?;
             let group_names = exec
                 .group_expr()
+                .expr()
                 .iter()
                 .map(|expr| expr.1.to_owned())
                 .collect();
```
thinkharderdev, line 815: Here we need to serialize/deserialize the `null_expr` and `groups` as well. This would change the structure of `GROUPING SET` queries. I had to do the same changes on our downstream fork. This seems to work: https://github.com/coralogix/arrow-ballista/commit/68aecb8e7e16ce7346dac3192153e000d3f0dfe4

ballista/rust/core/src/serde/physical_plan/mod.rs
```diff
@@ -235,7 +235,11 @@ impl AsExecutionPlan for PhysicalPlanNode {
             PhysicalPlanType::GlobalLimit(limit) => {
                 let input: Arc<dyn ExecutionPlan> =
                     into_physical_plan!(limit.input, registry, runtime, extension_codec)?;
-                Ok(Arc::new(GlobalLimitExec::new(input, limit.limit as usize)))
+                Ok(Arc::new(GlobalLimitExec::new(
```
thinkharderdev, line 238: Seems like here we should just update the protobuf struct to hold both skip and fetch. In our downstream I just used `0` to encode `None`. I _think_ that should work unless having `fetch = Some(0)` is meaningful in some scenario

thinkharderdev: See https://github.com/coralogix/arrow-ballista/commit/68aecb8e7e16ce7346dac3192153e000d3f0dfe4

thinkharderdev (commented):
A couple comments

ballista/rust/core/src/serde/physical_plan/mod.rs
thinkharderdev: Actually there was a bug in that version :) See the fix here https://github.com/coralogix/arrow-ballista/pull/1/commits/830b2d58a13710bb3e28337be5a641716155487e

andygrove:
Thanks @thinkharderdev I have applied those changes. Could  you take another look?

thinkharderdev (approved):

thinkharderdev:
> Thanks @thinkharderdev I have applied those changes. Could you take another look?

LGTM

PR #93, opened 2022-07-24
Add FlightSQL support

### LINKED ISSUE #92: Ballista should support Arrow FlightSQL

In order to integrate with the wide ecosystem of database tools (tableau, etc) Ballista should support (J/O)DBC. 

Directly implementing xDBC drivers would be wasted effort, since the Arrow FlightSQL protocol already has xDBC drivers. Therefor Ballista should simply implement support for Arrow FlightSQL and call it a day.

Directly implementing a JDBC driver.

avantgardnerio:

Closes #92.

 # Rationale for this change

To use off the shelf tools with Ballista.

A minimal FlightSQL service implementation.

They should be able to connect with off-the-shelf tools, given they have correctly installed and configured the FlightSQL JDBC or ODBC drivers.

### DIFF
--- a/ballista/rust/scheduler/src/flight_sql.rs
+++ b/ballista/rust/scheduler/src/flight_sql.rs
@@ -1,0 +1,195 @@
+use arrow_flight::{FlightData, FlightDescriptor, FlightInfo};
+use arrow_flight::flight_service_server::FlightService;
+use arrow_flight::sql::{ActionClosePreparedStatementRequest, ActionCreatePreparedStatementRequest, ActionCreatePreparedStatementResult, CommandGetCatalogs, CommandGetCrossReference, CommandGetDbSchemas, CommandGetExportedKeys, CommandGetImportedKeys, CommandGetPrimaryKeys, CommandGetSqlInfo, CommandGetTables, CommandGetTableTypes, CommandPreparedStatementQuery, CommandPreparedStatementUpdate, CommandStatementQuery, CommandStatementUpdate, SqlInfo, TicketStatementQuery};
+use arrow_flight::sql::server::FlightSqlService;
+use tonic::{Response, Status, Streaming};
+
+#[derive(Clone)]
+pub struct FlightSqlServiceImpl {}
+
+#[tonic::async_trait]
+impl FlightSqlService for FlightSqlServiceImpl {
+    type FlightService = FlightSqlServiceImpl;
+    // get_flight_info
+    async fn get_flight_info_statement(
+        &self,
+        _query: CommandStatementQuery,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(tonic::Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_prepared_statement(
+        &self,
+        _query: CommandPreparedStatementQuery,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_catalogs(
+        &self,
+        _query: CommandGetCatalogs,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_schemas(
+        &self,
+        _query: CommandGetDbSchemas,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_tables(
+        &self,
+        _query: CommandGetTables,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_table_types(
+        &self,
+        _query: CommandGetTableTypes,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_sql_info(
+        &self,
+        _query: CommandGetSqlInfo,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_primary_keys(
+        &self,
+        _query: CommandGetPrimaryKeys,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_exported_keys(
+        &self,
+        _query: CommandGetExportedKeys,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_imported_keys(
+        &self,
+        _query: CommandGetImportedKeys,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn get_flight_info_cross_reference(
+        &self,
+        _query: CommandGetCrossReference,
+        _request: FlightDescriptor,
+    ) -> Result<Response<FlightInfo>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    // do_get
+    async fn do_get_statement(
+        &self,
+        _ticket: TicketStatementQuery,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+
+    async fn do_get_prepared_statement(
+        &self,
+        _query: CommandPreparedStatementQuery,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_catalogs(
+        &self,
+        _query: CommandGetCatalogs,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_schemas(
+        &self,
+        _query: CommandGetDbSchemas,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_tables(
+        &self,
+        _query: CommandGetTables,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_table_types(
+        &self,
+        _query: CommandGetTableTypes,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_sql_info(
+        &self,
+        _query: CommandGetSqlInfo,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_primary_keys(
+        &self,
+        _query: CommandGetPrimaryKeys,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_exported_keys(
+        &self,
+        _query: CommandGetExportedKeys,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_imported_keys(
+        &self,
+        _query: CommandGetImportedKeys,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        Err(Status::unimplemented("Not yet implemented"))
+    }
+    async fn do_get_cross_reference(
+        &self,
+        _query: CommandGetCrossReference,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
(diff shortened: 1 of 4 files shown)


avantgardnerio:
I was able to retrieve query results from some custom Kotlin JDBC client code I'll post shortly:

```
[Test worker] INFO FlightSqlStatement - Execute SQL: SELECT 'keep alive'
[Test worker] INFO FlightSqlStatement - Got response: keep alive
```

avantgardnerio:
Working via custom JDBC driver in DataGrip:

Screenshot from 2022-07-24 17-05-50

avantgardnerio:
The official JDBC driver needs prepared statement support in order to work:

Screenshot from 2022-07-24 17-33-56

```
> select 'Hello, Apache Arrow Ballista2!'
[2022-07-24 17:33:31] Error while executing SQL "select 'Hello, Apache Arrow Ballista2!'": Implement do_action_create_prepared_statement
[2022-07-24 17:33:31] UNIMPLEMENTED: Implement do_action_create_prepared_statement
```

Also, it is necessary to set settings to:
1. no authentication
2. useEncryption=false

Screenshot from 2022-07-24 17-35-21

andygrove (approved):
Thanks @avantgardnerio this looks like a great start.

@thinkharderdev @yahoNanJing any thoughts?

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
+use std::collections::HashMap;
```
andygrove, line 1: Need an ASF header here to get the checks to pass

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+                let loc = Location { uri: format!("grpc+tcp://{}", authority) };
+                let buf = fetch.encode_to_vec();
+                let ticket = Ticket { ticket: buf };
+                let fiep = FlightEndpoint {
+                    ticket: Some(ticket),
+                    location: vec![loc],
+                };
+                fieps.push(fiep);
+            }
+            break fieps;
```
Dandandan, line 190: There seems no need to use a label here for the break?

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+                        stage_id: id.stage_id,
+                        partition_id: id.partition_id,
+                        path: loc.path.clone()
+                    };
+                    protobuf::Action {
+                        action_type: Some(protobuf::action::ActionType::FetchPartition(fetch)),
+                        settings: vec![],
+                    }
+                } else {
+                    Err(Status::internal(format!("Error getting partition it")))?
```
thinkharderdev, line 166:
```suggestion
                    Err(Status::internal(format!("Error getting partition ID")))?
```

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+                };
+                let authority = if let Some(ref md) = loc.executor_meta {
+                    // pub id: ::prost::alloc::string::String,
+                    format!("{}:{}", md.host, md.port)
```
thinkharderdev, line 170:
```suggestion
                    format!("{}:{}", md.host, md.port)
```

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+                } else {
+                    Err(Status::internal(format!("Error getting location")))?
```
thinkharderdev, line 172:
```suggestion
                    Err(Status::internal(format!("Invalid partition location, missing executor metadata")))?
```

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+                };
+                if let Some(ref stats) = loc.partition_stats {
+                    num_rows += stats.num_rows;
+                    num_bytes += stats.num_bytes;
+                    // pub num_batches: i64,
```
thinkharderdev, line 177:
```suggestion
                    num_rows += stats.num_rows;
                    num_bytes += stats.num_bytes;
```

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,402 @@
...
+use crate::scheduler_server::event::QueryStageSchedulerEvent;
+use ballista_core::serde::protobuf::job_status;
+use ballista_core::serde::protobuf::JobStatus;
+use ballista_core::serde::protobuf;
+use prost::Message;
+
+#[derive(Clone)]
+pub struct FlightSqlServiceImpl {
+    server: SchedulerServer<LogicalPlanNode, PhysicalPlanNode>,
+    statements: Arc<Mutex<HashMap<Uuid, LogicalPlan>>>,
```
thinkharderdev, line 34: If we need to save the logical plan then we'd need to eventually persist this to the state backend since the scheduler may be distributed.

thinkharderdev (approved):
Some minor nits but this looks awesome!

ballista/rust/scheduler/src/flight_sql.rs
avantgardnerio: It's pretty neat. You can actually return a value from a `loop`, and that's what this is doing. I wasn't aware of it either, until this PR. https://doc.rust-lang.org/rust-by-example/flow_control/loop/return.html

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+                job_status::Status::Failed(e) => Err(Status::internal(format!(
+                    "Error building plan: {}",
+                    e.error
+                )))?,
+                job_status::Status::Completed(comp) => comp,
+            };
+            let mut fieps: Vec<_> = vec![];
+            for loc in completed.partition_location.iter() {
+                let fetch = if let Some(ref id) = loc.partition_id {
+                    let fetch = protobuf::FetchPartition {
```
avantgardnerio, line 190: I built this directly using the prost generated structs, and it was very painful since all this lacked autocomplete in my IDE. It looked like it was also possible to create Ballista/DataFusion native versions of these structs then `From` or `Into` to convert to prost? @andygrove or anyone who knows, I'd love advice on how to do this better.

thinkharderdev: Most of these prost structs have a corresponding struct that is nicer to use (along with conversions from/to). For this case you can convert the prost struct into a `ballista_core::serde::scheduler::PartitionLocation` which is nicer to use (it will unwrap all the `Option` fields which protobuf insists on but are not really optional).

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+use datafusion::arrow::datatypes::Schema;
+use datafusion::arrow::ipc::writer::{IpcDataGenerator, IpcWriteOptions};
+use datafusion::logical_expr::LogicalPlan;
+use datafusion_proto::protobuf::LogicalPlanNode;
+use prost::Message;
+use tokio::time::sleep;
+use uuid::Uuid;
+
+pub struct FlightSqlServiceImpl {
+    server: SchedulerServer<LogicalPlanNode, PhysicalPlanNode>,
```
avantgardnerio, line 56: Cloning the whole server here seems... wrong. Perhaps I should have an `Arc<Mutex<>>` to it?

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+            _statements: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl FlightSqlService for FlightSqlServiceImpl {
+    type FlightService = FlightSqlServiceImpl;
+    // get_flight_info
+    async fn get_flight_info_statement(
```
avantgardnerio, line 73: The JDBC connection `ExecuteQuery()` maps directly to this call. It appears the JDBC `createStatement()` approach causes the driver to invoke `do_action_create_prepared_statement()` which we'll need to implement, and as @thinkharderdev points out, maybe the state will need to go into etcd.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+        // Run query
+        let config_builder = BallistaConfig::builder();
+        let config = config_builder
+            .build()
+            .map_err(|e| Status::internal(format!("Error building config: {}", e)))?;
+        let ctx = self
+            .server
+            .state
+            .session_manager
+            .create_session(&config)
```
avantgardnerio, line 89: Probably shouldn't create a session every time.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+                let msg =
+                    format!("Failed to send JobQueued event for {}: {:?}", job_id, e);
+                error!("{}", msg);
+                Status::internal(msg)
+            })?;
+
+        // let handle = Uuid::new_v4();
+        // let mut statements = self.statements.try_lock()
+        //     .map_err(|e| Status::internal(format!("Error locking statements: {}", e)))?;
+        // statements.insert(handle, plan.clone());
```
avantgardnerio, line 144: Not necessary for this, but needed for `do_action_create_prepared_statement()` and possibly others.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+
+        // poll for job completion
+        let mut num_rows = 0;
+        let mut num_bytes = 0;
+        let fieps = loop {
+            sleep(Duration::from_millis(100)).await;
```
avantgardnerio, line 150: The busy wait is a bit cheesy, but it appears Arrow FlightSQL expects synchronous execution? At least we're in a coroutine context and this isn't blocking a thread, thanks to Tokio, but the polling has been moved out of the client and into the scheduler. I'm not sure how else to do it, just something to be aware of.

ballista/rust/scheduler/src/flight_sql.rs
thinkharderdev: I think we can probably get away with an `Arc<SchedulerServer<>>`. The only time you need a `mut` reference is when you `init` the server which should only happen once.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,460 @@
...
+                    Err(Status::internal(
+                        "Invalid partition location, missing executor metadata"
+                            .to_string(),
+                    ))?
+                };
+                if let Some(ref stats) = loc.partition_stats {
+                    num_rows += stats.num_rows;
+                    num_bytes += stats.num_bytes;
+                } else {
+                    Err(Status::internal("Error getting stats".to_string()))?
```
avantgardnerio, line 217: There's a lot of untested branches here. I'm still trying to figure out how to test properly. If the Rust implementation of the Flight SQL client is far enough along, integration tests would be good. Otherwise (and perhaps in addition to), unit test would be helpful. But so much of this is interacting with other parts of the system, I'm not quite sure how to do it. I'll look over `test_poll_work()` as a reference.

ballista/rust/scheduler/src/main.rs
```diff
@@ -100,10 +102,15 @@ async fn start_server(
             let scheduler_grpc_server =
                 SchedulerGrpcServer::new(scheduler_server.clone());
 
+            let flight_sql_server = FlightServiceServer::new(FlightSqlServiceImpl::new(
+                scheduler_server.clone(),
```
avantgardnerio, line 106: Again, cloning the whole server seems wrong, but this PR isn't the first to do so. Is there something I'm missing here?

ballista/rust/scheduler/src/flight_sql.rs
avantgardnerio: It looks like I can store things trivially in `etcd` (or whatever backing store) using `server.state.session_manager.state`?

ballista/rust/scheduler/src/main.rs
thinkharderdev: Yeah, this probably needs to be cleaned up. The other places where we are cloning the scheduler server can probably be refactored to just take an `Arc<SchedulerState<>>`. Cloning the `EventLoop` might be more problematic though.

ballista/rust/scheduler/src/flight_sql.rs
thinkharderdev: Sort of. There is a bit more structure imposed by the `StateBackendClient` to try and avoid lock contention in the scheduler. So you have to define a new `Keyspace` enum for the new data to live in. The it's probably best to add methods in `TaskManager` to save/fetch the logical plans. That can wrap the serde logic and interaction with the state backend.

ballista/rust/scheduler/src/flight_sql.rs
thinkharderdev: In this case we may be able to just use a `watch` and let the state backend push status updates (instead of polling). You can wrap that in a `JobCompletionFuture`. Something like 

```
struct JobCompletionFuture {
    job_completion_watch: Watch,
}

impl Future for JobCompletionFuture {
    type Output = Result<JobStatus, BallistaError>;

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
            match self.job_completion_watch.poll_next_unpin(cx) {
                Poll::Ready(Some(WatchEvent::Put(_key, value))) => {
                   if let Ok(job_status) = decode_protobuf::<JobStatus>(&value) {
                        Poll::Ready(Ok(job_status))
                    } else {
                       Poll::Ready(Err(BallistaError::Internal("some error".to_owned())))
                   }
                }
                Poll::Ready(None) => Poll::Ready(Err(BallistaError::Internal("watch ended with no update".to_owned()))),
                _ => Poll::Pending,
            }
    }
}
```

It would have to be slightly more complicated than this since we would need to watch both the `CompletedJobs/<job id>` and `FailedJob/<job id>` keys but the basic premise would still work I think.

ballista/rust/scheduler/src/flight_sql.rs
avantgardnerio: I'm looking at the available parameters (`CommandStatementQuery` and `FlightDescriptor`) and not finding anything that would work particularly well as a session-identifier. I assume there must be such a thing lower down at the GRPC level? Anyone know where I can find something like that?

avantgardnerio:
I think the remaining problems I'm having with prepared statements are actually do to the implementation of the JDBC driver, as described here: https://github.com/apache/arrow/pull/12830/files#r929303757

avantgardnerio:
I'm undrafting this because it doesn't break anything and works against https://github.com/rafael-telles/arrow/pull/42 and now the code is a little cleaner. I'll probably keep improving it based on the feedback above, but that stuff could probably happen after merge.

PR #110, opened 2022-08-03
MINOR: Use `sqlparser` version exported by `datafusion` crates

andygrove:

N/A

 # Rationale for this change

Ballista should always use the same version of sqlparser as DataFusion.

- Remove Cargo dependency on `sqlparser`
- Use the version exported by DataFusion

No

### DIFF
--- a/ballista/rust/client/Cargo.toml
+++ b/ballista/rust/client/Cargo.toml
@@ -36,7 +36,6 @@ datafusion-proto = { git = "https://github.com/apache/arrow-datafusion", rev = "
 futures = "0.3"
 log = "0.4"
 parking_lot = "0.12"
-sqlparser = "0.18"
 tempfile = "3"
 tokio = "1.0"
 
--- a/ballista/rust/client/src/context.rs
+++ b/ballista/rust/client/src/context.rs
@@ -19,7 +19,6 @@
 
 use log::info;
 use parking_lot::Mutex;
-use sqlparser::ast::Statement;
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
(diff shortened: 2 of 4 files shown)


andygrove:
fyi @avantgardnerio

avantgardnerio (approved):
Is re-exporting a common practice? I see how it is nice in that it eliminates the possibility of accidentally importing two versions of these crates (if the user is aware), but it also makes it more likely that if they are unaware, they import it again and get a conflict.

I'm not certain of the alternative - I think just use `cargo tree` or a tool like mine to hunt down conflicting versions...

I'm not opposed per se, just bringing it up for debate. If you have any external links to best practices, or if this was discussed previously, I'd love to see those discussions as well.

All that said, :+1:

andygrove:
> If you have any external links to best practices, or if this was discussed previously, I'd love to see those discussions as well.

The best resource I can find is this: https://github.com/rust-lang/api-guidelines/discussions/176

avantgardnerio:
Thanks for providing the link @andygrove . If I understand correctly, I think I have to agree with this comment: https://github.com/rust-lang/api-guidelines/discussions/176#discussioncomment-232635

> no way of discerning that this doc is lifted from the futures crate

Except that it's not just about docs, it's about information loss. Cargo is able to maintian a DAG of dependencies and detect conflicts. If I understand correctly, if a dependency is re-exported, then this information is lost?

Also, if ballista were to add another important 3rd party dependency that followed the same practice (re-exporting their own SQL parser) it would be impossible to pass a `sqlparser::ast::Statement` between the two crates in a safe way?

andygrove:
> Also, if ballista were to add another important 3rd party dependency that followed the same practice (re-exporting their own SQL parser) it would be impossible to pass a `sqlparser::ast::Statement` between the two crates in a safe way?

I just learned this lesson rather painfully in another project. I am no longer a fan of re-exporting!

andygrove:
I may have closed this prematurely. I should have just used `[patch.crates-io]` to override versions.

PR #113, opened 2022-08-03
Add session token so we can register tables for FlightSQL

### LINKED ISSUE #112: Use ArrowFlight bearer token auth to create a session key for FlightSql clients

Since tables have to be registered per each session in Ballista, no one can query any tables unless we have a session identifier. The native Ballista wire protocol has a special field for this, but in FlightSQL the best we can do is use their [bearer token auth](https://arrow.apache.org/docs/format/Flight.html#authentication). 

Accept a username and password on handshake, return a session token, use that token to maintain persistent sessions.

I can't really think of one.

avantgardnerio:

Closes #112.

 # Rationale for this change

Since tables have to be registered to sessions, we need a session identifier or else each new FlightSql command runs on a clean slate with no tables. To fix this, we need a session identifier.

Implementation of a Basic Auth handshake, returning a UUID bearer token that the FlightSqlClient should send back with each request. Using this, I was successfully able to register the TPC-H customer table and select some data from it.

They can start doing interesting things with JDBC / ODBC.

### DIFF
--- a/ballista-cli/Cargo.toml
+++ b/ballista-cli/Cargo.toml
@@ -31,8 +31,8 @@ readme = "README.md"
 [dependencies]
 ballista = { path = "../ballista/rust/client", features = [], optional = false }
 clap = { version = "3", features = ["derive", "cargo"] }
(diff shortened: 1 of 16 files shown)


avantgardnerio:
I will follow up with another PR shortly.

PR #115, opened 2022-08-05
Add timeout and keep-alive settings for Grpc Client and Server

### LINKED ISSUE #114: Add timeout settings for Grpc Client

The current GRpc clients: SchedulerGrpcClient, ExecutorGrpcClient and FlightServiceClient do not configure any timeout settings.
We should have a default timeout settings for interactions between the system components.

The reason that there is no timeout setting is because we use the tonic build generate client code and method to connect to the
remote Grpc Server.

````
impl ExecutorGrpcClient<tonic::transport::Channel> {
        /// Attempt to create a new client by connecting to a given endpoint.
        pub async fn connect<D>(dst: D) -> Result<Self, tonic::transport::Error>
        where
            D: std::convert::TryInto<tonic::transport::Endpoint>,
            D::Error: Into<StdError>,
        {
            let conn = tonic::transport::Endpoint::new(dst)?.connect().await?;
            Ok(Self::new(conn))
        }
    }

````
We should pass in our own channel via new on the generated types that we can then set the timeout.

**Additional context**

### PR DESCRIPTION
**Pull request by mingmwang**

Closes #114.

 # Rationale for this change

### DIFF
--- a/ballista/rust/core/src/utils.rs
+++ b/ballista/rust/core/src/utils.rs
@@ -49,7 +49,10 @@ use std::io::{BufWriter, Write};
 use std::marker::PhantomData;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use std::time::Duration;
 use std::{fs::File, pin::Pin};
+use tonic::codegen::StdError;
+use tonic::transport::{Channel, Error};
 
 /// Stream data to disk in Arrow IPC format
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by mingmwang** [comment]
@thamht4190 @yahoNanJing 

Please help to take a look and review.

File: ballista/rust/core/src/utils.rs [hunk 1]
```diff
@@ -310,3 +313,32 @@ impl<T: 'static + AsLogicalPlan> QueryPlanner for BallistaQueryPlanner<T> {
...
+{
+    let endpoint = tonic::transport::Endpoint::new(dst)?
+        .connect_timeout(Duration::from_secs(20))
+        .timeout(Duration::from_secs(20))
+        // Disable Nagle's Algorithm since we don't want packets to wait
+        .tcp_nodelay(true)
+        .tcp_keepalive(Option::Some(Duration::from_secs(3600)))
+        .http2_keep_alive_interval(Duration::from_secs(300))
+        .keep_alive_timeout(Duration::from_secs(20))
+        .keep_alive_while_idle(true);
```
andygrove, line 332: It would be nice to make these options user-configurable via the recently added config mechanism that was added in DataFusion (but not integrated with Ballista yet).

andygrove (approved):
Looks reasonable to me but I have not been working with gRPC lately. Maybe @avantgardnerio could review as well since he is working on Flight SQL support.

ballista/rust/core/src/utils.rs
andygrove: I filed https://github.com/apache/arrow-ballista/issues/117 for this and I plan on picking this up

PR #124, opened 2022-08-10
Ballista Executor report plan/operators metrics to Ballista Scheduler 

### LINKED ISSUE #116: Ballista Executor report plan/operators metrics to Ballista Scheduler when task finish

In the current Ballista implementation, when the Executor finish the tasks, only the task status(Failed/Complete) is report back to the Scheduler, the plan metrics are not collected and further combined in the Scheduler, the Scheduler does not have an overall view of plan execution metrics, like how many rows are scanned, total elapse times of each operator etc.

1. In the Executor side, when the task is finished, need to collect all the plan operators metrics for that task.
2. Send the plan metrics back to Scheduler along with TaskStatus.
3. Scheduler need to further combine/aggregate the metrics from all the completed tasks and finally update the plan metrics to the physical plan hold by the Scheduler. 
4. It is better that DataFusion can add a unique id to ExecutionPlanMetricsSet, so that Ballista Scheduler can easily combine and aggregate metrics.
5. The metrics combine/aggregation overhead could be quite expensive, considering there might be tens of thousands of tasks for a heavy stage, the metrics aggregation should be handled in an async way and should not impact the major stage/task scheduling and task result processing flow and allow dropping the metrics if the metrics combine/aggregate speed can not catch up.

**Additional context**

mingmwang:

Partial closes #116.
There are still some pending (nice to have) changes need to be made to the DataFusion code base
Basically this PR covered those parts:
1) Enhance rpc protocols to report task metrics
``
rpc UpdateTaskStatus (UpdateTaskStatusParams) returns (UpdateTaskStatusResult) {}
``
2) When the task finish, collect the plan metrics and report back to Scheduler
3) In Scheduler side, merge and aggregate task metrics to stage metrics, update the ExecutionGraph.
4) When the stage finish, display the stage plan with combined/aggregated metrics.

 # Rationale for this change

### DIFF
--- a/ballista/rust/core/proto/ballista.proto
+++ b/ballista/rust/core/proto/ballista.proto
@@ -429,6 +429,7 @@ message ExecutionGraphStage {
   repeated TaskStatus task_statuses = 6;
   uint32 output_link = 7;
   bool resolved = 8;
+  repeated OperatorMetricsSet stage_metrics = 9;
 }
 
 message ExecutionGraph {
@@ -501,6 +502,41 @@ message ColumnStats {
   datafusion.ScalarValue max_value = 2;
   uint32 null_count = 3;
   uint32 distinct_count = 4;
+}
+
+message OperatorMetricsSet {
+  repeated OperatorMetric metrics = 1;
+}
+
+
+message NamedCount {
+  string name = 1;
+  uint64 value = 2;
+}
+
+message NamedGauge {
+  string name = 1;
+  uint64 value = 2;
+}
+
+message NamedTime {
+  string name = 1;
+  uint64 value = 2;
+}
+
+message OperatorMetric {
+  oneof metric {
+    uint64 output_rows = 1;
+    uint64 elapse_time = 2;
+    uint64 spill_count = 3;
+    uint64 spilled_bytes = 4;
+    uint64 current_memory_usage = 5;
+    NamedCount count = 6;
+    NamedGauge gauge = 7;
+    NamedTime time = 8;
+    int64 start_timestamp = 9;
+    int64 end_timestamp = 10;
+  }
 }
 
 // Used by scheduler
@@ -594,6 +630,7 @@ message TaskStatus {
     FailedTask failed = 3;
     CompletedTask completed = 4;
   }
+  repeated OperatorMetricsSet metrics = 5;
 }
 
 message PollWorkParams {
--- a/ballista/rust/scheduler/src/state/execution_graph.rs
+++ b/ballista/rust/scheduler/src/state/execution_graph.rs
@@ -20,7 +20,7 @@ use ballista_core::error::{BallistaError, Result};
 use ballista_core::execution_plans::{ShuffleWriterExec, UnresolvedShuffleExec};
 
 use ballista_core::serde::protobuf::{
-    self, CompletedJob, JobStatus, QueuedJob, TaskStatus,
+    self, CompletedJob, JobStatus, OperatorMetricsSet, QueuedJob, TaskStatus,
 };
 use ballista_core::serde::protobuf::{job_status, FailedJob, ShuffleWritePartition};
 use ballista_core::serde::protobuf::{task_status, RunningTask};
@@ -28,14 +28,16 @@ use ballista_core::serde::scheduler::{
     ExecutorMetadata, PartitionId, PartitionLocation, PartitionStats,
 };
 use datafusion::physical_plan::{
-    accept, ExecutionPlan, ExecutionPlanVisitor, Partitioning,
+    accept, ExecutionPlan, ExecutionPlanVisitor, Metric, Partitioning,
 };
 use log::debug;
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fmt::{Debug, Formatter};
 
+use ballista_core::utils::collect_plan_metrics;
 use datafusion::physical_plan::display::DisplayableExecutionPlan;
+use datafusion::physical_plan::metrics::{MetricValue, MetricsSet};
 use std::sync::Arc;
 
 /// This data structure collects the partition locations for an `ExecutionStage`.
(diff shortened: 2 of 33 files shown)


mingmwang:
@thinkharderdev @andygrove 

Please help to review my PR.

andygrove (approved):
This looks like a great improvement!

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -513,6 +574,40 @@ impl ExecutionGraph {
...
+                            // TODO the plan_metrics update above is a snapshot clone from the plan metrics.
+                            // TODO Need to modify DataFusion to return metricset reference, not clone.
+
+                            println!(
+                                "=== [{}/{}/{}] Stage finished, physical plan with metrics ===\n{}\n",
+                                job_id,
+                                stage_id,
+                                partition,
+                                DisplayableExecutionPlan::with_full_metrics(stage_plan.as_ref()).indent()
+                            );
```
thinkharderdev, line 608: I'm not sure we should be printing to stdout here. If we want to print something it should be controlled by the log level (i.e. people can turn it off if it's too noisy)

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -153,6 +157,7 @@ impl ExecutionStage {
             task_statuses: vec![None; num_tasks],
             output_link,
             resolved,
+            stage_metrics: None,
```
thinkharderdev, line 160: I'm not sure about storing the raw metrics in the `ExecutionGraph`. It could potentially be a lot of data (eg if you had a `ParquetExec` across many files then you can have a potentially very big set of metrics) and we have to read/write the graph to the state store a lot. 

I think it might be worth introducing a `MetricStore` trait where we can make the storage (or non-storage as the case may be) of metrics extensible.

thinkharderdev (commented):
Thanks @mingmwang this is great! I had one nit and one more substantive comment

ballista/rust/core/proto/ballista.proto
```diff
@@ -503,6 +504,41 @@ message ColumnStats {
...
+
+
+message NamedCount {
+  string name = 1;
+  uint64 value = 2;
+}
+
+message NamedGauge {
+  string name = 1;
+  uint64 value = 2;
```
liukun4515, line 519: From the metric system, the Gauge could be negative.
INT64 is better.

ballista/rust/scheduler/src/state/execution_graph.rs
mingmwang: I will change it to use log! here. But I realize that the println! is used in many other places like ExecutorMetricsCollector, execution_graph.resolve_shuffles() etc.

ballista/rust/scheduler/src/state/execution_graph.rs
mingmwang: @thinkharderdev Actually the `stage_metrics` stored here is not the raw metrics. It is combined/aggregated among all the tasks of that stage, that's why I name it to `stage_metrics`, not task metrics.

ballista/rust/core/proto/ballista.proto
mingmwang: Good point.

liukun4515: @andygrove @alamb 
Can we collation two side?
The `Gauge` use the signed integer to store the value.
https://github.com/apache/arrow-datafusion/pull/1682

mingmwang: I would prefer to handle the type alignment in another PR.

ballista/rust/scheduler/src/state/execution_graph.rs
thinkharderdev: Right, that helps but I still think that storing the metrics in the `ExecutionGraph` is not the right way to go. Even aggregated we're adding more data to a value that has to be read/written/decoded/encoded a lot. The other problem I think is that ideally we want to discard the full `ExecutionGraph` soon after the job is completed to prevent unbounded growth in the state store. But metrics may be something that we wish to preserve for longer (or even indefinitely). In addition, I think this may very well be an area where extensibility could be important. The `ExecutionGraph` is an internal implementation detail, but users may have an interest in storing and analyzing metrics and may wish to "export" them to a different system of their choosing (RDBMS, etc). Having an interface that allows for that would be helpful.

mingmwang: > 
> Even aggregated we're adding more data to a value that has to be read/written/decoded/encoded a lot

I agree with you that the `ExecutionGraph` is not the perfect place to store metrics. But since there is no in-memory stat store now. And I do not want to add another key space to sled/etcd to store the metrics, so just reuse the existing `ExecutionGraph`.  I think it is not that heavy compared to the existing task_statuses array If we have thousands of tasks.  

```
/// Status of each already scheduled task. If status is None, the partition has not yet been scheduled
 pub(crate) task_statuses: Vec<Option<task_status::Status>>,
```

Ideally those metrics should be stored in memory because currently they are just for displaying purpose.
For long terms preserving, both the task status and metrics should be persisted to other places.

mingmwang: I think we should still have different stores: in-memory stat store,  persisted stat store(sled/etcd) and long term persisted store(RDBMS, object store).

ballista/rust/core/proto/ballista.proto
alamb: I think changing `Gauge` in datafusion to use `AtomicI64` rather than `AtomicUsize` would make sense to me

alamb:
is there any other comments or concerns to be resolved before merging this PR? I haven't been following ballista closely so I don't know who has been merging the PRs recently

mingmwang:
@andygrove @thinkharderdev 
Can we merge this PR first? We can have ongoing PR to track the proper place for metrics and task status storage.

Issue #129, opened 2022-08-12
Improve ballista performance


yahoNanJing:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

To be an efficient distributed execution platform, there're still many aspects we need to improve.

**Describe the solution you'd like**

- For scheduler, efficient task scheduling
   - [ ] refine the scheduler state cache layer
   - [ ] use multiple threads for generating execution plan from logical plan
   - [ ] use round-robin policy for executor resource assignment

**Describe alternatives you've considered**

**Additional context**

andygrove:
I added https://github.com/apache/arrow-ballista/issues/243

yahoNanJing:
With all the PRs merged, we can achieve a relatively high throughput in RoundRobinLocal SlotsPolicy. For a 200ms query with 200 concurrency running 5 minutes, the average query latency is also around 200ms, which proves the task scheduling is very efficient and should not be a bottleneck in a high throughput system.

yahoNanJing:
Since Dan may not have time for reviewing on these PRs, Andy could you at other guys who may be interested on these PRs? Thanks very much.

PR #141, opened 2022-08-15
Ignore the previous job_id inside fill_reservations()

### LINKED ISSUE #138: Ignore the previous job_id inside fill_reservations()

Current strategy of filling reservation will leverage the previous job_id in the ExecutorReservation to give higher priority for that job scheduling. However, it will cause starvation issue when there are jobs with huge number of tasks.

It's better for the fill_reservations() to choose tasks fairly for each job and let the scheduler decides which job to be scheduled first which may be implemented by a few queues maintained by the scheduler in the future.

**Additional context**

yahoNanJing:

Closes #138.

 # Rationale for this change

### DIFF
--- a/ballista/rust/scheduler/src/state/execution_graph.rs
+++ b/ballista/rust/scheduler/src/state/execution_graph.rs
@@ -101,8 +101,8 @@ pub struct ExecutionStage {
     /// Status of each already scheduled task. If status is None, the partition has not yet been scheduled
     pub(crate) task_statuses: Vec<Option<task_status::Status>>,
     /// Stage ID of the stage that will take this stages outputs as inputs.
-    /// If `output_link` is `None` then this the final stage in the `ExecutionGraph`
-    pub(crate) output_link: Option<usize>,
+    /// If `output_links` is empty then this the final stage in the `ExecutionGraph`
+    pub(crate) output_links: Vec<usize>,
     /// Flag indicating whether all input partitions have been resolved and the plan
     /// has UnresovledShuffleExec operators resolved to ShuffleReadExec operators.
     pub(crate) resolved: bool,
@@ -136,7 +136,7 @@ impl ExecutionStage {
         stage_id: usize,
         plan: Arc<dyn ExecutionPlan>,
         output_partitioning: Option<Partitioning>,
-        output_link: Option<usize>,
+        output_links: Vec<usize>,
         child_stages: Vec<usize>,
     ) -> Self {
         let num_tasks = plan.output_partitioning().partition_count();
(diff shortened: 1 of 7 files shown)


yahoNanJing:
Hi @thinkharderdev, could you help review this PR?

andygrove:
@yahoNanJing It looks like this was building on another PR that has now been merged. Could you rebase / upmerge so we can see the changes specific to this PR?

yahoNanJing:
Thanks @andygrove. Just rebased. I will follow the order listed in #129 to raise PRs for a few performance improvements.

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -381,11 +381,21 @@ impl ExecutionPlanVisitor for ExecutionStageBuilder {
         } else if let Some(unresolved_shuffle) =
             plan.as_any().downcast_ref::<UnresolvedShuffleExec>()
         {
-            self.output_links
-                .insert(unresolved_shuffle.stage_id, self.current_stage_id);
+            if let Some(output_links) =
+                self.output_links.get_mut(&unresolved_shuffle.stage_id)
+            {
+                if !output_links.contains(&self.current_stage_id) {
```
andygrove, line 387: Would it be better to make `output_links` a `HashSet` instead of a `Vec`, given that duplicate values are not allowed?

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -381,11 +381,21 @@ impl ExecutionPlanVisitor for ExecutionStageBuilder {
...
+                    output_links.push(self.current_stage_id);
+                }
+            } else {
+                self.output_links
+                    .insert(unresolved_shuffle.stage_id, vec![self.current_stage_id]);
+            }
 
             if let Some(deps) = self.stage_dependencies.get_mut(&self.current_stage_id) {
-                deps.push(unresolved_shuffle.stage_id)
+                if !deps.contains(&unresolved_shuffle.stage_id) {
```
andygrove, line 396: Same question here - would `HashSet` make more sense?

andygrove (approved):
I left a couple questions but LGTM. Thanks @yahoNanJing!

ballista/rust/scheduler/src/state/execution_graph.rs
yahoNanJing: For deps and output_links, there will not be many elements. I think that's why it uses vec rather than the set 🤣 . I'm OK to change it to be set.

andygrove: I don't have a strong opinion on this. Set would reduce the code complexity slightly but possibly add some overhead. :man_shrugging:

PR #146, opened 2022-08-15
Add job cancellation

### LINKED ISSUE #139: Cancel a running query.

In current Ballista implementation, when a SQL is submitted,  there is no way to cancel it. We need to implement a way to cancel a running SQL, and if there are running tasks scheduled to run on Executors, those running tasks should be cancelled as well.
I'm not sure whether DataFusion can cancel a running query or not. For Ballista, it needs additional communication to cancel running queries.

1. Add cancel_query rpc call to Scheduler grpc protocol
  rpc CancelQuery (CancelQueryParams) returns (CancelQueryResult) {}

2. Add cancel job logic to query stage scheduler
3. Cancel the running task on Executor
4. Cancel the running Futures on Executor's Tokio queue
5. Check all the Datafusion async spawn logic, make sure the spawned future can be cancelled. 

**Additional context**

thinkharderdev:

Closes #139 

 # Rationale for this change

We need a way to cancel a running job. 

Add a new RPC to the scheduler and executors that will allow cancellation of jobs. A client can cancel an entire job by job ID. The scheduler will immediately update the job state to `Failed` with reason `Cancelled` and then determine all running tasks. It will then call an RPC on the executor side to cancel tasks that are currently in flight. In order to preempt running tasks the executor will have to launch tasks wrapped in an `futures_util::abortable::Abortable`. This will give us an `AbortHandle` which can be used to cancel the task. 

A new grpc endpoint for job cancellation. 

No


thinkharderdev:
@mingmwang

mingmwang:
This is awesome!!!

ballista/rust/scheduler/src/state/executor_manager.rs
```diff
@@ -194,6 +200,33 @@ impl ExecutorManager {
...
+
+            {
+                let mut clients = self.clients.write();
+                clients.insert(executor_id.to_owned(), client.clone());
+            }
+            Ok(client)
+        }
+    }
+
     /// Get a list of all executors along with the timestamp of their last recorded heartbeat
```
mingmwang, line 230: There is a new method in utils.rs to create Grpc client which handles the grpc connection timeout and keep-alive settings, you can use that.

````
pub async fn create_grpc_client_connection<D>(
    dst: D,
) -> std::result::Result<Channel, Error>
where
    D: std::convert::TryInto<tonic::transport::Endpoint>,
    D::Error: Into<StdError>,
{
    let endpoint = tonic::transport::Endpoint::new(dst)?
        .connect_timeout(Duration::from_secs(20))
        .timeout(Duration::from_secs(20))
        // Disable Nagle's Algorithm since we don't want packets to wait
        .tcp_nodelay(true)
        .tcp_keepalive(Option::Some(Duration::from_secs(3600)))
        .http2_keep_alive_interval(Duration::from_secs(300))
        .keep_alive_timeout(Duration::from_secs(20))
        .keep_alive_while_idle(true);
    endpoint.connect().await
}

File: ballista/rust/core/proto/ballista.proto [hunk 2]
```diff
@@ -732,6 +740,14 @@ message LaunchTaskResult {
   // TODO when part of the task set are scheduled successfully
 }
 
+message CancelTasksParams {
+  repeated PartitionId partition_id = 1;
+}
+
```
mingmwang, line 746: Can we name the partition_id to task_id here ?

````
  PartitionId task_id = 1;

File: ballista/rust/core/src/error.rs [hunk 3]
```diff
@@ -47,6 +48,7 @@ pub enum BallistaError {
     TonicError(tonic::transport::Error),
     GrpcError(tonic::Status),
     TokioError(tokio::task::JoinError),
+    Cancelled,
 }
```
mingmwang, line 52: How about Cancelled -> TaskCancelled ?

ballista/rust/executor/src/executor.rs
```diff
@@ -110,15 +121,224 @@ impl Executor {
...
+                PartitionId {
+                    job_id: job_id.clone(),
+                    stage_id,
+                    partition_id: part,
+                },
+                abort_handle,
+            );
+        }
+
+        let partitions = task.await??;
```
mingmwang, line 139: two ??

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -376,11 +375,78 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
+                    })
+                    .await?;
+            } else {
+                error!("Failed to get client for executor ID {}", executor_id)
+            }
+        }
+
+        Ok(())
+    }
+
```
mingmwang, line 435: The performance might be a problem here. Multiple cancel task requests are issued sequentially.  Seems launch_task has the same problem. We can spawn tokio tasks to cancel/launch tasks concurrently.

ballista/rust/executor/src/executor_server.rs
```diff
@@ -411,4 +411,31 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
...
+                .await
+            {
+                error!("Error cancelling task: {:?}", e);
+                cancelled = false;
+            }
+        }
+
+        Ok(Response::new(CancelTasksResult { cancelled }))
+    }
 }
```
mingmwang, line 441: The cancel_tasks RPC might take long time (wait for cancel).  Sometimes the operators are burning cpu and can not reach a yield point in short term. Maybe we should just return the rpc immediately with a status called 'Cancelling'. Whether the tasks/futures are finally cancelled do not matter here.  When the task/futures are really cancelled, we leverage the task status update RPC call to report back the finally status and the Scheduler can update the task related metadata and task slots later. If the task/future becomes zombie, the executor can also report back zombie task status to Scheduler.

andygrove:
LGTM. Another big step forward. Thanks @thinkharderdev

ballista/rust/executor/src/executor_server.rs
thinkharderdev: The cancellation only flips an atomic boolean and doesn't actually wait for the task to stop executing, so there should never be a case where the cancellation takes a long time. The underlying task may take some time before it reaches a yield point and actually stops executing but as far as I can tell there is no way to track that.

ballista/rust/executor/src/executor.rs
thinkharderdev: Yeah, when you wrap the inner future in an `Abortable` it adds another `Result` layer (to indicate whether the task was cancelled). So the first `?` is unwrapping the cancellation result (i.e. `Err` if the task was cancelled) and the second `?` is unwrapping the inner future's result.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: See below, the task cancellation itself is fast (only setting an `AtomicBoolean`). We're ultimately limited by the number of CPU cores available so I would worry that spawning thousands of tokio tasks to try and do the cancellations concurrently could actually cause a performance issue rather than alleviate one. If we did do it that way we would need a way to control the level of concurrency. In general, I think it's probably best to only try and optimize this once we've observed a performance issue related to it as I think the solution may be more complicated.

mingmwang: We will not spawn thousands of tokio tasks here in most cases, the concurrency level would  depends on how many executor instances we have in the system. I think sending the cancel request to multiple executors concurrently is better than sending the request sequentially. The same for assign/launch tasks to executors.  Usually in a data center, a remote RPC call could take 5ms to 20 ms.

ballista/rust/executor/src/executor_server.rs
mingmwang: Good to know.

andygrove:
@thinkharderdev There is a conflict that needs fixing before we can merge this one

andygrove:
@yahoNanJing @mingmwang @avantgardnerio This is the next PR to merge, but there are no approvals yet, and I'm sure if I should merge this yet.

ballista/rust/scheduler/src/state/task_manager.rs
avantgardnerio: Ultimately, might we want a `CancelTask(TaskIds[])` message so we can just send one message to each executor? Personally I'm not worried either way here since coroutines are just file handles - it's not like we're blocking a thread or spawning a thread.

avantgardnerio (approved):
LGTM. We have no way of cancelling tasks right now, and I don't want to let perfect be the enemy of good. 

I don't think we have other projects integrating with our protocol, so breaking changes to it are acceptable, and any performance concerns (if realized) can be addressed in future PRs.

ballista/rust/core/proto/ballista.proto
andygrove: I'd like to get this PR merged. Feel free to create a follow-on issue or PR to discuss this and other naming issues.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: @avantgardnerio This is what is happening already. We build a `HashMap<ExecutorId, Vec<Task>>` and then call each executor only once with all tasks that need to be cancelled on that executor.

PR #151, opened 2022-08-18
Stop Executor Impl, Executor Graceful Shutdown

mingmwang:

This PR covers below parts:
1. Implement Stop Executor grpc call
2. Graceful shutdown executor and related internal components.
    a). Graceful shutdown arrowflight and executor server grpc services.
    b). Graceful shutdown heartbeat loop, task runner loop and task status report loop
    c). Graceful shutdown and clean up shuffle data on the Executor

3. Better error handling and propagation during Executor startup.

In the current implementation, there are three cases could trigger the shutdown to the Executor
1.  Receive a stop executor grpc request
2. Receive Ctrl + c signal
3. Tonic errors during grpc service startup. 

 # Rationale for this change


mingmwang:
@thinkharderdev @yahoNanJing @andygrove  Please help to take a look.

yahoNanJing (approved):
Look great to me

ballista/rust/executor/src/executor_server.rs
```diff
@@ -310,25 +354,38 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskRunnerPool<T,
...
         //1. loop for task status reporting
         let executor_server = self.executor_server.clone();
+        let mut tasks_status_shutdown = graceful.subscribe_for_shutdown();
+        // Not used directly.
+        let tasks_status_complete = graceful.shutdown_complete_tx.clone();
         tokio::spawn(async move {
             info!("Starting the task status reporter");
-            loop {
+
+            // Drop and notifies the receiver half that the shutdown is complete
```
thinkharderdev, line 371:
```suggestion
            // Capture `task_status_complete`. This will be dropped when this background task returns, indicating that shutdown is complete
```

ballista/rust/executor/src/executor_server.rs
```diff
@@ -368,16 +425,31 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskRunnerPool<T,
 
         //2. loop for task fetching and running
         let executor_server = self.executor_server.clone();
+        let mut task_runner_shutdown = graceful.subscribe_for_shutdown();
+        // Not used directly.
+        let task_runner_complete = graceful.shutdown_complete_tx.clone();
         tokio::spawn(async move {
             info!("Starting the task runner pool");
+
+            // Drop and notifies the receiver half that the shutdown is complete
```
thinkharderdev, line 434:
```suggestion
            // Capture `task_runner_complete`. This will be dropped when this background task returns, indicating that shutdown is complete
```

ballista/rust/executor/src/main.rs
```diff
@@ -154,57 +162,156 @@ async fn main() -> Result<()> {
...
+                    default_codec,
+                    stop_send,
+                    &graceful,
+                )
+                .await?,
+            );
         }
         _ => {
-            tokio::spawn(execution_loop::poll_loop(
+            service_handlers.push(tokio::spawn(execution_loop::poll_loop(
```
thinkharderdev, line 221: Why does the poll loop not listen for shutdown signals?

thinkharderdev (commented):
Looks great! A few nits and questions

ballista/rust/executor/src/main.rs
```diff
@@ -154,57 +162,156 @@ async fn main() -> Result<()> {
...
+    )));
+
+    // Concurrently run the service checking and listen for the `shutdown` signal and wait for the stop request coming.
+    // The check_services runs until an error is encountered, so under normal circumstances, this `select!` statement runs
+    // until the `shutdown` signal is received or a stop request is coming.
+    tokio::select! {
+        service_val = check_services(&mut service_handlers) => {
+             info!("services stopped with reason {:?}", service_val);
+        },
+        _ = signal::ctrl_c() => {
```
avantgardnerio, line 240: My main concern is that `crtl_c` gets called when the process receives a SIGINT on linux. This is how kubernetes/docker will ask a container politely to shutdown. From the docs it does appear SIGINT will trigger `ctrl_c` but I just thought I'd mention something here to see if you tried that @mingmwang .

ballista/rust/executor/src/graceful.rs
```diff
@@ -0,0 +1,58 @@
...
+    ///
+    /// Tokio channels are closed once all `Sender` handles go out of scope.
+    /// When a channel is closed, the receiver receives `None`. This is
+    /// leveraged to detect all shutdown processing completing.
+    pub shutdown_complete_rx: mpsc::Receiver<()>,
+
+    pub shutdown_complete_tx: mpsc::Sender<()>,
+}
+
+impl Graceful {
```
avantgardnerio, line 36: `Graceful` is an adjective, and therefor not a great name. I would suggest `ShutdownNotifier` or similar.

ballista/rust/executor/src/main.rs
```diff
@@ -154,57 +162,156 @@ async fn main() -> Result<()> {
...
-                    error!("Ballista executor fail to clean_shuffle_data {:?}", e)
-                }
+            // Drop and notifies the receiver half that the shutdown is complete
+            let _shuffle_cleaner_complete = shuffle_cleaner_complete;
+            // As long as the shutdown notification has not been received
+            while !shuffle_cleaner_shutdown.is_shutdown() {
+                tokio::select! {
+                    _ = interval_time.tick() => {
+                            if let Err(e) = clean_shuffle_data_loop(&work_dir, cleanup_ttl as i64).await
+                        {
```
avantgardnerio, line 182: Superfluous scope?

ballista/rust/executor/src/executor_server.rs
```diff
@@ -289,18 +317,34 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> Heartbeater<T, U>
...
-    async fn start(&self) {
+    fn start(&self, graceful: &Graceful) {
         let executor_server = self.executor_server.clone();
+        let mut heartbeat_shutdown = graceful.subscribe_for_shutdown();
+        // Not used directly.
+        let heartbeat_complete = graceful.shutdown_complete_tx.clone();
         tokio::spawn(async move {
             info!("Starting heartbeater to send heartbeat the scheduler periodically");
-            loop {
+            // Drop and notifies the receiver half that the shutdown is complete
```
avantgardnerio, line 327: Maybe an explicit call to `drop` would be better in this case than a comment explaining that it will be dropped...

ballista/rust/executor/src/main.rs
```diff
@@ -154,57 +162,156 @@ async fn main() -> Result<()> {
...
+                            error!("Ballista executor fail to clean_shuffle_data {:?}", e)
+                        }
+                        },
```
avantgardnerio, line 185: There is some very funky formatting going on here. Could you run formatting on the sections that have been changed please? @andygrove any opposition to adding format checks to CI for Ballista?

avantgardnerio (changes requested):
Mostly minor nits. I do think `Graceful` should be renamed though. My main functional concern is verification shutdown is graceful in Docker, but I think we should merge this either way - even if SIGINT isn't handled in this PR, it certainly lays the groundwork for it in a following PR.

ballista/rust/executor/src/execution_loop.rs
```diff
@@ -108,7 +107,7 @@ pub async fn poll_loop<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                 }
             }
             Err(error) => {
-                warn!("Executor registration failed. If this continues to happen the executor might be marked as dead by the scheduler. Error: {}", error);
+                warn!("Executor poll work loop failed. If this continues to happen the Scheduler might be marked as dead. Error: {}", error);
```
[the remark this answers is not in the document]
mingmwang: If I understand correctly, currently in the Scheduler side, there is no logic to mark an Executor dead or blacklist an Executor. And I think in a distributed system, if a remote RPC call failed,  callers usually suspect that the other part might has a problem.

ballista/rust/executor/src/main.rs
mingmwang: Good catch. The reason I didn't add the shutdown logic here is just because this poll loop is also used by standalone Executor, I'm not sure whether a standalone Executor need graceful shutdown or not.

ballista/rust/executor/src/main.rs
mingmwang: I did not get chance to test it on kubernetes/docker containers yet. I see that some other systems have the similar logic. Let me do some test and will get back on this.

mingmwang: https://hyper.rs/guides/server/graceful-shutdown/

avantgardnerio (approved):
LGTM!

ballista/rust/executor/src/executor_server.rs
mingmwang: I will take @avantgardnerio suggestion and call the drop explicitly.

ballista/rust/executor/src/executor_server.rs
mingmwang: Calling the drop explicitly is better in most case, but if the internal loop has complex branches, we will have to call the drop in all the possible branches which is error prone.

avantgardnerio: That's a well-reasoned explanation, thank you :+1:

mingmwang:
@alamb @andygrove 
]Please help to approve and merge the PR.

ballista/rust/executor/src/main.rs
andygrove: I didn't realize that we were missing CI checks here. +1 for adding them.

andygrove:
Thanks @mingmwang for the work and thank you, @thinkharderdev @yahoNanJing and @avantgardnerio for reviewing. I will merge this now.

Issue #142, opened 2022-08-15
Save encoded execution plan in the ExecutionStage to reduce cost of task serialization and deserialization


yahoNanJing:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently we save the execution plan by its decoded format. When scheduling tasks to executors, we encode them. When there're thousands of tasks for a stage, it will cost too much for the encoding process.

**Describe the solution you'd like**

It's better for us to just save the execution plan by its encoded format, since there are few places that we need the decoded execution plan in the scheduler.

**Describe alternatives you've considered**

**Additional context**

PR #153, opened 2022-08-24
Introduce CuratorTaskManager for make an active job be curated by only one scheduler

### LINKED ISSUE #130: Introduce CuratorTaskManager for make an active job be curated by only one scheduler

After introducing #59, the previous cache layer is almost ineffective, which will downgrade the task scheduling performance very much especially for scheduling thousands of tasks.

It's better to introduce CuratorTaskManager for make an active job be curated by only one scheduler. Then we can leverage cache for the active jobs to avoid serialization and deserialization cost.

To achieve this, we need the following things:
1. Introduce scheduler id for execution graph as its curator
2. Extract task status from *ExecutionStage* 
3. Extract job status from *ExecutionGraph* 
4. Introduce cache for the active execution graph in *TaskManager*
5. Make the executor grpc server able to know which scheduler the requests are from
6. Make the executor able to update task status to its curator scheduler
7. Error handling:
    - When one scheduler is dead and executors fail to update tasks to this scheduler, at the first stage, when other scheduler receives such task status update request, it can simply mark its related job failed. Later we can improve by stage-based recovering. 

**Additional context**

yahoNanJing:

Closes #130.

 # Rationale for this change

### DIFF
--- a/ballista/rust/scheduler/src/scheduler_server/event.rs
+++ b/ballista/rust/scheduler/src/scheduler_server/event.rs
@@ -36,8 +36,8 @@ pub enum QueryStageSchedulerEvent {
         plan: Box<LogicalPlan>,
     },
     JobSubmitted(String),
-    // For a job fails without its execution graph
-    JobFailed(String, String),
+    // For a job which failed during planning
+    JobPlanningFailed(String, String),
     JobFinished(String),
     // For a job fails with its execution graph setting failed
     JobRunningFailed(String),
--- a/ballista/rust/scheduler/src/scheduler_server/event.rs
+++ b/ballista/rust/scheduler/src/scheduler_server/event.rs
@@ -36,6 +36,10 @@ pub enum QueryStageSchedulerEvent {
         plan: Box<LogicalPlan>,
     },
     JobSubmitted(String),
-    JobFinished(String),
+    // For a job fails without its execution graph
     JobFailed(String, String),
+    JobFinished(String),
+    // For a job fails with its execution graph setting failed
+    JobRunningFailed(String),
+    JobUpdated(String),
 }
--- a/ballista/rust/scheduler/src/scheduler_server/grpc.rs
+++ b/ballista/rust/scheduler/src/scheduler_server/grpc.rs
@@ -391,7 +391,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
                     .and_then(|m| {
                         m.try_into_logical_plan(
                             session_ctx.deref(),
-                            self.codec.logical_extension_codec(),
+                            self.state.codec.logical_extension_codec(),
                         )
                     })
                     .map_err(|e| {
@@ -535,16 +535,15 @@ mod test {
     #[tokio::test]
     async fn test_poll_work() -> Result<(), BallistaError> {
         let state_storage = Arc::new(StandaloneClient::try_new_temporary()?);
-        let namespace = "default";
         let scheduler: SchedulerServer<LogicalPlanNode, PhysicalPlanNode> =
             SchedulerServer::new(
+                "localhost:50050".to_owned(),
                 state_storage.clone(),
-                namespace.to_owned(),
                 BallistaCodec::default(),
             );
         let exec_meta = ExecutorRegistration {
             id: "abc".to_owned(),
-            optional_host: Some(OptionalHost::Host("http://host:8080".to_owned())),
+            optional_host: Some(OptionalHost::Host("http://localhost:8080".to_owned())),
             port: 0,
             grpc_port: 0,
             specification: Some(ExecutorSpecification { task_slots: 2 }.into()),
@@ -564,7 +563,6 @@ mod test {
         let state: SchedulerState<LogicalPlanNode, PhysicalPlanNode> =
             SchedulerState::new(
                 state_storage.clone(),
-                namespace.to_string(),
                 default_session_builder,
                 BallistaCodec::default(),
             );
@@ -580,7 +578,7 @@ mod test {
         assert_eq!(stored_executor.grpc_port, 0);
         assert_eq!(stored_executor.port, 0);
         assert_eq!(stored_executor.specification.task_slots, 2);
-        assert_eq!(stored_executor.host, "http://host:8080".to_owned());
+        assert_eq!(stored_executor.host, "http://localhost:8080".to_owned());
 
         let request: Request<PollWorkParams> = Request::new(PollWorkParams {
             metadata: Some(exec_meta.clone()),
@@ -598,7 +596,6 @@ mod test {
         let state: SchedulerState<LogicalPlanNode, PhysicalPlanNode> =
             SchedulerState::new(
                 state_storage.clone(),
-                namespace.to_string(),
                 default_session_builder,
                 BallistaCodec::default(),
             );
@@ -614,7 +611,7 @@ mod test {
         assert_eq!(stored_executor.grpc_port, 0);
         assert_eq!(stored_executor.port, 0);
         assert_eq!(stored_executor.specification.task_slots, 2);
-        assert_eq!(stored_executor.host, "http://host:8080".to_owned());
+        assert_eq!(stored_executor.host, "http://localhost:8080".to_owned());
 
         Ok(())
     }
--- a/ballista/rust/scheduler/src/scheduler_server/grpc.rs
+++ b/ballista/rust/scheduler/src/scheduler_server/grpc.rs
@@ -561,7 +561,7 @@ mod test {
         // no response task since we told the scheduler we didn't want to accept one
         assert!(response.task.is_none());
         let state: SchedulerState<LogicalPlanNode, PhysicalPlanNode> =
-            SchedulerState::new(
+            SchedulerState::new_with_default_scheduler_name(
                 state_storage.clone(),
                 default_session_builder,
                 BallistaCodec::default(),
@@ -594,7 +594,7 @@ mod test {
         // still no response task since there are no tasks in the scheduler
         assert!(response.task.is_none());
         let state: SchedulerState<LogicalPlanNode, PhysicalPlanNode> =
-            SchedulerState::new(
+            SchedulerState::new_with_default_scheduler_name(
                 state_storage.clone(),
                 default_session_builder,
                 BallistaCodec::default(),
--- a/ballista/rust/scheduler/src/state/execution_graph.rs
+++ b/ballista/rust/scheduler/src/state/execution_graph.rs
@@ -15,61 +15,40 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::display::DisplayableBallistaExecutionPlan;
-use crate::planner::DistributedPlanner;
+use std::collections::HashMap;
+use std::convert::TryInto;
+use std::fmt::{Debug, Formatter};
+use std::sync::Arc;
+
+use datafusion::physical_plan::display::DisplayableExecutionPlan;
+use datafusion::physical_plan::{
+    accept, ExecutionPlan, ExecutionPlanVisitor, Partitioning,
+};
+use datafusion::prelude::SessionContext;
+use datafusion_proto::logical_plan::AsLogicalPlan;
+use log::{error, warn};
+
 use ballista_core::error::{BallistaError, Result};
 use ballista_core::execution_plans::{ShuffleWriterExec, UnresolvedShuffleExec};
-
 use ballista_core::serde::protobuf::{
-    self, CompletedJob, JobStatus, OperatorMetricsSet, QueuedJob, TaskStatus,
+    self, execution_graph_stage::StageType, CompletedJob, JobStatus, QueuedJob,
+    TaskStatus,
 };
 use ballista_core::serde::protobuf::{job_status, FailedJob, ShuffleWritePartition};
 use ballista_core::serde::protobuf::{task_status, RunningTask};
 use ballista_core::serde::scheduler::{
     ExecutorMetadata, PartitionId, PartitionLocation, PartitionStats,
 };
-use datafusion::physical_plan::{
-    accept, ExecutionPlan, ExecutionPlanVisitor, Metric, Partitioning,
-};
-use log::{debug, info};
-use std::collections::HashMap;
-use std::convert::TryInto;
-use std::fmt::{Debug, Formatter};
-
-use ballista_core::serde::physical_plan::from_proto::parse_protobuf_hash_partitioning;
-use ballista_core::serde::scheduler::to_proto::hash_partitioning_to_proto;
 use ballista_core::serde::{AsExecutionPlan, BallistaCodec};
-use ballista_core::utils::collect_plan_metrics;
-use datafusion::physical_plan::display::DisplayableExecutionPlan;
-use datafusion::physical_plan::metrics::{MetricValue, MetricsSet};
-use datafusion::prelude::SessionContext;
-use datafusion_proto::logical_plan::AsLogicalPlan;
-use std::sync::Arc;
 
-/// Represents the basic unit of work for the Ballista executor. Will execute
-/// one partition of one stage on one task slot.
-#[derive(Clone)]
-pub struct Task {
-    pub session_id: String,
-    pub partition: PartitionId,
-    pub plan: Arc<dyn ExecutionPlan>,
-    pub output_partitioning: Option<Partitioning>,
-}
+use crate::display::print_stage_metrics;
+use crate::planner::DistributedPlanner;
+use crate::state::execution_graph::execution_stage::{
+    CompletedStage, ExecutionStage, FailedStage, ResolvedStage, RunningStage,
+    UnResolvedStage,
+};
 
-impl Debug for Task {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        let plan = DisplayableExecutionPlan::new(self.plan.as_ref()).indent();
-        write!(
-            f,
-            "Task[session_id: {}, job: {}, stage: {}, partition: {}]\n{}",
-            self.session_id,
-            self.partition.job_id,
-            self.partition.stage_id,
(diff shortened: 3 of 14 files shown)


yahoNanJing:
Hi @thinkharderdev, @avantgardnerio and @andygrove, could you help review this PR?

A few other PRs will be proposed after this PR finishes.

yahoNanJing:
retest please

ballista/rust/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -390,7 +390,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
                     .and_then(|m| {
                         m.try_into_logical_plan(
                             session_ctx.deref(),
-                            self.codec.logical_extension_codec(),
+                            self.state.codec.logical_extension_codec(),
```
avantgardnerio, line 393: Sorry, I'm daft - why does the codec move out of state? It needs to be unique to the scheduler instance for some reason?

ballista/rust/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -514,16 +514,15 @@ mod test {
...
             SchedulerServer::new(
+                "localhost:50050".to_owned(),
                 state_storage.clone(),
-                namespace.to_owned(),
                 BallistaCodec::default(),
             );
         let exec_meta = ExecutorRegistration {
             id: "abc".to_owned(),
-            optional_host: Some(OptionalHost::Host("http://host:8080".to_owned())),
+            optional_host: Some(OptionalHost::Host("http://localhost:8080".to_owned())),
```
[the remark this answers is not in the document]
avantgardnerio: oh, I see this is just for testing n/m

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -410,87 +594,33 @@ impl ExecutionGraph {
...
-            let partitions: usize = stage.partitions as usize;
-
-            let mut task_statuses: Vec<Option<task_status::Status>> =
-                vec![None; partitions];
-
-            for status in stage.task_statuses {
-                if let Some(task_id) = status.task_id.as_ref() {
-                    task_statuses[task_id.partition_id as usize] = status.status
+        for graph_stage in proto.stages {
+            let stage_type = graph_stage.stage_type.expect("Unexpected empty stage");
```
avantgardnerio, line 598: I'd love to see this turned into a `?`.

avantgardnerio (approved):
I'm not super familiar with the Ballista protocol and lifecycle of the execution graph and stages yet, but at a high level this makes sense to me - if we can keep a job localized to a scheduler, we can cache plans and that has helped performance at ebay?

If this is code that has been in use and is just getting upstreamed now, then I think this PR is pretty safe to merge. The code and comments all look like an improvement, and the logical change makes sense to me.

I'd love to hear if @andygrove or @thinkharderdev have any deeper observations, but this LGTM! Thank you for contributing back upstream, your help is very much appreciated as my company will be soon running into these same issues as we go into production!

andygrove:
I can only review non-trivial Ballista PRs like this in my free time since my employer has no interest in the project, and I don't have much free time right now. I am happy to approve and merge based on consensus from the people that are actually using the project. Based on the description this sounds like a good improvement.

thinkharderdev:
I didn't have spare cycles today to look at this in depth. I'll review more carefully tomorrow.

ballista/rust/scheduler/src/scheduler_server/event.rs
```diff
@@ -36,6 +36,10 @@ pub enum QueryStageSchedulerEvent {
         plan: Box<LogicalPlan>,
     },
     JobSubmitted(String),
-    JobFinished(String),
+    // For a job fails without its execution graph
     JobFailed(String, String),
```
thinkharderdev, line 40:
```suggestion
    // For a job which failed during planning
    JobPlanningFailed(String, String),
```

ballista/rust/scheduler/src/state/execution_graph/execution_stage.rs
```diff
@@ -0,0 +1,914 @@
...
+///                                ↓
+///                         CompletedStage
+#[derive(Clone)]
+pub(super) enum ExecutionStage {
+    UnResolved(UnResolvedStage),
+    Resolved(ResolvedStage),
+    Running(RunningStage),
+    Completed(CompletedStage),
+    Failed(FailedStage),
+}
```
thinkharderdev, line 56: I really like representing this as an enum!

ballista/rust/scheduler/src/state/execution_graph/execution_stage.rs
```diff
@@ -0,0 +1,914 @@
...
+            ExecutionStage::Running(running_stage) => running_stage.fmt(f),
+            ExecutionStage::Completed(completed_stage) => completed_stage.fmt(f),
+            ExecutionStage::Failed(failed_stage) => failed_stage.fmt(f),
+        }
+    }
+}
+
+/// For a stage whose input stages are not all completed, we say it's a unresolved stage
+#[derive(Clone)]
+pub(super) struct UnResolvedStage {
```
thinkharderdev, line 72:
```suggestion
pub(super) struct UnresolvedStage {
```

ballista/rust/scheduler/src/state/execution_graph/execution_stage.rs
```diff
@@ -0,0 +1,914 @@
...
+    pub(super) plan: Arc<dyn ExecutionPlan>,
+    /// Status of each already scheduled task. If status is None, the partition has not yet been scheduled
+    pub(super) task_statuses: Vec<Option<task_status::Status>>,
+    /// Combined metrics of the already finished tasks in the stage, If it is None, no task is finished yet.
+    pub(super) stage_metrics: Option<Vec<MetricsSet>>,
+    /// Error message
+    pub(super) error_message: String,
+}
+
+impl UnResolvedStage {
```
thinkharderdev, line 174:
```suggestion
impl UnresolvedStage {
```

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -55,44 +58,62 @@ pub struct TaskManager<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
         self.state
             .put(
                 Keyspace::ActiveJobs,
                 job_id.to_owned(),
-                self.encode_execution_graph(graph)?,
+                self.encode_execution_graph(graph.clone())?,
             )
             .await?;
 
+        graph.revive();
```
thinkharderdev, line 102: Why would we need to call `revive` here?

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -15,61 +15,40 @@
 // specific language governing permissions and limitations
```
thinkharderdev, line 15: Should this entire file move to src/state/execution_graph/mod.rs ?

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -106,123 +127,63 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
-                    }
-                }
+            let graph = self.get_active_execution_graph(&job_id).await;
+            let job_event = if let Some(graph) = graph {
+                let mut graph = graph.write().await;
+                graph.update_task_status(executor, statuses)?
+            } else {
+                // TODO Deal with curator changed case
+                error!("Fail to find job {} in the active cache and it may not be curated by this scheduler", job_id);
+                None
```
thinkharderdev, line 165: For now it seems like we should explicitly fail the job here so the job state gets updated.

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -37,16 +37,19 @@ use ballista_core::serde::{AsExecutionPlan, BallistaCodec};
...
 use rand::{thread_rng, Rng};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::default::Default;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 use tonic::transport::Channel;
 
 type ExecutorClients = Arc<RwLock<HashMap<String, ExecutorGrpcClient<Channel>>>>;
+type ExecutionGraphCache = Arc<RwLock<HashMap<String, Arc<RwLock<ExecutionGraph>>>>>;
```
thinkharderdev, line 52: What do you think about making this a `Arc<RwLock<VecDequeue<(String,ExecutionGraph)>>>`? Lookup by job ID becomes `O(n)` but the number of active jobs shouldn't be prohibitive and you get a proper queue in return so we can do FIFO scheduling of new jobs.

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -37,16 +37,19 @@ use ballista_core::serde::{AsExecutionPlan, BallistaCodec};
...
 
 #[derive(Clone)]
 pub struct TaskManager<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> {
```
thinkharderdev, line 55: This requires changes to the executors as well right? Currently the executor will just have a single connection to the scheduler but it will need to accept the scheduler URL as part of the task definition for this to work correctly.

thinkharderdev (commented):
Awesome work! This should be a big improvement. Had a few questions and some minor nits.

ballista/rust/scheduler/src/scheduler_server/grpc.rs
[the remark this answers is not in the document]
yahoNanJing: Thanks for your suggestion. Agree that `127.0.0.1` is much safer

ballista/rust/scheduler/src/scheduler_server/grpc.rs
yahoNanJing: Since the codec is already a field of the state and is a necessary field for the state, I think it's better for the state to own the codec rather than the server. And it's better to keep unique instance of the codec in case that we change it in the future.

ballista/rust/scheduler/src/state/execution_graph.rs
yahoNanJing: The `stage_type` is an option. Can we apply `?` to an option.

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -175,145 +154,226 @@ impl ExecutionGraph {
...
     /// An ExecutionGraph is complete if all its stages are complete
     pub fn complete(&self) -> bool {
-        self.stages.values().all(|s| s.complete())
+        self.stages
+            .values()
+            .all(|s| matches!(s, ExecutionStage::Completed(_)))
+    }
+
+    /// Revive the execution graph by converting the resolved stages to running changes
+    /// If existing such change, return true; else false.
```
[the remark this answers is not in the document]
yahoNanJing: Agree.

ballista/rust/scheduler/src/scheduler_server/event.rs
yahoNanJing: Agree

ballista/rust/scheduler/src/state/execution_graph/execution_stage.rs
yahoNanJing: Agree

ballista/rust/scheduler/src/state/execution_graph/execution_stage.rs
yahoNanJing: Agree

ballista/rust/scheduler/src/state/task_manager.rs
yahoNanJing: Running stages will not be persisted to the backend. Before cache the `execution_graph`, we need to revive it by converting all revolved stages to the running stages. 

And it's better to make the plan in the running stages be the encoded one to avoid encoding cost when creating hundreds of task definitions as mentioned in #142. Later I'll propose a PR for it.

ballista/rust/scheduler/src/state/task_manager.rs
yahoNanJing: Since now we haven't implemented the multi-scheduler task updating logic in the executor, I prefer to leave the handling logic empty here.

ballista/rust/scheduler/src/state/task_manager.rs
yahoNanJing: I don't think it's a good idea 🤣. If we dequeue the `ExecutionGraph`, and meanwhile we lookup the graph, we can get nothing. And the number of active jobs may be large. In our case, it may be thousands.

ballista/rust/scheduler/src/state/task_manager.rs
yahoNanJing: You are correct. For this part, I'll propose another PR.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: Ok, we can sort it out in a future PR. But I think we do want to take advantage of this change to introduce proper queueing for the jobs for fair scheduling. It could be as simple as another `VecDequeue<String>` that holds represents the queue which we could use to indicate priority.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: I think we would have to implement that as part of this PR or else any multi-scheduler deployment would break.

yahoNanJing: I don't think we can achieve multiple scheduler now, since there's only one scheduler in the executor.

thinkharderdev: The executor only has one scheduler connection but the scheduler's can (currently) assign tasks to any executor since the executor pool is a shared global state. So they would currently report status to only one scheduler instance (unless the scheduler endpoint was pointing to a load balancer or something) but there is no guarantee that they would have received the task assignment from that scheduler. 

I think it should be relatively straightforward though. The `TaskAssignment` would just need a scheduler URL in it and the executor would report status back to that URL (instead of maintaining a single global connection).

yahoNanJing: If there's only one scheduler in the executor, how can the executors register themselves to multiple schedulers?

yahoNanJing: And I don't think it's a good idea to include scheduler url to the task definition. We can achieve the scheduler url in the grpc request like the scheduler does for the executors.

ballista/rust/scheduler/src/scheduler_server/grpc.rs
avantgardnerio: Oh, I'm sorry, I got the direction wrong here: it went into state - a mistake with my diff tool :+1:

ballista/rust/scheduler/src/state/execution_graph.rs
avantgardnerio: Trivia: I have learned recently that the `?` operator works with `Option<>`s as well as `Results<>` if the outer function itself also returns an option.

Here it would need to be mapped to a `Result<>` though:

```suggestion
            let stage_type = graph_stage.stage_type.ok_or(BallistaError::Internal("Unexpected empty stage"))?;
```

ballista/rust/scheduler/src/state/execution_graph.rs
avantgardnerio: I think `git mv` should keep it intact? Definitely if it's a single commit and we fast-forward (vs squash) merge it will.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: The executor only has one scheduler connection but when it registers it that one scheduler will add the executor to the global state. At that point, all schedulers will see that resource and be able to launch tasks on it. For pull-based scheduling this will be fine but for push-based scheduling it would break down.

andygrove:
@yahoNanJing This is the next PR to merge but there are some conflicts now. Could you fix them?

ballista/rust/scheduler/src/state/task_manager.rs
andygrove: I need to get back to speed with the current architecture before I can help with this review. I won't have time to do that until this weekend, unfortunately. 

It would be good if we had a doc in this repo with the architecture we are working towards to make sure we are all on the same page. I can maybe try and start that if nobody beats me to it.

yahoNanJing: For push-based task scheduling, if all of the executors report task update to only one scheduler, when the whole cluster is busy, I think there's only one scheduler will be busy scheduling tasks and other schedulers will be idle due to there're no `ExecutorReservation` left for them.

yahoNanJing: I'm OK to add another commit of the multiple scheduler change in the executors to this PR. However, it may take a few days.

yahoNanJing:
Thanks @andygrove. Just solved the conflicts by rebasing.

ballista/rust/scheduler/src/state/task_manager.rs
thinkharderdev: What we do is configure the scheduler endpoint in the executor to be the kubernetes service endpoint. So this will load balance executor registrations across the schedulers  but any given scheduler should maintain a persistent connection to only the scheduler it registered with.

yahoNanJing: Got it. I'll propose the commit in early next week.

andygrove:
Thanks for the PR @yahoNanJing and thanks for the reviews @thinkharderdev and @avantgardnerio. It looks like there is consensus here, so I will go ahead and merge this.

PR #157, opened 2022-08-25
[Python] Fix Ballista Python bindings 

andygrove:

N/A

 # Rationale for this change

We can now run queries in Python again!

```
>>> import ballista
>>> ctx = ballista.BallistaContext("localhost", 50050)
>>> df = ctx.sql("SELECT 1")
>>> df.show()
+----------+
| Int64(1) |
+----------+
| 1        |
+----------+
```

- Copy latest from `arrow-datafusion-python` 
- Change names from datafusion to ballista
- Add BallistaContext from https://github.com/apache/arrow-ballista/pull/58
- Update CI to no longer use an old nightly Rust

No


andygrove:
@thinkharderdev @yahoNanJing @avantgardnerio This fixes the Python bindings for Ballista. This will help make Ballista easy to try out for new users and make it easier to write tutorials.

thinkharderdev:
LGTM

thinkharderdev (approved):

avantgardnerio:
LGTM. The one thing I was really hoping to see I didn't: can we remove the `exclude = ["python"]`? I think this means it doesn't build in CI, which is how it gets broken?

avantgardnerio (approved):

andygrove:
> LGTM. The one thing I was really hoping to see I didn't: can we remove the `exclude = ["python"]`? I think this means it doesn't build in CI, which is how it gets broken?

There is a GitHub action that builds the Python module: https://github.com/apache/arrow-ballista/runs/8038213148?check_suite_focus=true

andygrove:
It would be good to add to the workspace too. I will add that to this PR.

andygrove:
> It would be good to add to the workspace too. I will add that to this PR.

That broke CI so will handle in a follow-on PR.

andygrove:
Thanks for the reviews @thinkharderdev and @avantgardnerio!

PR #161, opened 2022-08-27
Make the user guide about Ballista, not DataFusion

andygrove:

Part of https://github.com/apache/arrow-ballista/issues/160

 # Rationale for this change

The user guide in this repo was the DataFusion user guide and there were no links to the Ballista content.

This is the first step in getting the docs up-to-date. There is much more work to do after this.

When we release 0.8 we can publish this user guide to the Arrow web site.

- Delete DataFusion content
- Add Ballista content to index

Preview:

docs

No


andygrove:
@thinkharderdev @yahoNanJing @avantgardnerio please review when you have time. This is just the first step to restore the old Ballista content, and I will follow up with smaller PRs to update the content.

avantgardnerio (approved):
I didn't actually clone it to verify links, but LGTM. I don't know much about python either, so I can't speak to that part. You said this was a base for other changes? Let's get it in then!

andygrove:
> I didn't actually clone it to verify links, but LGTM. I don't know much about python either, so I can't speak to that part. You said this was a base for other changes? Let's get it in then!

Thanks! Yes, this is old content that was approved a very long time ago. The PRs following on from this will be more interesting.

PR #184, opened 2022-09-01
Executor lost handling

### LINKED ISSUE #143: Executor Lost handling

Need to correctly handle the executor Lost case.
1. If Executor is lost and there are running tasks on the Executor, need to resubmit those failed tasks.
2. If Executor is lost and there are pending shuffle reading on the Executor, need to recompute the missing shuffle data.
3. Executor graceful shutdown

**Additional context**

mingmwang:

Closes #143. 
This PR covers the basic Executor lost handling in Scheduler side. 
Generally there are four cases could trigger the Executor lost handling, this PR covers the first two cases;
1) Executor graceful shutdown, Executor will call RPC to notify the Scheduler it is stopped.
2) Scheduler expired the dead Executors that do not receive heartbeats from them.
3) Shuffle Reader failure due to connectivity issue in reduce tasks(Not implemented, will cover this in following task failure related PRs).
4) Grpc connection lost detect(Not implemented).

 # Rationale for this change


mingmwang:
@yahoNanJing @thinkharderdev @andygrove 
Please help to take a look.

ballista/rust/scheduler/src/state/executor_manager.rs
```diff
@@ -383,15 +430,42 @@ impl ExecutorManager {
         Ok(())
     }
 
-    /// Initialize the set of executor heartbeats from storage
-    pub(crate) async fn init_executor_heartbeats(&self) -> Result<()> {
+    pub(crate) async fn save_dead_executor_heartbeat(
```
thinkharderdev, line 433: I'm a little confused as to why we keep the dead executors around in the state.

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -485,8 +490,139 @@ impl ExecutionGraph {
         self.output_locations.clone()
     }
 
+    /// Reset running and completed stages on a given executor
```
thinkharderdev, line 493: There is a lot of subtle logic here. I went through it the best I could and it seems ok but we should really have some test coverage for this.

thinkharderdev (commented):
Awesome work! The logic is quite complicated here so I think we could really use some test coverage.

ballista/rust/scheduler/src/state/execution_graph.rs
mingmwang: Yes, I already added two UTs to test the reset logic.

````
async fn test_reset_completed_stage()

async fn test_reset_resolved_stage()

````

Comment (thinkharderdev) [reply ballista/rust/scheduler/src/state/execution_graph.rs:493]: Sorry I missed those (it's a rather large PR :)). I think it might be worth adding some coverage for different graph topologies though (union/aggregation/coalesce). 

This may be a bit more involved but it may also be worthwhile to test the end-to-end flow from the grpc endpoint. There are quite a few moving parts here and given this is an exceptional situation it will be hard to track down issues that pop up in production.

File: ballista/rust/scheduler/src/planner.rs [hunk 3]
```diff
@@ -246,6 +246,31 @@ pub fn remove_unresolved_shuffles(
     Ok(with_new_children_if_necessary(stage, new_children)?)
 }
 
+pub fn rollback_resolved_shuffles(
```
thinkharderdev, line 249: A comment here might be helpful as to what this function is doing

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -581,6 +726,54 @@ impl ExecutionGraph {
         }
     }
 
+    /// Convert running stage to be unresolved
+    fn rollback_running_stage(&mut self, stage_id: usize) -> Result<bool> {
```
thinkharderdev, line 730: I'm a little confused by this. If we have tasks that are already running then they will either fail (if they cannot fetch their input partitions from the lost executor) or have already fetched the partitions from the lost executor in which case we shouldn't roll back. I guess I'm not clear on how this process interacts with the incoming task updates. Shouldn't the rollback be in response to task failure instead of just the executor missing a heartbeat?

thinkharderdev (commented):
Had a chance to review more carefully and I'm not sure I understand how this will interact with task statuses that are arriving from the executors. I haven't traced all the way through the logic but from what I can tell, if a stage is rolled back due to an executor being lost and a task status comes in after the rollback (or during the rollback but is queued waiting for the write lock on the job) then it will error out since the stage is no longer running. But this will cause all task updates in that batch to fail.  This is not an error we currently handle so it can cause some cascading issues (including causing other jobs to stall out entirely since task updates are effectively dropped completely). 

I think trying to immediately rollback once an executor stops heart-beating may be too extreme. Perhaps the logic could be simpler (and less prone to causing collateral damage) if we did any stage rollback in response to actual task failure updates from executors. This does leave the issue of executors lost due to a SIGKILL or panic which causes a sudden failure but that seems like the smaller issue to me.

ballista/rust/scheduler/src/state/execution_graph.rs
mingmwang: Sure, I can add UT to test from the grpc endpoint.

ballista/rust/scheduler/src/state/execution_graph.rs
mingmwang: Yes, in this PR, If it decides to rollback, the enter stage will be rollbacked and rescheduled.  Even there are completed tasks, they will be rollback and rescheduled. This approach is simple and safe.  Because its input stages need to be recomputed, we are not sure whether the re-computation result is deterministic or not.

Regarding when to trigger the executor lost handling, as I noted in the PR's description, there could be four cases.  I will implement the shuffle reader failure check in a seperate PR which will focus on the task failure handing case.

1) Executor graceful shutdown, Executor will call RPC to notify the Scheduler it is stopped.
2) Scheduler expired the dead Executors that do not receive heartbeats from them.
3) Shuffle Reader failure due to connectivity issue in reduce tasks(Not implemented, will  cover this in following task failure related PRs).
4) Grpc connection lost detect(Not implemented).

I think we still need the Scheduler to expire the dead executors actively. For example if we have heavy tasks in the map stage and the executor is lost, no task updates come back from the lost executor, in this case we can only rely on the Scheduler expiration logic.
 
We can have the expiration time longer than the executor's heartbeat time interval, so only after missing two or three consecutive heartbeats, the Scheduler consider it is dead.

mingmwang:
> Had a chance to review more carefully and I'm not sure I understand how this will interact with task statuses that are arriving from the executors. I haven't traced all the way through the logic but from what I can tell, if a stage is rolled back due to an executor being lost and a task status comes in after the rollback (or during the rollback but is queued waiting for the write lock on the job) then it will error out since the stage is no longer running. But this will cause all task updates in that batch to fail. This is not an error we currently handle so it can cause some cascading issues (including causing other jobs to stall out entirely since task updates are effectively dropped completely).
> 

Good catching. There is some gap here, I will take a look and fix it.
In this PR, the executor lost handling and task statuses update will not interleave with each other, they should be no lock
contentions.  All the updates to the TaskManager can only be triggered through QueryStageSchedulerEvents. The events are queued and processed one by one.  Either QueryStageSchedulerEvent::ExecutorLost happens first or QueryStageSchedulerEvent::TaskUpdating happens first. 
If executor lost event is happened first, the QueryStageSchedulerEvent::TaskUpdating from that executor should be ignored. I already have some check in the task_status rpc call, there is some gap in the TaskManager::update_task_status, I will take a look.

thinkharderdev:
> > Had a chance to review more carefully and I'm not sure I understand how this will interact with task statuses that are arriving from the executors. I haven't traced all the way through the logic but from what I can tell, if a stage is rolled back due to an executor being lost and a task status comes in after the rollback (or during the rollback but is queued waiting for the write lock on the job) then it will error out since the stage is no longer running. But this will cause all task updates in that batch to fail. This is not an error we currently handle so it can cause some cascading issues (including causing other jobs to stall out entirely since task updates are effectively dropped completely).
> 
> Good catching. There is some gap here, I will take a look and fix it. In this PR, the executor lost handling and task statuses update will not interleave with each other, they should be no lock contentions. All the updates to the `TaskManager` can only be triggered through `QueryStageSchedulerEvents`. The events are queued and processed one by one. Either `QueryStageSchedulerEvent::ExecutorLost `happens first or `QueryStageSchedulerEvent::TaskUpdating `happens first. If executor lost event is happened first, the `QueryStageSchedulerEvent::TaskUpdating` from that executor should be ignored. I already have some check in the task_status rpc call, there is some gap in the `TaskManager::update_task_status`, I will take a look.

The scenario I am concerned about is a task update coming from a different, still-alive executor. So the issue wouldn't be interleaving but rather that we may consider a valid status update to be an error because it applies to a stage that is no longer running (since it has been rolled back). I think we need to make sure that scenario is handled in the task status update and we ignore them instead of erroring. 

Thinking about this a bit more. I think it would be a good idea to make the failure handling configurable. It may not be desired to roll back and recompute results. In an interactive query system for example, it may be desirable to just fail queries when an executor is lost rather than consume more cluster resources attempting to recompute results. Not for this PR of course, but what do you think?

ballista/rust/scheduler/src/state/execution_graph.rs
thinkharderdev: > We can have the expiration time longer than the executor's heartbeat time interval, so only after missing two or three consecutive heartbeats, the Scheduler consider it is dead.

I think this is a good idea. We probably want to be pretty conservative in marking an executor dead due to missing heartbeats.

mingmwang:
> > > Had a chance to review more carefully and I'm not sure I understand how this will interact with task statuses that are arriving from the executors. I haven't traced all the way through the logic but from what I can tell, if a stage is rolled back due to an executor being lost and a task status comes in after the rollback (or during the rollback but is queued waiting for the write lock on the job) then it will error out since the stage is no longer running. But this will cause all task updates in that batch to fail. This is not an error we currently handle so it can cause some cascading issues (including causing other jobs to stall out entirely since task updates are effectively dropped completely).
> > 
> > 
> > Good catching. There is some gap here, I will take a look and fix it. In this PR, the executor lost handling and task statuses update will not interleave with each other, they should be no lock contentions. All the updates to the `TaskManager` can only be triggered through `QueryStageSchedulerEvents`. The events are queued and processed one by one. Either `QueryStageSchedulerEvent::ExecutorLost `happens first or `QueryStageSchedulerEvent::TaskUpdating `happens first. If executor lost event is happened first, the `QueryStageSchedulerEvent::TaskUpdating` from that executor should be ignored. I already have some check in the task_status rpc call, there is some gap in the `TaskManager::update_task_status`, I will take a look.
> 
> The scenario I am concerned about is a task update coming from a different, still-alive executor. So the issue wouldn't be interleaving but rather that we may consider a valid status update to be an error because it applies to a stage that is no longer running (since it has been rolled back). I think we need to make sure that scenario is handled in the task status update and we ignore them instead of erroring.
> 
> Thinking about this a bit more. I think it would be a good idea to make the failure handling configurable. It may not be desired to roll back and recompute results. In an interactive query system for example, it may be desirable to just fail queries when an executor is lost rather than consume more cluster resources attempting to recompute results. Not for this PR of course, but what do you think?

Agree, we should make it configurable.  Fast fail queries do make sense in some scenario.  In future we will definitely re-implement the 'All-At-Once' scheduling and it should fast failed the queries.

mingmwang:
https://github.com/apache/arrow-ballista/issues/189

mingmwang:
@thinkharderdev @yahoNanJing 
Could you please help to take a look and approve the PR ?

thinkharderdev:
> Hi @thinkharderdev, are there still any concerns for this PR?

Hey, I think it's almost there but I believe @mingmwang was still looking at the issue with task updates. As is, this may cause task updates to fail for jobs unrelated to the dead executor (because we fail is any update in the batch fails).

mingmwang:
> xecutor (because we fail is any update in the

@thinkharderdev

It is already fixed in the PR. I also added a UT and verified.

```
 #[tokio::test]
 async fn test_task_update_after_reset_stage() 
```

thinkharderdev (approved):

thinkharderdev:
> > xecutor (because we fail is any update in the
> 
> @thinkharderdev
> 
> It is already fixed in the PR. I also added a UT and verified.
> 
> ```
>  #[tokio::test]
>  async fn test_task_update_after_reset_stage() 
> ```

Ah, sorry must have missed the update :)

LGTM

Issue #185, opened 2022-09-04
Cleanup job/stage status from TaskManager and clean up shuffle data after a period after JobFinished


mingmwang:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
(This section helps Arrow developers understand the context and *why* for this feature, in addition to  the *what*)

Today, there is no clean up logic to remove those job/stage status from StateBackend, the disk space might be exhausted quickly in a busy cluster. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

mingmwang:
https://github.com/apache/arrow-ballista/issues/9

PR #188, opened 2022-09-05
Clean up job data on both Scheduler and Executor

### LINKED ISSUE #9: Need clean up intermediate data in Ballista
We need to check whether the states saved in the sled is consumed by UI or not.
if not consumed by UI, we can clean the job/task data when the SQL is finished.

If they are consumed by UI, we can choose either LRU based policy like Spark or time based eviction policy.

Regarding shuffle files, we also need to implement a way to clean them. This is a little bit complex because we need to clean up the files on all the hosts. ~~We might need to add new RPCs~~ .

mingmwang:

Closes #9 and #185.

 # Rationale for this change


mingmwang:
@thinkharderdev @yahoNanJing @Ted-Jiang @andygrove

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -519,4 +589,54 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
+        let mut active_graph_cache = active_job_cache.write().await;
+        active_graph_cache.remove(&job_id);
+
+        let keyspace = if failed {
+            Keyspace::FailedJobs
+        } else {
+            Keyspace::CompletedJobs
+        };
+
+        let lock = state.lock(keyspace.clone(), "").await?;
```
Ted-Jiang, line 609: I think this will lock the whole 🤔 in standalone

```
 async fn lock(&self, keyspace: Keyspace, key: &str) -> Result<Box<dyn Lock>> {
        let mut mlock = self.locks.lock().await;
        let lock_key = format!("/{:?}/{}", keyspace, key);
        if let Some(lock) = mlock.get(&lock_key) {
            Ok(Box::new(lock.clone().lock_owned().await))
        } else {
            let new_lock = Arc::new(Mutex::new(()));
            mlock.insert(lock_key, new_lock.clone());
            Ok(Box::new(new_lock.lock_owned().await))
        }
    }
```

Ted-Jiang: Oh, this lock only works in func ` async fn lock(&self, keyspace: Keyspace, key: &str)`, will release after this func quick call.

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -519,4 +589,54 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
+        executor_manager
+            .map(|em| async { Self::clean_up_executors_data(job_id.clone(), em).await });
+        Ok(())
+    }
+
+    async fn clean_up_executors_data(job_id: String, executor_manager: ExecutorManager) {
+        let alive_executors = executor_manager.get_alive_executors_within_one_minute();
+        for executor in alive_executors {
+            let job_id_clone = job_id.to_owned();
+            let executor_manager_clone = executor_manager.clone();
```
Ted-Jiang, line 621: I think each SQL sends an RPC to all executors is not a good idea (for interactive query finish in ms)🤔

Ted-Jiang: Maybe we can do some improvement later !

ballista/rust/scheduler/src/state/task_manager.rs
```diff
@@ -299,6 +320,23 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
             }
         }
 
+        // spawn a delayed future to clean up job data on both Scheduler and Executors
+        let state = self.state.clone();
+        let job_id_str = job_id.to_owned();
+        let active_job_cache = self.active_job_cache.clone();
+        tokio::spawn(async move {
+            tokio::time::sleep(Duration::from_secs(CLEANUP_FINISHED_JOB_DELAY_SECS))
```
Ted-Jiang, line 328: (for interactive query finish in ms)  i think this will store a lot feature in heap. Just some opinion.

ballista/rust/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -481,16 +482,20 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
-            .task_manager
-            .cancel_job(&job_id, &self.state.executor_manager)
+        self.query_stage_event_loop
+            .get_sender()
+            .map_err(|e| {
+                let msg = format!("Get query stage event loop error due to {:?}", e);
+                error!("{}", msg);
+                Status::internal(msg)
+            })?
+            .post_event(QueryStageSchedulerEvent::JobCancel(job_id))
```
yahoNanJing, line 492: Nice refinement. It's better to have only one entrance to modify the scheduler state and all of the state changes should have a related event and be dealt with in the event loop.

andygrove:
I am constantly filling my disk up with shuffle files so would love to see us get this merged before the 0.9.0 release.

@mingmwang Could you rebase when you get a change and I will test this out and review the PR as well.

mingmwang:
> I am constantly filling my disk up with shuffle files so would love to see us get this merged before the 0.9.0 release.
> 
> @mingmwang Could you rebase when you get a chance and I will test this out and review the PR as well.

Sure, working on it.

ballista/rust/scheduler/src/state/task_manager.rs
mingmwang: Yes, in current code base, the TaskManager/ExecutionGraph do not track the tasks are executed by which executors.

mingmwang:
@andygrove @yahoNanJing @Ted-Jiang @yahoNanJing 

BTW, in this PR, the job data in the state store will also be deleted after 300s.
I think we need a following PR to move the completed(Success or Failed) job data from state store to ObjectStore 
for long time storing purpose, and Scheduler UI can read from the ObjectStore.

Please share your thoughts.

````
const CLEANUP_FINISHED_JOB_DELAY_SECS: u64 = 300;

async fn clean_up_job_data(
        state: Arc<dyn StateBackendClient>,
        active_job_cache: ExecutionGraphCache,
        failed: bool,
        job_id: String,
        executor_manager: Option<ExecutorManager>,
    ) -> Result<()> {
        let mut active_graph_cache = active_job_cache.write().await;
        active_graph_cache.remove(&job_id);

        let keyspace = if failed {
            Keyspace::FailedJobs
        } else {
            Keyspace::CompletedJobs
        };

        let lock = state.lock(keyspace.clone(), "").await?;
        with_lock(lock, state.delete(keyspace, &job_id)).await?;

        executor_manager
            .map(|em| async { Self::clean_up_executors_data(job_id.clone(), em).await });
        Ok(())
    }

File: ballista/rust/scheduler/src/state/task_manager.rs [hunk 5]
```diff
@@ -43,17 +43,17 @@ use rand::{thread_rng, Rng};
...
 use std::default::Default;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::RwLock;
-use tonic::transport::Channel;
 
-type ExecutorClients = Arc<RwLock<HashMap<String, ExecutorGrpcClient<Channel>>>>;
 type ExecutionGraphCache = Arc<RwLock<HashMap<String, Arc<RwLock<ExecutionGraph>>>>>;
 
+const CLEANUP_FINISHED_JOB_DELAY_SECS: u64 = 300;
```
andygrove, line 51: We should make this configurable. Some of the queries I am testing take much longer than 300 seconds. We already have the ability to set configs on the context.

andygrove: Never mind, this is a delay _after_ the job completes. I would still like to see this configurable but we could do that as a follow in PR.

andygrove (approved):
LGTM. Thanks @mingmwang

andygrove:
@mingmwang could you fix the conflicts here when you have the time so that we can merge this?

mingmwang:
Resolved conflicts.

andygrove:
Thanks again @mingmwang

PR #202, opened 2022-09-11
Executor uses all available cores by default + add tuning guide

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/218

 # Rationale for this change

By using all available cores by default, I see a ~4x improvement in performance when testing locally compared to the previous default of 4. This makes us more comparable to Spark, which also uses all available cores by default.

The tuning guide also explains the importance of configuring the number of shuffle partitions (our default for that is very low).

- Executor uses all available cores by default
- Starting point for a tuning guide that we can expand over time.

No


andygrove:
cc @thinkharderdev @yahoNanJing @avantgardnerio

andygrove:
@yahoNanJing @thinkharderdev Do you have any additional tips that we can add to this tuning guide based on your experiences scaling Ballista?

thinkharderdev:
> @yahoNanJing @thinkharderdev Do you have any additional tips that we can add to this tuning guide based on your experiences scaling Ballista?

Sorry, I've been a bit swamped this week. I will review this evening or tomorrow.

docs/source/user-guide/tuning-guide.md
```diff
@@ -0,0 +1,61 @@
...
+
+Each executor instance has a fixed number of tasks that it can process concurrently. This is specified by passing a
+`concurrent_tasks` command-line parameter. The default value is 4.
+
+Increasing this configuration setting will increase the number of tasks that each executor can run concurrently but
+this will also mean that the executor will use more memory. If executors are failing due to out-of-memory errors then
+decreasing the number of concurrent tasks can help.
+
+In the future, Ballista will have better support for tracking memory usage and allocating tasks based on available
+memory, as well as supporting spill-to-disk to reduce memory pressure.
```
thinkharderdev, line 61: I think the task concurrency is generally more constrained by CPU than memory. Maybe indicate here that best practice is to allocate one task slot per CPU core?

thinkharderdev (approved):
Looks great!

docs/source/user-guide/tuning-guide.md
andygrove: Thanks for the review @thinkharderdev. I added that and also changed the executor to use all available cores by default.

andygrove:
@yahoNanJing I made the change here for the executor to use all cores by default

yahoNanJing:
Thanks @andygrove. The patch is good for me. Another config influences the performance much is the shuffle partition number, *ballista.shuffle.partitions*. Currently it's just 2, which may be too small for some cases. We can improve it later.

yahoNanJing (approved):

PR #204, opened 2022-09-11
[Python] Allow some config options to be set when creating context

andygrove:

Part of https://github.com/apache/arrow-ballista/issues/203

 # Rationale for this change

Allow jobs to be tuned from Python

Allow some basic configs to be set from Python

No

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -43,7 +43,7 @@ import pyarrow
 f = ballista.functions
 
 # create a context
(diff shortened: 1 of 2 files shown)


andygrove:
I'll go ahead and merge this since it is a trivial change, and I don't think any of the existing committers are using these bindings? If anyone has any objections to me merging Python PRs without review and wants to help with reviews, please let me know!

PR #210, opened 2022-09-14
Add serde support for all binary operators in physical plan, including `StringConcat`

andygrove:

Related to https://github.com/apache/arrow-ballista/issues/201

 # Rationale for this change

This is just a quick fix to add the missing operators but https://github.com/apache/arrow-ballista/issues/201 is for improving the situation and avoiding the duplicate code.

### DIFF
--- a/ballista/rust/core/src/serde/mod.rs
+++ b/ballista/rust/core/src/serde/mod.rs
@@ -213,13 +213,23 @@ pub(crate) fn from_proto_binary_op(op: &str) -> Result<Operator, BallistaError> 
         "Modulo" => Ok(Operator::Modulo),
         "Like" => Ok(Operator::Like),
         "NotLike" => Ok(Operator::NotLike),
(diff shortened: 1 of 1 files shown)


ballista/rust/core/src/serde/mod.rs
```diff
@@ -213,13 +213,23 @@ pub(crate) fn from_proto_binary_op(op: &str) -> Result<Operator, BallistaError>
         "Modulo" => Ok(Operator::Modulo),
         "Like" => Ok(Operator::Like),
         "NotLike" => Ok(Operator::NotLike),
+        "IsDistinctFrom" => Ok(Operator::IsDistinctFrom),
```
andygrove, line 216: Note that this method is copied from the private `from_proto_binary_op` method in DataFusion. It would be better to make that public rather than maintain a copy here.

thinkharderdev (approved):

PR #212, opened 2022-09-15
Make executor logging less verbose

andygrove:

N/A

 # Rationale for this change

I am starting to do some performance profiling and tuning and I cannot read the plans because they are so huge.

- Change some logging from `info!` to `debug!`
- Makes the output much less verbose for shuffle reader exec:

```
2022-09-15T02:03:17.337693Z  INFO tokio-runtime-worker ThreadId(49) ballista_scheduler::display: === [Xl4tDeB/15] Stage finished, physical plan with metrics ===
ShuffleWriterExec: None, metrics=[output_rows=100, input_rows=100, repart_time=1ns, write_time=534.050341ms]
  GlobalLimitExec: skip=0, fetch=100, metrics=[output_rows=100, elapsed_compute=1.839828ms, spill_count=0, spilled_bytes=0, mem_used=0]
    SortExec: [c_customer_id@0 ASC NULLS LAST], metrics=[output_rows=8192, elapsed_compute=518.201936ms, spill_count=0, spilled_bytes=0]
      CoalescePartitionsExec, metrics=[output_rows=772575, elapsed_compute=51.948µs, spill_count=0, spilled_bytes=0, mem_used=0]
        ShuffleReaderExec: partitions=24, metrics=[fetch_time=392.660551ms]
```

Note that `ShuffleWriterExec` was already concise. This makes `ShuffleReaderExec` more consistent with that.

Do we still have a need to output all the paths to all the location being read? It was helpful early on in development for sure.

Ideally I would make this configurable but there does not seem to be an easy way to hook this into the config system?

Yes, easier to read logs, less disk space used.

### DIFF
--- a/ballista/rust/core/src/execution_plans/shuffle_reader.rs
+++ b/ballista/rust/core/src/execution_plans/shuffle_reader.rs
@@ -136,29 +136,7 @@ impl ExecutionPlan for ShuffleReaderExec {
     ) -> std::fmt::Result {
         match t {
             DisplayFormatType::Default => {
-                let loc_str = self
-                    .partition
-                    .iter()
-                    .enumerate()
-                    .map(|(partition_id, locations)| {
-                        format!(
-                            "[partition={} paths={}]",
-                            partition_id,
-                            locations
-                                .iter()
-                                .map(|l| l.path.clone())
-                                .collect::<Vec<String>>()
-                                .join(",")
-                        )
-                    })
-                    .collect::<Vec<String>>()
-                    .join(", ");
-                write!(
-                    f,
-                    "ShuffleReaderExec: partition_locations({})={}",
-                    self.partition.len(),
-                    loc_str
-                )
+                write!(f, "ShuffleReaderExec: partitions={}", self.partition.len(),)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


andygrove:
@yahoNanJing @thinkharderdev would be good to get your opinions on this

yahoNanJing:
Totally agree with you.

yahoNanJing:
And for the datafusion repo, we have similar verbose things in *datafusion/core/src/physical_optimizer/repartition.rs* and *datafusion/core/src/physical_plan/file_format/parquet.rs*. It's not necessary to display the partition detailed info

ballista/rust/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -136,29 +136,7 @@ impl ExecutionPlan for ShuffleReaderExec {
     ) -> std::fmt::Result {
         match t {
             DisplayFormatType::Default => {
-                let loc_str = self
```
andygrove, line 139: This was generating pages of data

avantgardnerio (approved):
I 100% endorse this change. Info messages should be something that a system administrator can parse manually understand and take action to tune the system. Metrics like this are even debatable at the debug level - they feel more like they should be exposed via an API like Prometheus that is meant for collecting stats.

PR #220, opened 2022-09-15
Propagate Ballista configs to DataFusion context in scheduler

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/213

 # Rationale for this change

I would like to be able to change DataFusion config settings to tune queries

Pass configs into DataFusion context

Yes, but we already documented this capability in the new tuning guide


andygrove:
cc @yahoNanJing @thinkharderdev @avantgardnerio

avantgardnerio (approved):
Seems like a good idea...

thinkharderdev (approved):

PR #222, opened 2022-09-15
MINOR: Increase default shuffle.partitions to 16

andygrove:

N/A

 # Rationale for this change

Get better performance with default settings

### DIFF
--- a/ballista/rust/core/src/config.rs
+++ b/ballista/rust/core/src/config.rs
@@ -157,7 +157,7 @@ impl BallistaConfig {
         let entries = vec![
             ConfigEntry::new(BALLISTA_DEFAULT_SHUFFLE_PARTITIONS.to_string(),
                              "Sets the default number of partitions to create when repartitioning query stages".to_string(),
-                             DataType::UInt16, Some("2".to_string())),
+                             DataType::UInt16, Some("16".to_string())),
             ConfigEntry::new(BALLISTA_DEFAULT_BATCH_SIZE.to_string(),
                              "Sets the default batch size".to_string(),
                              DataType::UInt16, Some("8192".to_string())),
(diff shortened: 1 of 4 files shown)


avantgardnerio (approved):
IMO 2 is clearly not a good choice for multi-core machines in 2022. 16 sounds much more practice in an age where the lowest core count will likely be 8, 16 with hyperthreading.

Dandandan:
Might be good to provide an automated way of creating the number of partitions based on a heuristic on startup if the configuration is not set like DataFusion does?
E.g. `SUM(core_count)`.

andygrove:
> Might be good to provide an automated way of creating the number of partitions based on a heuristic on startup if the configuration is not set like DataFusion does? E.g. `SUM(core_count)`.

This is challenging because it is the scheduler process that needs to use this setting, and it has no way today of knowing how many cores are in the cluster AFAIK. The scheduler may start up before any executors as well.

Eventually, I would expect us to dynamically re-partition and coalesce partitions in a similar way to Spark AQE, based on the size and number of partitions coming out of each query stage.

andygrove:
> IMO 2 is clearly not a good choice for multi-core machines in 2022. 16 sounds much more practice in an age where the lowest core count will likely be 8, 16 with hyperthreading.

Thanks for the sanity check on this arbitrary number that I picked. For reference, Spark's default is 200. I didn't want to go that far, at least for now.

avantgardnerio:
I agree with this comment:

```
message ExecutorResource {
  // TODO add more resources
  oneof resource {
    uint32 task_slots = 1;
  }
}
```

Maybe @yahoNanJing has some input? Do you have more advanced resource reporting on your fork?

yahoNanJing:
Hi @avantgardnerio, actually at our side, we set the partition number depending on different kinds of tables at the client side. In the future, we can adjust the number based on some stats. However, currently it's better to enlarge it from 2 to 16

PR #232, opened 2022-09-16
Fix integration tests

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/231

 # Rationale for this change

### DIFF
--- a/dev/docker/ballista.dockerfile
+++ b/dev/docker/ballista.dockerfile
@@ -44,6 +44,18 @@ RUN cargo install cargo-chef --version 0.1.34
 #RUN cargo chef cook $RELEASE_FLAG --recipe-path recipe.json
 
 FROM base as builder
+
+RUN mkdir /tmp/protoc
+WORKDIR /tmp/protoc
+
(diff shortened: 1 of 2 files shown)


dev/docker/ballista.dockerfile
```diff
@@ -44,6 +44,18 @@ RUN cargo install cargo-chef --version 0.1.34
 #RUN cargo chef cook $RELEASE_FLAG --recipe-path recipe.json
 
 FROM base as builder
+
+RUN mkdir /tmp/protoc
+WORKDIR /tmp/protoc
+
+RUN export PROTO_ZIP="protoc-21.4-linux-x86_64.zip" && \
+  curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v21.4/$PROTO_ZIP && \
```
avantgardnerio, line 52: It might be good to interpolate `21.4` into it's own variable, but better to fix these than not fix!

avantgardnerio (approved):

andygrove:
One day we'll get these tests running in CI per issue https://github.com/apache/arrow-ballista/issues/24

PR #233, opened 2022-09-16
Fix some issues found when testing rc1

andygrove:

Closes #.

 # Rationale for this change


avantgardnerio (approved):
Agree `arrow` only has two `r`s :+1:

PR #238, opened 2022-09-17
Add scheduler UI to ballista-scheduler Docker image

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/11

 # Rationale for this change

With this PR, I can connect to the scheduler web UI. The only functionality so far is being able to see a list of executors but with this in place we can iterate on adding new functionality such as viewing jobs, stages, and metrics.

- Create separate images for scheduler vs executor
- Prefix image names with `apache/arrow-`
- Include scheduler UI in scheduler image
- Use `ENTRYPOINT` instead of `CMD` inside scheduler and executor images

No

### DIFF
--- a/dev/docker/ballista-scheduler.dockerfile
+++ b/dev/docker/ballista-scheduler.dockerfile
@@ -26,8 +26,9 @@ RUN yarn
 RUN yarn build
 
 FROM apache/arrow-ballista:$VERSION
-COPY --from=ui-build /app/build /usr/share/nginx/html
 RUN apt -y install nginx
+RUN rm -rf /var/www/html/*
+COPY --from=ui-build /app/build /var/www/html
 
 ENV RUST_LOG=info
 ENV RUST_BACKTRACE=full
--- a/dev/docker/ballista-scheduler.dockerfile
+++ b/dev/docker/ballista-scheduler.dockerfile
@@ -1,0 +1,18 @@
+ARG VERSION=0.8.0
+FROM ballista:$VERSION
+
+# Scheduler UI
+RUN apt -y install nodejs npm nginx
+RUN npm install --global yarn
+RUN npm install --save node-releases
+RUN mkdir /tmp/ballista-ui
+WORKDIR /tmp/ballista-ui
+
+COPY ballista/ui/scheduler ./
+RUN yarn
+RUN yarn build
+
+ENV RUST_LOG=info
+ENV RUST_BACKTRACE=full
+
+CMD ["/scheduler"]
--- a/dev/docker/ballista-scheduler.dockerfile
+++ b/dev/docker/ballista-scheduler.dockerfile
@@ -1,5 +1,22 @@
-ARG VERSION=0.8.0
-FROM ballista:$VERSION
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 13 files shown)


dev/docker/ballista-scheduler.dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /tmp/ballista-ui
+
+COPY ballista/ui/scheduler ./
+RUN yarn
+
+# TODO this fails with:
+# Error: Cannot find module '../../data/browsers'
+#RUN yarn build
+
+# TODO start nginx in background to serve the UI
```
andygrove, line 35: I need help with this part as I have zero experience with node or yarn :weary:

andygrove: cc @msathis

andygrove:
@yahoNanJing @thinkharderdev @avantgardnerio This PR is now ready for review

avantgardnerio (approved):
LGTM

thinkharderdev (approved):
Nice!

avantgardnerio:
Pictures or it didn't happen :)

yahoNanJing (commented):
LGTM

yahoNanJing:
Maybe better to add an example command to leverage these docker files. For example, like follows:
{code}
./dev/build-ballista-docker.sh

docker run -p 10.249.75.38:80:80 -p 10.249.75.38:50050:50050 -it apache/arrow-ballista-scheduler:0.8.0 --external-host 10.249.75.38 -s push-staged

docker run -p 10.249.75.38:50051:50051 -p 10.249.75.38:50052:50052 -it apache/arrow-ballista-executor:0.8.0 --external-host 10.249.75.38 --scheduler-host 10.249.75.38 -s push-staged
{code}

And it seems we need to fix something in file *executor_server.rs* to change the following part 
{code}
        let addr = format!(
            "{}:{}",
            executor_meta
                .optional_host
                .map(|h| match h {
                    OptionalHost::Host(host) => host,
                })
                .unwrap_or_else(|| String::from("0.0.0.0")),
            executor_meta.grpc_port
        );
{code}
to
{code}
let addr = format!(
            "{}:{}",
            "0.0.0.0",
            executor_meta.grpc_port
        );
{code}

andygrove:
> Pictures or it didn't happen :)

ballista-ui

The UI says "scheduler hasn't started" now, but I saw this working previously. Once this PR is merged, I will start learning more about the UI code and see if I can start adding the new functionality.

andygrove:
> Maybe better to add an example command to leverage these docker files.

I have now updated the relevant sections in the user guide.

avantgardnerio:
> add an example command

I personally favor `docker-compose.yml` because it can build the images and store all the runtime information like ports in a formal way.

andygrove:
> > add an example command
> 
> I personally favor `docker-compose.yml` because it can build the images and store all the runtime information like ports in a formal way.

Here is the user guide documentation for using docker-compose:

https://github.com/apache/arrow-ballista/blob/master/docs/source/user-guide/deployment/docker-compose.md

avantgardnerio:
> documentation for using docker-compose

I don't understand `./dev/build-ballista-docker.sh`? Why not simply 
```
docker-compose up --build
```
?

andygrove:
> > documentation for using docker-compose
> 
> I don't understand `./dev/build-ballista-docker.sh`? Why not simply
> 
> ```
> docker-compose up --build
> ```
> 
> ?

I'm not familiar with that feature. Do you want to PR to add that? We will still need a way to create the images for other uses, like k8s.

andygrove:
> Also, why is out `docker-compose.yml` inside a `readme.md` vs just checked in?

This is for the user guide that will soon be published to the Arrow site. It is common practice for user guides to have sample yaml along with an explanation (see Spark's docs for example).

Ideally, we would have the yaml checked into the repo and tested and then generate that into the docs. I would like to get to that point eventually.

We do have one tested docker-compose.yaml checked into the `benchmarks` directory, and this is tested by the integration test script.

avantgardnerio:
I was just in the process of making PRs out of all our changes now that DataFusion 12 was released. This is the PR: https://github.com/apache/arrow-ballista/pull/249

avantgardnerio:
We also have a helm chart I'd like to PR in as well to handle kubernetes.

PR #245, opened 2022-09-19
Use bind host rather than the external host for starting a local executor service

### LINKED ISSUE #244: Use bind host rather than the external host for starting a local executor service
**Describe the bug**

When using the external host for starting a local executor service, it will fail if we run the docker container to start an executor.

**To Reproduce**

**Expected behavior**

We should use the bind host rather than the external host.

**Additional context**

yahoNanJing:

Closes #244.

 # Rationale for this change


andygrove (approved):
LGTM. Thanks @yahoNanJing

PR #249, opened 2022-09-19
Add a docker-compose file and dockerfiles

### LINKED ISSUE #248: Add docker build files

When I clone this project, I would expect to run

```
find . -name *.Dockerfile
```

And [see the appropriate](https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles) Dockerfiles. Currently if I try this, I see no results.

Add Dockerfiles for building the scheduler and executor, add a docker-compose file so that users can build them with a single command. This file also serves as documentation on the necessary ports and things to expose in kubernetes or helm.

- Not having Dockerfiles
- Putting the same info in readmes and having the users transcribe it manually

avantgardnerio:

Closes #248 .

 # Rationale for this change

Described in the issue.

- Add dockerfiles
- Add a docker-compose file

They should find dockerfiles and be happy.


avantgardnerio:
@andygrove for your review queue

.dockerignore
```diff
@@ -5,7 +5,9 @@ dev/dist
 dev/release
 python
 **/docs
-**/target
+target/
```
avantgardnerio, line 8: The proper way to declare a folder

.dockerignore
```diff
@@ -5,7 +5,9 @@ dev/dist
...
 **/CHANGELOG.md
 **/tests
 **/data
+!target/release/ballista-scheduler
+!target/release/ballista-executor
```
avantgardnerio, line 13: Can't exclude the executables if we want them to go into the docker image.

.gitignore
```diff
@@ -96,3 +96,6 @@ venv/*
 
 # apache release artifacts
 dev/dist
+
+# logs
+logs/
```
avantgardnerio, line 101: Running Ballista in any way seems to put logs here that aren't gitignored, which is a pain for developers to not accidentally get them into PRs.

scheduler.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+FROM ubuntu:22.04
+
+COPY ./target/release/ballista-scheduler .
+
+EXPOSE 50050
+
+CMD ./ballista-scheduler
```
andygrove, line 24: What is the goal of these dockerfiles versus the existing ones in `dev/docker`? These are for faster dev testing as opposed to building releases?

We need to start the web ui as well, which we have with the existing dockerfiles under `dev/docker`

avantgardnerio: > What is the goal of these dockerfiles

Oh, as stated in the issue, it wasn't discoverable due to 

```find . -name *.Dockerfile```

Returning no results. I see now that it returns `./dev/docker/ballista-arm64.Dockerfile` because it is capitalised according to [the standard](https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles).

I'll amend this PR to use those instead, and capitalize them as most devops engineers would expect.

**Comment by avantgardnerio** [comment]
For future reference, the command was 
```
find . -name "*.dockerfile" | sed -rn 's/(.*)\.dockerfile/\1.dockerfile \1.Dockerfile/p' | xargs -r -I {} bash -c "git mv {}"
```

File: ballista/rust/client/src/context.rs [hunk 5]
```diff
@@ -274,7 +274,14 @@ impl BallistaContext {
         path: &str,
         options: CsvReadOptions<'_>,
     ) -> Result<()> {
-        match self.read_csv(path, options).await?.to_logical_plan()? {
+        let plan = self
+            .read_csv(path, options)
+            .await
+            .map_err(|e| {
+                DataFusionError::Context(format!("Can't read CSV: {}", path), Box::new(e))
```
avantgardnerio, line 281: Say file location when we can't load it

ballista/ui/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
avantgardnerio, line 1: Best practice to check this in, just like cargo.lock

benchmarks/tpch-gen.sh
```diff
@@ -21,14 +21,13 @@
...
-docker build -t ballista-tpchgen:$BALLISTA_VERSION -f tpchgen.dockerfile .
 
 # Generate data into the ./data directory if it does not already exist
 FILE=./data/supplier.tbl
 if test -f "$FILE"; then
     echo "$FILE exists."
 else
   mkdir data 2>/dev/null
-  docker run -v `pwd`/data:/data -it --rm ballista-tpchgen:$BALLISTA_VERSION
+  docker run -v `pwd`/data:/data -it --rm ghcr.io/databloom-ai/tpch-docker:main -vf -s 1
```
avantgardnerio, line 31: Use off-the-shelf data generator to go faster. No need to re-invent this wheel.

dev/build-ballista-docker-arm64.sh
```diff
@@ -19,7 +19,7 @@
 
 if [ -z "${DOCKER_REPO}" ]; then
   echo "DOCKER_REPO env var must be set"
-  exit -1
+  exit 1
```
avantgardnerio, line 22: Negative error codes aren't actually valid?

dev/build-ballista-docker.sh
```diff
@@ -19,9 +19,13 @@
 
 set -e
 
+docker build -t ballista-builder --build-arg EXT_UID="$(id -u)" -f dev/docker/ballista-builder.Dockerfile .
```
avantgardnerio, line 22: Make a build container so that WSL & mac users are happy

dev/build-ballista-docker.sh
```diff
@@ -19,9 +19,13 @@
...
+
+docker run -v $(pwd):/home/builder/workspace ballista-builder
```
avantgardnerio, line 24: Build in the build container, but don't use ephemeral storage and wipe everything so we have faster build & test cycles, but CI is still pure.

dev/build-ballista-docker.sh
```diff
@@ -19,9 +19,13 @@
...
+
+docker-compose build
```
avantgardnerio, line 26: docker-compose is our new build script

dev/build-ballista-docker.sh
```diff
@@ -19,9 +19,13 @@
...
+
 . ./dev/build-set-env.sh
-docker build -t apache/arrow-ballista-base:$BALLISTA_VERSION -f dev/docker/ballista-base.dockerfile .
-docker build --build-arg VERSION=$BALLISTA_VERSION -t apache/arrow-ballista:$BALLISTA_VERSION -f dev/docker/ballista.dockerfile .
-docker build --build-arg VERSION=$BALLISTA_VERSION -t apache/arrow-ballista-executor:$BALLISTA_VERSION -f dev/docker/ballista-executor.dockerfile .
-docker build --build-arg VERSION=$BALLISTA_VERSION -t apache/arrow-ballista-scheduler:$BALLISTA_VERSION -f dev/docker/ballista-scheduler.dockerfile .
-docker build --build-arg VERSION=$BALLISTA_VERSION -t apache/arrow-ballista-benchmarks:$BALLISTA_VERSION -f dev/docker/ballista-benchmarks.dockerfile .
\ No newline at end of file
+docker tag apache/arrow-ballista-executor "apache/arrow-ballista-executor:$BALLISTA_VERSION"
```
avantgardnerio, line 29: We only need the BALLISTA_VERSION when tagging, not buiding.

dev/build-set-env.sh
```diff
@@ -17,4 +17,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-export BALLISTA_VERSION=$(awk -F'[ ="]+' '$1 == "version" { print $2 }' ballista/rust/core/Cargo.toml)
+cd ballista/rust/core/
+export BALLISTA_VERSION=$(cargo pkgid | cut '-d@' -f2)
```
avantgardnerio, line 21: The simplest version of this (that is correct) I could find.

dev/docker/ballista-benchmarks.Dockerfile
```diff
@@ -15,14 +15,21 @@
 # specific language governing permissions and limitations
 # under the License.
 
-ARG VERSION
-FROM apache/arrow-ballista:$VERSION
+FROM ubuntu:22.04
```
avantgardnerio, line 18: Use an OS that developers can install themselves for dev/prod parity.

dev/docker/ballista-benchmarks.Dockerfile
```diff
@@ -15,14 +15,21 @@
...
-COPY benchmarks/queries/ /queries/
+ARG RELEASE_FLAG=release
 
+ENV RELEASE_FLAG=${RELEASE_FLAG}
 ENV RUST_LOG=info
 ENV RUST_BACKTRACE=full
 
-CMD ["/run.sh"]
\ No newline at end of file
+COPY target/$RELEASE_FLAG/ballista-scheduler /root/ballista-scheduler
```
avantgardnerio, line 26: Copy in binaries to separate code build from docker build.

dev/docker/ballista-builder.Dockerfile
```diff
@@ -15,18 +15,34 @@
...
+ARG EXT_UID
+
+ENV RUST_LOG=info
+ENV RUST_BACKTRACE=full
+ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y git build-essential
+    apt-get -y install libssl-dev openssl zlib1g zlib1g-dev libpq-dev cmake protobuf-compiler netcat curl unzip \
+    nodejs npm && \
```
avantgardnerio, line 28: We probably want a newer node at some point, but I ran into a lot of trouble with every non-apt option.

andygrove:
You can exclude `yarn.lock` from the RAT checks by adding it to `dev/release/rat_exclude_files.txt`

andygrove:
I just tried this out on Ubuntu 20.04 after running `docker system prune -a` to remove any previous images and I ran into a couple issues that I was able to fix with the following path:

```
diff --git a/dev/build-ballista-docker.sh b/dev/build-ballista-docker.sh
index d8af64b5..5efbb152 100755
--- a/dev/build-ballista-docker.sh
+++ b/dev/build-ballista-docker.sh
@@ -23,9 +23,9 @@ docker build -t ballista-builder --build-arg EXT_UID="$(id -u)" -f dev/docker/ba
 
 docker run -v $(pwd):/home/builder/workspace ballista-builder
 
-docker-compose build
+docker compose build
 
 . ./dev/build-set-env.sh
-docker tag apache/arrow-ballista-executor "apache/arrow-ballista-executor:$BALLISTA_VERSION"
-docker tag apache/arrow-ballista-scheduler "apache/arrow-ballista-scheduler:$BALLISTA_VERSION"
-docker tag apache/arrow-ballista-benchmarks "apache/arrow-ballista-benchmarks:$BALLISTA_VERSION"
+docker tag ballista-executor "apache/arrow-ballista-executor:$BALLISTA_VERSION"
+docker tag ballista-scheduler "apache/arrow-ballista-scheduler:$BALLISTA_VERSION"
+docker tag ballista-benchmarks "apache/arrow-ballista-benchmarks:$BALLISTA_VERSION"
diff --git a/docker-compose.yml b/docker-compose.yml
index cbfb400c..1865db69 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,7 +48,7 @@ services:
     depends_on:
       - ballista-scheduler
   ballista-client:
-    image: ballista-client
+    image: ballista-benchmarks
     build:
       dockerfile: dev/docker/ballista-benchmarks.Dockerfile
       context: .
```

This got me further - the images were built and tagged - but the integration test failed with:

```
Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/run.sh": stat /run.sh: no such file or directory: unknown

```

avantgardnerio:
> +docker compose build

I don't understand how this works for you. I tested in docker in Ubuntu 20.04: 

```
root@6b1248d5c2bd:/# docker --help | grep compose

root@6b1248d5c2bd:/# cat /etc/*-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
...

root@6b1248d5c2bd:/# docker compose

docker: 'compose' is not a docker command.
See 'docker --help'

root@6b1248d5c2bd:/# docker-compose
Define and run multi-container applications with Docker.

Usage:
```

The command is `docker-compose` unless I'm going crazy.

avantgardnerio:
```
bgardner@bg-tensorbook:~$ docker --version
Docker version 20.10.12, build 20.10.12-0ubuntu4
```

and in docker ubuntu 20.04:

```
root@46a85dc6e8a1:/# docker --version
Docker version 20.10.12, build 20.10.12-0ubuntu2~20.04.1
```

avantgardnerio:
@andygrove I take it back. I think this would be good to merge as soon as you feel the issues you listed have been addressed. Unfortunately I do not know how to reproduce, therefor fix, the one you are seeing with `docker-compose` not working.

avantgardnerio:
Everything works now on a system with `docker-compose` including the `integration-tests.sh`

.dockerignore
andygrove: Is this change still needed now that the executables are build in Docker?

andygrove: Never mind ... after reviewing  the rest of the PR I understand why we have this

andygrove (approved):
Thanks @avantgardnerio! This looks great. I tested locally after removing all Ballista images again and it worked perfectly.

It would be good if @yahoNanJing and @thinkharderdev could also test this out in their environments.

andygrove:
@yahoNanJing @thinkharderdev ok if I merge this, or do you want to test first?

andygrove:
Thanks for the review @thinkharderdev. I will go ahead and merge this now.

PR #260, opened 2022-09-21
Introduce the datafusion-objectstore-hdfs in datafusion-contrib as an object store feature

### LINKED ISSUE #259: Introduce the datafusion-objectstore-hdfs in datafusion-contrib as an object store feature

Introduce the datafusion-objectstore-hdfs in datafusion-contrib as an object store feature based on #258 

**Additional context**

yahoNanJing:

Closes #259, #6.

 # Rationale for this change

### DIFF
--- a/ballista/rust/core/src/utils.rs
+++ b/ballista/rust/core/src/utils.rs
@@ -43,6 +43,8 @@ use datafusion::physical_plan::metrics::MetricsSet;
 use datafusion::physical_plan::projection::ProjectionExec;
 use datafusion::physical_plan::sorts::sort::SortExec;
 use datafusion::physical_plan::{metrics, ExecutionPlan, RecordBatchStream};
+#[cfg(feature = "hdfs")]
+use datafusion_objectstore_hdfs::object_store::hdfs::HadoopFileSystem;
 use datafusion_proto::logical_plan::{
     AsLogicalPlan, DefaultLogicalExtensionCodec, LogicalExtensionCodec,
 };
@@ -82,7 +84,16 @@ pub struct FeatureBasedObjectStoreProvider;
 impl ObjectStoreProvider for FeatureBasedObjectStoreProvider {
     /// Detector a suitable object store based on its url if possible
     /// Return the key and object store
-    fn get_by_url(&self, _url: &Url) -> Option<Arc<dyn ObjectStore>> {
+    #[allow(unused_variables)]
+    fn get_by_url(&self, url: &Url) -> Option<Arc<dyn ObjectStore>> {
+        #[cfg(feature = "hdfs")]
+        {
+            let store = HadoopFileSystem::new(url.as_str());
+            if let Some(store) = store {
+                return Some(Arc::new(store));
+            }
+        }
+
         None
     }
 }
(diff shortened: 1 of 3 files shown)


yahoNanJing:
Will be ready after #258 merged.

ballista/rust/core/src/utils.rs
```diff
@@ -54,9 +58,47 @@ use std::time::Duration;
...
+
+/// An object store detector based on which features are enable for different kinds of object stores
+pub struct FeatureBasedObjectStoreProvider;
+
+impl ObjectStoreProvider for FeatureBasedObjectStoreProvider {
+    /// Detector a suitable object store based on its url if possible
+    /// Return the key and object store
+    #[allow(unused_variables)]
+    fn get_by_url(&self, url: &Url) -> Option<Arc<dyn ObjectStore>> {
+        #[cfg(feature = "hdfs")]
```
tustvold, line 89: Being able to enable/disable object stores based on features sounds sensible, but I would recommend supporting multiple based on the scheme. Having to recompile different binaries for different backends is unfortunate. It's possible you intend to do this anyway, but thought I would mention

yahoNanJing: For a standalone system, the binaries for the backend should be fixed before running. How would you use the capability of one crate without compiling them.

yahoNanJing: For a standalone system, we should know which object stores to be used before running it. Therefore, it's OK to make the decision at the compiling phase.

tustvold: How would you distribute binaries for such a system?

yahoNanJing: We run it in K8S with replicas.

ballista/rust/core/src/utils.rs
```diff
@@ -54,9 +58,47 @@ use std::time::Duration;
...
+        {
+            let store = HadoopFileSystem::new(url.as_str());
+            if let Some(store) = store {
+                return Some(Arc::new(store));
+            }
+        }
+
+        None
```
tustvold, line 97: Following the current construction this could return

"FeatureBasedObjectStoreProvider failed to create store for url {}"

This is already an improvement as it tells us that the provider was registered and used.

As described above, I think it would be even better if it could provide insight into why nothing was found, e.g. not compiled with the right feature, unsupported scheme, etc..

ballista/rust/core/src/utils.rs
tustvold: Let me phrase it differently, how would the ballista project distribute binaries/docker images for general consumption in arbitrary environments. I presume we want to provide this?

yahoNanJing:
We'll change the related ObjectStoreProvider interface after we upgrading to latest datafusion due to [#3595](https://github.com/apache/arrow-datafusion/pull/3595), [#3619](https://github.com/apache/arrow-datafusion/pull/3619)

ballista/rust/core/src/utils.rs
yahoNanJing: We made an agreement that it's supposed to be able to find all of the object stores from the *ObjectStoreProvider*. Otherwise, an error should be thrown. We'll change the related ObjectStoreProvider interface after we upgrading to latest datafusion.

PR #261, opened 2022-09-21
Task level retry and Stage level retry

### LINKED ISSUE #140: Failed task retry

When a task is failed on Ballista Executor, due to different failure reasons, sometimes the failed tasks need to be resubmitted to another Executor and rerun the failed tasks. Need to enhance the Ballista Scheduler to support this.
Need to enhance `FailedTask` rpc message to indicate whether the failure task should retry or not.
A good example of failed task that should retry is Executor lost. And on the other hand, user cancelled queries/tasks should not retry. A shuffle read failure should not retry the task either.

**Additional context**

mingmwang:

Closes #140.

 # Rationale for this change

**1. Task level retry,** 
a) Add task attempt num, use a job level unique task_id to represent the Task
b) Define couple of failure reasons for FailedTask

````
pub enum FailedReason {
        #[prost(message, tag="4")]
        ExecutionError(super::ExecutionError),
        #[prost(message, tag="5")]
        FetchPartitionError(super::FetchPartitionError),
        #[prost(message, tag="6")]
        IoError(super::IoError),
        #[prost(message, tag="7")]
        ExecutorLost(super::ExecutorLost),
        /// A successful task's result is lost due to executor lost
        #[prost(message, tag="8")]
        ResultLost(super::ResultLost),
        #[prost(message, tag="9")]
        TaskKilled(super::TaskKilled),
    }

````
c) Reasoning the real task failure reason when return the TaskStatus back to Scheduler
d) Based on the failure reason, scheduler decide to reschedule the task and bump task attempt number of failed task.

**2. Stage level retry and shuffle read failure handling** 
a) Add stage attempt num.
b) When there are shuffle partition fetch failures, the current running stage will be rolled back and the map stages will be resubmit.
c) handling of delayed fetch failure task updates
d) Cancel the running tasks if Scheduler decide to fail the stage/job.

And below two items are not covered in this PR and we may need to revisit the related logic in future:
1) If the plans/expressions in the Stage is not deterministic, need to revisit the resubmit logic.
2) If we have a map stage whose shuffle output can be reused by multiple reduce stages, need to revisit the stage retry logic.

### CODE REVIEW

**Comment by mingmwang** [comment]
@thinkharderdev @andygrove @yahoNanJing
Please help to review my PR.

**Comment by mingmwang** [comment]
Added 10+ UTs to cover different cases.

**Comment by yahoNanJing** [comment]
With this PR, the state machine for the stage becomes as follows:
StageStateMachine

The transitions marked in red are newly added ones to deal with error task recovery.

**Comment by yahoNanJing** [comment]
For example, one SQL stage graph snapshot is as follows:

           							Stage 4(Resolved)
                                                  ↗                                                                       ↘
Stage 1(Successful)                										Stage 5(Unresolved)
                                       ↘                                                                                   ↗
           				      Stage 2(Successful)  ->  Stage 3(Running) 

Both of Stage1 and Stage2 has output shuffle data resides on Executor1. Unluckily, the Executor1 gets lost. Then tasks for the running Stage3 will fail due to not being able to fetch shuffle data. For this kind of scenario and error, by this PR, the ExecutionGraph will be able to continue running with proper task reset.

File: ballista/rust/scheduler/src/state/execution_graph.rs [hunk 1]
```diff
@@ -296,18 +442,105 @@ impl ExecutionGraph {
...
+                            "TID {} {}/{}.{}/{}",
+                            task_status.task_id,
+                            job_id,
+                            stage_id,
+                            task_stage_attempt_num,
+                            partition_id
+                        );
+                        let mut should_ignore = true;
+                        // handle delayed failed tasks if the stage's next attempt is still in UnResolved status.
+                        if let Some(task_status::Status::Failed(failed_task)) =
```
yahoNanJing, line 471: Here, only failed tasks will be dealt with. Once a running stage converted into an unresolved, it will discard all of its tasks info. Then to update the successful task info is meaningless.

thinkharderdev:
@mingmwang I'm pretty swamped this week so won't have time to review this until this weekend.

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -202,88 +228,208 @@ impl ExecutionGraph {
...
+                                );
                             }
                         }
                     }
-                    let is_completed = running_stage.is_completed();
-                    if is_completed {
-                        events.push(StageEvent::StageCompleted(stage_id));
-                        // if this stage is completed, we want to combine the stage metrics to plan's metric set and print out the plan
+                    let is_final_successful = running_stage.is_successful()
+                        && !reset_running_stages.contains_key(&stage_id);
```
yahoNanJing, line 429: When this running stage is converted from successful stage, there may be in flight running tasks of the stages depending on this stage. When we find those tasks fails due to fetching data failure, then we also need to reset this running stage. Therefore, here, the reset_running_stages check is necessary.

yahoNanJing:
Later, maybe we can make this error task recovery feature configurable.

yahoNanJing:
The overall design of this PR is good for me and it will be really useful to deal with executor lost issue and executor bad disk issue. @thinkharderdev, @andygrove, @avantgardnerio, could you help review this PR?

andygrove:
Thanks, @mingmwang. I plan on starting to review this tomorrow.

ballista/rust/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -252,48 +261,71 @@ fn send_fetch_partitions(
...
-                ),
-                other => other,
-            })?;
-
-    ballista_client
-        .fetch_partition(&metadata.id, partition_id, &location.path, host, port)
-        .await
+/// Partition reader Trait, different partition reader can have
+#[async_trait]
+trait PartitionReader: Send + Sync + Clone {
```
yahoNanJing, line 266: Trait is much more extensible than condition compiling 👍

ballista/rust/core/proto/ballista.proto
```diff
@@ -424,54 +424,87 @@ message ExecutionGraph {
   uint64 output_partitions = 5;
   repeated PartitionLocation output_locations = 6;
   string scheduler_id = 7;
+  uint32 tid_gen = 8;
```
andygrove, line 427: nit: `tid_gen` isn't a descriptive name. I am guessing this is short for `task_id_gen`?

ballista/rust/core/proto/ballista.proto
```diff
@@ -521,12 +554,12 @@ message FetchPartition {
...
 }
 
-// Mapping from partition id to executor id
 message PartitionLocation {
-  PartitionId partition_id = 1;
-  ExecutorMetadata executor_meta = 2;
-  PartitionStats partition_stats = 3;
-  string path = 4;
+  uint32 map_partition_id = 1;
+  PartitionId partition_id = 2;
```
andygrove, line 559: Could you add comments explaining these different partition ids?

andygrove:
@mingmwang I have taken a first pass through this PR and I think it looks good. I am going to spend some time testing out the PR locally. It would be good to wait for Dan to review as well before we merge this.

ballista/rust/core/proto/ballista.proto
```diff
@@ -424,54 +424,87 @@ message ExecutionGraph {
...
-  PhysicalHashRepartition output_partitioning = 3;
-  repeated uint32 output_links = 4;
-  repeated  GraphStageInput inputs = 5;
-  bytes plan = 6;
-  repeated TaskStatus task_statuses = 7;
-  repeated OperatorMetricsSet stage_metrics = 8;
+message ResolvedStage {
+  uint32 stage_id = 1;
+  uint32 stage_attempt_num = 2;
+  uint32 partitions = 3;
```
andygrove, line 458: There are breaking changes here because this PR is changing the meaning of fields. For example field 3 was `PhysicalHashRepartition` and now it is `uint32`. We should assign new numbers to new fields and not modify numbers for existing fields.

ballista/rust/core/proto/ballista.proto
```diff
@@ -424,54 +424,87 @@ message ExecutionGraph {
...
-  uint32 partitions = 2;
-  PhysicalHashRepartition output_partitioning = 3;
-  repeated uint32 output_links = 4;
-  bytes plan = 5;
-  repeated TaskStatus task_statuses = 6;
-  repeated OperatorMetricsSet stage_metrics = 7;
-  string error_message = 8;
+  uint32 stage_id = 1;
+  uint32 stage_attempt_num = 2;
+  uint32 partitions = 3;
```
andygrove, line 481: Same issue here with changing tag numbers.

ballista/rust/core/proto/ballista.proto
andygrove: These changes meant that I could not use my benchmark client without recompiling it against this branch, which should not be necessary.

andygrove:
I tested this locally, and it worked really well!

I ran one scheduler, and two executors and I could kill one executor and still see a query complete successfully. This is not the case in the master branch.

I am happy to approve this once feedback has been addressed.

ballista/rust/core/proto/ballista.proto
mingmwang: Sure,  will change to `task_id_gen`

ballista/rust/core/proto/ballista.proto
mingmwang: > These changes meant that I could not use my benchmark client without recompiling it against this branch, which should not be necessary.

Well, I thought those are private structures and only used internally by the TaskManager.

mingmwang:
> I tested this locally, and it worked really well!
> 
> I ran one scheduler, and two executors and I could kill one executor and still see a query complete successfully. This is not the case in the master branch.
> 
> I am happy to approve this once feedback has been addressed.

Thank you. We will also start the chaos-monkey testing in next month to verify the recent changes.

ballista/rust/core/proto/ballista.proto
mingmwang: Sure, will add more comments in the code. Here the new added `map_partition_id` is the partition_id of the map stage who produce this shuffle data.  But the original `PartitionId partition_id` has different meanings in different places.
Sometimes it stands for a task_id, and here it stands for a shuffle partition id(composition of map_stage_id + reduce partition id + job_id), a mixed up of map and reduce infos together.
So if we do not consider the backward compatibility, I would suggest to unnest this and make the PartitionLocation a plan struct like blow:

````
message PartitionLocation {
  string job_id = 1;
  uint32 map_stage_id = 2;
  uint32 map_partition_id = 3;
  uint32 partition_id = 4;
  ExecutorMetadata executor_meta = 5;
  PartitionStats partition_stats = 6;
  string path = 7;
}
````

File: ballista/rust/core/proto/ballista.proto (hunk 6 above)
Comment (mingmwang) [reply ballista/rust/core/proto/ballista.proto:458]: I will change the tag num back.  But still there are backward incompatible changes here,  The `stage_id` was changed from `uint64 stage_id` to `uint32 stage_id` to have the same type with other stage_ids in this proto file.

Comment (andygrove) [reply ballista/rust/core/proto/ballista.proto]: Yes, after reviewing again, I don't think we can avoid breaking changes given the extent of these changes.

File: ballista/rust/core/proto/ballista.proto (hunk 5 above)
Comment (andygrove) [reply ballista/rust/core/proto/ballista.proto]: It is late here for me, but I think that makes sense. I will review it again tomorrow.

**Review by andygrove — APPROVED** [review]
Thanks @mingmwang. This is a very nice improvement in stability.

**Review by thinkharderdev — APPROVED** [review]
Had a few questions but this looks really great! Thanks for your work on this @mingmwang!

File: ballista/rust/core/src/error.rs [hunk 8]
```diff
@@ -182,13 +200,78 @@ impl Display for BallistaError {
...
+                    count_to_failures: true,
+                    failed_reason: Some(FailedReason::IoError(IoError {})),
+                }
+            }
+            BallistaError::DataFusionError(DataFusionError::IoError(io)) => {
+                FailedTask {
+                    error: format!("Task failed due to DataFusion IO error: {:?}", io),
+                    // IO error is considered to be temporary and retryable
+                    retryable: true,
+                    count_to_failures: true,
```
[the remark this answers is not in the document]
mingmwang: Yes, have both `retyrable` `and count_to_failures` is to support the case that we might have some specific error that we want it to retry forever until it is successful.

ballista/rust/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -250,48 +261,72 @@ fn send_fetch_partitions(
...
+
+        ballista_client
+            .fetch_partition(&metadata.id, partition_id, &location.path, host, port)
+            .await
+    }
 }
 
+#[allow(dead_code)]
+// TODO
+struct LocalPartitionReader {}
```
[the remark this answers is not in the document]
mingmwang: > Nice! I assume where this is going is making the executors smart enough to read partitions on the same physical machine directly from disk.

Yes, exactly.

ballista/rust/scheduler/src/state/execution_graph.rs
```diff
@@ -110,6 +114,20 @@ pub struct ExecutionGraph {
     output_partitions: usize,
     /// Locations of this `ExecutionGraph` final output locations
     output_locations: Vec<PartitionLocation>,
+    /// Task ID generator, generate unique TID in the execution graph
+    task_id_gen: usize,
+    /// Failed stage attempts, record the failed stage attempts to limit the retry times.
+    /// Map from Stage ID -> Set<Stage_ATTPMPT_NUM>
+    failed_stage_attempts: HashMap<usize, HashSet<usize>>,
```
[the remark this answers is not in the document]
mingmwang: > I don't get why we are saving a `HashSet` of attempts. Shouldn't it just be `usize`?

The purpose of using HashSet is to record what are the exact distinct failed attempts.

ballista/rust/core/proto/ballista.proto
```diff
@@ -664,15 +698,48 @@ message RunningTask {
...
+}
+
+message IOError {
+}
+
+message ExecutorLost {
+}
+
+message ResultLost {
+}
```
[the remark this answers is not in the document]
mingmwang: > Not sure I understand the difference between these two errors

Yes, good question. In the current code base, both the two errors are not used directly by the executor tasks.
They are used by the Scheduler. When we see a 'FetchPartitionError' task update from the reduce task, the related map task's status is changed to  'ResultLost'.  Of cause most of the time `ResultLost` should be caused by `ExecutorLost`.

yahoNanJing:
Thanks @mingmwang for this huge step of error recovering. Thanks @andygrove and @thinkharderdev for reviewing this PR. Since we all approved this PR, I'll merge it first.

yahoNanJing (approved):

PR #269, opened 2022-09-23
Store sessions so users can register tables and query them through flight

### LINKED ISSUE #112: Use ArrowFlight bearer token auth to create a session key for FlightSql clients

avantgardnerio:

Closes #112.

 # Rationale for this change

Allow users to do more than `select 1;` thought Flight SQL.

- Handshake / auth endpoint to create a session UUID
- Map of contexts to session UUIDs in scheduler memory (TODO: etcd)
- Proxying of flights from the scheduler to the executors (because JDBC driver in arrow repo ignores alt EPs)
- Add host & port information to FlightInfo so they can be canonically found

Users can now register tables and query them through Flight SQL (and JDBC).

### DIFF
--- a/ballista/rust/core/proto/ballista.proto
+++ b/ballista/rust/core/proto/ballista.proto
@@ -517,6 +517,8 @@ message FetchPartition {
   uint32 stage_id = 2;
   uint32 partition_id = 3;
   string path = 4;
+  string host = 5;
+  uint32 port = 6;
 }
 
 // Mapping from partition id to executor id
--- a/ballista/rust/executor/src/flight_service.rs
+++ b/ballista/rust/executor/src/flight_service.rs
@@ -17,6 +17,7 @@
 
 //! Implementation of the Apache Arrow Flight protocol that wraps an executor.
 
+use std::convert::TryFrom;
 use std::fs::File;
 use std::pin::Pin;
 
@@ -35,7 +36,7 @@ use datafusion::arrow::{
     record_batch::RecordBatch,
 };
 use futures::{Stream, StreamExt};
-use log::{debug, warn};
+use log::{debug, info, warn};
 use std::io::{Read, Seek};
 use tokio::sync::mpsc::channel;
 use tokio::{
@@ -43,6 +44,7 @@ use tokio::{
     task,
 };
 use tokio_stream::wrappers::ReceiverStream;
+use tonic::metadata::MetadataValue;
 use tonic::{Request, Response, Status, Streaming};
 
 type FlightDataSender = Sender<Result<FlightData, Status>>;
@@ -135,7 +137,23 @@ impl FlightService for BallistaFlightService {
         &self,
         _request: Request<Streaming<HandshakeRequest>>,
     ) -> Result<Response<Self::HandshakeStream>, Status> {
-        Err(Status::unimplemented("handshake"))
+        let token = uuid::Uuid::new_v4();
+        info!("do_handshake token={}", token);
+
+        let result = HandshakeResponse {
+            protocol_version: 0,
+            payload: token.as_bytes().to_vec(),
+        };
+        let result = Ok(result);
+        let output = futures::stream::iter(vec![result]);
+        let str = format!("Bearer {}", token);
+        let mut resp: Response<
+            Pin<Box<dyn Stream<Item = Result<_, Status>> + Sync + Send>>,
+        > = Response::new(Box::pin(output));
+        let md = MetadataValue::try_from(str)
+            .map_err(|_| Status::invalid_argument("authorization not parsable"))?;
+        resp.metadata_mut().insert("authorization", md);
+        Ok(resp)
     }
 
     async fn list_flights(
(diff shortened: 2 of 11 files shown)


avantgardnerio:
@andygrove this is the big one that we've been sitting on until Arrow upgraded to 22.

ballista/rust/executor/src/flight_service.rs
```diff
@@ -135,7 +137,23 @@ impl FlightService for BallistaFlightService {
         &self,
         _request: Request<Streaming<HandshakeRequest>>,
     ) -> Result<Response<Self::HandshakeStream>, Status> {
-        Err(Status::unimplemented("handshake"))
+        let token = uuid::Uuid::new_v4();
```
avantgardnerio, line 140: When working properly (i.e. supporting alternate endpoints) the Flight SQL JDBC driver will try to talk to the executor with the same credentials as it used with the scheduler. Let's just accept any credentials for now, so that when the issue is fixed in the `arrow` repo this will still work.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -52,22 +60,26 @@ use datafusion::prelude::SessionContext;
...
 use prost::Message;
 use tokio::time::sleep;
+use tonic::codegen::futures_core::Stream;
+use tonic::metadata::MetadataValue;
 use uuid::Uuid;
 
 pub struct FlightSqlServiceImpl {
     server: SchedulerServer<LogicalPlanNode, PhysicalPlanNode>,
     statements: Arc<Mutex<HashMap<Uuid, LogicalPlan>>>,
+    contexts: Arc<Mutex<HashMap<Uuid, Arc<SessionContext>>>>,
```
avantgardnerio, line 70: Store a map of `SessionContext` to `UUID` so we can look up the session for the user and keep the tables registered.

TODO: move this into etcd?

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -309,33 +362,140 @@ impl FlightSqlServiceImpl {
...
+            .map_err(|_| Status::invalid_argument("authorization not parsable"))?;
+        let parts: Vec<_> = str.split(":").collect();
+        if parts.len() != 2 {
+            Err(Status::invalid_argument(format!(
+                "Invalid authorization header"
+            )))?;
+        }
+        let user = parts[0];
+        let pass = parts[1];
+        if user != "admin" || pass != "password" {
```
avantgardnerio, line 403: TODO: authentication providers? (i.e. oauth, active directory, etc)

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -309,33 +362,140 @@ impl FlightSqlServiceImpl {
...
+            Err(Status::unauthenticated("Invalid credentials!"))?
+        }
+
+        let token = self.create_ctx().await?;
+
+        let result = HandshakeResponse {
+            protocol_version: 0,
+            payload: token.as_bytes().to_vec(),
```
avantgardnerio, line 411: We don't actually care about credentials at all, we just need auth to be active in order to get this handshake request and give them back a session token.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -309,33 +362,140 @@ impl FlightSqlServiceImpl {
...
+            let connection =
+                create_grpc_client_connection(addr.clone())
+                    .await
+                    .map_err(|e| {
+                        Status::internal(format!(
+                        "Error connecting to Ballista scheduler or executor at {}: {:?}",
+                        addr, e
+                    ))
+                    })?;
+            let mut flight_client = FlightServiceClient::new(connection);
```
avantgardnerio, line 453: Proxy the flight to the correct executor instance.

ballista/rust/scheduler/src/flight_sql.rs
```diff
@@ -309,33 +362,140 @@ impl FlightSqlServiceImpl {
...
 
     async fn get_flight_info_prepared_statement(
         &self,
         handle: CommandPreparedStatementQuery,
-        _request: Request<FlightDescriptor>,
+        request: Request<FlightDescriptor>,
     ) -> Result<Response<FlightInfo>, Status> {
-        let ctx = self.create_ctx().await?;
+        debug!("get_flight_info_prepared_statement");
+        let ctx = self.get_ctx(&request)?;
```
avantgardnerio, line 492: Use the UUID token to get the cached `SessionContext`.

ballista/rust/core/proto/ballista.proto
```diff
@@ -517,6 +517,8 @@ message FetchPartition {
   uint32 stage_id = 2;
   uint32 partition_id = 3;
   string path = 4;
+  string host = 5;
+  uint32 port = 6;
```
avantgardnerio, line 521: Add host & port to flights so that when the scheduler gets a flight request, it can go find the source. Architecturally, I see this as a good canonicalization of the source of Flight data, but I'm happy to discuss here!

andygrove (approved):
Thanks @avantgardnerio! I pulled this PR locally and tested it out. I was able to register tables and query them. This is a huge leap forward.

PR #271, opened 2022-09-23
Also run yarn build to catch JavaScript errors in CI

### LINKED ISSUE #270: There are bugs in the yarn build github misses but break our internal build
**Describe the bug**

Github actions to not yet perform the `yarn build` step that is now required to make a fully functioning Ballista scheduler artifact. Due to this, there are warnings that are treated as errors in CI that break S&T internal circleci but are not caught by github actions.

**To Reproduce**

Set up your own CI that has `yarn build`.

**Expected behavior**

Both CIs work.

avantgardnerio:

Closes #270.

 # Rationale for this change

Treat the yarn build as a first-class citizen.

A CI job to build yarn.

TypeScript code should always at least compile if not work.

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -1,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


avantgardnerio:
@andygrove would love to get this running in CI so other PRs don't silently break things.

andygrove (approved):
Thanks @avantgardnerio

PR #274, opened 2022-09-24
User Guide improvements

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/272

 # Rationale for this change

Trying to improve the user guide

- Spit content into more sections (deployment, clients, reference)
- Add page on scheduler
- Add info in tuning guide on generating query plan images
- Fix a bunch of things that were wrong or referring to DataFusion instead of Ballista


andygrove:
@yahoNanJing @thinkharderdev @avantgardnerio PTAL when you can

andygrove:
I'm going to go ahead and merge this since it is only docs, and it unblocks the next set of changes. We can always revert this if there are objections later.

avantgardnerio (commented):
A little late, but LGTM.

PR #285, opened 2022-09-26
[arrow-ballista] enable delete log periodically (#280)

### LINKED ISSUE #280: [feat] Enable enable delete log periodically.
For now we are using `tokio-tracing` (also used in influx-iox, datafuse) to manage our tracing log. But there is not a feature like sl4j in java to delete our outdated log on disk. For long running application , we need manually delete them 😂, So before `tokio-tracing` support it, i think we need this feature.

**Additional context**

Ted-Jiang:

Closes #280.

add param `cleanup_log_ttl` in both executor and scheduler, means the number of hours to retain log files on each node, set zero means disable.

 # Rationale for this change

For now we are using tokio-tracing (also used in influx-iox, datafuse) to manage our tracing log. But there is not a feature like sl4j in java to delete our outdated log on disk. For long running application , we need manually delete them 😂, So before tokio-tracing support it, i think we need this feature.


Ted-Jiang:
@andygrove PTAL

andygrove:
I'm not sure that we should be building log management features into Ballista. Perhaps it would be better to deploy a separate process alongside the executor to do this? Different companies will likely have different requirements here?

Ted-Jiang:
> I'm not sure that we should be building log management features into Ballista. Perhaps it would be better to deploy a separate process alongside the executor to do this? Different companies will likely have different requirements here?

Yes, this is a temporary solution, it should be better handle by `tokio-tracing`(by now not). So i set default to `0` to disable this feature.

PR #288, opened 2022-09-28
Fix documentation example

onthebridgetonowhere:

There is no issue related to this PR

 # Rationale for this change
Outdated or wrong documentation. Given that these are the main examples linked from the main README.md, I think it's worth fixing them. 

One code snippet for the distributed SQL example is wrong (likely a copy paste error), and the other example uses the New York Taxi and Limousine Commission data, which are now only provided as Parquet files. Thus, this PR updates the example to use Parquet instead.

No

### DIFF
--- a/ballista/rust/client/README.md
+++ b/ballista/rust/client/README.md
@@ -84,14 +84,14 @@ To build a simple ballista example, add the following dependencies to your `Carg
 
 ```toml
 [dependencies]
-ballista = "0.6"
-datafusion = "7.0"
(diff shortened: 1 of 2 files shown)


andygrove (approved):
Thanks @onthebridgetonowhere!

onthebridgetonowhere:
> @onthebridgetonowhere It looks like you need to run `prettier` .. please see contributors guide for more info

Sorry about that, it should be fixed now (at least on my machine it does not complain anymore).

PR #290, opened 2022-09-28
Add support for S3 data sources

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/10

Builds on https://github.com/apache/arrow-ballista/pull/289

 # Rationale for this change

For users on AWS, Ballista is not an option unless it supports S3.

```
$ ./target/release/ballista-cli --host localhost --port 50050
Ballista CLI v0.8.0
❯ CREATE EXTERNAL TABLE test (a int, b int) STORED AS CSV LOCATION 's3://andygrove-benchmark-data/test.csv';
0 rows in set. Query took 0.000 seconds.
❯ select * from test;
+---+---+
| a | b |
+---+---+
| 1 | 2 |
+---+---+
1 row in set. Query took 0.716 seconds.
```

No


yahoNanJing (approved):
LGTM

PR #304, opened 2022-10-01
Add support for SortPreservingMergeExec; fix LIMIT bug

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/300

 # Rationale for this change

Queries with `LIMIT` may now include `SortPreservingMergeExec` and we did not support that in ballista.proto

- Add serde code for `SortPreservingMergeExec`
- Improve error when serde code finds an unsupported exec
- Fixed a bug with encoding `GlobalLimitExec`

### DIFF
--- a/ballista/rust/core/proto/ballista.proto
+++ b/ballista/rust/core/proto/ballista.proto
@@ -375,6 +375,8 @@ message LocalLimitExecNode {
 message SortExecNode {
   PhysicalPlanNode input = 1;
   repeated PhysicalExprNode expr = 2;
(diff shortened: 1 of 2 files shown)


Dandandan:
@andygrove this probably is added in plans after the `order by limit`. `SortExec` should probably be updated as well - that node has a new *fetch* parameter.

andygrove:
Thanks @Dandandan. I pushed a fix for SortExec.

Dandandan (approved):

thinkharderdev (approved):

PR #309, opened 2022-10-02
[UI] Use tabbed pane with Queries and Executors tabs

andygrove:

N/A

 # Rationale for this change

The list of executors could be very large so it does not make sense to show in the header. 

- Use tabbed interface with Jobs tab and Executors tab.

## Jobs Tab

tab0

## Executors Tab

tab

### DIFF
--- a/ballista/ui/scheduler/src/App.tsx
+++ b/ballista/ui/scheduler/src/App.tsx
@@ -84,7 +84,7 @@ const App: React.FunctionComponent<any> = () => {
                 <QueriesList queries={jobs} />
               </TabPanel>
               <TabPanel>
-                {/*<ExecutorsList executors={executors} />*/}
+                <ExecutorsList executors={executors} />
               </TabPanel>
             </TabPanels>
           </Tabs>
--- a/ballista/ui/scheduler/src/App.tsx
+++ b/ballista/ui/scheduler/src/App.tsx
@@ -16,17 +16,18 @@
 // under the License.
 
 import React, { useState, useEffect } from "react";
-import { Box, Grid, VStack } from "@chakra-ui/react";
+import {Box, Grid, Tab, TabList, TabPanel, TabPanels, Tabs, VStack} from "@chakra-ui/react";
 import { Header } from "./components/Header";
 import { Summary } from "./components/Summary";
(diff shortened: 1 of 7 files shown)


andygrove:
@KenSuenobu @onthebridgetonowhere Could you review this PR when you get a chance?

ballista/ui/scheduler/src/App.tsx
```diff
@@ -50,20 +60,43 @@ const App: React.FunctionComponent<any> = () => {
...
-  console.log(JSON.stringify(schedulerState));
-
   return (
     <Box>
       <Grid minH="100vh">
         <VStack alignItems={"flex-start"} spacing={0} width={"100%"}>
           <Header schedulerState={schedulerState} />
           <Summary schedulerState={schedulerState} />
-          <QueriesList queries={jobs} />
+          <Tabs>
```
onthebridgetonowhere, line 86: We probably want to have here: `<Tabs width={"100%"}>`

ballista/ui/scheduler/src/components/ExecutorsList.tsx
```diff
@@ -56,16 +61,12 @@ const columns: Column<any>[] = [
...
-
-export const NodesList: React.FunctionComponent<NodesListProps> = ({
-  nodes = [],
+export const ExecutorsList: React.FunctionComponent<ExecutorsListProps> = ({
+  executors = [],
 }) => {
   return (
     <Box flex={1}>
-      <DataTable maxW={960} columns={columns} data={nodes} pageSize={4} />
+      <DataTable maxW={960} columns={columns} data={executors} pageSize={4} />
```
onthebridgetonowhere, line 69: The `maxW` property is not really useful, unless we need this width size restriction. Removing it makes the table look better.

onthebridgetonowhere (changes requested):

onthebridgetonowhere:
> @KenSuenobu @onthebridgetonowhere Could you review this PR when you get a chance?

@andygrove it looks great. I have only two comments regarding look&feel, see above.

andygrove:
Thanks for the review @onthebridgetonowhere. I addressed those two points and it looks much better now.

onthebridgetonowhere (approved):

PR #312, opened 2022-10-02
Add ability to specify job name and have it show in the job listing page in the UI

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/277

 # Rationale for this change

Easier to find my jobs now:

query-names

Users can set the `ballista.job.name` config in their context and that name appears in UI.


avantgardnerio (approved):
Due to the overhead of handling conditionals in Rust, I'd opt to just have the absence of a `job_name` be represented by an empty `String`, but it's not a huge deal.

andygrove:
Thanks for the review @avantgardnerio

PR #319, opened 2022-10-05
Atomic support for enhancement

metesynnada:

Closes [101](https://github.com/apache/arrow-ballista/issues/101) and address multiple TODOs inside the codebase.

 # Rationale for this change
If a large number of states undergo updates, using `RwLock` for the scheduler's in-memory states at the global level will reduce the overall performance. Additionally, some state updates are not atomic and were listed as TODOs. The modifications in this PR makes concurrency more efficient during heavily loaded in-memory updates and more reliable during state persistence.

- Changed `RwLock<HashMap<K, V>>` into a `DashMap`. The [dashmap](https://docs.rs/dashmap) crate provides an implementation of a more sophisticated, high-performance concurrent hash map. For reference, DashMap stats are

  | Stars      | Forks      | Used By | Contributors | License     |
  |------------|------------|---------|--------------|-------------|
  | 1.8k stars | 104  forks | 42.8k   | 34           | MIT license |

- Now performing atomic transactions on `TaskManager` and `ExecuterManager`, resolving `TODO`s.
  - Change in `trait StateBackendClient`, converted method `put_txn` to a more general `apply_txn`.
  - Rollback if the state persisting is not successful.

No

### DIFF
--- a/ballista/rust/scheduler/src/state/mod.rs
+++ b/ballista/rust/scheduler/src/state/mod.rs
@@ -274,11 +274,12 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
     }
 }
 
-pub async fn with_lock<Out, F: Future<Output = Out>>(lock: Box<dyn Lock>, op: F) -> Out {
-    let mut lock = lock;
+pub async fn with_lock<Out, F: Future<Output = Out>>(
+    mut lock: Box<dyn Lock>,
+    op: F,
+) -> Out {
     let result = op.await;
     lock.unlock().await;
-
     result
 }
 /// It takes multiple locks and reverse the order for releasing them to prevent a race condition.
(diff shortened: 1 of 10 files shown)


ballista/rust/scheduler/src/state/mod.rs
```diff
@@ -281,6 +281,19 @@ pub async fn with_lock<Out, F: Future<Output = Out>>(lock: Box<dyn Lock>, op: F)
...
     result
 }
+/// It takes multiple locks and reverse the order for releasing them to prevent a race condition.
+pub async fn with_locks<Out, F: Future<Output = Out>>(
+    locks: Vec<Box<dyn Lock>>,
+    op: F,
+) -> Out {
+    let mut locks = locks;
+    let result = op.await;
+    locks.reverse();
```
thinkharderdev, line 291: nit

```suggestion
pub async fn with_locks<Out, F: Future<Output = Out>>(
    mut locks: Vec<Box<dyn Lock>>,
    op: F,
) -> Out {
    let result = op.await;
    locks.reverse();
```

thinkharderdev (approved):
LGTM

ozankabak:
Thanks for the review, just sent a commit that incorporates your suggestion.

PR #329, opened 2022-10-08
Downgrade `docker-compose.yaml` to version 3.3 so that we can support Ubuntu 20.04.4 LTS

andygrove:

N/A

 # Rationale for this change

I could not build Docker images on Ubuntu 20.04.4 LTS

Specify version 3.3 in docker-compose.yaml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-version: '3.8'
+version: '3.3'
 services:
   etcd:
     image: quay.io/coreos/etcd:v3.4.9
@@ -54,8 +54,7 @@ services:
     volumes:
       - ./benchmarks/data:/data
     depends_on:
-      ballista-scheduler:
-        condition: service_healthy
+      - ballista-scheduler
     healthcheck:
       test: ["CMD", "nc", "-z", "ballista-executor", "50051"]
       interval: 5s
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,9 +46,10 @@ services:
     build:
       dockerfile: dev/docker/ballista-executor.Dockerfile
       context: .
-    command: "--bind-host 0.0.0.0 --scheduler-host ballista-scheduler"
+    command: "--bind-host 0.0.0.0 --scheduler-host ballista-scheduler --scheduler-connect-timeout-seconds 15"
     deploy:
       replicas: 2
+    restart: always
     environment:
       - RUST_LOG=ballista=debug,info
     volumes:
--- a/.github/workflows/dev_pr.yml
+++ b/.github/workflows/dev_pr.yml
@@ -36,7 +36,7 @@ jobs:
           github.event_name == 'pull_request_target' &&
             (github.event.action == 'opened' ||
              github.event.action == 'synchronize')
-        uses: actions/labeler@4.0.2
+        uses: actions/labeler@v4.0.2
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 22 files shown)


docker-compose.yml
```diff
@@ -54,8 +54,7 @@ services:
     volumes:
       - ./benchmarks/data:/data
     depends_on:
-      ballista-scheduler:
-        condition: service_healthy
```
andygrove, line 58: We should be able to start executors before the scheduler runs, but I haven't tested this in a long time. If we require the scheduler to be running first, then we should file a bug for that.

andygrove: I tested this and the executor fails if it cannot connect to the scheduler. In theory, this should be fine in an environment such as k8s because the pod will just keep restarting until it succeeds.

onthebridgetonowhere: The restart config option could be added to the compose file. https://docs.docker.com/compose/compose-file/#restart
```
restart:always
```

andygrove: I am not sure what to do about this. We could have a config option for having the executor itself retry connecting to the scheduler and use that config option when using docker compose.

andygrove: WDYT @avantgardnerio ?

andygrove: Executor crashes if it starts before the scheduler is fully running (which seems to be the default case for me) and never restarts.

andygrove: I am tempted to add a `--retry-timeout--seconds 10` option to the executor, which would fix this. Ballista is targeted at enterprise users, and they are often not on the latest distro, so would be nice for this to work out of the box for as many people as possible.

onthebridgetonowhere: > I am tempted to add a `--retry-timeout--seconds 10` option to the executor, which would fix this. Ballista is targeted at enterprise users, and they are often not on the latest distro, so would be nice for this to work out of the box for as many people as possible.

I was thinking of the same option given that the application does not seem to have a hard constraint to be readily available within milliseconds on startup. I second the part with having it working out of the box for different users.

onthebridgetonowhere: @andygrove - I am not too familiar with the codebase yet, but I saw that if the scheduler is killed while executor is running, the executor tries to reconnect forever. Once the scheduler is restarted, the executor is fine. Probably a silly question, but couldn't this behaviour be there from the start? Why does it first need to connect to an actual running instance of the scheduler, instead of keep trying the default/configured one until  it succeeds?

andygrove: `TaskRunnerPool` loops forever, polling schedulers to fetch tasks to run. This code opens a new connection to the scheduler each time, which is inefficient but does mean that it eventually recovers from scheduler downtime.

andygrove: I pushed a commit to add the new config. This fixes it for me.

```
ballista-executor_1   | 2022-10-15T15:44:09.490225Z  WARN main ThreadId(01) ballista_executor: Failed to connect to scheduler at http://ballista-scheduler:50050 (Could not connect to scheduler); retrying ...    
ballista-executor_1   | 2022-10-15T15:44:09.991817Z  INFO main ThreadId(01) ballista_executor: Connected to scheduler at http://ballista-scheduler:50050    
```

avantgardnerio (approved):
LGTM

andygrove:
Thanks for the reviews @avantgardnerio and @onthebridgetonowhere

docker-compose.yml
onthebridgetonowhere: > `TaskRunnerPool` loops forever, polling schedulers to fetch tasks to run. This code opens a new connection to the scheduler each time, which is inefficient but does mean that it eventually recovers from scheduler downtime.

I see, I thought it might be the case. Thanks for taking the time to explain!

PR #331, opened 2022-10-09
Enable more benchmark serde tests

andygrove:

None

 # Rationale for this change

Tying to prevent any more regressions

More tests


benchmarks/src/bin/tpch.rs
```diff
@@ -1500,20 +1500,6 @@ mod tests {
             // test logical plan round trip
             let plans = create_logical_plans(&ctx, n)?;
             for plan in plans {
-                let proto: datafusion_proto::protobuf::LogicalPlanNode =
```
andygrove, line 1503: We cant do serde on unoptimized plans because serde does not support subqueries yet

avantgardnerio (approved):

PR #335, opened 2022-10-10
Check executor id consistency when receive stop executor request

### LINKED ISSUE #333: Executor get stopped unexpected
**Describe the bug**

When restart an executor in a running ballista cluster, the executor's ip and port will not be changed. Only its id will be changed. After it gets started, the executor's old one will be detected of losing heartbeat and a stop executor request will be sent to this newly started executor. And this executor will be stopped unexpectedly.

The executor should check whether the executor id in the request is the same as its id. If so, then stop itself. Otherwise, it should do nothing.

**To Reproduce**

**Expected behavior**

**Additional context**

yahoNanJing:

Closes #333 .

 # Rationale for this change

### DIFF
--- a/ballista/rust/core/proto/ballista.proto
+++ b/ballista/rust/core/proto/ballista.proto
@@ -845,10 +845,11 @@ message HeartBeatResult {
 }
 
 message StopExecutorParams {
+  string executor_id = 1;
   // stop reason
-  string reason = 1;
+  string reason = 2;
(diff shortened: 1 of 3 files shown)


thinkharderdev (approved):

thinkharderdev:
Looks like we have a bad version for the labeler action.

andygrove (approved):

Issue #265, opened 2022-09-23
[EPIC] Scheduler UI


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
We have a very basic scheduler UI. This issue is for tracking the work we need to do to make it useful.

**Describe the solution you'd like**

- [x] https://github.com/apache/arrow-ballista/pull/241
- [x] https://github.com/apache/arrow-ballista/issues/276
- [x] https://github.com/apache/arrow-ballista/issues/275
- [x] View query stages in a job
- [x] Cancel job

**Describe alternatives you've considered**
None

**Additional context**
None

onthebridgetonowhere:
Looks like the Cancel job was merged, so this issue/epic could be successfully closed.  :fireworks:

Dandandan:
Looks like the cancelling the job is not enabled yet in the UI - so reopening

onthebridgetonowhere:
Good catch, sorry about that!

andygrove:
Looks like we really can close this now. Thanks @Dandandan  and @onthebridgetonowhere

PR #340, opened 2022-10-11
Add `cancel_job` REST API

tfeda:
Hi all! Hope to get more involved with the project over the next few weeks. Thought id start with some easier items.

part of #265

 # Rationale for this change

Moved grpc cancellation logic to `SchedulerState` so that it could be reused by the REST interface. Kept error handling behavior the same, but added a 404 return when canceling a job that doesn't exist.

For any long running tasks, users can execute a *PATCH* request on the job. for example: 

`curl -X PATCH -v -H 'Accept: application/json'  http://localhost:50050/api/job/25UPl4w | jq`

### _NOTE_
When testing out this addition, I found that cancelling the job causes a few errors:

1. The scheduler always fails its grpc call to the executor to `cancel_tasks()`, and returns with an `Unipmlemented` code. From looking at the executor logs, it doesn't seem to even reach them.

3. The scheduler goes into an infinite loop 
       
image

I think the bugs are unrelated. I'm happy to help debug, but I thought I'd post what I had since the changes are useful for debugging with `curl`

### DIFF
--- a/ballista/rust/scheduler/src/api/handlers.rs
+++ b/ballista/rust/scheduler/src/api/handlers.rs
@@ -54,6 +54,11 @@ pub struct JobResponse {
 }
 
 #[derive(Debug, serde::Serialize)]
+struct CancelJobResponse {
+    pub cancelled: bool,
+}
+
+#[derive(Debug, serde::Serialize)]
 pub struct QueryStageSummary {
     pub stage_id: String,
     pub stage_status: String,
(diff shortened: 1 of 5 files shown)


andygrove:
This is looking great. Thanks @tfeda. Could you also update the REST API documentation at `docs/source/user-guide/scheduler.md` ?

I also see that infinite loop issue when jobs fail. I just filed https://github.com/apache/arrow-ballista/issues/341 to track this.

andygrove (approved):

mingmwang:
Ah, this is a known issue, actually I added the failed job check in the pop_next_task() loop.
Without such check, the scheduler loop will try to schedule the pending tasks from failed job which would be worse !!

@yahoNanJing 
Please take a look and have a fix.

PR #343, opened 2022-10-13
Benchmark & UI improvements

andygrove:

N/A

 # Rationale for this change

Elapsed time is now shown for queries:

quertime

- Set job name so we can see which query is running in the UI
- Infer schema when running against Parquet sources
- Show elapsed time for successful queries


andygrove:
cc @mingmwang @avantgardnerio

Dandandan (approved):

PR #347, opened 2022-10-14
Using local shuffle reader avoid flight rpc call.

### LINKED ISSUE #346: Using local shuffle reader avoid flight rpc call.

There is a situation when shuffler reader and writer  are on the same machine. So we can read from local disk without call rpc and arrow-flight decoding.

**Additional context**

Ted-Jiang:

Closes #346.

There is a situation when shuffler reader and writer are on the same machine. So we can read from directly from local disk without calling RPC and arrow-flight decoding.

 # Rationale for this change

### DIFF
--- a/ballista/rust/core/src/execution_plans/shuffle_reader.rs
+++ b/ballista/rust/core/src/execution_plans/shuffle_reader.rs
@@ -18,20 +18,26 @@
 use async_trait::async_trait;
 use std::any::Any;
 use std::collections::HashMap;
+use std::fs::File;
+use std::pin::Pin;
 use std::result;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
 use crate::client::BallistaClient;
 use crate::serde::scheduler::{PartitionLocation, PartitionStats};
 
 use datafusion::arrow::datatypes::SchemaRef;
-use datafusion::arrow::error::ArrowError;
+use datafusion::arrow::error::{ArrowError, Result as ArrowResult};
+use datafusion::arrow::ipc::reader::FileReader;
+use datafusion::arrow::record_batch::RecordBatch;
 
 use datafusion::error::{DataFusionError, Result};
 use datafusion::physical_plan::expressions::PhysicalSortExpr;
 use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricsSet};
 use datafusion::physical_plan::{
-    DisplayFormatType, ExecutionPlan, Partitioning, SendableRecordBatchStream, Statistics,
+    DisplayFormatType, ExecutionPlan, Partitioning, RecordBatchStream,
+    SendableRecordBatchStream, Statistics,
 };
 use futures::{Stream, StreamExt, TryStreamExt};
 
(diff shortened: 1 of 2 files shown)


ballista/rust/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -498,36 +639,37 @@ mod tests {
         Schema::new(vec![Field::new("id", DataType::Int32, false)])
     }
 
-    #[derive(Clone)]
-    struct MockPartitionReader {}
```
Ted-Jiang, line 502: this will replace by  `PartitionReaderEnum::local`

Ted-Jiang:
@mingmwang @andygrove @yahoNanJing PTAL😊

ballista/rust/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -232,35 +267,53 @@ impl Stream for AbortableReceiverStream {
...
+                drop(permit);
+            });
+            join_handles.push(join_handle);
+        }
     }
 
     AbortableReceiverStream::create(response_receiver, join_handles)
 }
 
+fn check_is_local_location(host: &str) -> bool {
```
andygrove, line 311: An alternate approach here would be to just look on disk and see if the shuffle files exist, then this can work in containerized environments, such as k8s.

andygrove:
Thanks @Ted-Jiang! I will test this over the weekend.

ballista/rust/core/src/execution_plans/shuffle_reader.rs
andygrove: Filed https://github.com/apache/arrow-ballista/issues/356 so we can discuss this as a follow on

andygrove (approved):
This is awesome! Thanks @Ted-Jiang. I see a nice 30% speedup when running queries locally. :rocket:

ballista/rust/core/src/execution_plans/shuffle_reader.rs
Ted-Jiang: Sounds like a better approach🤔

Ted-Jiang:
> This is awesome! Thanks @Ted-Jiang. I see a nice 30% speedup when running queries locally. 🚀

@andygrove Cool! We can care the locality of shuffle in the future!👍

mingmwang:
@Ted-Jiang Could you please add some log info to indicate how many shuffle partitions are read from local and how many are ready from remote?

mingmwang:
@Ted-Jiang 

Please pay attention to the error handling, both remote shuffle read and local shuffle read should return the same error when encounter issues, so that the Scheduler can reschedule tasks/stages.

`BallistaError::FetchFailed`

Ted-Jiang:
> @Ted-Jiang
> 
> Please pay attention to the error handling, both remote shuffle read and local shuffle read should return the same error when encounter issues, so that the Scheduler can reschedule tasks/stages.
> 
> `BallistaError::FetchFailed`

Sure!

PR #367, opened 2022-10-16
Fix integration test and add more queries

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/360

 # Rationale for this change

Integration tests are better when they pass

- Sleep 10 (open to better ideas but this fixed it for me)
- Run more benchmark queries in the test


Dandandan (approved):
LGTM

A follow-up could use something like https://github.com/vishnubob/wait-for-it

PR #370, opened 2022-10-16
Switch to mimalloc and enable by default in executor

### LINKED ISSUE #369: Switch to mimalloc and enable by default
datafusion-cli uses mimalloc by default instead of sn-malloc.
This is generally faster.

Use mimalloc and enable by default.

**Additional context**

Dandandan:

Closes #369

 # Rationale for this change

Helps with performance, like in DataFusion. Also keep benchmark config in sync with executor.


Dandandan:
Going to merge it - seems non controversial as we do it in DataFusion already.

PR #378, opened 2022-10-17
Avoid calling scheduler when the executor cannot accept new tasks

### LINKED ISSUE #377: Avoid calling scheduler when the executor cannot accept new tasks
Currently, we send a request even if there is no available task slot in the executor. This does some unneeded calls. 

Check the task slots first.

n/a

**Additional context**
n/a

Dandandan:

Closes #377 

 # Rationale for this change

Improves performance, especially when using a more limited number of task slots by avoiding a longer sleep between runs.

Running with 1 executor and task slots=1:

Timings(ms):

| Q | Old | New |
|---|-----|-----|
| 1 | 15818.86  | 11097.58   |
| 2 |   50434.39 |  27457.48  |
| 3 |  19201.45  | 12227.47    |
| 4 |  12864.86  |  13034.73   |
| 5 |   39042.86  |  24073.82   |
| 6 |   3823.15  |  2750.53   |
| 7 |   59200.60  |  36749.42   |

### DIFF
--- a/ballista/executor/src/execution_loop.rs
+++ b/ballista/executor/src/execution_loop.rs
@@ -63,10 +63,6 @@ pub async fn poll_loop<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
     info!("Starting poll work loop with scheduler");
 
     loop {
-        // Keeps track of whether we received task in last iteration
-        // to avoid going in sleep mode between polling
-        let mut active_job = false;
-
         let can_accept_task = available_tasks_slots.load(Ordering::SeqCst) > 0;
 
         // Don't poll for work if we can not accept any tasks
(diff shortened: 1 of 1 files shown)


ballista/executor/src/execution_loop.rs
```diff
@@ -63,20 +63,28 @@ pub async fn poll_loop<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
-
         // Keeps track of whether we received task in last iteration
         // to avoid going in sleep mode between polling
         let mut active_job = false;
 
+        let can_accept_task = available_tasks_slots.load(Ordering::SeqCst) > 0;
+
+        // Don't poll for work if we can not accept any tasks
+        if !can_accept_task {
+            tokio::time::sleep(Duration::from_millis(1)).await;
```
andygrove, line 74: Is there any reasoning behind picking `1 ms` versus any other value? Would it be less overhead to sleep for longer here, maybe `10 ms`? Should we make this configurable?

andygrove (approved):
LGTM. Thanks @Dandandan

ballista/executor/src/execution_loop.rs
Dandandan: I think checking the value is really cheap (e.g. compared to grpc requests), so I did use the minimum to reduce the latency as much as possible when a job is running.

I wonder if we can use some lock or channel instead of this loop in the future to rewrite this without `sleep` 🤔 and use resources more efficiently...

PR #379, opened 2022-10-17
Fix latest commit

Dandandan:

n/a
 # Rationale for this change

Somehow the latest commit from https://github.com/apache/arrow-ballista/pull/378 has a side effect of having timeouts :/


Dandandan:
Merging this when green I'm not sure what changes that last commit to fail the job but it seems not really necessary either.

PR #380, opened 2022-10-17
Use dedicated executor in execution loop

Dandandan:

Closes #.

 # Rationale for this change

### DIFF
--- a/ballista/executor/src/execution_loop.rs
+++ b/ballista/executor/src/execution_loop.rs
@@ -22,6 +22,7 @@ use ballista_core::serde::protobuf::{
     TaskDefinition, TaskStatus,
 };
 
+use crate::cpu_bound_executor::DedicatedExecutor;
 use crate::executor::Executor;
 use crate::{as_task_status, TaskExecutionTimes};
 use ballista_core::error::BallistaError;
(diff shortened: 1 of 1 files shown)


ballista/executor/src/execution_loop.rs
```diff
@@ -162,7 +165,7 @@ async fn run_received_tasks<T: 'static + AsLogicalPlan, U: 'static + AsExecution
         task_id, job_id, stage_id, stage_attempt_num, partition_id, task_attempt_num
     );
     info!("Received task {}", task_identity);
-    available_tasks_slots.fetch_sub(1, Ordering::SeqCst);
+    let permit = available_task_slots.clone().acquire_owned().await.unwrap();
```
tfeda, line 168: I'd be careful acquiring a permit _after_ retrieving a task from the scheduler. I'm imagining a scenario where the executor retrieves a task to execute, but then that task sits waiting for a permit to open up. could we move this call up to `poll_loop`, before `scheduler.poll_work()` is called?

Dandandan: I can't see whether that happens, as we also wait until there is a slot available before polling (so at this point there should be at least one available.

But I can see if I can make it not need the first check at all.

tfeda: Sorry about that, I didn't see the first check above. The problem I described wouldn't happen.

Another option is to pass the permit from the first check into `run_received_tasks()`

Dandandan: Yeah I did something like that, but after polling.

I would prefer to see if we can keep it like this (acquiring+release before poll instead of acquiring directly), as I want to add the possibility to retrieve multiple tasks from the scheduler based on `semaphore.available_permits()` (and then acquire those permits later based on the nr. of tasks that are returned).

Dandandan: https://github.com/apache/arrow-ballista/issues/400

PR #398, opened 2022-10-19
Make the scheduler event loop buffer size configurable

### LINKED ISSUE #397: Make the scheduler event loop buffer size configurable

**Additional context**

yahoNanJing:

Closes #397.

 # Rationale for this change

Since the channel in the event loop is a bounded one, sometimes when there're many events produced in a short time, it will block the caller when sending events to the channel. If the buffer size is not larger enough, it will downgrade the performance, especially in a high throughput system.

### DIFF
--- a/ballista/scheduler/src/standalone.rs
+++ b/ballista/scheduler/src/standalone.rs
@@ -38,6 +38,7 @@ pub async fn new_standalone_scheduler() -> Result<SocketAddr> {
             "localhost:50050".to_owned(),
             Arc::new(client),
             BallistaCodec::default(),
+            100000,
         );
     scheduler_server.init().await?;
     let server = SchedulerGrpcServer::new(scheduler_server.clone());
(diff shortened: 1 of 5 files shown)


yahoNanJing:
Hi @thinkharderdev and @andygrove, could you help review this PR?

ballista/scheduler/src/standalone.rs
```diff
@@ -38,6 +38,7 @@ pub async fn new_standalone_scheduler() -> Result<SocketAddr> {
             "localhost:50050".to_owned(),
             Arc::new(client),
             BallistaCodec::default(),
+            100000,
```
andygrove, line 41: Is it intentionaly to use `100,000` here and not `10,000` like other places?

yahoNanJing: It's not intentional. It's better to set it to be the same as other places. I'll change it.

yahoNanJing:
For production usage, we set the buffer size to be one million so that large queries with tens of thousands tasks will not be blocked.

PR #403, opened 2022-10-19
Enable scheduler UI in docker builds

### LINKED ISSUE #362: Incorrect documentation for building Ballista on Linux when using docker-compose
**Describe the bug**
The following pages have incorrect instructions for building and running with docker-compose:

The page at https://github.com/apache/arrow-ballista/blob/master/benchmarks/README.md has:

```
cargo build --release
docker-compose up --build
```

The page at https://github.com/apache/arrow-ballista/blob/master/docs/source/user-guide/flightsql.md has these specific instructions for Linux:

```
cargo build --release --all --features flight-sql
docker-compose up --build
```

Using the latter as an example, docker-compose fails to start the scheduler:

```
Step 9/14 : COPY ballista/ui/scheduler/build /var/www/html
COPY failed: file not found in build context or excluded by .dockerignore: stat ballista/ui/scheduler/build: file does not exist
```

**To Reproduce**
Clone the repo into a new directory and follow the instructions above

**Expected behavior**

**Additional context**

### LINKED ISSUE #402: The Ballista Scheduler Dockerfile copies a file that no longer exists
**Describe the bug**
https://github.com/spaceandtimelabs/arrow-ballista/blob/master/dev/docker/ballista-scheduler.Dockerfile#L31

The line above refers to a file that no longer exists in the master branch.

**To Reproduce**

**Expected behavior**

**Additional context**

iajoiner:

Closes #362.

 # Rationale for this change

### DIFF
--- a/dev/docker/ballista-scheduler.Dockerfile
+++ b/dev/docker/ballista-scheduler.Dockerfile
@@ -28,7 +28,6 @@ RUN apt-get update && apt-get install -y nginx netcat
 
 COPY target/$RELEASE_FLAG/ballista-scheduler /root/ballista-scheduler
 
-COPY ballista/ui/scheduler/build /var/www/html
 COPY dev/docker/nginx.conf /etc/nginx/sites-enabled/default
 
 # Expose Ballista Scheduler web UI port
--- a/dev/docker/ballista-scheduler.Dockerfile
+++ b/dev/docker/ballista-scheduler.Dockerfile
@@ -24,10 +24,20 @@ ENV RUST_LOG=info
 ENV RUST_BACKTRACE=full
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y nginx netcat
+RUN apt-get update &&\
+    apt-get install -y nginx netcat nodejs npm
+    
+COPY ballista/ui /root/ballista/ui
+
+RUN cd /root/ballista/ui/scheduler &&\
+    npm install -g yarn &&\
+    yarn add node-releases &&\
+    yarn install &&\
+    CI=false yarn build
 
 COPY target/$RELEASE_FLAG/ballista-scheduler /root/ballista-scheduler
 
+COPY ballista/ui/scheduler/build /var/www/html
 COPY dev/docker/nginx.conf /etc/nginx/sites-enabled/default
 
 # Expose Ballista Scheduler web UI port
@@ -38,3 +48,4 @@ EXPOSE 50050
 
 COPY dev/docker/scheduler-entrypoint.sh /root/scheduler-entrypoint.sh
 ENTRYPOINT ["/root/scheduler-entrypoint.sh"]
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,7 @@ services:
       context: .
     command: "--config-backend etcd --etcd-urls etcd:2379 --bind-host 0.0.0.0"
     ports:
-      - "80:80"
+      - "55000:80"
       - "50050:50050"
     environment:
       - RUST_LOG=info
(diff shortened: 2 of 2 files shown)


andygrove (changes requested):
This is the scheduler UI, and we need it in the image. I think you are running into https://github.com/apache/arrow-ballista/issues/362

iajoiner:
@andygrove @avantgardnerio This version works although locally I'm using a different port for the UI since my 80 port is taken.

dev/docker/ballista-scheduler.Dockerfile
```diff
@@ -24,7 +24,16 @@ ENV RUST_LOG=info
 ENV RUST_BACKTRACE=full
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y nginx netcat
+RUN apt-get update &&\
+    apt-get install -y nginx netcat nodejs npm
+    
+COPY ballista/ui /root/ballista/ui
+
+RUN cd /root/ballista/ui/scheduler &&\
```
avantgardnerio, line 32: We deliberately got away from building inside the deployment container. If you run https://github.com/apache/arrow-ballista/blob/master/dev/build-ballista-docker.sh

It should run the build https://github.com/apache/arrow-ballista/blob/231d1e3c9936d44c435d1d2fef54178c77e137d7/dev/build-ballista-docker.sh#L22

who's entrypoint calls yarn build https://github.com/apache/arrow-ballista/blob/231d1e3c9936d44c435d1d2fef54178c77e137d7/dev/docker/ballista-builder.Dockerfile#L29

avantgardnerio (changes requested):

dev/docker/ballista-scheduler.Dockerfile
iajoiner: @avantgardnerio Really thanks for explaining! This works. So I guess the real issue is really the docs which I can fix. `. dev/build-ballista-docker.sh` is what people should run.

andygrove: The ARM docker build is removed in https://github.com/apache/arrow-ballista/pull/413

iajoiner: @andygrove Thanks for letting me know! So docker build doesn't really work by running any script if one is on ARM and for AMD people `. dev/build-ballista-docker.sh` should be used, right?

avantgardnerio: I'd like to not have any `.sh` files at all. The build should not be complicated:

```
git clone
cargo build
npm build
docker build
```

Every time I see shell scripts, then I see tech debt go into them instead of just making the builds as easy as above. So the only issue here is just that `npm build` didn't go into all the places in the docs where it should, which is my main complaint against docs as well. The CI scripts are the canonical place to look to see how to run a build, and they are under test.

avantgardnerio: Apologies the previous two comments were not productive. I think the way forward is one that @andygrove suggested: generate the docs from the scripts somehow. I'll have a think on that for a bit. Happy to hear about ways to do that if anyone has any advice.

andygrove: > The CI scripts are the canonical place to look to see how to run a build, and they are under test.

For expert developers such as yourself, I can see how this makes sense, but I want to make sure that we are inclusive for all levels and backgrounds. Some developers would feel intimidated if they were expected to learn how CI works just to be able to build the project locally and would prefer to have documentation and scripts, especially with a project this complex, with multiple languages & technologies (Python, Rust, React, Docker). I know I have given up on projects if I couldn't figure out how to build them locally without going through new learning curves.

I'll file an issue for Windows support. It would be great to support that.

iajoiner:
Closed as the bug #402 doesn't in fact exist.

PR #414, opened 2022-10-20
scheduler now verifies that `file://` ListingTable URLs are accessible

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/353

 # Rationale for this change

My query now fails rather than silently producing the wrong results (empty result set).

```
Error planning job JfrneIG: 
General(\"logical plan refers to path that is not accessible in scheduler file system: /mnt/bigdata/tpcds/sf100-parquet/store_returns.parquet/: IoError(Os { code: 2, kind: NotFound, message: \\\"No such file or directory\\\" })\")"))))
```

- Scheduler verifies files exist
- Logging changes

### DIFF
--- a/ballista/scheduler/src/planner.rs
+++ b/ballista/scheduler/src/planner.rs
@@ -229,6 +229,7 @@ pub fn remove_unresolved_shuffles(
                     .iter()
                     .map(|c| c
                         .iter()
+                        .filter(|l| !l.path.is_empty())
                         .map(|l| l.path.clone())
                         .collect::<Vec<_>>()
                         .join(", "))
--- a/ballista/scheduler/src/scheduler_server/grpc.rs
+++ b/ballista/scheduler/src/scheduler_server/grpc.rs
@@ -38,7 +38,7 @@ use datafusion::datasource::file_format::parquet::ParquetFormat;
 use datafusion::datasource::file_format::FileFormat;
 use datafusion_proto::logical_plan::AsLogicalPlan;
 use futures::TryStreamExt;
-use log::{debug, error, info, warn};
+use log::{debug, error, info, trace, warn};
 
 use std::ops::Deref;
 use std::sync::Arc;
@@ -71,7 +71,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
             task_status,
         } = request.into_inner()
         {
-            debug!("Received poll_work request for {:?}", metadata);
+            trace!("Received poll_work request for {:?}", metadata);
             // We might receive buggy poll work requests from dead executors.
             if self
                 .state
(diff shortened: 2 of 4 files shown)


andygrove:
@avantgardnerio PTAL

ballista/scheduler/src/state/mod.rs
```diff
@@ -256,11 +260,46 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
         plan: &LogicalPlan,
     ) -> Result<()> {
         let start = Instant::now();
+
+        // optimizing the plan here is redundant because the physical planner will do this again
+        // but it is helpful to see what the optimized plan will be
```
yahoNanJing, line 265: Nice catch. How about changing this to the following:

`
        if log::max_level() >= log::Level::Debug {
            let optimized_plan = session_ctx.optimize(plan)?;
            debug!("Optimized plan: {}", optimized_plan.display_indent());
        }
`
so that we can avoid unnecessary optimization when the log level in set to be info.

ballista/scheduler/src/state/mod.rs
```diff
@@ -256,11 +260,46 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
+                plan: &LogicalPlan,
+            ) -> std::result::Result<bool, Self::Error> {
+                if let LogicalPlan::TableScan(scan) = plan {
+                    let provider = source_as_provider(&scan.source)?;
+                    if let Some(table) = provider.as_any().downcast_ref::<ListingTable>()
+                    {
+                        for url in table.table_paths() {
+                            // remove file:/// prefix and verify that the file is accessible
+                            let url = url.as_str();
+                            let url = url.strip_prefix("file:///").unwrap_or(url);
```
yahoNanJing, line 284: Will it check files on S3 or HDFS? Sometimes, the table_paths may be of tens of thousands. Here, the check logic may be very time consuming.

ballista/scheduler/src/state/mod.rs
andygrove: That's a good point. Maybe we just need to check the first path.

ballista/scheduler/src/state/mod.rs
andygrove: I pushed a change so that this check is only performed for files on the local file system (starting with `file:///`) and it now only checks the first file

andygrove:
Thanks for the review @yahoNanJing. I have pushed changes to address your feedback.

ballista/scheduler/src/planner.rs
```diff
@@ -229,6 +229,7 @@ pub fn remove_unresolved_shuffles(
                     .iter()
                     .map(|c| c
                         .iter()
+                        .filter(|l| !l.path.is_empty())
```
andygrove, line 232: unrelated change but this avoids printing out lots of newlines in the log

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -71,7 +71,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
             task_status,
         } = request.into_inner()
         {
-            debug!("Received poll_work request for {:?}", metadata);
+            trace!("Received poll_work request for {:?}", metadata);
```
andygrove, line 74: This was too verbose to be `debug`

andygrove:
I am going to go ahead and merge this so that I can cut the RC today.

PR #442, opened 2022-10-24
Add optional flag which advertises host for Arrow Flight SQL #418

### LINKED ISSUE #418: Add optional flag which advertises host for Arrow Flight SQL
We have the ability to proxy flights through the Scheduler, but whether or not the client tries to get the Flight from the Scheduler or the Executor is entirely up to us and the IP address we advertise as having the Flight.

Many deployments will be made easier if we provide the ability to proxy all flights through the Scheduler. However, some users may prefer the ability to have the Client retrieve flights directly from Executors, so a flag seems like the appropriate middle ground.

Since there are two potential modes of operation, we should use command line flags (or environment variables) to determine which mode to use. Currently, the executor has command line arguments called `--bind-host` and `--external-host`. Following this pattern, this issue proposes we add a command line argument to the scheduler called --advertise-host. If specified, all Flights will be altered to use that address. If not specified, the IPs of the executors will be left intact.

- Force everyone to proxy via the Scheduler
- Don't ever proxy through the Scheduler

**Additional context**

DaltonModlin:
- Update scheduler_config_spec.toml to include new flag 'advertise_host'
- Add advertise_host member variable to SchedulerServer
- Add advertise_host argument to new, new_with_policy, new_with_builder, and new_with_state in order to propagate flag
- Add None argument to relevant method calls

Add optional flag which advertises host for Arrow Flight SQL #418

- Update logic in job_to_fetch_part to use advertise-host flag when it exists
- Remove default from advertise_host in scheduler_config_spec.toml
- Wrap scheduler_server advertise_host variable in Option
- Update scheduler's main.rs to reflect advertise_host being wrapped in Option

Utilize executor IP for routing to flight results in job_to_fetch_part even when advertise-host flag is set.

Add missing variable and ownership stuff

Remove unnecessary output from do_get_fallback

Closes #418

 # Rationale for this change

Adds member variable to SchedulerServer, adds argument to relevant constructors as necessary.

Adds advertise-host flag for proxying via scheduler as described in original issue.

### DIFF
--- a/ballista/scheduler/scheduler_config_spec.toml
+++ b/ballista/scheduler/scheduler_config_spec.toml
@@ -25,6 +25,11 @@ name = "version"
 doc = "Print version of this executable"
 
 [[param]]
+name = "advertise_endpoint"
+type = "String"
+doc = "Route for proxying flight results via scheduler. Should be of the form 'IP:PORT'"
+
+[[param]]
 abbr = "b"
 name = "config_backend"
 type = "ballista_scheduler::state::backend::StateBackend"
--- a/ballista/scheduler/src/flight_sql.rs
+++ b/ballista/scheduler/src/flight_sql.rs
@@ -235,26 +235,41 @@ impl FlightSqlServiceImpl {
     ) -> Result<Vec<FlightEndpoint>, Status> {
         let mut fieps: Vec<_> = vec![];
         for loc in completed.partition_location.iter() {
-            let (host, port) = if let Some(ref md) = loc.executor_meta {
+            let (exec_host, exec_port) = if let Some(ref md) = loc.executor_meta {
                 (md.host.clone(), md.port)
             } else {
                 Err(Status::internal(
-                    "Invalid partition location, missing executor metadata".to_string(),
+                    "Invalid partition location, missing executor metadata and advertise_endpoint flag is undefined.".to_string(),
                 ))?
             };
+
+            let (host, port) = match &self.server.advertise_endpoint {
+                Some(endpoint) => {
+                    let advertise_endpoint_vec: Vec<&str> = endpoint.split(":").collect();
+                    match advertise_endpoint_vec.as_slice() {
+                        [host_ip, port] => {
+                            (String::from(*host_ip), FromStr::from_str(*port).expect("Failed to parse port from advertise-endpoint."))
+                        }
+                        _ => {
+                            Err(Status::internal("advertise-endpoint flag has incorrect format. Expected IP:Port".to_string()))?
+                        }
+                    }
+                }
+                None => (exec_host.clone(), exec_port.clone()),
+            };
+
             let fetch = if let Some(ref id) = loc.partition_id {
                 let fetch = protobuf::FetchPartition {
                     job_id: id.job_id.clone(),
                     stage_id: id.stage_id,
                     partition_id: id.partition_id,
                     path: loc.path.clone(),
-                    host: host.clone(),
-                    port,
+                    // Use executor ip:port for routing to flight result
+                    host: exec_host.clone(),
+                    port: exec_port,
                 };
                 protobuf::Action {
-                    action_type: Some(protobuf::action::ActionType::FetchPartition(
-                        fetch,
-                    )),
+                    action_type: Some(FetchPartition(fetch)),
                     settings: vec![],
                 }
             } else {
@@ -266,7 +281,7 @@ impl FlightSqlServiceImpl {
             } else {
                 Err(Status::internal("Error getting stats".to_string()))?
             }
-            let authority = format!("{}:{}", &host, &port); // TODO: my host & port
+            let authority = format!("{}:{}", &host, &port);
             let loc = Location {
                 uri: format!("grpc+tcp://{}", authority),
             };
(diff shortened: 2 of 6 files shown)


DaltonModlin:
I've validated my changes have the expected outcome via writing a test in the flight portion of the arrow repo.
```
@Test
public void testBallistaFlag() throws Exception {
    String url = "jdbc:arrow-flight://127.0.0.0:50050";
    java.util.Properties props = new java.util.Properties();
    props.setProperty("useEncryption", "false");
    props.setProperty("user", "admin");
    props.setProperty("password", "password");
    Connection con = DriverManager.getConnection(url, props);
    // statement
    java.sql.Statement stmt = con.createStatement();
    String sql = "create external table customer STORED AS CSV WITH HEADER ROW LOCATION 'arrow-datafusion/datafusion/core/tests/tpch-csv/customer.csv';\n";
    assertTrue(stmt.execute(sql));
}
```
Running this test in conjunction with a local scheduler (having passing it `--advertise-host="127.0.0.2:50050"`) and a local executor shows the request for flight results is proxied via the Scheduler rather than going directly to the executor IP.

avantgardnerio:
> flight results is proxied

I'm actually a little surprised that query returns a flight since it is DML. 

@andygrove I think we'll want to merge this PR because we presently rely on a bug in the FlightSQL JDBC driver where Ballista only works because it always re-uses the connection to the scheduler. This PR should make it work even after that issue is fixed.

ballista/scheduler/src/flight_sql.rs
```diff
@@ -196,26 +196,41 @@ impl FlightSqlServiceImpl {
...
+            let (host, port) = match self.server.advertise_host {
+                Some(_) => {
+                    let advertise_host_flag: Vec<&str> = self
+                        .server
+                        .advertise_host
+                        .as_ref()
+                        .unwrap()
+                        .split(":")
+                        .collect();
+                    (advertise_host_flag[0].to_string(), advertise_host_flag[1].parse().unwrap())
```
avantgardnerio, line 217: Same... I think there is a safe way to index into the array: https://adventures.michaelfbryan.com/posts/daily/slice-patterns/

ballista/scheduler/scheduler_config_spec.toml
```diff
@@ -24,6 +24,11 @@ conf_file_param = "config_file"
 name = "version"
 doc = "Print version of this executable"
 
+[[param]]
+name = "advertise_host"
```
avantgardnerio, line 28: I love the use of the word `advertise` as that is what it is commonly called. I dislike the word `host`, because this is hostname & port. Sadly the term for that appears to be "socket", which I think no one uses outside OS libraries https://stackoverflow.com/questions/34955747/whats-the-combination-of-an-ip-address-and-port-number-called

avantgardnerio (approved):
I'd say the unwraps and unsafe array indexing need resolution. The others are just nits.

Nice job! :partying_face:

ballista/scheduler/scheduler_config_spec.toml
andygrove: How about `advertise_endpoint` ?

andygrove:
Approach LGTM. Agree on the unwraps. I would like to test this out as well before approving. I'll try and get that to that today.

ballista/scheduler/src/flight_sql.rs
```diff
@@ -196,26 +196,48 @@ impl FlightSqlServiceImpl {
...
                 ))?
             };
+
+            let (host, port) = match self.server.advertise_endpoint {
+                Some(_) => {
+                    let advertise_endpoint_vec: Vec<&str> = self
+                        .server
+                        .advertise_endpoint
+                        .as_ref()
+                        .expect("Failed to parse advertise_endpoint value")
```
andygrove, line 213: The endpoint is already available from the pattern matching so the `expect` code is unreachable. This can be simplified:

```suggestion
            let (host, port) = match &self.server.advertise_endpoint {
                Some(endpoint) => {
                    let advertise_endpoint_vec: Vec<&str> = endpoint
```

avantgardnerio (approved):

andygrove (approved):
LGTM. Thanks @DaltonModlin!

ballista/scheduler/src/flight_sql.rs
```diff
@@ -196,26 +196,41 @@ impl FlightSqlServiceImpl {
...
-                    "Invalid partition location, missing executor metadata".to_string(),
+                    "Invalid partition location, missing executor metadata and advertise_endpoint flag is undefined.".to_string(),
                 ))?
             };
+
+            let (host, port) = match &self.server.advertise_endpoint {
+                Some(endpoint) => {
+                    let advertise_endpoint_vec: Vec<&str> = endpoint.split(":").collect();
+                    match advertise_endpoint_vec.as_slice() {
+                        [host_ip, port] => {
```
avantgardnerio, line 211: :heart:

andygrove:
@DaltonModlin There is a clippy failure .. I will merge once that is resolved

DaltonModlin:
It looks like clippy is worried because there are 8 argument to the function `new_with_policy`. It might be worth using a Builder pattern to construct the SchedulerServer rather than the current pattern of many slightly differing constructors.

andygrove:
@DaltonModlin Looks like there is conflict?

PR #468, opened 2022-10-28
Add some validation for remove_job_data in the executor server

### LINKED ISSUE #467: Add some validation for the remove_job_data grpc service

To avoid bad or malicious request for the remove_job_data grpc service, we need to check the legality of the job_id. For example, - it should not be an empty string 
- it should not be like "../../"
- etc.

**Additional context**

yahoNanJing:

Closes #467.

 # Rationale for this change

### DIFF
--- a/ballista/executor/src/executor_server.rs
+++ b/ballista/executor/src/executor_server.rs
@@ -19,7 +19,7 @@ use ballista_core::BALLISTA_VERSION;
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::ops::Deref;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


yahoNanJing:
Hi @andygrove @avantgardnerio, could you help review this PR?

mingmwang:
Why we return errors in this RPC?

yahoNanJing:
Although currently we do nothing for the result, it's better to know the detailed error which may be helpful for the scheduler to take future actions.

ballista/executor/src/executor_server.rs
```diff
@@ -720,13 +720,45 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
         request: Request<RemoveJobDataParams>,
     ) -> Result<Response<RemoveJobDataResult>, Status> {
         let job_id = request.into_inner().job_id;
+        info!("Remove data for job {:?}", job_id);
+
+        // Verify whether it's a legal job id
+        {
```
avantgardnerio, line 726: Why the redundant scope?

ballista/executor/src/executor_server.rs
```diff
@@ -720,13 +720,45 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
...
+            }
+        }
+
         let work_dir = self.executor.work_dir.clone();
         let mut path = PathBuf::from(work_dir);
-        path.push(job_id.clone());
-        if path.is_dir() {
-            info!("Remove data for job {:?}", job_id);
-            std::fs::remove_dir_all(&path)?;
+        path.push(job_id);
```
avantgardnerio, line 742: Love the `work_dir` subdirectory check... that's the critical vulnerability here.

ballista/executor/src/executor_server.rs
```diff
@@ -720,13 +720,45 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
...
+        info!("Remove data for job {:?}", job_id);
+
+        // Verify whether it's a legal job id
+        {
+            if job_id.is_empty() {
+                return Err(Status::internal(
+                    "Job id should not be empty!!!".to_string(),
+                ));
+            }
+            if job_id.contains('.') {
```
avantgardnerio, line 732: This is to prevent files with extensions? Or parent directories?

ballista/executor/src/executor_server.rs
```diff
@@ -720,13 +720,45 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
...
+            }
+            if !path.is_dir() {
+                return Err(Status::internal(format!(
+                    "Path {:?} is not for a directory!!!",
+                    path
+                )));
+            }
         }
+
+        std::fs::remove_dir_all(&path)?;
```
avantgardnerio, line 760: According to [the docs](https://doc.rust-lang.org/std/path/struct.PathBuf.html) we have a vulnerability here: 

```
if self has a verbatim prefix (e.g. \\?\C:\windows) and path is not empty, the new path is normalized: all references to . and .. are removed.
```

I can bypass the `.` string check on Windows by doing a `\\?`. It is impossible to validate directories at a string level. The missing check is to fully normalize this directory, then check if the normalized directory is a subdirectory of the `work_dir`.

avantgardnerio (changes requested):
I think this has a critical vulnerability of allowing deletion of paths outside of the workdir.

avantgardnerio:
[This](https://users.rust-lang.org/t/check-whether-a-directory-is-a-subdirectory-of-another-directory/69018) (downvoted) solution is correct.

ballista/executor/src/executor_server.rs
yahoNanJing: It's for prevent deleting parent directory. Generally, a directory should not contain '.'

ballista/executor/src/executor_server.rs
yahoNanJing: It's unnecessary. Just for making the verification steps more clearly

ballista/executor/src/executor_server.rs
yahoNanJing: Agree. I'll refine it to use a more standard and robust way for the subdirectory check.

yahoNanJing:
Hi @avantgardnerio, could you help review again for the refined commit?

avantgardnerio:
> help review again

Looks great, thank you @yahoNanJing !

yahoNanJing:
Thanks @avantgardnerio. Then I'll merge it.

PR #472, opened 2022-10-28
 Add SchedulerConfig for the scheduler configurations, like event_loop_buffer_size, finished_job_data_clean_up_interval_seconds, finished_job_state_clean_up_interval_seconds

### LINKED ISSUE #469: Make the delayed time interval for cleanup job data in both scheduler and executor configurable

Currently it's hard coded by 5 minutes. It's not flexible for different situations and environment. And it's better to distinguish the delayed time interval for shuffle data and job metadata in sled by different values. For example, 
- for the shuffle data, it's OK to delete them in a very short delayed time. 5 minutes is OK.
- for the job metadata in sled, it's better to keep it longer. Maybe hours is better.

**Additional context**

yahoNanJing:

Closes #469.

 # Rationale for this change


yahoNanJing:
Hi @andygrove, @Dandandan, @avantgardnerio, could you help review this PR which refactors the scheduler configurations so that we can add more scheduler related configurations easily in the future?

ballista/core/src/config.rs
```diff
@@ -59,68 +58,51 @@ impl ConfigEntry {
...
-            _data_type,
+            data_type: _data_type,
             default_value,
         }
     }
 }
 
-/// Ballista configuration builder
-pub struct BallistaConfigBuilder {
-    settings: HashMap<String, String>,
```
avantgardnerio, line 70: Oh, it looks like this was already a HashMap based thing. I guess this PR is refactoring that pattern?

avantgardnerio (approved):
1. I reviewed the config changes, they seem to solve an immediate problem without making the existing pattern any worse
2. I didn't review the functional change - cleaning up job stuff, I assume that works
3. Let's discuss the runtime `HashMap` based config somewhere and if we all agree on a resolution, update [this issue](https://github.com/apache/arrow-ballista/issues/479)

yahoNanJing:
Thanks @avantgardnerio for your comment. I'll learn about  https://github.com/softprops/envy later. For the concern of using `HashMap`, I think we can raise another issue or PR if serde is better.

ballista/core/src/config/query.rs
```diff
@@ -0,0 +1,204 @@
...
+pub const BALLISTA_REPARTITION_AGGREGATIONS: &str = "ballista.repartition.aggregations";
+pub const BALLISTA_REPARTITION_WINDOWS: &str = "ballista.repartition.windows";
+pub const BALLISTA_PARQUET_PRUNING: &str = "ballista.parquet.pruning";
+pub const BALLISTA_WITH_INFORMATION_SCHEMA: &str = "ballista.with_information_schema";
+/// give a plugin files dir, and then the dynamic library files in this dir will be load when scheduler state init.
+pub const BALLISTA_PLUGIN_DIR: &str = "ballista.plugin_dir";
+
+/// Ballista configuration, mainly for the query
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BallistaConfig {
```
andygrove, line 39: What is the intent of moving the configs into a `query` namespace? Not all configs will be related to the execution of a single query.

yahoNanJing: The existing configurations seem to be all related to the query execution, which may be used by both the scheduler and executor. The reason to add the query namespace is to distinguish it from the new added SchedulerConfig, since the name `BallistaConfig` may be too general and may give users wrong impression that it includes the SchedulerConfig.

yahoNanJing:
Hi @andygrove, do you have any more concerns?

andygrove:
@yahoNanJing I don't see any updates in the user guide page that covers configuration. I think it would make the review easier if I can see the new docs explaining how to use these configs.

yahoNanJing:
Hi @andygrove, just add the scheduler configurations to the user guide.

Hi @avantgardnerio and @DaltonModlin, with the current configuration refactoring, just add the advertise_endpoint to the scheduler config. Could you help review the related changes?

yahoNanJing:
Hi @andygrove, the scheduler config just be added to the user guide.

avantgardnerio (commented):
Does this turn the command line argument into `--advertise-flight-result-route-endpoint`? That seems overly verbose for a command line argument name. `--help` will show the full description. I think we could safely get away with `--advertise-flight-endpoint` at most.

docs/source/user-guide/configs.md
```diff
@@ -58,3 +60,24 @@ In addition to Ballista-specific configuration settings, the following DataFusio
...
+_Example: Specifying configuration options when starting the scheduler_
+
+```shell
+./ballista-scheduler --scheduler-policy push-staged --event-loop-buffer-size 1000000 --executor-slots-policy
+round-robin-local
+```
+
+| key                                            | type      | default      | description                                                                                                                                                                      |
+|------------------------------------------------|-----------|--------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| scheduler_policy                               | Utf8      | pull-staged  | Sets the scheduing policy for the scheduler, possible values: pull-staged, push-staged.                                                                                          |
```
andygrove, line 78: The table is using underscore (`scheduler_policy`) but the example shell command is using hyphen (`scheduler-policy`).

ballista/scheduler/src/main.rs
```diff
@@ -250,15 +254,34 @@ async fn main() -> Result<()> {
...
+            &opt.event_loop_buffer_size.to_string(),
+        )
+        .set(
+            BALLISTA_FINISHED_JOB_DATA_CLEANUP_DELAY_SECS,
+            &opt.finished_job_data_clean_up_interval_seconds.to_string(),
+        )
+        .set(
+            BALLISTA_FINISHED_JOB_STATE_CLEANUP_DELAY_SECS,
+            &opt.finished_job_state_clean_up_interval_seconds.to_string(),
+        );
```
andygrove, line 269: I don't understand the intention here of creating key-value pairs for these configs. It looks like these configs are just command-line arguments to the scheduler and there is no need to serialize them or send them over the network?

Shouldn't the scheduler config just be a simple struct like this?

```rust
struct SchedulerConfig {
  event_loop_buffer_size: ...,
  finished_job_data_clean_up_interval_seconds: ...,
  ...
  }
```

yahoNanJing: Thanks @andygrove for your comments. Agree with you. It's a bit over designed. I'll change it back and use a normal struct to indicate the scheduler config.

yahoNanJing:
> Does this turn the command line argument into `--advertise-flight-result-route-endpoint`? That seems overly verbose for a command line argument name. `--help` will show the full description. I think we could safely get away with `--advertise-flight-endpoint` at most.

Hi @avantgardnerio, is it only for the result or both the result and the intermediate shuffle data? If it's only for the result, it's better to include 'result' in the configuration name. How about `advertise-flight-result-endpoint`

avantgardnerio:
> advertise-flight-result-endpoint

How about `--advertise-flightsql-endpoint`, which should remove any ambiguity as this is only related to FlightSQL, not regular flights between clients?

A tangent not for this PR: we probably just want separate `--advertise-host` and `--flightsql-port` at some point, as I imagine other services may eventually need to know what host is listening. And now that I think about it, we shouldn't need the port to be specified as that shouldn't get remapped by anything.

andygrove (approved):
Thanks @yahoNanJing! That is looking much cleaner now and easier to review.

PR #474, opened 2022-10-28
Check for expected benchmark results in integration test

andygrove:

Related to https://github.com/apache/arrow-ballista/issues/330

 # Rationale for this change

The integration test has little value if it does not check that query results are correct.

This is implemented now but disabled for most queries because we do not produce the correct results because of issues with decimal precision and also because the test relies on casting the expected results (in csv format) to decimal and we do not support casting utf8 to decimal yet.

check query results

No

### DIFF
--- a/benchmarks/run.sh
+++ b/benchmarks/run.sh
@@ -23,5 +23,5 @@ set -x
 
 for query in 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 17 18 19 21
 do
-  /root/tpch benchmark ballista --host ballista-scheduler --port 50050 --query $query --path /data --format tbl --iterations 1 --debug --verify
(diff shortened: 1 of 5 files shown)


andygrove:
I cannot fully implement this until https://github.com/apache/arrow-datafusion/issues/3478 is fixed

Dandandan (approved):
Super to have those. I think we should get them in even without the decimal setup :+1: 

Would be nice to have something similar for DataFusion as well to catch regressions there sooner.

andygrove:
> Would be nice to have something similar for DataFusion as well to catch regressions there sooner.

Agreed. It would also be nice if I didn't have to keep making the same changes in both repos :roll_eyes: but I don't have a good plan for how to do that

PR #501, opened 2022-11-08
Automatically register tables if env var specified

### LINKED ISSUE #500: Use ListingSchema in Ballista

Ever since [this PR](https://github.com/apache/arrow-datafusion/pull/4112) was merged, DataFusion has had the ability to automatically register tables for a default schema, and Ballista users have be jealous. This issue represents that envy.

Treat Ballista users like first class citizens and allow them to have the nice things present in DataFusion.

Grin and bear it.

avantgardnerio:

Closes #500.

 # Rationale for this change

Described in issue.

1. upgrade of datafusion & arrow
2. config plumbing
3. conversion of `default_session_builder` to a supserset of functionality (now with state)
4. default table factories
5. a bunch of flight sql work for database introspection (more required in future PRs)
6. 

### DIFF
--- a/ballista-cli/Cargo.toml
+++ b/ballista-cli/Cargo.toml
@@ -29,12 +29,12 @@ rust-version = "1.59"
 readme = "README.md"
 
 [dependencies]
-ballista = { path = "../ballista/client", version = "0.9.0", features = [
-    "standalone",
-] }
+ballista = { path = "../ballista/client", version = "0.9.0", features = ["standalone"] }
+ballista-core = { path = "../ballista/core", version = "0.9.0" }
 clap = { version = "3", features = ["derive", "cargo"] }
-datafusion = "14.0.0"
-datafusion-cli = "14.0.0"
+dashmap = "5.4.0"
+datafusion = { git = "https://github.com/apache/arrow-datafusion.git", rev = "7b5842b91ebd00a2c7f894fcad797bea68a56d0f" }
+datafusion-cli = { git = "https://github.com/apache/arrow-datafusion.git", rev = "7b5842b91ebd00a2c7f894fcad797bea68a56d0f" }
 dirs = "4.0.0"
 env_logger = "0.9"
 mimalloc = { version = "0.1", default-features = false }
--- a/ballista/client/src/context.rs
+++ b/ballista/client/src/context.rs
@@ -124,6 +124,7 @@ impl BallistaContext {
                 scheduler_url,
                 remote_session_id,
                 state.config(),
+                HashMap::new(),
             )
         };
 
@@ -137,13 +138,18 @@ impl BallistaContext {
     pub async fn standalone(
         config: &BallistaConfig,
         concurrent_tasks: usize,
+        session_builder: Option<Arc<dyn ballista_core::utils::SessionBuilder>>,
     ) -> ballista_core::error::Result<Self> {
         use ballista_core::serde::protobuf::PhysicalPlanNode;
         use ballista_core::serde::BallistaCodec;
+        use ballista_core::utils::DefaultSessionBuilder;
+        use ballista_scheduler::standalone::new_standalone_scheduler;
 
         log::info!("Running in local mode. Scheduler will be run in-proc");
 
-        let addr = ballista_scheduler::standalone::new_standalone_scheduler().await?;
+        let session_builder =
+            session_builder.unwrap_or_else(|| Arc::new(DefaultSessionBuilder {}));
+        let addr = new_standalone_scheduler(session_builder.clone()).await?;
         let scheduler_url = format!("http://localhost:{}", addr.port());
         let mut scheduler = loop {
             match SchedulerGrpcClient::connect(scheduler_url.clone()).await {
@@ -183,6 +189,7 @@ impl BallistaContext {
                 scheduler_url,
                 remote_session_id,
                 config,
+                HashMap::new(),
             )
         };
 
@@ -438,9 +445,10 @@ mod tests {
     #[cfg(feature = "standalone")]
     async fn test_standalone_mode() {
         use super::*;
-        let context = BallistaContext::standalone(&BallistaConfig::new().unwrap(), 1)
-            .await
-            .unwrap();
+        let context =
+            BallistaContext::standalone(&BallistaConfig::new().unwrap(), 1, None)
+                .await
+                .unwrap();
         let df = context.sql("SELECT 1;").await.unwrap();
         df.collect().await.unwrap();
     }
@@ -452,9 +460,10 @@ mod tests {
         use std::fs::File;
         use std::io::Write;
         use tempfile::TempDir;
-        let context = BallistaContext::standalone(&BallistaConfig::new().unwrap(), 1)
-            .await
-            .unwrap();
+        let context =
+            BallistaContext::standalone(&BallistaConfig::new().unwrap(), 1, None)
+                .await
+                .unwrap();
 
         let data = "Jorge,2018-12-13T12:12:10.011Z\n\
                     Andrew,2018-11-13T17:11:10.011Z";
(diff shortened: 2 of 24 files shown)


avantgardnerio:
@andygrove this allows users to see tables in ballista from DataGrip (and probably tableau):

image

andygrove:
I'm excited about this feature! I will test this out later this week.

avantgardnerio:
image

We can now see columns

yahoNanJing:
Thanks @avantgardnerio. I'll read carefully for this PR tomorrow.

ballista/core/src/utils.rs
```diff
@@ -53,25 +56,86 @@ use log::error;
...
+        let state =
+            SessionState::with_config_rt(config, Arc::new(RuntimeEnv::new(rt_cfg)?));
+        Ok(state)
+    }
+}
+
+/// Session builder with custom table providers
+#[derive(Clone)]
+pub struct TableProviderSessionBuilder {
+    table_factories: Arc<Mutex<HashMap<String, Arc<dyn TableProviderFactory>>>>,
```
yahoNanJing, line 91: Should we use `DashMap` to replace the `Mutex<HashMap>`, since we have done the similar change in other places and it will improve the performance under high throughput scenarios.

ballista/client/src/context.rs
```diff
@@ -137,13 +139,17 @@ impl BallistaContext {
     pub async fn standalone(
         config: &BallistaConfig,
         concurrent_tasks: usize,
+        table_factories: HashMap<String, Arc<dyn TableProviderFactory>>,
```
yahoNanJing, line 142: Should we use `TableProviderSessionBuilder` here? And for the standalone case, can the table_factories be determined by the ballista config? Maybe we don't need this additional parameter for this interface.

avantgardnerio: > can the table_factories be determined by the ballista config?

Unfortunately, I don't think so:

```
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct BallistaConfig {
```

I don't think it is logical for `TableProviderFactories` to implement `PartialEq` or `Eq`, and without making another DataFusion PR, they don't impl `Debug` at present either. Everything else in the `BallistaConfig` is a key-value pair, and extending it to include stateful runtime objects seems like a questionable choice.

I can however swap out the Map for an `Arc<dyn SessionBuilder>`, but then `client` needs to have a new dependency on `core`.

avantgardnerio: I made it `Option`al with a default. Hopefully passing `None` will be less burdensome.

avantgardnerio: Another option I just thought of: create a new struct (i.e. `BallistaSettings`) which wraps `BallistaConfig` but also adds `concurrent_tasks` and `table_factories` fields.

avantgardnerio:
@stuartcarnie you might be interested in this as well. I can start copying you on FlightSQL related PRs if you like.

stuartcarnie:
@avantgardnerio great, thank you!

ballista-cli/Cargo.toml
```diff
@@ -29,12 +29,12 @@ rust-version = "1.59"
...
-ballista = { path = "../ballista/client", version = "0.9.0", features = [
-    "standalone",
-] }
+ballista = { path = "../ballista/client", version = "0.9.0", features = ["standalone"] }
+ballista-core = { path = "../ballista/core", version = "0.9.0" }
 clap = { version = "3", features = ["derive", "cargo"] }
-datafusion = "14.0.0"
-datafusion-cli = "14.0.0"
+dashmap = "5.4.0"
+datafusion = { git = "https://github.com/apache/arrow-datafusion.git", rev = "7b5842b91ebd00a2c7f894fcad797bea68a56d0f" }
```
andygrove, line 36: If we need features added since 14.0.0, then we either need to delay merging this PR until after we release Ballista 0.10.0, or we can merge now and then wait for DataFusion 15.0.0 before releasing Ballista 0.10.0. I don't have a strong preference either way.

ballista/core/src/config.rs
```diff
@@ -87,6 +87,17 @@ impl BallistaConfigBuilder {
         Self { settings }
     }
 
+    pub fn load_env(&self) -> Self {
+        let mut settings = self.settings.clone();
+        if let Ok(it) = env::var("DATAFUSION_CATALOG_LOCATION") {
+            settings.insert("datafusion.catalog.location".to_string(), it);
```
andygrove, line 93: The Flight SQL page in the user guide should be updated with information about these env vars, and the configs.md page should be updated with these DataFusion settings.

andygrove:
I tried this out and did run into an issue when connecting from DataGrip:

```
INTERNAL: Failed to create SessionContext: Context("Could not create table for file:///mnt/bigdata/tpch/sf10-parquet/partsupp.parquet at /home/andy/.cargo/git/checkouts/arrow-datafusion-71ae82d9dec9a01c/7b5842b/datafusion/core/src/catalog/listing_schema.rs:116", ObjectStore(Generic { store: "LocalFileSystem", source: UnableToReadBytes { source: Os { code: 21, kind: IsADirectory, message: "Is a directory" }, path: "/mnt/bigdata/tpch/sf10-parquet/partsupp.parquet" } })).
```

The code assumes that files in `DATAFUSION_CATALOG_LOCATION` will be single files and not directories containing files.

ballista-cli/Cargo.toml
avantgardnerio: I see we have done the git-ref thing in the past, so there's precedent.

For FlightSql in particular, I've had to wait a long time for arrow to release, then datafusion, then ballista. I've been rebasing changes for months and our fork continues to have changes that are still waiting (in this case on delta-rs, but now they are way behind datafusion).

For these reasons, I'd strongly prefer to do a ballista release shortly after the datafusion release, but then go back to git-ref tracking, so we can keep PRs flowing. 

Just my $0.02

avantgardnerio:
> I tried this out

Thank you so much for your review! I realize you cloned and did manual testing and that probably took a fair bit of time, which I really appreciate :)

> did run into an issue

It definitely works with csvs with headers, and there's a test for that. I'll look into parquet directory support tomorrow. If it's not a big deal, let's get it in this PR. If it isn't trivial though, I'd prefer to document it in another issue and get this PR merged - with the logic that getting the framework in place for auto-registration is the bigger deal, and once it's in we can extend it for all the permutations of parameters.

ballista/core/src/config.rs
avantgardnerio: > Flight SQL page in the user guide should be updated

I can do that, but I think it should work across the board - even if you connect from a regular ballista client. I'll test that.

> configs.md page should be updated with these DataFusion settings

Thanks, I was unaware. I'll go do that.

r4ntix:
great feature! 👍 
Is it possible to support custom `CatalogProvider` in the next step?

andygrove (approved):
LGTM. It isn't fully working for my use case but since this is new functionality I am fine with merging and then iterating on it after it is merged.

andygrove:
> Thank you so much for your review! I realize you cloned and did manual testing and that probably took a fair bit of time, which I really appreciate :)

Yes, this is one of the challenges with this project. Many PRs need to be tested end-to-end and this does take time, which is why I mostly only review larger PRs at weekends.

avantgardnerio:
> support custom `CatalogProvider`

@r4ntix could you explain your use-case? It may already be covered. For example, when this gets merged, [deltatables](https://github.com/spaceandtimelabs/delta-rs/blob/cb553eac000ab491f8b553b8f3e20bd791dc10bf/rust/src/delta_datafusion.rs#L834) should work out of the box, as long as there are `TableProviderFactory`s registered for them.

r4ntix:
> > support custom `CatalogProvider`
> 
> @r4ntix could you explain your use-case? It may already be covered. For example, when this gets merged, [deltatables](https://github.com/spaceandtimelabs/delta-rs/blob/cb553eac000ab491f8b553b8f3e20bd791dc10bf/rust/src/delta_datafusion.rs#L834) should work out of the box, as long as there are `TableProviderFactory`s registered for them.

@avantgardnerio according to the current implementation, the automatically registered tables are put into the `datafusion.default.{table}` datafusion schema. 
In my case, for tenant isolation, I need to be able to dynamically register tables with different schema or catalog, eg: `datafusion.{tenant}.{table}` or `{tenant}.default.{table}`. 
So is it possible to support `CatalogProvider` in `SessionBuilder` to support such case?

avantgardnerio:
> is it possible to support CatalogProvider in SessionBuilder

I don't see why not. You could follow the patterns in this PR but with a CatalogProvider instead.

r4ntix:
> > is it possible to support CatalogProvider in SessionBuilder
> 
> I don't see why not. You could follow the patterns in this PR but with a CatalogProvider instead.

Ok, I will try submit a RP for this, thx

PR #504, opened 2022-11-08
Prometheus metrics

### LINKED ISSUE #427: Provide integration with Grafana / Prometheus
I would like to be able to monitor scheduler and runtime metrics real time

Provide integration with Grafana / Prometheus

**Additional context**

thinkharderdev:

Closes #427

 # Rationale for this change

We should expose metrics from the scheduler in a standard format. This PR adds an initial integration with prometheus to expose some basic scheduler metrics

1. Add a new trait `SchedulerMetricsCollector` which is plugged into the core event loop to capture metrics
2. Add some additional metadata to the job events so we can track durations of the different parts of the query lifecycle (queued, planning, execution) separately. 
3. Add an implementation of `SchedulerMetricsCollector` for prometheus
4. Add an API endpoint `/api/metrics` which will expose metrics. 
5. Add some baseline metrics

In addition, I did some refactoring to make the core event loop more testable. Currently you can't really test it rigorously at all which is not great since it is a rather critical piece of code. I added a new trait `TaskLauncher` which can control how the `TaskManager` which actually launch tasks and various crate-private methods to plug in a custom launcher. I don't imagine this will ever be something exposed through a public API but it is very useful for testing. 

Using the `TaskLauncher` I added some additional utilities to write scheduler test more succinctly. 

Users can plug in their own metrics collector if they want. 

No

### DIFF
--- a/ballista/scheduler/src/metrics/prometheus.rs
+++ b/ballista/scheduler/src/metrics/prometheus.rs
@@ -18,13 +18,17 @@
 use crate::metrics::SchedulerMetricsCollector;
 use ballista_core::error::{BallistaError, Result};
 use hyper::header::CONTENT_TYPE;
+use once_cell::sync::OnceCell;
 use prometheus::{
     register_counter_with_registry, register_gauge_with_registry,
     register_histogram_with_registry, Counter, Gauge, Histogram, Registry,
 };
 use prometheus::{Encoder, TextEncoder};
+use std::sync::Arc;
 
 use warp::Reply;
+
+static COLLECTOR: OnceCell<Arc<dyn SchedulerMetricsCollector>> = OnceCell::new();
 
 pub struct PrometheusMetricsCollector {
     execution_time: Histogram,
@@ -112,6 +116,16 @@ impl PrometheusMetricsCollector {
             submitted,
             pending_queue_size,
         })
+    }
+
+    pub fn current() -> Result<Arc<dyn SchedulerMetricsCollector>> {
+        COLLECTOR
+            .get_or_try_init(|| {
+                let collector = Self::new(::prometheus::default_registry())?;
+
+                Ok(Arc::new(collector) as Arc<dyn SchedulerMetricsCollector>)
+            })
+            .map(|arc| arc.clone())
     }
 }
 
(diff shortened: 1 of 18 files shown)


ballista/scheduler/src/metrics/prometheus.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let planning_time = register_histogram_with_registry!(
+            "planning_time_ms",
+            "Histogram of query planning time in milliseconds",
+            vec![1.0_f64, 5.0_f64, 25.0_f64, 100.0_f64, 500.0_f64],
+            registry
+        )
+        .map_err(|e| {
+            BallistaError::Internal(format!("Error registering metric: {:?}", e))
+        })?;
```
thinkharderdev, line 42: These histograms are hard to generalize since use cases could vary wildly. We can either:
1. Make these configurable
2. Try and choose sensible defaults and if someone needs to customize then they can implement a custom collector

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -405,19 +478,19 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
         };
 
-        let _res = if let Some(graph) = self.remove_active_execution_graph(job_id).await {
+        if let Some(graph) = self.remove_active_execution_graph(job_id).await {
             let mut graph = graph.write().await;
             let previous_status = graph.status();
             graph.fail_job(failure_reason);
-            let value = self.encode_execution_graph(graph.clone())?;
+
+            let value = encode_protobuf(&graph.status())?;
```
thinkharderdev, line 486: First bug uncovered while testing the event loop proper :) When a job would fail we would save the entire graph into `FailedJobs` here but in all other scenarios we only save the status

andygrove:
This is awesome! Thanks @thinkharderdev. I will make time to try this out later this week.

thinkharderdev:
I think this is ready for review. I had to tweak how we do the API routing slightly because the prometheus exporter probably won't send an `application/json` `Accept` header and I'm not sure it is guaranteed to send any particular `Accept` header so changed the routing to look at the path instead. But I was able to verify the metrics are exported.

andygrove:
I filed https://github.com/apache/arrow-ballista/issues/507 for writing documentation in the user guide for this feature.

andygrove:
> > 
> 
> Hmm... I haven't seen this locally but is looks like something when building the prometheus-rs crate. What OS are you using?

Ubuntu 20.04.4 LTS

thinkharderdev:
@andygrove This PR is superseded by https://github.com/apache/arrow-ballista/pull/511 which has some tweaks to make the plug-ability more useful so if we're good with the other PR we should just merge that one and I can close this one. I'll leave it open for the moment in case it's easier to review them separately.

PR #520, opened 2022-11-17
configure_me_codegen retroactively reserved on our `bind_host` parame…

### LINKED ISSUE #519: build broken: configure_me_codegen retroactively reserved `bind_host`
**Describe the bug**

The master build is broken because configure_me_codegen released a breaking change somewhere after `0.4.0` which is causing our builds to fail because they "reserved" the `bind_host` parameter name (which we use).

**To Reproduce**

```
rm Cargo.lock
cargo clean
cargo build
```

**Expected behavior**

The build works.

**Additional context**

We should follow up by dropping this crate, or renaming our parameter and unpinning the old version.

avantgardnerio:

Closes #519.

 # Rationale for this change

Described in issue.

Pin an old version until we can fix properly.

The build works.

### DIFF
--- a/ballista/executor/Cargo.toml
+++ b/ballista/executor/Cargo.toml
@@ -62,7 +62,7 @@ uuid = { version = "1.0", features = ["v4"] }
 [dev-dependencies]
 
 [build-dependencies]
-configure_me_codegen = "0.4.0"
+configure_me_codegen = "=0.4.0"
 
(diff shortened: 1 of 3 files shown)


andygrove (approved):

avantgardnerio:
This is 1 breakage less than master... perhaps we can merge this?

PR #521, opened 2022-11-18
Ballista gets a docker image!!!

avantgardnerio:

N/A

 # Rationale for this change

Now that we know github actions can publish images, let's only publish images on tags by certified apache committers.

described above

described above


andygrove (approved):

dev/docker/ballista-standalone.Dockerfile
```diff
@@ -33,6 +33,9 @@ RUN apt-get -qq update && apt-get install -qq -y nginx netcat wget
 COPY target/$RELEASE_FLAG/ballista-scheduler /root/ballista-scheduler
 COPY target/$RELEASE_FLAG/ballista-executor /root/ballista-executor
 
+RUN chmod a+x /root/ballista-scheduler && \
+    chmod a+x /root/ballista-executor
```
Jimexist, line 37: ideally this should be placed into `/usr/local/bin` or `/usr/bin` so that others can use.

using `/root/` isn't user friendly for non-root users

andygrove: I'd like to merge this to unblock cutting the RC, but it would be good to discuss this in another PR

PR #523, opened 2022-11-18
Provide a memory StateBackendClient

### LINKED ISSUE #505: Provide an in-memory StateBackend

In some scenarios, there's no requirement of multiple schedulers. And when a scheduler is down, there's no requirement of task recovering and cluster topology recovering. In such cases, an in-memory StateBackend is enough and we don't need to persist the scheduler to the external storage. It's better to provide such kind of option.

**Additional context**

yahoNanJing:

Closes #505.

 # Rationale for this change


yahoNanJing:
Hi @andygrove, @thinkharderdev, @Ted-Jiang, could you help review this PR?

ballista/scheduler/src/state/backend/utils/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[allow(dead_code)]
+mod sled_oneshot;
+#[allow(dead_code)]
+pub(crate) mod sled_subscriber;
```
thinkharderdev, line 21: Confused about the naming here. Why is this called `sled_subscriber`? It's used by the memory backend right?

ballista/scheduler/src/state/backend/memory.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        if let Some(from_space_state) = self.states.get(&from_space_key) {
+            if let Some(state) = from_space_state.value().get(key) {
+                let ops = vec![
+                    (Operation::Delete, from_keyspace, key.to_owned()),
+                    (Operation::Put(state.clone()), to_keyspace, key.to_owned()),
+                ];
+                self.apply_txn(ops).await
+            } else {
+                // TODO should this return an error?
```
thinkharderdev, line 188: I think this should be consistent with the Sled and Etcd backends which I believe would both return an error in this scenario

ballista/scheduler/src/state/backend/memory.rs
```diff
@@ -0,0 +1,395 @@
...
+    async fn lock(&self, keyspace: Keyspace, key: &str) -> Result<Box<dyn Lock>> {
+        let flat_key = Self::get_flat_key(&keyspace, key);
+        if let Some(lock) = self.locks.get_mut(&flat_key) {
+            Ok(Box::new(lock.value().clone().lock_owned().await))
+        } else {
+            let new_lock = Arc::new(Mutex::new(()));
+            self.locks.insert(flat_key, new_lock.clone());
+            Ok(Box::new(new_lock.lock_owned().await))
+        }
+    }
```
thinkharderdev, line 214: Hmm, I think this might allow multiple locks. 

Can we use the entry API here? 

```rust
let new_lock = Arc::new(Mutex::new(()));

let lock = self.locks.entry(&flat_key).or_insert_with(|| Arc::new(Mutex::new(())));
Ok(Box::new(lock.value().clone().lock_owned().await))
```

thinkharderdev (commented):
Very nice! A few comments.

ballista/scheduler/src/state/backend/utils/mod.rs
yahoNanJing: Hi @thinkharderdev, thanks for pointing it out. Here, I just hope to indicate these mods are the modified version from the sled crate, which may cause a bit confusing here. I'll remove the `sled_` prefix.

ballista/scheduler/src/state/backend/memory.rs
yahoNanJing: Actually, it's consistent with the `sled` implementation

ballista/scheduler/src/state/backend/memory.rs
yahoNanJing: Good suggestion. I'll refine it.

thinkharderdev (approved):

yahoNanJing:
Thanks @thinkharderdev. I'll merge this PR, since it's also been verified on our internal cluster.

PR #557, opened 2022-12-09
Make it concurrently to launch tasks to executors

### LINKED ISSUE #544: Make it concurrently to launch tasks to executors

For a large cluster with hundreds of executors, it's better to make it concurrently to launch tasks to executors rather than sequentially. Support for every launch tasks it costs 10ms, then for a cluster with one hundred executors, there may be 1s delay if we do the launching tasks sequentially.

**Additional context**

yahoNanJing:

Closes #544.

 # Rationale for this change

### DIFF
--- a/ballista/scheduler/src/state/mod.rs
+++ b/ballista/scheduler/src/state/mod.rs
@@ -281,9 +281,21 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
                     });
                     join_handles.push(join_handle);
                 }
-                for join_handle in join_handles {
-                    unassigned_reservations.append(&mut join_handle.await?);
-                }
+
+                let unassigned_executor_reservations =
+                    futures::future::join_all(join_handles)
+                        .await
+                        .into_iter()
+                        .collect::<std::result::Result<
+                        Vec<Vec<ExecutorReservation>>,
+                        tokio::task::JoinError,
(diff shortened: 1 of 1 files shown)


yahoNanJing:
Hi @thinkharderdev and @andygrove, could you help review this PR?

ballista/scheduler/src/state/mod.rs
```diff
@@ -236,46 +236,53 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
+                                ExecutorReservation::new_free(executor_id.clone(),);
+                                n_tasks
+                            ]
                         }
-                    }
+                    });
+                    join_handles.push(join_handle);
+                }
+                for join_handle in join_handles {
+                    unassigned_reservations.append(&mut join_handle.await?);
```
thinkharderdev, line 285: Not sure I understand, wouldn't this still process the work sequentially? I would we would want do do a `futures::future::join_all` on the collection of futures to actually execute them in parallel

yahoNanJing: Hi @thinkharderdev, thanks for your advice. The blocking part is calling the method _launch_multi_task_ which has already been invoked in parallel. However, `futures::future::join_all` is a much more simple way. I'll refine it.

thinkharderdev (approved):

PR #560, opened 2022-12-09
Cluster state refactor part 1

### LINKED ISSUE #554: Refactor `StateBackendClient` to be a higher-level interface

Currently, persistent state is managed using the `StateBackendClient` trait which is very low-level interface over a key-value store. This is quite flexible but has a couple of drawbacks:

1. It requires serde overhead even when it is not required. For instance in the recently added `MemoryBackendState` everything is kept in memory so there is no real reason why anything needs to be serialized to `Vec<u8>`
2. Much more importantly, it fixes access patterns in a way that prevents specific implementation from using backend-specific features such as batch operations, atomic counters, key expiration etc. This has the effect of forcing the use of distributed locking in cases where it can potentially be avoided.

For example, it's possible with a Redis as a backend to do global task slot allocation atomically and without any distributed locking using HMap and a little bit of server-side scripting. But there is really no way to express that using the current interface. 

Break the current `StateBackendClient` trait into two separate higher-level interfaces:

```rust
trait JobState {
  async fn save_job(&self, job_id: &str, graph: ExecutionGraph) -> Result<()>;
  async fn fail_job(&self, job_id: &str, reason: String, graph: Option<ExecutionGraph>) -> Result<()>; 
  async fn remove_job(&self, job_id: &str) -> Result<()>;
  async fn save_session(&self, ctx: &Arc<SessionContext>) -> Result<()>;
  async fn get_session(&self, session_id: &str) -> Result<Arc<SessionContext>>;
}

trait ClusterState {
  async fn reserve_slots(&self, num_slots: u32) -> Result<Vec<ExecutorReservation>>;
  /// Either reserve all `num_slots` slots or none
  async fn reserve_slots_exact(&self, num_slots: u32) -> Result<Vec<ExecutorReservation>>;
  async fn cancel_reservations(&self, reservations: Vec<ExecutorReservations>) -> Result<()>;
  async fn register_executor(&self, metadata: ExecutorMetadata, spec: ExecutorSpecification, reserve: bool) -> Result<Vec<ExecutorReservation>>;
  async fn remove_executor(&self, executor_id: &str) -> Result<()>;
}
```

This is only a sketch and these interfaces may need to be tweaked in various ways, but I think the benefits of the general approach are significant:

1. It pushes locking down as an implementation detail in the state backend so we can leverage advanced features of particular data stores
2. It makes application code more concise as we don't have to deal with the fussy details of the low-level interface there
3. Implementations can "pick and choose" what needs to be visible globally (eg to all schedulers in a cluster) so we have more flexibility in deciding where we want to make tradeoffs between resiliency and performance. 
4. For things where we only need to store state in memory we can get rid of the serialization overhead. 

Note that the above traits don't have a way to `Watch` keys. I think the only place we use that is for listening to executor heartbeats and I think that can be internalized into the `ClusterState` implementation but if not it should be easy to work into this design. 

We could not do this

**Additional context**

As briefly mentioned above this is mostly coming from the desire to make Ballista truly highly available. The basic outline of the HA solution is pretty straightforward:

1. Run multiple schedulers which manage all job state in memory.
2. Handle sharding of jobs across schedulers on the client side using consistent hashing. 
3. Use Redis as a backend for task slots/executor data so we can do atomic, lock-free slot allocation.
4. Profit!

thinkharderdev:

Closes #554 

 # Rationale for this change

See issue for rationale

This is part 1 which creates a `ClusterState` interface for a higher-level abstraction to manage executor resources. A default implementation is provided which just uses the `StateBackendClient` for the underlying implementation so functionally this should not change anything for existing users. 

A new CLI config `-c/--cluster-backend` is available to set the backend for `ClusterState` which will default to `Sled`. Right now there is no way to use a separate Sled dir (or etcd cluster for that matter) for the regular `StateBackendClient` and the `ClusterState` because I though that would be weird use case and mostly just make things confusing. 

No


thinkharderdev:
@yahoNanJing @mingmwang @andygrove Can you guys help review this?

andygrove:
I took a quick skim through this and looks reasonable to me. I will try and find time soon for a more in-depth review.

ballista/core/src/serde/generated/ballista.rs
```diff
@@ -1985,7 +1985,7 @@ pub mod executor_grpc_client {
 pub mod scheduler_grpc_server {
     #![allow(unused_variables, dead_code, missing_docs, clippy::let_unit_value)]
     use tonic::codegen::*;
-    /// Generated trait containing gRPC methods that should be implemented for use with SchedulerGrpcServer.
+    ///Generated trait containing gRPC methods that should be implemented for use with SchedulerGrpcServer.
```
yahoNanJing, line 1988: Is it necessary to remove the blank after the ///

ballista/scheduler/scheduler_config_spec.toml
```diff
@@ -36,6 +36,13 @@ type = "ballista_scheduler::state::backend::StateBackend"
 doc = "The configuration backend for the scheduler, possible values: etcd, memory, sled. Default: sled"
 default = "ballista_scheduler::state::backend::StateBackend::Sled"
 
+[[param]]
+abbr = "c"
```
yahoNanJing, line 40: Do we need both `config_backend` and `cluster_backend`? Personally, I prefer `cluster_backend`

thinkharderdev: Right now, no. But once part 2 is done, the goal is to be able to have different backends for cluster state and config (eg job) state. But maybe we expose that in the second PR so it is not so confusing.

ballista/core/src/serde/generated/ballista.rs
thinkharderdev: No, not sure why it was removed in the generated code.

ballista/scheduler/scheduler_config_spec.toml
thinkharderdev: @yahoNanJing are you okay to merge this as is?

ballista/core/src/serde/generated/ballista.rs
avantgardnerio: Can you remind me again why the generated code is checked in? Where did we land on that?

avantgardnerio (approved):
I'm not as familiar with the overall Ballista architecture as I'd like to be, but this PR still makes a lot of sense.

I _love_ seeing this stuff come out of the `executor_manager` which seemed like it didn't have a clear boundary on what it's role was and go into a more self-contained `cluster.rs`.

I like that it encapsulates all the HA session management stuff, because I think we'll need options like Redis and others in different enterprise environments.

The only thing that makes me sad is seeing 2 values (state + backend) get passed around everywhere instead of 1. That's really not the fault of this PR however, as mentioned in my other comment, I think we need to clean the plumbing up around config in general.

ballista/core/src/serde/generated/ballista.rs
thinkharderdev: I think it was so the project played nicely with IDEs (some of which do not evaluate proc macros)

thinkharderdev:
I'll plan on merging this tomorrow morning

PR #576, opened 2022-12-22
Use lz4 for shuffle compression

Dandandan:

Closes #.

 # Rationale for this change

### DIFF
--- a/ballista/executor/src/flight_service.rs
+++ b/ballista/executor/src/flight_service.rs
@@ -21,6 +21,7 @@ use std::convert::TryFrom;
 use std::fs::File;
 use std::pin::Pin;
 
+use arrow::ipc::CompressionType;
 use arrow_flight::SchemaAsIpc;
 use ballista_core::error::BallistaError;
 use ballista_core::serde::decode_protobuf;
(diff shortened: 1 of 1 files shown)


andygrove:
LGTM but I am curious @Dandandan if you observed any performance improvement from this change?

andygrove (approved):

yahoNanJing:
Hi @Dandandan, could we make this feature optional? 

And I think the benefits of compression will be reflected at two aspects. The network throughput will be exactly much higher than before. While for the overall time costing, the  transforming speed will be faster. However, it may also cost more CPUs for compression and decompression. So whether the overall total time cost can be reduced may depend on the size of the shuffle data. How about making the decision depending on some size of the shuffling data?

thinkharderdev (approved):

Dandandan:
Hey @yahoNanJing - you're right, it's a trade-off between CPU and bandwidth memory.

Ideally we also need to do this only once (in the shuffle writer), and only need to stream the files from disk instead of re-encoding (and recompressing) them to IPC. This would both save network and disk IO (write+read) while paying the CPU overhead only once! 
LZ4 is much more light on the CPU than ZSTD, so would be great if we could configure this too!

yahoNanJing:
> LZ4 is much more light on the CPU than ZSTD, so would be great if we could configure this too!

I think by default, the compression is not enabled, right? So at least, it's better for us to add a config to indicate whether to enable the compression feature.

PR #614, opened 2023-01-18
Refactor executor main

andygrove:

Part of https://github.com/apache/arrow-ballista/issues/613

 # Rationale for this change

Make it easier to build custom versions of Ballista with extensions and to wrap the executor in Python.

Refactor existing code:

- Add new `ExecutorProcessConfig` struct
- Move logic from `main.rs` to `executor_process.rs`
- Move `main.rs` out of lib and into `src/bin/main.rs`

No

### DIFF
--- a/ballista/client/Cargo.toml
+++ b/ballista/client/Cargo.toml
@@ -41,9 +41,9 @@ tempfile = "3"
 tokio = "1.0"
 
 [features]
+azure = ["ballista-core/azure"]
 default = []
 hdfs = ["ballista-core/hdfs"]
 hdfs3 = ["ballista-core/hdfs3"]
 s3 = ["ballista-core/s3"]
-azure = ["ballista-core/azure"]
 standalone = ["ballista-executor", "ballista-scheduler"]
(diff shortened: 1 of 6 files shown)


andygrove:
@avantgardnerio You have experience in building custom executors that extend Ballista. Is this approach reasonable as a starting point to simplify doing this?

@jdye64 Would be good to get your thoughts too

jdye64 (approved):
I like this. Moving the `start_executor_process(...)` to executor_process.rs makes it more clean to delegate the starting of the executor from a custom driver program.

avantgardnerio (approved):

avantgardnerio:
> Is this approach reasonable as a starting point to simplify doing this?

Yes, I definitely think we should make this change.

That being said, it can be painful extending the dependency graph to this many layers deep:

```
                         +-> delta-rs -+
                         |             |
                         |             |
arrow-rs ----> datafusion -------------+--> ballista --> custom-app
```

So it's my hope at S&T that we can get as much functionality merged upstream as possible to minimize TTL on arrow-rs changes.

PR #615, opened 2023-01-18
Refactor scheduler main

andygrove:

Part of https://github.com/apache/arrow-ballista/issues/613

 # Rationale for this change

Make it easier to build custom versions of Ballista with extensions and to wrap the scheduler in Python.

- Move `main.rs` to `src/bin/main.rs`
- Move some logic from `main.rs` into `scheduler_process.rs` making it available for re-use

No

### DIFF
--- a/ballista/client/Cargo.toml
+++ b/ballista/client/Cargo.toml
@@ -41,9 +41,9 @@ tempfile = "3"
 tokio = "1.0"
 
 [features]
+azure = ["ballista-core/azure"]
 default = []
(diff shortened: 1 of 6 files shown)


andygrove:
cc @jdye64

jdye64 (approved):
LGTM, similar to the executor refactor this makes it much easier to write a custom driver implementation that registers certain items and configs the scheduler in the desired manner before delegating the actual starting of the scheduler server to the new `scheduler_process.rs` file.

avantgardnerio (approved):

PR #623, opened 2023-01-22
Implement new release process

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/622

 # Rationale for this change

Development of Ballista has been lagging behind DataFusion because we have to wait four weeks to get the next DataFusion release before we can update this repo.

I would like to change our release process and branch policy so that we can keep the main branch up-to-date with the latest DataFusion using GitHub dependencies and then create release branches when we are ready to create a new release.

New documentation

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -19,22 +19,31 @@
 
 # Release Process
 
-## Branching
+Development happens on the `main` branch, and most of the time, we depend on DataFusion using GitHub dependencies
+rather than using an official release from crates.io. This allows us to pick up new features and bug fixes frequently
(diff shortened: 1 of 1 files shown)


andygrove:
@thinkharderdev @Dandandan @yahoNanJing @avantgardnerio Please review and let me know if you are ok for me to implement this.

thinkharderdev (approved):
Makes sense to me. It will definitely be helpful to be able to to keep up to date with DF between releases

avantgardnerio (approved):
This looks great! I think it will be a big step forward. I think we'll probably want to release Ballista the instant DataFusion is released though, lest someone PR Ballista to a post release DataFusion commit and then we need to undo functionality on the release branch.

PR #649, opened 2023-02-02
Add executor self-registration mechanism in the heartbeat service

### LINKED ISSUE #648: Add executor self-registration mechanism in the heartbeat service

Currently when a scheduler restarts, we need to restart all of the executors, which is not so convenient.

 Add executor self-registration mechanism in the heartbeat service so that executors can register itself to the schedulers automatically through its heartbeat. 

**Additional context**

yahoNanJing:

Closes #648.

 # Rationale for this change


thinkharderdev (approved):
LGTM

PR #659, opened 2023-02-08
Minor refactor to reduce duplicate code

andygrove:

N/A

 # Rationale for this change

Drive by cleanup.

Extract method


ballista/scheduler/src/planner.rs
```diff
@@ -16,8 +16,6 @@
 // under the License.
 
 //! Distributed query execution
-//!
-//! This code is EXPERIMENTAL and still under development
```
thinkharderdev, line 20: I guess we're official now :)

thinkharderdev (approved):

Dandandan (approved):

PR #664, opened 2023-02-11
Upgrade to DataFusion 18.0.0-rc1

andygrove:

Closes #.

 # Rationale for this change

### DIFF
--- a/ballista/scheduler/src/planner.rs
+++ b/ballista/scheduler/src/planner.rs
@@ -326,17 +326,15 @@ mod test {
     use datafusion::physical_plan::aggregates::{AggregateExec, AggregateMode};
     use datafusion::physical_plan::coalesce_batches::CoalesceBatchesExec;
     use datafusion::physical_plan::joins::HashJoinExec;
+    use datafusion::physical_plan::projection::ProjectionExec;
     use datafusion::physical_plan::sorts::sort::SortExec;
-    use datafusion::physical_plan::{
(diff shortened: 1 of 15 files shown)


ballista/scheduler/src/planner.rs
```diff
@@ -375,19 +373,19 @@ mod test {
...
+              AggregateExec: mode=FinalPartitioned, gby=[l_returnflag@0 as l_returnflag], aggr=[SUM(lineitem.l_extendedprice * Int64(1))]
+                CoalesceBatchesExec: target_batch_size=8192
+                  UnresolvedShuffleExec
 
         ShuffleWriterExec: None
-          SortExec: [l_returnflag@0 ASC]
-            CoalescePartitionsExec
-              UnresolvedShuffleExec
+          SortPreservingMergeExec: [l_returnflag@0 ASC NULLS LAST]
+            UnresolvedShuffleExec
```
andygrove, line 388: This query just go more efficient and sorts data in parallel now :rocket:

Dandandan: This depends on how many partitions it sorts into.
As `SortPreservingMergeExec` is itself single threaded, after a certain number of partitions it will be faster to use `SortExec`.

Dandandan (approved):

ballista/scheduler/src/planner.rs
andygrove: Sounds like we need to start customizing the optimization rules we use in Ballista

PR #666, opened 2023-02-11
Enable physical plan round-trip tests

andygrove:

Closes #.

 # Rationale for this change

We have physical plan serde tests but previously they only ran if the `TPCH_DATA` env var was set, which does not make much sense.

Always run these tests.

### DIFF
--- a/benchmarks/src/bin/tpch.rs
+++ b/benchmarks/src/bin/tpch.rs
@@ -1671,8 +1671,8 @@ mod tests {
                     )
                     .unwrap();
                 assert_eq!(
-                    format!("{physical_plan:?}"),
-                    format!("{round_trip:?}"),
+                    format!("{}", displayable(physical_plan.as_ref()).indent()),
+                    format!("{}", displayable(round_trip.as_ref()).indent()),
                     "physical plan round trip failed"
                 );
             }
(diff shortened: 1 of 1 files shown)


benchmarks/src/bin/tpch.rs
```diff
@@ -1611,65 +1611,137 @@ mod tests {
...
-                    let round_trip: Arc<dyn ExecutionPlan> = proto
-                        .try_into_physical_plan(
-                            &ctx,
-                            runtime.deref(),
-                            codec.physical_extension_codec(),
-                        )
-                        .unwrap();
-                    assert_eq!(
-                        format!("{physical_plan:?}"),
-                        format!("{round_trip:?}"),
```
andygrove, line 1634: debug output is non-deterministic (alias maps print in different order) so changed to use `displayable` instead

avantgardnerio: Makes sense...

benchmarks/src/bin/tpch.rs
```diff
@@ -1611,65 +1611,137 @@ mod tests {
...
+                let provider = ListingTable::try_new(config)?;
+                ctx.register_table(table, Arc::new(provider))?;
+            }
+
+            // test logical plan round trip
+            let plans = create_logical_plans(&ctx, n).await?;
+            for plan in plans {
+                let plan = session_state.optimize(&plan)?;
 
                 // test physical plan roundtrip
```
avantgardnerio, line 1657: Seems straightforward

avantgardnerio (approved):

PR #667, opened 2023-02-12
Add executor terminating status for graceful shutdown

### LINKED ISSUE #655: Issues with lost executor handling

While working on the cluster state refactor and observing some issues with handling of lost executors in our system, I think there are a couple of issues with the current approach:

1. In multi-scheduler deployments, an executor will register itself and send heartbeats with a particular scheduler.
2. Any scheduler can scheduler tasks on the executor
3. However, in the case of a `SIGTERM` the executor will call `executor_stopped` on the scheduler from 1. 

This doesn't work particularly well with the curated task architecture since the scheduler from step 1 may not be the owner of the jobs which the executor is running when it gets the `SIGTERM`. At best this may cause the lost executor handling to not work correctly (or be delayed since the owning scheduler has to wait for a timeout). At worst it could corrupt the job state and cause unpredictable errors in the scheduler.   

More broadly, resetting tasks immediately when an executor gets a `SIGTERM` is a bit limiting. The executor may have anywhere from 30s - 2 minutes (or maybe more) to cleanup and finish any existing work after receiving a `SIGTERM` and ideally we could take advantage of that to minimize disruptions. 

We have an existing mechanism to broadcast executor state to all schedulers using the executor heartbeat. These will also go to a particular scheduler but can be broadcast fairly easily through `ClusterState`. So I would propose the following:

1. Remove the `executor_stopped` rpc.
2. Instead of sending the `executor_stopped` rpc on `SIGTERM` the executor can just send a heartbeat where it's state changes from `Active` to `Dead` (or maybe something more descriptive like `ShuttingDown`).
3. When receiving that heartbeat the scheduler can put that executor in a "quarantine" state which means:
   a. Stop scheduling new tasks on it
   b. Wait for some (configurable) interval before resetting tasks

This can ensure that we handle job updates on the appropriate scheduler and that we can attempt to finish outstanding work before potentially re-computing tasks/stages. 

The executor could maybe track all schedulers for which it has active tasks and send an `executor_stopped` rpc to all of them but the rpc itself seems somewhat duplicative of the existing heartbeat mechanism

**Additional context**

thinkharderdev:

Closes #655 

 # Rationale for this change

See #655 issues description. 

1. Add a new executor status `Terminating` to indicate that an executor has started the termination process. 
2. On receiving a `SIGTERM` the executor will immediately send a heartbeat to update it's status to `Terminating`
3. After receiving a `SIGTERM` the executor should attempt to complete any tasks currently running before terminating
4. The scheduler should consider executors in `Active` status to be live (i.e. eligible to schedule tasks on)
5. When receiving an `executor_stopped` rpc, the scheduler will wait a configurable amount of time before removing the executor. This can be set through the new `--executor-termination-grace-period` cli flag and has a default value of 30s. 
6. When periodically checking for dead executors, the scheduler should also remove any executors in `Terminating` status for longer than `executor_termination_grace_period`.

All together this tries to ensure:
1. When an executor is terminated it has a chance to finish any in-flight tasks to minimize disruption
2. The termination of an executor will be propagated to all schedulers in the cluster through the heartbeat mechanism
3. The scheduler will give the terminating executor a grace period to complete it's tasks before resetting tasks/stages and potentially recomputing results. 

A new cli flag `--executor-termination-grace-period` for the scheduler with a default of 30 seconds.

### DIFF
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -18,6 +18,7 @@
 //! Ballista Executor Process
 
 use std::net::SocketAddr;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 use std::{env, io};
@@ -41,18 +42,21 @@ use datafusion_proto::protobuf::{LogicalPlanNode, PhysicalPlanNode};
 
 use ballista_core::config::{LogRotationPolicy, TaskSchedulingPolicy};
 use ballista_core::error::BallistaError;
+use ballista_core::serde::protobuf::executor_resource::Resource;
+use ballista_core::serde::protobuf::executor_status::Status;
 use ballista_core::serde::protobuf::{
     executor_registration, scheduler_grpc_client::SchedulerGrpcClient,
-    ExecutorRegistration, ExecutorStoppedParams,
+    ExecutorRegistration, ExecutorResource, ExecutorSpecification, ExecutorStatus,
+    ExecutorStoppedParams, HeartBeatParams,
 };
-use ballista_core::serde::scheduler::ExecutorSpecification;
 use ballista_core::serde::BallistaCodec;
 use ballista_core::utils::{
     create_grpc_client_connection, create_grpc_server, with_object_store_provider,
 };
 use ballista_core::BALLISTA_VERSION;
 
-use crate::executor::Executor;
+use crate::executor::{Executor, TasksDrainedFuture};
+use crate::executor_server::TERMINATING;
 use crate::flight_service::BallistaFlightService;
 use crate::metrics::LoggingMetricsCollector;
 use crate::shutdown::Shutdown;
@@ -155,12 +159,11 @@ pub async fn start_executor_process(opt: ExecutorProcessConfig) -> Result<()> {
             .map(executor_registration::OptionalHost::Host),
         port: opt.port as u32,
         grpc_port: opt.grpc_port as u32,
-        specification: Some(
-            ExecutorSpecification {
-                task_slots: concurrent_tasks as u32,
-            }
-            .into(),
-        ),
+        specification: Some(ExecutorSpecification {
+            resources: vec![ExecutorResource {
+                resource: Some(Resource::TaskSlots(concurrent_tasks as u32)),
+            }],
+        }),
     };
 
     let config = with_object_store_provider(
@@ -295,6 +298,8 @@ pub async fn start_executor_process(opt: ExecutorProcessConfig) -> Result<()> {
         shutdown_noti.subscribe_for_shutdown(),
     )));
 
+    let tasks_drained = TasksDrainedFuture(executor);
+
     // Concurrently run the service checking and listen for the `shutdown` signal and wait for the stop request coming.
     // The check_services runs until an error is encountered, so under normal circumstances, this `select!` statement runs
     // until the `shutdown` signal is received or a stop request is coming.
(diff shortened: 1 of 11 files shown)


ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -85,6 +85,10 @@ impl ExecutorReservation {
 /// to be dead.
 pub const DEFAULT_EXECUTOR_TIMEOUT_SECONDS: u64 = 180;
 
+// TODO move to configuration file
+/// Interval check for expired or dead executors
+pub const EXPIRE_DEAD_EXECUTOR_INTERVAL_SECS: u64 = 15;
```
thinkharderdev, line 90: This is changed to poll for dead executors more frequently

ballista/executor/src/executor_process.rs
```diff
@@ -319,7 +324,41 @@ pub async fn start_executor_process(opt: ExecutorProcessConfig) -> Result<()> {
         },
     };
 
+    // Set status to fenced
+    info!("setting executor to TERMINATING status");
+    TERMINATING.store(true, Ordering::Release);
+
     if notify_scheduler {
+        // Send a heartbeat to update status of executor to `Fenced`. This should signal to the
+        // scheduler to no longer scheduler tasks on this executor
```
avantgardnerio, line 333:
```suggestion
        // scheduler to no longer schedule tasks on this executor
```

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -760,15 +763,22 @@ mod test {
...
             .expect("getting executor");
 
+        let is_stopped = await_condition(Duration::from_millis(10), 5, || {
+            futures::future::ready(Ok(state.executor_manager.is_dead_executor("abc")))
+        })
+        .await?;
+
         // executor should be marked to dead
-        assert!(state.executor_manager.is_dead_executor("abc"));
+        assert!(is_stopped, "Executor not marked dead after 50ms");
```
avantgardnerio, line 772: Nice test

ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -216,83 +220,113 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T
...
+
+                    let terminating = matches!(
+                        expired
+                            .status
+                            .as_ref()
+                            .and_then(|status| status.status.as_ref()),
+                        Some(ballista_core::serde::protobuf::executor_status::Status::Terminating(_))
+                    );
+
+                    let stop_reason = if terminating {
```
avantgardnerio, line 243: Seems logical...

ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -216,83 +220,113 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T
...
         reason: Option<String>,
-    ) -> Result<()> {
-        // Update the executor manager immediately here
-        executor_manager
-            .remove_executor(executor_id, reason.clone())
-            .await?;
+        wait_secs: u64,
+    ) {
+        let executor_id = executor_id.to_owned();
+        tokio::spawn(async move {
```
avantgardnerio, line 311: I was worried that spawning every time without backpressure from blocking or an idempotency check could cause a storm of `ExecutorLost` events, but it seems like this is handled by:

```
self.cluster_state.remove_executor(executor_id).await?;
...
self.executors_heartbeat.remove(&executor_id);
...
self.dead_executors.insert(executor_id);
```

?

It's not always easy to get a picture of the mental swim lane diagram of how all these pieces of state, messages, and events interact.

avantgardnerio (approved):
LGTM

mingmwang:
Sorry for late response,  I will take a look at your PR later.   I like the design and introduce a new `Terminating` status.

ballista/executor/src/executor_process.rs
```diff
@@ -319,7 +324,41 @@ pub async fn start_executor_process(opt: ExecutorProcessConfig) -> Result<()> {
...
+            })
+            .await
+        {
+            error!("error sending heartbeat with fenced status: {:?}", error);
+        }
+
+        // TODO we probably don't need a separate rpc call for this....
         if let Err(error) = scheduler
             .executor_stopped(ExecutorStoppedParams {
                 executor_id,
```
mingmwang, line 364: I think you can remove this RPC call (in the following PR), since this RPC is only used here.

ballista/scheduler/src/scheduler_server/mod.rs
mingmwang: I think if it the `wait_secs` = 0, there is no need to spawn.

ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -614,27 +623,62 @@ impl ExecutorManager {
...
+                    heartbeat
+                        .status
+                        .as_ref()
+                        .and_then(|status| status.status.as_ref()),
+                    Some(executor_status::Status::Active(_))
+                );
+                let live = heartbeat.timestamp > last_seen_ts_threshold;
+
+                (active && live).then(|| exec.clone())
             })
```
mingmwang, line 637: This method's comment need to be changed to reflect the new logic.
 
```rust
/// Retrieve the set of all executor IDs where the executor has been observed in the last
/// `last_seen_ts_threshold` seconds.
pub(crate) fn get_alive_executors(
        &self,
        last_seen_ts_threshold: u64,
    ) 
```

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -521,13 +522,14 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
+
+        Self::remove_executor(
+            executor_manager,
+            event_sender,
+            &executor_id,
+            Some(reason),
+            self.executor_termination_grace_period,
+        );
 
         Ok(Response::new(ExecutorStoppedResult {}))
```
mingmwang, line 534: Maybe we should rename the method name to `remove_executor_in_secs()`

mingmwang:
Nice improvement!

thinkharderdev:
Thanks @mingmwang will create a follow-up PR for your comments

PR #687, opened 2023-02-26
Add ExecutionEngine abstraction

andygrove:

N/A

 # Rationale for this change

Add an abstraction for shuffle writer as a step towards supporting pluggable execution engines. See https://github.com/apache/arrow-ballista-python/pull/19 for more context on how I intend to use this.

- Add `ExecutionEngine` and `QueryStageExecutor` traits
- Update Executor to delegate to the new traits

No

### DIFF
--- a/ballista/core/src/execution_plans/shuffle_writer.rs
+++ b/ballista/core/src/execution_plans/shuffle_writer.rs
@@ -22,7 +22,6 @@
 
 use datafusion::physical_plan::expressions::PhysicalSortExpr;
 
-use async_trait::async_trait;
 use std::any::Any;
 use std::fmt::Debug;
 use std::future::Future;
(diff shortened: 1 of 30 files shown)


ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -139,7 +154,7 @@ impl ShuffleWriterExec {
         self.shuffle_output_partitioning.as_ref()
     }
 
-    pub fn execute_shuffle_write(
+    fn execute_shuffle_write_internal(
```
andygrove, line 157: @tustvold You changed the signature of this method in https://github.com/apache/arrow-ballista/commit/b7bb2cfba13cc04a08c2f687102dd14a8dedc7b6 to change it from `async` to return an `impl Future`, and I am now calling it from an `async` method so just wanted to check that I am not undoing anything here.

Dandandan (approved):
Trait change looks good (not sure about async question)

ballista/core/src/execution_plans/shuffle_writer.rs
tustvold: So there are two changes here

* [`async_trait`] will box the future, so polling the future now goes through a dyn dispatch. This probably doesn't matter given this method is likely performing non-trivial IO
* The returned future borrows self, i.e. can't outlive the borrow of self. This could matter, but given the pervasive use of `Arc` and `async move` it is probably easy enough to work around

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -288,6 +305,22 @@ impl ShuffleWriterExec {
     }
 }
 
+#[async_trait]
+impl ShuffleWriter for ShuffleWriterExec {
```
thinkharderdev, line 309: This seems inverted to me. Shouldn't `ShuffleWriterExec` have a `Arc<dyn ShuffleWriter>` instead of impl the trait? 

I know this is a step toward the pluggable execution engine, but also thinking about how this fits with allowing for different shuffle methods (write to disk, write to object store, stream to network, etc.) and seems like we would need a second layer of indirection for that.

andygrove: That's a good point. Maybe want I want for now is just an `ExecutionEngine` trait with an `execute` method that returns the metadata with shuffle file locations. I will give this some more thought.

andygrove: @thinkharderdev I have renamed the trait and made some other naming changes. I also added the ExecutionEngine abstraction in this PR. PTAL when you have time.

andygrove:
@Dandandan I know you already approved this PR, but it has now changed significantly. Could you take another look?

Dandandan (approved):
Makes sense

PR #700, opened 2023-03-06
Allow accessing s3 locations in client mode

luckylsk34:
Partially addressess #6 .

Closes #.

 # Rationale for this change
Was able to read data from S3 in standalone mode but in client mode, it gives the error `DataFusion error: Execution("No object store available for s3://path/to/folder/")`

Added `S3` feature to `Scheduler` and `Executor`. Not having this feature in either will again give the above error.

User will be access S3 paths in client mode.


avantgardnerio (approved):
I have no issue including the s3 feature...

andygrove (approved):
Thanks @luckylsk34

PR #720, opened 2023-03-25
Fix parameter '--config-backend' to '--cluster-backend'

### LINKED ISSUE #719: Starting scheduler with docker-compose on `main` throws error: Error: An unknown argument '--config-backend' was specified.
**Describe the bug**
Starting the deployment in the `main` branch with the docker-compose option ttriggers this error:

```
arrow-ballista-ballista-scheduler-1  | Starting nginx to serve Ballista Scheduler web UI on port 80
arrow-ballista-ballista-scheduler-1  | Error: An unknown argument '--config-backend' was specified.
```

**To Reproduce**
Execute `docker-compose up --build` from main branch.

**Expected behavior**
Expected services to execute successfully.

**Additional context**
When looking at the source code, specifically `ballista/scheduler/src/bin/main.rs` I could not find the relevant `etcd` options.

Changing the `docker-compose.yml` line 29 from

```
    command: "--config-backend etcd --etcd-urls etcd:2379 --bind-host 0.0.0.0"
```

To
```
    command: " --bind-host 0.0.0.0"
```

Fixes the error and deploys successfully the containers. It does not seem like a proper solution here, as we're not effectively using the `etcd` service. Is there an unfinished feature in `main` branch?

paolorechia:

Closes #719 .

 # Rationale for this change
It seems the parameter `config-backend` passed in `docker-compose.yml` is wrong and is now called `cluster-backend`.

The default parameter for the scheduler is now `--cluster-backend etcd`.

No, except that now the `docker-compose.yml` should work with no changes from end users.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
     build:
       dockerfile: dev/docker/ballista-scheduler.Dockerfile
       context: .
-    command: "--config-backend etcd --etcd-urls etcd:2379 --bind-host 0.0.0.0"
+    command: "--cluster-backend etcd --etcd-urls etcd:2379 --bind-host 0.0.0.0"
     ports:
       - "80:80"
       - "50050:50050"


r4ntix (approved):

yahoNanJing (approved):

yahoNanJing:
Thanks @paolorechia paolorechia

PR #727, opened 2023-03-28
Upgrade DataFusion to 21.0.0

r4ntix:

Upgrade DataFusion to 21.0.0

 # Rationale for this change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,22 +16,14 @@
 # under the License.
 
 [workspace]
-members = [
-    "ballista-cli",
-    "ballista/client",
-    "ballista/core",
-    "ballista/executor",
-    "ballista/scheduler",
-    "benchmarks",
-    "examples",
-]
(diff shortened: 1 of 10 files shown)


Dandandan (approved):

yahoNanJing:
Thanks @r4ntix. How about extracting the feature based store into a single method, like the following does 
https://github.com/yahoNanJing/arrow-ballista/blob/84a98d73a164b655b80f4391611de4c499b27f73/ballista/core/src/utils.rs#L105-L179?

r4ntix:
> Thanks @r4ntix. How about extracting the feature based store into a single method, like the following does https://github.com/yahoNanJing/arrow-ballista/blob/84a98d73a164b655b80f4391611de4c499b27f73/ballista/core/src/utils.rs#L105-L179? Then we can separate the **_get_** and **_register_** logic.

@yahoNanJing thanks for the feedback. i will submit a new commit for it a little later.

yahoNanJing (approved):
LGTM

PR #732, opened 2023-04-02
tpch conversion is failling with ArrowError(CsvError)

zhzy0077:

This is - to me - a utility change, so I don't create a bug.
Closes #.

 # Rationale for this change

benchmarks/src/bin/tpch.rs is failing to convert TPC-H's tbl file to parquet file.
because there's a tailing '|' in the tbl file like this:
```
1|goldenrod lavender spring chocolate lace|Manufacturer#1|Brand#13|PROMO BURNISHED COPPER|7|JUMBO PKG|901.00|ly. slyly ironi|
```
So Arrow believes the schema doesn't match CSV itself and throws `Error: ArrowError(CsvError("incorrect number of fields for line 1, expected 9 got more than 9"))`

Adds an null-able trailing field to the schema so CSV parsing won't fail.

No

### DIFF
--- a/benchmarks/src/bin/tpch.rs
+++ b/benchmarks/src/bin/tpch.rs
@@ -874,6 +874,7 @@ fn get_schema(table: &str) -> Schema {
             Field::new("p_container", DataType::Utf8, false),
             Field::new("p_retailprice", DataType::Decimal128(15, 2), false),
             Field::new("p_comment", DataType::Utf8, false),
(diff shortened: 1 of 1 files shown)


XiangpengHao:
Strange that the same benchmark in DataFusion works fine. https://github.com/apache/arrow-datafusion/blob/main/benchmarks/src/tpch.rs#L69

andygrove (approved):

PR #735, opened 2023-04-03
Support graceful shutdown for scheduler.

### LINKED ISSUE #565: Scheduler graceful shutdown

We currently have graceful shutdown for the Ballista executors. We should have a similar behavior for the scheduler. 

When the scheduler process receives a SIGTERM (or the equivalent on non-Linux), it should initiate a graceful shutdown. This would consist of:

1. Stop accepting new jobs. If a new job is submitted return an error indicating the scheduler is shutting down. 
2. Continue to accept job status requests
3. Do not shutdown until either
   i. All active jobs on the scheduler are completed
   ii. Some configured timeout (default 30s?) elapses

This could be extended in the future to try and "hand off" scheduling to other schedulers but that is not necessary for now. 

We could do nothing

**Additional context**

zhzy0077:

Closes #565.

 # Rationale for this change

Setting the terminating flag when receiving terminate signals and shutdown the server gracefully.  

No.

N.B. I'm really a newbie to Rust. Appreciate in advance for your reviewing the PR and raising comments.


mingmwang:
@liukun4515

PR #750, opened 2023-04-21
[log] Print long running task status.

### LINKED ISSUE #749: [LOG] Print long running task status.
I use a `atomicU64` as a counter to calculate all events cost time in while loop

https://github.com/Ted-Jiang/arrow-ballista/blob/0b6b28f8d2bdeca8f2d66e5453008dca11974dae/ballista/core/src/event_loop.rs#L78-L88

I run 5 mins load test got `2023-04-14T07:52:45.743867Z  WARN tokio-runtime-worker ThreadId(16) ballista_scheduler::scheduler_server::query_stage_scheduler: [METRICS] total event cost 278 955 949 us`

So the events processing is the hot path and bottleneck in scheduler. we need log the long running event status.

Add a config `scheduler_event_expected_processing_duration` if event processing time more than this warn in log.

**Additional context**

Ted-Jiang:

Closes #749.

 # Rationale for this change

### DIFF
--- a/ballista/scheduler/scheduler_config_spec.toml
+++ b/ballista/scheduler/scheduler_config_spec.toml
@@ -140,10 +140,16 @@ default = "ballista_core::config::LogRotationPolicy::Daily"
 name = "job_resubmit_interval_ms"
 type = "u64"
 default = "0"
-doc = "If job is not able to be scheduled on submission, wait for this interval and resubmit. Default value of 0 indicates that job shuuld not be resubmitted"
+doc = "If job is not able to be scheduled on submission, wait for this interval and resubmit. Default value of 0 indicates that job should not be resubmitted"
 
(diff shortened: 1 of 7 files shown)


avantgardnerio (commented):
Personally, I've found metrics endpoints like the Prometheus one we have in the scheduler to generally be a more robust solution to this problem as I understand it. Would that be an alternative solution for the case you are trying to cover?

Ted-Jiang:
| Would that be an alternative solution for the case you are trying to cover?
Thanks for remind me this. Yes,  For user like us, us ES to collect metrics, directly print it out is enough. btw this feature is default off.

yahoNanJing (approved):
LGTM

r4ntix (approved):

PR #794, opened 2023-05-29
Upgrade node version to fix build in `main`

avantgardnerio:

Fixes the build.

 # Rationale for this change

We want the build to work.

Node version upgrade in github actions

No


avantgardnerio:
I guess I'll get a YOLO badge for this, but I think this is worth merging to fix main and not gum up the works for other PRs.

PR #743, opened 2023-04-12
Only decode plan in `LaunchMultiTaskParams`  once

Dandandan:
# Which issue does this PR close?

Closes #742

 # Rationale for this change

When starting many tasks on single executors, the execution plan is decoded for every task/partition.
This can take quite a bit of time for larger plans, delaying execution and consumes quite some memory as well for executors that start a lot of tasks for the same query stage at the same time.

# What changes are included in this PR?

Decode the plan only once and share the `Arc<dyn ExecutionPlan>`.

# Are there any user-facing changes?

PR #817, opened 2023-06-26
Refactor the TaskDefinition by changing encoding execution plan to the decoded one

### LINKED ISSUE #793: The decoded execution plan at the executor side cannot be directly shared.
**Describe the bug**

It's caused by the #743. Only unchanged things can be shared. However, the execution plan is not. Its metrics will be changed during task execution. Since the decoded execution plan is directly shared by multiple tasks, the plan metrics is also shared across multiple tasks, which is not correct.

**To Reproduce**

**Expected behavior**

**Additional context**

yahoNanJing:

Closes #793.

 # Rationale for this change

I think it's better for the plan decoding process inside the grpc interface so that we can detect failure as quickly as possible. And it's kind of meaningless to send a task with an illegal plan to the channel for future processing.

1. The patch of #743 is reverted
2. The encoded execution plan in the `TaskDefinition` is changed to the decoded one.
3. The error handling of run_task in the executor_server is refined

### DIFF
--- a/ballista/core/src/serde/scheduler/from_proto.rs
+++ b/ballista/core/src/serde/scheduler/from_proto.rs
@@ -16,10 +16,15 @@
 // under the License.
 
 use chrono::{TimeZone, Utc};
+use datafusion::common::tree_node::{Transformed, TreeNode};
+use datafusion::execution::runtime_env::RuntimeEnv;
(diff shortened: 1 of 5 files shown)


yahoNanJing:
Hi @thinkharderdev and @Dandandan, could you help review this PR?

PR #842, opened 2023-07-11
Reduce the number of calls to `create_logical_plan`

jonahgao:

None

 # Rationale for this change
There is a duplicate invocation of `create_logical_plan` within `ctx.sql(sql)`.
Since we already have a logical plan at hand, perhaps we can execute it directly.

No


yahoNanJing:
Good catch. Thanks @jonahgao.

PR #877, opened 2023-09-08
refactor: port get_scan_files to Ballista

alamb:

Related to https://github.com/apache/arrow-datafusion/issues/7357

 # Rationale for this change

We are sorting out how to split up DataFusion into smaller modules, and we hit https://github.com/apache/arrow-datafusion/issues/7357

There is a tension between knowing about datasources and separating out physical plan and datasources (see commentary on https://github.com/apache/arrow-datafusion/issues/7357 and linked PRs). 

We thus propose to move this API upstream into Ballista. See  https://github.com/apache/arrow-datafusion/pull/7487 

1. Move this API upstream into Ballista:  https://github.com/apache/arrow-datafusion/pull/7487 

No

### DIFF
--- a/ballista/scheduler/src/cluster/kv.rs
+++ b/ballista/scheduler/src/cluster/kv.rs
@@ -17,7 +17,7 @@
 
 use crate::cluster::storage::{KeyValueStore, Keyspace, Lock, Operation, WatchEvent};
 use crate::cluster::{
-    bind_task_bias, bind_task_consistent_hash, bind_task_round_robin,
+    bind_task_bias, bind_task_consistent_hash, bind_task_round_robin, get_scan_files,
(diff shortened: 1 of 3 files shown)


ballista/scheduler/src/cluster/kv.rs
```diff
@@ -39,7 +39,6 @@ use ballista_core::serde::protobuf::{
 use ballista_core::serde::scheduler::{ExecutorData, ExecutorMetadata};
 use ballista_core::serde::BallistaCodec;
 use dashmap::DashMap;
-use datafusion::datasource::physical_plan::get_scan_files;
```
alamb, line 42: The point of this PR is to remove this dependency (so we can remove it from DataFusion)

Dandandan (approved):

alamb:
Thanks @Dandandan

PR #896, opened 2023-10-20
Update base64 lib and fix compilation failure in flight_sql.rs

ahmedriza:

Closes https://github.com/apache/arrow-ballista/issues/895

Update `base64` lib to version 0.21 and fix compilation error in `flight_sql.rs`

No


ahmedriza:
Updated to_proto.rs where there were build failures due to deprecations from the use of timestamp_nanos() which is now deprecated in chrono crate.

Replaced with timestamp_nanos_opt()

PR #945, opened 2023-12-22
Fix regression in DataFrame.write_xxx

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/894

Status

- [x] Add failing test
- [x] Upgrade to DataFusion version that supports CopyTo serde (https://github.com/apache/arrow-datafusion/pull/8618)
- [x] Upgrade to DataFusion version that supports CopyTo serde with support for WriterProperties (https://github.com/apache/arrow-datafusion/pull/8627 and https://github.com/apache/arrow-datafusion/pull/8641)
- [x] Upgrade to DataFusion version that supports serde for Csv and Parquet FileSinkExec (https://github.com/apache/arrow-datafusion/pull/8646)
- [x] Test passes
- [x] Upgrade to DataFusion main branch version
- [x] Clippy fixes for latest Rust version

 # Rationale for this change

Fix a major regression where we can no longer save the results of a query. 

- Upgrade to more recent DataFusion
- Add test

### DIFF
--- a/ballista/cache/src/listener/loading_cache.rs
+++ b/ballista/cache/src/listener/loading_cache.rs
@@ -145,7 +145,7 @@ where
     fn listen_on_get_if_present(&self, k: Self::K, v: Option<Self::V>) {
         if self.listeners.len() == 1 {
             self.listeners
-                .get(0)
+                .first()
                 .unwrap()
                 .listen_on_get_if_present(k, v);
         } else {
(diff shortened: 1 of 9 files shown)


ballista/cache/src/listener/loading_cache.rs
```diff
@@ -145,7 +145,7 @@ where
     fn listen_on_get_if_present(&self, k: Self::K, v: Option<Self::V>) {
         if self.listeners.len() == 1 {
             self.listeners
-                .get(0)
+                .first()
```
andygrove, line 148: Clippy

Dandandan (approved):

ballista/client/src/context.rs
```diff
@@ -462,12 +462,15 @@ impl BallistaContext {
 }
 
 #[cfg(test)]
-mod tests {
-    #[cfg(feature = "standalone")]
+#[cfg(feature = "standalone")]
+mod standalone_tests {
```
andygrove, line 466: Drive-by cleanup to declare that the whole module requires the `standalone` feature rather than declare this on each import and each test.

PR #956, opened 2024-01-16
Prepare for 0.12.0-rc2

andygrove:

Closes https://github.com/apache/arrow-ballista/issues/926

 # Rationale for this change

0.12.0-rc1 failed because we are pinned to a non-release version of DataFusion. We will need to wait for the imminent 35.0.0 release. 

Pin to newer version. I will update this PR once DF 35.0.0 is released.


ballista/core/Cargo.toml
```diff
@@ -41,7 +41,6 @@ force_hash_collisions = ["datafusion/force_hash_collisions"]
 hdfs = ["datafusion-objectstore-hdfs/hdfs"]
 hdfs3 = ["datafusion-objectstore-hdfs/hdfs3"]
 s3 = ["object_store/aws"]
-simd = ["datafusion/simd"]
```
andygrove, line 44: DataFusion no longer has a simd feature

andygrove:
@avantgardnerio @Dandandan I plan on merging this when CI passes and will then cut a new RC

ballista/core/Cargo.toml
avantgardnerio: What happened?

andygrove: simd feature was removed from arrow-rs in:

```
commit d41e90e31d07bc2ad2f05ffc091f171e6c846ddf
Author: Raphael Taylor-Davies <1781103+tustvold@users.noreply.github.com>
Date:   Fri Dec 8 08:49:46 2023 +0000

    Remove SIMD Feature (#5184)

    * Remove SIMD feature

    * Clippy

    * Tweak WASM features

    * Fix test

    * Exclude pyarrow from general build

```

There are commits prior to this with statements such as:

```
    * Refactor numeric aggregation kernels to make better use of
    auto-vectorization.

    Remove the explicit simd implementations since the autovectorized
    versions are faster on average.
```

PR #989, opened 2024-02-28
docs: list protoc dependency

davidwilemski:
# Rationale for this change

While following the [quick start] guide, I ran into a compilation error when prost tried to build the proto files. Installing the protobuf compiler allowed me to follow the rest of the quick start guide.

I also noticed the manual setup section of the contributors guide had a similar list of dependencies that was missing the protobuf compiler.

[quick start]: https://arrow.apache.org/ballista/user-guide/deployment/quick-start.html

Since installing the protobuf compiler was sufficient for me to build the project, I thought it made sense to document this so others following the docs would be aware of the dependency.

No -- only documentation changes.


andygrove (approved):
Thanks @davidwilemski

PR #998, opened 2024-04-03
Fix shuffle writer test

Jefffrey:

Closes #.

 # Rationale for this change

Running test locally on clean checkout had failing test suite. Looks like values of `1` and `2` were hashing to same partition causing only 1 partition, where test expects 2. Changing `2` to `3` seems sufficient to ensure it is hashed to separate partition.


andygrove (approved):
Thanks @Jefffrey I saw the same failure locally and this fixes it for me too.

Issue #974, opened 2024-02-09
Ballista 43.0.0 Release


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

We have a milestone tracking items for the next release: https://github.com/apache/datafusion-ballista/milestone/4

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

andrewwebber:
It would bring great confidence to the community and or unsure observers if there was a regular release cadence. Would you have an update regarding this ticket?

milenkovicm:
As balista has tight coupling with datafusion, and we cant really mix and match different versions, would it make sense to have same version as datafusion, like datafusion-python does? it would help users to pick right datafusion version (probably we should re-export datafusion in `ballista::datafusion`)

main reason we can't have same version as datafusion is that ballista is still not as mature as df is, so we should not use major version > 0. Nothing stopping us to use keep minor version in sync with df major version. This would users with decision which versions to choose

in short, proposal is: 

- use `0.43.0` for next release and keep minor version in sync with datafusion major
- use same version as datafusion once ballista get promoted to version 1

wdyt @andygrove ?

andygrove:
I don't have a strong opinion, but if we are going to change the versioning scheme, I would rather do it once than twice. Why don't we just jump straight to 43.0.0 for the next release. This would be consistent with the other subprojects.

milenkovicm:
I have no objections. Other projects put the bar very hight, we might get under pressure 😀

andygrove:
Sure, lets do it. Can you create a PR against `main` to update version numbers and add the changelog?

PR #1011, opened 2024-04-29
feat(scheduler): add csv support for get_file_metadata grpc method

etolbakov:

Closes #.
I was exploring the code and came across the `TODO` about the `csv` file format support. So decided to address it.

 # Rationale for this change
The functionality extension.

- add `csv` / `tlb` file format support
- fix minor typos

No

### DIFF
--- a/ballista/scheduler/src/scheduler_server/grpc.rs
+++ b/ballista/scheduler/src/scheduler_server/grpc.rs
@@ -653,6 +653,8 @@ mod test {
 
     use datafusion_proto::protobuf::LogicalPlanNode;
     use datafusion_proto::protobuf::PhysicalPlanNode;
+    use object_store::path::Path;
+    use tempfile::NamedTempFile;
     use tonic::{Code, Request};
 
     use crate::config::SchedulerConfig;
@@ -858,17 +860,21 @@ mod test {
             );
         scheduler.init().await?;
 
+        let mut test_file = NamedTempFile::new()?;
+        let source_location_str =
+            Path::from(test_file.as_ref().to_str().unwrap()).to_string();
+
         let requests = vec![
             Request::new(GetFileMetadataParams {
-                path: "/tmp/file.parquet".to_string(),
+                path: source_location_str.clone(),
                 file_type: "parquet".to_string(),
             }),
             Request::new(GetFileMetadataParams {
-                path: "/tmp/file.csv".to_string(),
+                path: source_location_str.clone(),
                 file_type: "csv".to_string(),
             }),
             Request::new(GetFileMetadataParams {
-                path: "/tmp/file.tbl".to_string(),
+                path: source_location_str.clone(),
                 file_type: "tbl".to_string(),
             }),
         ];
@@ -891,7 +897,7 @@ mod test {
 
         let request: Request<GetFileMetadataParams> =
             Request::new(GetFileMetadataParams {
-                path: "/tmp/file.avro".to_string(),
+                path: source_location_str,
                 file_type: "avro".to_string(),
             });
         let avro_response = scheduler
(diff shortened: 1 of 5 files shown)


ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -297,13 +298,24 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
         // TODO shouldn't this take a ListingOption object as input?
 
         let GetFileMetadataParams { path, file_type } = request.into_inner();
-        let file_format: Arc<dyn FileFormat> = match file_type.as_str() {
+        let file_format: Result<Arc<dyn FileFormat>, Status> = match file_type.as_str() {
             "parquet" => Ok(Arc::new(ParquetFormat::default())),
```
etolbakov, line 302: I've noticed that in the [benchmark](https://github.com/apache/datafusion-ballista/blob/04766d5fd7cbb15d583f736979c11de06c720928/benchmarks/src/bin/tpch.rs#L830) there's slightly different `default` for parquet:
```rust
ParquetFormat::default().with_enable_pruning(Some(true))
```
I wonder if we need consistency or it's ok to leave it as is?

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -297,13 +298,24 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
+            "tbl" => {
+                let format = CsvFormat::default()
+                    .with_delimiter(b'|')
+                    .with_has_header(false);
+                Ok(Arc::new(format))
+            }
             _ => Err(tonic::Status::unimplemented(
                 "get_file_metadata unsupported file type",
             )),
-        }?;
```
etolbakov, line 306: I had to drop the "short circuit"  from here to resolve the rust compilation error:
```sh
Type mismatch [E0308] expected `Result<Arc<CsvFormat>, Status>`, but found `Result<Arc<ParquetFormat>, Status>` 
```
please let me know if there's a better way of doing things?

etolbakov:
@andygrove 
Could you please take a look at this change when you have  time? 
Provided that there were no open ticket (but only todo in the code) I understand that maybe this change is of no use.

etolbakov:
@andygrove just a ping in case this notification is lost.
Could you please take a look?

ballista/scheduler/src/scheduler_server/grpc.rs
Dandandan: I think pruning is default already, so in practice it will be the same.

etolbakov:
Thank you for the feedback @Dandandan 
I'll resolve the conflicts and address your comments in the next couple of days!

milenkovicm:
@Dandandan @etolbakov  `get_file_metadata` has been removed from latest code (#/1110). it was not clear what it was used for, no documentation nor any of the internal ballista code was using it. 

can you please elaborate what was the point of it ?

etolbakov:
@milenkovicm @Dandandan 
sorry for the confusion, as per description I was exploring the code and came across the TODO about the csv file format support. So decided to address it.
If it doesn't seem to be useful enough or creates more hassle we can close the ticket. 
I'm happy to look into something else.

milenkovicm:
As project was not actively maintained and lot of work accumulated we decide to change project's scope and remove code so maintainers have smaller code base to maintain ( some details in #1067 ). This specific functionality has been removed as there is no clear benefit of keeping it. 

Currently, there is active work ongoing towards #974

@etolbakov if you wish to contribute please join discord channel, there are ongoing discussions on tasks and priority

PR #1024, opened 2024-06-18
Fix job hangs when partition count of plan is zero

lewiszlw:

Closes #.

 # Rationale for this change

When partition count of execution plan is 0, the job will hang because available tasks of running stage will be empty, then `get_running_stage_id` method of `ExecutionGraph` will return None. https://github.com/apache/datafusion-ballista/blob/e7cb3d57f61900ea56107c6477c5754429ba821e/ballista/scheduler/src/state/execution_graph.rs#L969
So the running stage will keep running and won't be converted to successful/failed stage.


andygrove:
Thanks @lewiszlw. Could you rebase/upmerge then I can review

andygrove (approved):
LGTM. Thanks @lewiszlw

PR #1039, opened 2024-07-12
Fixes Setting Job Name Not Reflected in Ballista UI

### LINKED ISSUE #1019: Setting job name is not reflected in the Web UI
**Describe the bug**

I'm setting the job name with:

```rust
    let start = SystemTime::now();
    let since_the_epoch = start
        .duration_since(UNIX_EPOCH)
        .expect("Time went backwards");

    let n: usize = 10;
    let job_name = format!("top_{}_{}", n, since_the_epoch.as_secs());

    let config = BallistaConfig::builder()
        .set("ballista.job.name", &job_name)
        .build()?;
```

But it isn't shown in the Web UI (`None` is shown as the job name).

This is the same both locally and in a GKE cluster.

**To Reproduce**

Run a job with the above config.

**Expected behavior**

The job name should appear in the Web UI

**Additional context**

It seems the relevant code is https://github.com/apache/datafusion-ballista/blob/04766d5fd7cbb15d583f736979c11de06c720928/ballista/core/src/config.rs#L30 but I can't see anything wrong.

athultr1997:
Closes #1019  

 # Rationale for this change
Bug of Job Name not being displayed in Ballista UI even after its explicitly set.

Changes in `ballista/core/src/execution_plans/distributed_query.rs` to list the configs passed for the job.

Indirectly yes. The Job Name section in the Ballista UI shows empty currently even if the job name is set. This PR fixes the issue and shows the actual name being set.

### DIFF
--- a/ballista/core/src/execution_plans/distributed_query.rs
+++ b/ballista/core/src/execution_plans/distributed_query.rs
@@ -21,7 +21,7 @@ use crate::serde::protobuf::execute_query_params::OptionalSessionId;
 use crate::serde::protobuf::{
     execute_query_params::Query, execute_query_result, job_status,
     scheduler_grpc_client::SchedulerGrpcClient, ExecuteQueryParams, GetJobStatusParams,
-    GetJobStatusResult, PartitionLocation,
+    GetJobStatusResult, KeyValuePair, PartitionLocation,
 };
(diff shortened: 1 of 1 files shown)


andygrove (approved):
Thanks @athultr1997. LGTM pending CI.

PR #1045, opened 2024-09-02
Fix cargo build

andygrove:

Closes #.

 # Rationale for this change

New lints in Cargo caused build failure.


andygrove:
Cargo build is fixed with this PR but the docker build still fails. 

```
./dev/build-ballista-docker.sh: line 26: docker-compose: command not found
```

PR #1059, opened 2024-09-22
Upgrade to Datafusion 42

palaska:
Upgrades datafusion and certain libraries to the equivalent versions of what datafusion uses (mainly arrow and it's crates, which depend on prost, tonic, hyper etc). `warp` was still on hyper v0.14 and v1 breaks quite a lot of stuff so I replaced it with `axum`.

Closes #.

 # Rationale for this change


ballista/core/src/serde/mod.rs
```diff
@@ -114,6 +118,100 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> BallistaCodec<T,
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct BallistaLogicalExtensionCodec {
```
palaska, line 122: had to create a `BallistaLogicalExtensionCodec` which combines all the file format codecs, not sure if there is a better way atm

ballista/core/src/utils.rs
```diff
@@ -62,13 +63,14 @@ use tonic::transport::{Channel, Error, Server};
 
 /// Default session builder using the provided configuration
 pub fn default_session_builder(config: SessionConfig) -> SessionState {
-    SessionState::new_with_config_rt(
-        config,
-        Arc::new(
+    SessionStateBuilder::new()
+        .with_default_features()
```
palaska, line 67: previous method did not have the default_features which seemed necessary to me, please check :)

ballista/core/src/utils.rs
```diff
@@ -252,15 +254,16 @@ pub fn create_df_ctx_with_ballista_query_planner<T: 'static + AsLogicalPlan>(
     let session_config = SessionConfig::new()
         .with_target_partitions(config.default_shuffle_partitions())
         .with_information_schema(true);
-    let mut session_state = SessionState::new_with_config_rt(
-        session_config,
-        Arc::new(
+    let session_state = SessionStateBuilder::new()
+        .with_default_features()
```
palaska, line 258: same here

ballista/executor/src/executor.rs
```diff
@@ -103,12 +105,22 @@ impl Executor {
         concurrent_tasks: usize,
         execution_engine: Option<Arc<dyn ExecutionEngine>>,
     ) -> Self {
+        let scalar_functions = all_default_functions()
```
palaska, line 108: since we have these here, maybe we can remove the code that manually registers functions in the execution_loop?

ballista/scheduler/src/api/mod.rs
```diff
@@ -13,126 +13,39 @@
...
-        .or(route_cancel_job)
-        .or(route_query_stages)
-        .or(route_job_dot)
-        .or(route_query_stage_dot)
-        .or(route_job_dot_svg)
-        .or(route_scheduler_metrics);
-    routes.boxed()
+use std::sync::Arc;
+
+pub fn get_routes<
```
palaska, line 22: axum routes

ballista/scheduler/src/cluster/mod.rs
```diff
@@ -1020,14 +1024,10 @@ mod test {
...
         };
 
-        if let Some(task) = graph.pop_next_task(&executor.id)? {
-            let task_status = mock_completed_task(task, &executor.id);
-            graph.update_task_status(&executor, vec![task_status], 1, 1)?;
-        }
-
-        graph.revive();
+        // complete first stage
+        revive_graph_and_complete_next_stage(&mut graph)?;
```
palaska, line 1028: this helper method just finds the next stage that has an available task and completes the stage. This was required because we create scans or memtables with 2 partitions, which result in 2 tasks in the 1st stage. [This](https://github.com/apache/datafusion/issues/12611) only causes regression when the input doesn't have multiple partitions

ballista/scheduler/src/test_utils.rs
```diff
@@ -783,6 +783,47 @@ pub fn assert_failed_event(job_id: &str, collector: &TestMetricsCollector) {
     assert!(found, "{}", "Expected failed event for job {job_id}");
 }
 
+pub fn revive_graph_and_complete_next_stage(graph: &mut ExecutionGraph) -> Result<usize> {
```
palaska, line 786: this helper method just finds the next stage that has an available task and completes the stage. This was required because we create scans or memtables with 2 partitions, which result in 2 tasks in the 1st stage. [This](https://github.com/apache/datafusion/issues/12611) only causes regression when the input doesn't have multiple partitions

ballista/core/src/utils.rs
andygrove: That explains a lot. Thanks for catching that!

ballista/executor/src/executor.rs
andygrove: Yes, we should no longer need that

andygrove:
Thanks @palaska. This looks great. I wasn't able to run any queries though with these changes.

I built with `cargo build --release`.

I started the scheduler:

```
$ ./target/release/ballista-scheduler 
2024-09-25T12:57:29.760309Z  INFO main ThreadId(01) ballista_scheduler::cluster: Initializing Sled database in temp directory    
2024-09-25T12:57:29.769023Z  INFO main ThreadId(01) ballista_scheduler::scheduler_process: Ballista v0.12.0 Scheduler listening on 0.0.0.0:50050    
2024-09-25T12:57:29.769038Z  INFO main ThreadId(01) ballista_scheduler::scheduler_process: Starting Scheduler grpc server with task scheduling policy of PullStaged    
2024-09-25T12:57:29.769137Z  INFO main ThreadId(01) ballista_scheduler::cluster::kv: Initializing heartbeat listener    
2024-09-25T12:57:29.769155Z  INFO main ThreadId(01) ballista_scheduler::scheduler_server::query_stage_scheduler: Starting QueryStageScheduler    
2024-09-25T12:57:29.769175Z  INFO tokio-runtime-worker ThreadId(31) ballista_core::event_loop: Starting the event loop query_stage  
```

I started the executor:

```
$ ./target/release/ballista-executor -c 8
2024-09-25T12:57:34.575905Z  INFO main ThreadId(01) ballista_executor::executor_process: Running with config:    
2024-09-25T12:57:34.575919Z  INFO main ThreadId(01) ballista_executor::executor_process: work_dir: /tmp/.tmpnbFKWN    
2024-09-25T12:57:34.575921Z  INFO main ThreadId(01) ballista_executor::executor_process: concurrent_tasks: 8    
2024-09-25T12:57:34.576719Z  INFO tokio-runtime-worker ThreadId(32) ballista_executor::executor_process: Ballista v0.12.0 Rust Executor Flight Server listening on 0.0.0.0:50051    
2024-09-25T12:57:34.576759Z  INFO tokio-runtime-worker ThreadId(30) ballista_executor::execution_loop: Starting poll work loop with scheduler    
```

The executor immediately started producing errors:

```
2024-09-25T12:57:34.577270Z  WARN tokio-runtime-worker ThreadId(30) ballista_executor::execution_loop: Executor poll work loop failed. If this continues to happen the Scheduler might be marked as dead. Error: status: Cancelled, message: "h2 protocol error: http2 error", details: [], metadata: MetadataMap { headers: {} }    
2024-09-25T12:57:34.678711Z  WARN tokio-runtime-worker ThreadId(02) ballista_executor::execution_loop: Executor poll work loop failed. If this continues to happen the Scheduler might be marked as dead. Error: status: Cancelled, message: "h2 protocol error: http2 error", details: [], metadata: MetadataMap { headers: {} }    
```

I then went back to the scheduler process and saw errors there as well:

```
thread 'tokio-runtime-worker' panicked at /home/andy/git/apache/datafusion-ballista/ballista/scheduler/src/scheduler_server/grpc.rs:95:56:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at /home/andy/git/apache/datafusion-ballista/ballista/scheduler/src/scheduler_server/grpc.rs:95:56:
called `Option::unwrap()` on a `None` value
```

Here is the backtrace:

```
thread 'tokio-runtime-worker' panicked at /home/andy/git/apache/datafusion-ballista/ballista/scheduler/src/scheduler_server/grpc.rs:95:56:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: ballista_scheduler::scheduler_server::grpc::<impl ballista_core::serde::generated::ballista::scheduler_grpc_server::SchedulerGrpc for ballista_scheduler::scheduler_server::SchedulerServer<T,U>>::poll_work::{{closure}}
   5: <<ballista_core::serde::generated::ballista::scheduler_grpc_server::SchedulerGrpcServer<T> as tower_service::Service<http::request::Request<B>>>::call::PollWorkSvc<T> as tonic::server::service::UnaryService<ballista_core::serde::generated::ballista::PollWorkParams>>::call::{{closure}}
   6: <ballista_core::serde::generated::ballista::scheduler_grpc_server::SchedulerGrpcServer<T> as tower_service::Service<http::request::Request<B>>>::call::{{closure}}
```

andygrove:
I managed to get it working. In the scheduler `poll_work` method, you need to change ..

```
        let remote_addr = request.remote_addr();
```

to 

```
        let remote_addr = request
            .extensions()
            .get::<ConnectInfo<SocketAddr>>()
            .cloned();
```

I saw a similar change in register_executor

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -70,7 +72,10 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
                 "Bad request because poll work is not supported for push-based task scheduling",
             ));
         }
-        let remote_addr = request.remote_addr();
+        let remote_addr = request
+            .extensions()
+            .get::<ConnectInfo<SocketAddr>>()
+            .cloned();
```
andygrove, line 78: nit: this code block is repeated a few times so it may be worth extracting into a method

andygrove:
I have been testing with TPC-H, and many queries work. However, some queries, such as queries 2, 7, and 8, never complete, and I do not see any errors logged.

I won't have time to debug this issue until the weekend, and perhaps I should instead focus on adding some CI tests for TPC-H so that we can catch regressions like this in CI.

palaska:
> I have been testing with TPC-H, and many queries work. However, some queries, such as queries 2, 7, and 8, never complete, and I do not see any errors logged.
> 
> I won't have time to debug this issue until the weekend, and perhaps I should instead focus on adding some CI tests for TPC-H so that we can catch regressions like this in CI.

I've just looked at this, I think the never completing queries just try to return too many rows and they take too long. That is because the logical optimizer pushes down Limit to Sort but on logical plan deserialization, we seem to be ignoring it. This might be a quick fix: https://github.com/apache/datafusion/pull/12626.

andygrove:
> I've just looked at this, I think the never completing queries just try to return too many rows and they take too long. That is because the logical optimizer pushes down Limit to Sort but on logical plan deserialization, we seem to be ignoring it. This might be a quick fix: [apache/datafusion#12626](https://github.com/apache/datafusion/pull/12626).

It works fine with the main branch though. Was this bug not present in earlier DataFusion versions?

palaska:
> > I've just looked at this, I think the never completing queries just try to return too many rows and they take too long. That is because the logical optimizer pushes down Limit to Sort but on logical plan deserialization, we seem to be ignoring it. This might be a quick fix: [apache/datafusion#12626](https://github.com/apache/datafusion/pull/12626).
> 
> It works fine with the main branch though. Was this bug not present in earlier DataFusion versions?

I think this bug was always there but Datafusion v42's logical optimizer removes Limit logical node after pushing down it's `fetch` while v41 doesn't remove it. Even though we haven't been deserializing `Sort` properly, having the wrapping `Limit` was saving us.

andygrove:
Would it make sense just to upgrade to DF 41 in this PR?

palaska:
> Would it make sense just to upgrade to DF 41 in this PR?

I will try

> Would it make sense just to upgrade to DF 41 in this PR?

Raised a separate PR that upgrades Datafusion to 41: https://github.com/apache/datafusion-ballista/pull/1062, please see the last commit on that PR that reverts from 42 to 41. I had to disable two tests failing due to other bugs that are fixed on v42. All tpch queries seem to be working.

andygrove:
I'm fine with pinning to a revision of DataFusion once your PR is merged over there.

palaska:
> I'm fine with pinning to a revision of DataFusion once your PR is merged over there.

Just explored this, pinning `datafusion` to a rev makes the `python` crate break since we cannot pin `datafusion-python`'s `datafusion` dependency at the same time. Maybe we can just merge the upgrade to v41 and wait for v43 for now? 😅

andygrove:
This upgrade has now happened in another PR. Thanks for starting this @palaska

Issue #1067, opened 2024-10-11
EPIC: Deprecating, removing and making functionalities optional features


milenkovicm:
As per our discussion in #1066 I'm opening this epic to break down PR #1066 to set of smaller chunks. 

- [x] #1069 - datafusion python is the main effort, we should focus making it run on ballista
- [x] #1071- this can be contrib project
- [x] #1074 - ci works without it, could not relate content to anything in the repo
- [x] #1070 - never finalized, dangling code
- [x] #1076- this can be a contrib project, we could cover it with pluggable scheduler strategies
- [x] #1073 - library too old, there are newer library to be used
- [x] #1072 - can be a contrib project
- [x] #1079 
- [x] #1084
- [x] #1077 -  only memory store remains, we may want to take different approach, like checkpointing or similar
- [x] #1078 
- [x] #1075 
- [x] #1080
- [x] #1085
- [x] #1109
- [x] #1089
- [x] #1103 & #1102 
- [x] #1114
- [x] #1104 
- [x] #1105
- [x]  #1110
- [x] #1119  (depends on: #1091, #1100, and previous)

For `could be contrib` functionalities I suggest just to remove them at this point, who ever needs them can make contrib project from them

milenkovicm:
@andygrove the other few PR will be pushed once we finish with current backlog, they may introduce merge conflicts

PR #1066, opened 2024-10-09
Ballista reloaded - proposed changes to core ballista 

milenkovicm:
# Which issue does this PR close?

Closes #.

 # Rationale for this change

First of all, apologies for a gigantic PR, but I did not see better way of making this proposal.

With this proposal I'd like to propose one way forward for ballista, direction is not set in stone so we could discuss alternatives.

I'd like to set the stage with driving idea, it may help understanding PR better.

I believe if we want to revive ballista we need to make it a library rather than full fledge application with all batteries included. If we continue developing it as an application, core code will just grow, putting maintainers on pressure to maintain code which may not be related to core ballista mission.
Making ballista as library, which could bootstrap schedulers and executors with few lines of code, with possibility to plug in additional functionalities will keep use-case specific code out of ballista. We should make it as simple as bootstrapping datafusion application.

Second goal, I believe,  should be keeping ballista functionalities on par with datafusion releases. At the moment ballista is missing a lot of functionalities, like writes, sinks, to name a few. The main, short term, goal would be to get writes, inserts supported.

Lastly, client-scheduler, and scheduler-executor interaction should be improved. Client-scheduler at the moment communicate using custom made gprc protocol, which could get support for UDF(s), configuration ... current client-scheduler protocol can be made an optional component, so users can bring their own frontend such as spark connect, or similar. scheduler-executor as well can be extended to support session configuration information, ...

So with those high level goals in mind, I have done a bit of code trimming and maintenance. I have removed some code which I personally believe should/could be an "add-on" to ballista. 

# What changes are included in this PR?

This PR removes:

- python support (datafusion python is the main effort, we should focus making it run on ballista)
- helm support (this can be contrib project)
- good part of ci directory (ci works without it)
- plugin sub-system (never finalized)
- caching support (this can be a contrib project)
- support for hdfs (library too old, there are newer library to be used)
- UI (can be a contrib project)
- made keda, rest-api, flight-sql optional (we could move them to contrib project)
- key value store is gone, only memory store remains
- removes some dependencies which were detected as not used

This PR adds no functionality, but updates most of the dependencies.

# Are there any user-facing changes?

PR #1077, opened 2024-10-12
Remove key-value stores for scheduler persistence

milenkovicm:
in-memory store only left. if HA required another strategy like check-pointing should be implemented

Relates to: #1066 & #1067

Closes #.

 # Rationale for this change


milenkovicm:
PR has few `allow(dead_code)` as properties may be needed when we start publishing scheduler events

andygrove (approved):

PR #1084, opened 2024-10-15
Make rest-api optional feature ...

milenkovicm:
... disabled by default

relates to  #1067

Closes #.

 # Rationale for this change

### DIFF
--- a/ballista/scheduler/src/scheduler_server/mod.rs
+++ b/ballista/scheduler/src/scheduler_server/mod.rs
@@ -64,6 +64,7 @@ pub struct SchedulerServer<T: 'static + AsLogicalPlan, U: 'static + AsExecutionP
     pub start_time: u128,
     pub state: Arc<SchedulerState<T, U>>,
     pub(crate) query_stage_event_loop: EventLoop<QueryStageSchedulerEvent>,
+    #[cfg(feature = "rest-api")]
     query_stage_scheduler: Arc<QueryStageScheduler<T, U>>,
     config: Arc<SchedulerConfig>,
 }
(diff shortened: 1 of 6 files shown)


ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -155,7 +156,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T
     pub fn running_job_number(&self) -> usize {
         self.state.task_manager.running_job_number()
     }
-
+    #[allow(dead_code)] // used by rest-api feature
```
andygrove, line 159: can we not do `#[cfg(feature = "rest-api")]` here?

milenkovicm: we absolutluy could, will fix

ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -64,6 +64,7 @@ pub struct SchedulerServer<T: 'static + AsLogicalPlan, U: 'static + AsExecutionP
     pub start_time: u128,
     pub state: Arc<SchedulerState<T, U>>,
     pub(crate) query_stage_event_loop: EventLoop<QueryStageSchedulerEvent>,
+    #[allow(dead_code)] // used by rest-api feature
```
andygrove, line 67: Are the remaining uses of `#[allow(dead_code)]` still necessary?

milenkovicm: they are not, my bad, should be fixed now

mmooyyii:
When rest-api is disabled and I request "http://127.0.0.1:50050/api/jobs" or others api. it will return 200 ok with empty body. 
maybe it should return 400 and some error message?

milenkovicm:
> When rest-api is disabled and I request "http://127.0.0.1:50050/api/jobs" or others api. it will return 200 ok with empty body. maybe it should return 400 and some error message?

issue created https://github.com/apache/datafusion-ballista/issues/1223

PR #1085, opened 2024-10-15
fix issue with not building python package ...

milenkovicm:
... and add a python crate test in the workflows.

relates #1067 

Closes #.

 # Rationale for this change


andygrove (approved):
Thanks @milenkovicm

PR #1096, opened 2024-10-22
Ergonomic way to setup/configure `SessionContextExt`

### LINKED ISSUE #1092: Ergonomic way to setup/configure `SessionContextExt`

`SessionContextExt` at the moment does not expose methods which could be used to set up SessionState and/or
configuration.

```rust
let ctx = SessionContext::remote(url &str)
let ctx = SessionContext::standalone()
```

Extend `SessionStateExt` implementation to support configuring `SessionContext` like DataFusion does. Provide a methods like

```rust
let state : SessionState = SessionStateBuilder::build()

let ctx = SessionContext::remote_with_state(url &str)
let ctx = SessionContext::standalone_with_state()
```

to enable custom session state setup, as well as providing custom ballista related configuration as part of the SessionState.

To provide ballista specific configuration we could provide a `SessionConfigExt`  which would.
expose `with_ballista_config(&str, &str)`. Hopefully this way we would not need to expose `BallistaConfig` to
end user (making it private and simplifying it).

An alternative approach would be to provide `SessionConfigExt::new_with_ballista()`, this
way we could set ballista config with helper method which would:

```rust
let state = ctx.state_ref();
let mut state = state.write();
let config = state.config_mut().options_mut();

config.set(&key.key, key.value.as_deref().unwrap_or(""));
```

later approach could be better with python support. We would need to try them out

An alternative approach would be keeping `BallistaConfiguration` which would slightly increase code base,
and have a bit different approach than vanilla datafusion context set up. This would imply slidhtly different
user facing `SessionContextExt` API where additional configuration property `config: BallistaConfig` should be exposed.

**Additional context**

If effort is successful we would need to expose similar `SessionState` configuration API (session state factory), after which we could deprecate `ballista_core::object_store::with_object_store_registry(config: RuntimeConfig)` giving ability to users to set up their own context. Something very similar to `ballista-core::utils::default_session_builder(config: SessionConfig)`. 
If we go extra mile, maybe for datafusion api we would be able to provide a `cloudpickle` serialised session state factory where users would be able to configure their onw SessionsState across cluster

task relates to #1088 and #1091 (kind off)

milenkovicm:

Closes #1092.

# Rationale for this change

This change provides ergonomic way to configure SessionContextExt, similar to DataFusion context setup from using provided SessionStore.

This change provides two new methods in `SessionContextExt` `async fn standalone_with_state(state: SessionState)` and `async fn remote_with_state(url: &str,state: SessionState)`
Which accepts pre configured `SessionState` as parameter. `SessionState` should be configured in the same way like when `SessionContext` is configured in DataFusion.

```rust
let state = SessionStateBuilder::new().with_default_features().build();
let ctx: SessionContext = SessionContext::remote_with_state(&url, state).await?;
```

This change also exposes a `BallistaSessionConfigExt` which provides method to configure ballista specific settings like, `BallistaConfiguration`, codecs or even `QueryPlanner`.

```rust
use ballista_client::extension::BallistaSessionConfigExt;

let session_config = SessionConfig::new_with_ballista()
    .with_information_schema(true)
    .set_str(BALLISTA_JOB_NAME, "Super Cool Ballista App");

let state = SessionStateBuilder::new()
    .with_default_features()
    .with_config(session_config)
    .build();

let ctx: SessionContext = SessionContext::remote_with_state(&url, state).await?;
```

`LogicalExtensionCodec` and `PhysicalExtensionCodec` can be changed as well:

```rust
use ballista_client::extension::BallistaSessionConfigExt;
let logical_codec = Arc::new(BadLogicalCodec::default());
let physical_codec = Arc::new(MockPhysicalCodec::default());
let session_config = SessionConfig::new_with_ballista()
    .with_information_schema(true)
    .with_ballista_physical_extension_codec(physical_codec.clone())
    .with_ballista_logical_extension_codec(logical_codec.clone())
    ;
let state = SessionStateBuilder::new()
    .with_default_features()
    .with_config(session_config)
    .build();

let ctx: SessionContext = SessionContext::standalone_with_state(state).await?;
```

In this case logical and physical codec will be also be propagated to standalone.

Lastly, BallistaQueryPlanner can be replaced: 

```rust
let session_config = SessionConfig::new_with_ballista()
    .with_information_schema(true)
    .set_str(BALLISTA_PLANNER_OVERRIDE, "false");

let state = SessionStateBuilder::new()
    .with_default_features()
    .with_config(session_config)
    .with_query_planner(Arc::new(BadPlanner::default()))
    .build();

let ctx: SessionContext = SessionContext::standalone_with_state(state).await?;
```

At the moment there is a hacky way telling ballista not to override provided planner with `.set_str(BALLISTA_PLANNER_OVERRIDE, "false");` as it is not possible to detect if the planner is changed.

Introduction of two new methods and one extension to new functionality, no braking change

Notes: 

- Object store is not be set automatically anymore, will look into `ballista_core::object_store_registry::with_object_store_registry` deprecation in follow up commits when we expose configuration on scheduler and executor API
- `standalone_with_state` may change slightly adding executor related configuration

### DIFF
--- a/ballista/client/Cargo.toml
+++ b/ballista/client/Cargo.toml
@@ -43,6 +43,8 @@ tokio = { workspace = true }
 url = { version = "2.5" }
 
 [dev-dependencies]
+ballista-executor = { path = "../executor", version = "0.12.0" }
+ballista-scheduler = { path = "../scheduler", version = "0.12.0" }
 ctor = { version = "0.2" }
 env_logger = { workspace = true }
 
--- a/ballista/client/src/extension.rs
+++ b/ballista/client/src/extension.rs
@@ -15,15 +15,17 @@
 // specific language governing permissions and limitations
 // under the License.
 
+pub use ballista_core::utils::BallistaSessionConfigExt;
 use ballista_core::{
     config::BallistaConfig,
     serde::protobuf::{
         scheduler_grpc_client::SchedulerGrpcClient, CreateSessionParams, KeyValuePair,
     },
-    utils::{create_df_ctx_with_ballista_query_planner, create_grpc_client_connection},
+    utils::{create_grpc_client_connection, BallistaSessionStateExt},
 };
-use datafusion::{error::DataFusionError, prelude::SessionContext};
-use datafusion_proto::protobuf::LogicalPlanNode;
+use datafusion::{
+    error::DataFusionError, execution::SessionState, prelude::SessionContext,
+};
 use url::Url;
 
 const DEFAULT_SCHEDULER_PORT: u16 = 50050;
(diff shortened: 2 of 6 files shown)


andygrove (approved):
Looks nice. Thanks @milenkovicm

milenkovicm:
could this change get merged @andygrove , i have follow up #1099 on top of this one

milenkovicm:
Thanks @andygrove

PR #1112, opened 2024-11-11
support window functions

### LINKED ISSUE #542: Support window function in ballista

As https://github.com/apache/arrow-datafusion/issues/361 all feature has been done in datafusion. We can bring window function in ballista ✌️

I have implement a simple version in my repo. But we need do some change in datafusion.

- [x] https://github.com/apache/arrow-datafusion/issues/4479
- [x] https://github.com/apache/arrow-datafusion/issues/4480
- [x] https://github.com/apache/arrow-datafusion/issues/4509
- [ ]  ballista related...

After we use the related version datafusion, will implement some window function   `Physical plan`  encode and decode features.

I have checked with datafusion,  If the window function  `PARTITION BY` and `ORDER BY`, the physical  planner of datafusion has helped to add `repation` and `sort`. So we no need to deal `window_func` in `plan_query_stages_internal`.

**Additional context**

onursatici:

Closes #542 .

 # Rationale for this change

As stated in #542, datafusion adds `SortExec` and `RepartitionExec` based on the window clauses `partition by` and `order by` (See `enforce_distribution.rs` in datafusion and `WindowAggExec::required_input_distribution` .
From what I can see, this restriction on window functions in ballista was put way back when datafusion was in the same repo as ballista and window function support was still in its infancy. Now that datafusion enforces the right hash distribution by placing repartition nodes, I believe ballista currently already supports window functions.

I have tested partitioned, unpartitioned and sorted window expressions with ballista with this PR and compared them with datafusion, it seems to be working as intended. Parititioned window functions get repartitioned by their partition expressions and are parallelised.

Not sure what is the best way to test this given distribution handling is done by datafusion here.

Allow queries with window expressions

### DIFF
--- a/ballista/scheduler/src/planner.rs
+++ b/ballista/scheduler/src/planner.rs
@@ -298,15 +298,20 @@ mod test {
     use crate::planner::DistributedPlanner;
     use crate::test_utils::datafusion_test_context;
     use ballista_core::error::BallistaError;
-    use ballista_core::execution_plans::UnresolvedShuffleExec;
+    use ballista_core::execution_plans::{ShuffleWriterExec, UnresolvedShuffleExec};
     use ballista_core::serde::BallistaCodec;
+    use datafusion::arrow::compute::SortOptions;
+    use datafusion::physical_expr::expressions::Column;
     use datafusion::physical_plan::aggregates::{AggregateExec, AggregateMode};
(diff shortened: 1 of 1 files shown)


onursatici:
@milenkovicm should be good to go. I also enabled a previously ignored test, as the fix it mentions is merged for a while now

milenkovicm (approved):
LGTM
we wold need to wait for somebody who can actually merge this code

PR #1138, opened 2024-11-26
refactor: SessionBuilder to return Result<_>

milenkovicm:

Closes None.

 # Rationale for this change

Session builder method was returning SessionState, some of the example functions we have would use unwrap() internally. 

Change method signature to return Result<SessionState>

### DIFF
--- a/ballista/core/src/utils.rs
+++ b/ballista/core/src/utils.rs
@@ -68,7 +68,7 @@ pub fn default_session_builder(
     Ok(SessionStateBuilder::new()
         .with_default_features()
         .with_config(config)
-        .with_runtime_env(Arc::new(RuntimeEnv::new(RuntimeConfig::default()).unwrap()))
+        .with_runtime_env(Arc::new(RuntimeEnv::new(RuntimeConfig::default())?))
         .build())
 }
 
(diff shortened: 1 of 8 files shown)


ballista/core/src/utils.rs
```diff
@@ -62,12 +62,14 @@ use tonic::codegen::StdError;
...
 /// Default session builder using the provided configuration
-pub fn default_session_builder(config: SessionConfig) -> SessionState {
-    SessionStateBuilder::new()
+pub fn default_session_builder(
+    config: SessionConfig,
+) -> datafusion::common::Result<SessionState> {
+    Ok(SessionStateBuilder::new()
         .with_default_features()
         .with_config(config)
         .with_runtime_env(Arc::new(RuntimeEnv::new(RuntimeConfig::default()).unwrap()))
```
andygrove, line 71:
```suggestion
        .with_runtime_env(Arc::new(RuntimeEnv::new(RuntimeConfig::default())?))
```

milenkovicm:
missed that one, fixed now, thanks !

andygrove (approved):
Thanks @milenkovicm

PR #1139, opened 2024-11-26
feat: Add option to limit number of executor worker threads

milenkovicm:
`executor_cores` configuration option is similar to `spark.executor.cores`,  it sets number of tokio worker threads limiting executor cpu utilisation.

Closes none .

 # Rationale for this change

provide ability to users to limit number of cores which executor can utilise limiting overall resource utilisation 

- new configuration option 
- tokio executors got their own name 

new configuration option exposed to user

### DIFF
--- a/ballista/executor/Cargo.toml
+++ b/ballista/executor/Cargo.toml
@@ -51,13 +51,7 @@ log = { workspace = true }
 mimalloc = { workspace = true, optional = true }
 parking_lot = { workspace = true }
 tempfile = { workspace = true }
-tokio = { workspace = true, features = [
-    "macros",
-    "rt",
-    "rt-multi-thread",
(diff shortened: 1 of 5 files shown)


milenkovicm:
there is also `concurrent_tasks` which configuration option, i believe if `executor_cores` is set it `concurrent_tasks` should be <= to `executor_cores`

milenkovicm:
will open another PR to remove `cache_` executor config options

milenkovicm:
this PR may make no sense as there is dedicated executor which has limited number of runners

milenkovicm:
DedicatedExecutor will limit number of threads based on `concurrent_tasks`, maybe we should document it better. Closing this task, apologies for the noise

PR #1136, opened 2024-11-25
updated maturin version and ccargo build to build yml

tbar4:
# Which issue does this PR close?

Closes #1135 

 # Rationale for this change

This change will allow us to create python wheel files for several different platforms. 

# What changes are included in this PR?

A github build file for python wheels will be added and we will also be updating maturin from 0.6.0 t0 1.6.0

# Are there any user-facing changes?

There are no user facing changes

PR #1141, opened 2024-11-27
chore: Update python files

milenkovicm:
update pip dependencies, pyproject and cargo, also added fix few small issues.

Closes #.

 # Rationale for this change

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -16,7 +16,7 @@
 # under the License.
 
 [build-system]
-requires = ["maturin>=0.15,<0.16"]
+requires = ["maturin>=1.5.1,<1.6.0"]
 build-backend = "maturin"
 
 [project]
@@ -24,7 +24,7 @@ name = "ballista"
 description = "Python client for Apache Arrow Ballista Distributed SQL Query Engine"
 readme = "README.md"
 license = {file = "LICENSE.txt"}
-requires-python = ">=3.6"
+requires-python = ">=3.7"
 keywords = ["ballista", "sql", "rust", "distributed"]
 classifier = [
     "Development Status :: 2 - Pre-Alpha",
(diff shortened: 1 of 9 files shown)


python/pyproject.toml
```diff
@@ -16,15 +16,15 @@
...
+requires = ["maturin>=1.5.1,<1.6.0"]
 build-backend = "maturin"
 
 [project]
 name = "ballista"
 description = "Python client for Apache Arrow Ballista Distributed SQL Query Engine"
 readme = "README.md"
 license = {file = "LICENSE.txt"}
-requires-python = ">=3.6"
+requires-python = ">=3.7"
```
tbar4, line 27: This would conclict with #1136 so I will change it really quick and repush my branch.

tbar4: updated #1136 to require python 3.7 or greater

milenkovicm: Datafusion python has python > 3.7

python/src/lib.rs
```diff
@@ -66,7 +63,11 @@ impl PyBallistaBuilder {
     /// Construct the standalone instance from the SessionContext
     pub fn standalone(&self, py: Python) -> PyResult<PySessionContext> {
         // Build the config
-        let config: SessionConfig = SessionConfig::from_string_hash_map(&self.conf)?;
+        let mut config: SessionConfig = SessionConfig::new_with_ballista();
```
tbar4, line 66: Did you update something in `new_with_ballista`? When I initially tried this I was getting an error. If it works thatn that is great!

python/src/utils.rs
tbar4, line 1: This was initially used when awaiting the old BallistaContext. I think we could remove if needed. I left it in because I was unsure if we would need it again, but at this point I think we could just delete.

python/src/lib.rs
milenkovicm: Changed nothing

tbar4 (approved):
Looks good. Only callout is potentially removing `to_pyerr`

tbar4 (approved):

PR #1144, opened 2024-11-30
add partitioning scheme for unresolved shuffle and shuffle reader exec

onursatici:

Closes https://github.com/apache/datafusion-ballista/issues/16.

 # Rationale for this change

unresolved shuffle and shuffle reader exec does not have partitioning info. This becomes a problem when a stage has `InterleaveExec`, which requires all its children to have hash partitioning on the same expressions. Currently even though the partitioning satisfies InterleaveExec's requirements, because children show `UnknownPartitioning` planning such plans into stages fail.

Carry partitioning scheme through shuffles

### DIFF
--- a/ballista/core/src/serde/mod.rs
+++ b/ballista/core/src/serde/mod.rs
@@ -32,6 +32,7 @@ use datafusion_proto::logical_plan::file_formats::{
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_hash_partitioning;
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_partitioning;
 use datafusion_proto::physical_plan::to_proto::serialize_partitioning;
+use datafusion_proto::physical_plan::DefaultPhysicalExtensionCodec;
 use datafusion_proto::protobuf::proto_error;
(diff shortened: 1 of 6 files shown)


ballista/core/src/serde/mod.rs
```diff
@@ -291,8 +294,11 @@ impl PhysicalExtensionCodec for BallistaPhysicalExtensionCodec {
                 )?))
             }
             PhysicalPlanType::ShuffleReader(shuffle_reader) => {
+                let default_codec =
```
milenkovicm, line 297: would it make sense to make default_codec property `BallistaPhysicalExtensionCodec`, so it can be overridden if needed? `BallistaLogicalExtensionCodec` does similar thing. (also there is few other places where `DefaultPhysicalCodec` is created)

onursatici: good point, thanks!

Dandandan:
Changes I think look good. Can we add a test for it?

onursatici:
@Dandandan should be ready for another look

andygrove (approved):

Dandandan:
Thanks @onursatici

PR #1150, opened 2024-12-10
chore: dependency cleanup

milenkovicm:

Closes none.

 # Rationale for this change

There are a bit too many dependencies, some of which are duplicated and can be replaced with some other we use, some of which are not maintained. As users can create their own binaries there is no need to depend on dependencies which our binaries use. 

There were cases where log and trace has been used in the same files, I have removed trace but we can revisit that decision in subsequent PRs

- remove few dependencies (anyhow, parse_arg)
- make few dependencies optional. dependencies used in executor and scheduler binaries are optional now 
- remove use of trace in few files where log was used as well

No


ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
tbar4, line 1: Can you help me understand why we are removing the tracing? I know you have a good reason, just I am missing the context.

tbar4: Thanks for the explanation!

andygrove (approved):

PR #1181, opened 2025-02-15
feat: notify scheduler when a task fails

milenkovicm:

Closes #.

 # Rationale for this change

One case where task fails, error does not get propagated to scheduler which leaves it hanging, waiting for task competition notification

handling Error branch in executor loop 

No


andygrove (approved):
Thanks @milenkovicm

PR #1192, opened 2025-03-01
chore: minor release script fix

andygrove:

N/A

 # Rationale for this change

Improve release process

- Remove the redundant check for `GH_TOKEN` env var
- Modify email template to include vote for the email sender

### DIFF
--- a/dev/release/create-tarball.sh
+++ b/dev/release/create-tarball.sh
@@ -53,11 +53,6 @@ if [ "$#" -ne 2 ]; then
     exit
 fi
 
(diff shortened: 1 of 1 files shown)


dev/release/create-tarball.sh
```diff
@@ -53,11 +53,6 @@ if [ "$#" -ne 2 ]; then
     exit
 fi
 
-if [[ -z "${GH_TOKEN}" ]]; then
-    echo "Please set personal github token through GH_TOKEN environment variable"
-    exit
-fi
```
andygrove, line 59: This script does not use git

Issue #1194, opened 2025-03-05
TPCH unit tests failure


vmingchen:
**Describe the bug**
All unit tests in benchmarks/tpch.rs are failing with errors like this:

> Error: External(ArrowError(CsvError("incorrect number of fields for line 1, expected 16 got 17"), None))

**To Reproduce**
```shell
cd <BALLISTA_HOME>/benchmarks
./tpch-gen.sh
TPCH_DATA=data cargo test
```

**Expected behavior**
All tests should pass with `TPCH_DATA=data cargo test`

milenkovicm:
@vmingchen is this issue closed with #1195?

vmingchen:
Yes. Fixed with https://github.com/apache/datafusion-ballista/pull/1195.

PR #1195, opened 2025-03-05
Fix unit tests in tpch.rs

vmingchen:

https://github.com/apache/datafusion-ballista/issues/1194

 # Rationale for this change

Fix unit test in tpch unit tests. Now benchmark test `TPCH_DATA=data/ cargo test` is passing.

There are a list of changes each of which is in a separate commit:

- https://github.com/apache/datafusion-ballista/pull/1195/commits/4cb5a75f8a013490a82c4b84abd2f8e80f23adb6: Use  `get_tbl_tpch_table_schema` to special treat the`.tbl` files that are generated by `tpch-gen.sh` and have an additional trailing column.
- https://github.com/apache/datafusion-ballista/pull/1195/commits/007d148b94f2dff97509e3d876e109e3359793aa: Add missing limits to the tpch queries. Note that this may affect some of the benchmark results.
- https://github.com/apache/datafusion-ballista/pull/1195/commits/76b86079d6dd0a481bdb31d2d64a59e772ed5b04: Ignore `q15()` because it contains multiple statements and is not supported and similarly ignored by `run_q15()`.
- https://github.com/apache/datafusion-ballista/pull/1195/commits/9634d5c9666623a235c0d6332e0a06460d99acbe: Use Int32 for "*year" data types to match the actual query data type.
-  https://github.com/apache/datafusion-ballista/pull/1195/commits/8f6f1a003195857f776969352410cde7447e382a: Normalize query results before compare them with the expected results loaded from CSV file.
- https://github.com/apache/datafusion-ballista/pull/1195/commits/5f16e6aeca4863b14c69ac01a52415be68144cb8: Skip a query results mismatch that seems to be slightly off in the expected results in the tpch data.
- https://github.com/apache/datafusion-ballista/pull/1195/commits/7b80efd1d9055d1ab6fd705aded1231a966d89ca: Format only.

No

### DIFF
--- a/benchmarks/src/bin/tpch.rs
+++ b/benchmarks/src/bin/tpch.rs
@@ -1562,6 +1562,16 @@ mod tests {
         Ok(())
     }
 
+    // Our queries do not have the LIMIT clause, so we need to apply it
+    // before comparing the results with the expected results.
+    const QUERY_LIMITS: [(usize, usize); 5] = [
+        (2, 100),
+        (3, 10),
+        (10, 20),
+        (18, 100),
+        (21, 100),
+    ];
+
     // We read the expected results from CSV files so we need to normalize the
     // query results before we compare them with the expected results for the
     // following reasons:
@@ -1576,13 +1586,17 @@ mod tests {
     // 4. Rename columns using the expected schema to make schema matching
     // because, for q18, we have aggregate field `sum(l_quantity)` that is
     // called `sum_l_quantity` in the expected results.
-    async fn normalize_columns(
+    //
+    // 5. If there is a missing LIMIT clause, apply it to the results.
+    async fn normalize_for_verification(
         batches: Vec<RecordBatch>,
-        expected_schema: Schema,
+        n: usize,
     ) -> Result<Vec<RecordBatch>> {
         if batches.is_empty() {
             return Ok(vec![]);
         }
+        let expected_schema = get_answer_schema(n);
+        let limit = QUERY_LIMITS.iter().find(|&&x| x.0 == n).map(|x| x.1);
         let ctx = SessionContext::new();
         let schema = batches[0].schema();
         let df = ctx.read_batches(batches)?;
@@ -1624,7 +1638,7 @@ mod tests {
                     }
                 })
                 .collect::<Vec<Expr>>(),
-        )?;
+        )?.limit(0, limit)?;
         df.collect().await
     }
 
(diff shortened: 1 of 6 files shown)


milenkovicm:
@andygrove is right person for this review. 
I have one observation, q2 without limit helped us to find some issues with underlaying infrastructure, I wonder if we can keep those unlimited queries as some kind of stress test suite ?

andygrove:
Thanks @vmingchen I will try and review this over the weekend

benchmarks/src/bin/tpch.rs
```diff
@@ -1534,6 +1562,86 @@ mod tests {
         Ok(())
     }
 
+    // Our queries do not have the LIMIT clause, so we need to apply it
+    // before comparing the results with the expected results.
```
andygrove, line 1566: I don't remember why the queries don't contain limit clauses but I think they should, to match the TPC-H specification.

vmingchen: Make sense. I have reapplied the limit and reverted the normalization-limit step.

vmingchen: To provide more context, the ":n 100" lines in the official TPCH sql files are called flags. They are handled by the `qsub` function in qgen.c file:

```c
     /*
      * FUNCTION qsub(char *qtag, int flags)
      *
      * based on the settings of flags, and the template file $QDIR/qtag.sql
      * make the following substitutions to turn a query template into EQT
      *
      *  String      Converted to            Based on
      *  ======      ============            ===========
      *  first line  database <db_name>;      -n from command line
      *  second line set explain on;         -x from command line
      *   :<number>  parameter <number>
      *  :k          set number
      *  :o          output to outpath/qnum.snum----
      *                                      -o from command line, SET_OUTPUT
      *  :s          stream number
      *  :b          BEGIN WORK;             -a from command line, START_TRAN
      *  :e          COMMIT WORK;            -a from command line, END_TRAN
      *  :q          query number
      *  :n<number>                          sets rowcount to be returned
      */
     void
     qsub(char *qtag, int flags)
```

milenkovicm:
Looking at the https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-H_v3.0.1.pdf changes make sense @vmingchen, thanks a lot. 

I believe there are integration tests which verify this but they are disabled, it would be great if we can enable them after this get merged as a follow up. 

If @andygrove is ok with this we can merge it

milenkovicm (approved):

andygrove:
> Looking at the https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-H_v3.0.1.pdf changes make sense @vmingchen, thanks a lot. 
> 
> 
> 
> I believe there are integration tests which verify this but they are disabled, it would be great if we can enable them after this get merged as a follow up. 
> 
> 
> 
> If @andygrove is ok with this we can merge it 

LGTM

milenkovicm:
thanks for contributing @vmingchen

PR #1216, opened 2025-03-28
feat: Improve fetch partition performance, support skip validation arrow ipc files

### LINKED ISSUE #1189: Consider using `with_skip_validation` for shuffle file reading

Improve shuffle reading, with https://github.com/apache/arrow-rs/pull/7120 merged consider using it for shuffle files.

Change shuffle reader to utilise new api.

Keep everything as it is 

**Additional context**

westhide:

Closes #1189.

 # Rationale for this change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -181,9 +181,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "54.2.1"
+version = "54.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc208515aa0151028e464cc94a692156e945ce5126abd3537bb7fd6ba2143ed1"
+checksum = "84ef243634a39fb6e9d1710737e7a5ef96c9bacabd2326859ff889bc9ef755e5"
 dependencies = [
  "arrow-arith",
  "arrow-array",
@@ -202,9 +202,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-arith"
-version = "54.2.1"
+version = "54.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e07e726e2b3f7816a85c6a45b6ec118eeeabf0b2a8c208122ad949437181f49a"
+checksum = "8f420c6aef51dad2e4a96ce29c0ec90ad84880bdb60b321c74c652a6be07b93f"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -216,9 +216,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-array"
-version = "54.2.1"
+version = "54.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2262eba4f16c78496adfd559a29fe4b24df6088efc9985a873d58e92be022d5"
+checksum = "24bda5ff6461a4ff9739959b3d57b377f45e3f878f7be1a4f28137c0a8f339fa"
 dependencies = [
  "ahash",
  "arrow-buffer",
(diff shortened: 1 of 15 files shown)


milenkovicm:
Thanks @westhide this PR makes sense, 

It also brings configurable validation option, which I did not really think about.
I have few questions, for which I do not have definitive answer, so I'd like to hear your opinion:

1. In which cases do we want to enable validation?
2. If we want to have option to enable/disable it, should that be decision of who ever operates scheduler/executor rather than connected client?

westhide:
> Thanks @westhide this PR makes sense,
> 
> It also brings configurable validation option, which I did not really think about. I have few questions, for which I do not have definitive answer, so I'd like to hear your opinion:
> 
> 1. In which cases do we want to enable validation?
> 2. If we want to have option to enable/disable it, should that be decision of who ever operates scheduler/executor rather than connected client?

Q1:
As the `BallistaFlightService` keep listenning on each Executor, it allow client to send a `do_get` request, and without check `FetchPartition` action's `path` is created by shuffle writer, so the client can try to read any file on the executor. In this scene, we can enable validation.

Q2:
I'm not sure. As currently we just read ipc file created by `ShuffleWriterExec`, it's safe to skip all validation.

milenkovicm:
> Q1: As the `BallistaFlightService` keep listenning on each Executor,writting it allow client to send a `do_get` request, and without check `FetchPartition` action's `path` is created by shuffle writer, so the client can try to read any file on the executor. In this scene, we can enable validation.

I guess we can safely assume that only shuffle files are accessed, not a random files. 

> Q2: I'm not sure. As currently we just read ipc file created by `ShuffleWriterExec`, it's safe to skip all validation.

what I was having in mind is enabling unsafe (without validation) by default but having a executor configuration switch which could revert this. It might be easy to cover case with arrow flight but a bit harder when ShuffleReader reads local shuffle file 

> ### Addition
> Should we consider the power down scene? It may cause Arrow ipc file broken if `ShuffleWriterExec` is writting. Maybe we can support Job recover and reuse the partition file in the future.

Im not sure this case can be achieved with this validation? how can we be sure that file is written fully? I guess we'd need some kind of checksum for this scenario.

westhide:
> > Q1: As the `BallistaFlightService` keep listenning on each Executor,writting it allow client to send a `do_get` request, and without check `FetchPartition` action's `path` is created by shuffle writer, so the client can try to read any file on the executor. In this scene, we can enable validation.
> 
> I guess we can safely assume that only shuffle files are accessed, not a random files.
> 
> > Q2: I'm not sure. As currently we just read ipc file created by `ShuffleWriterExec`, it's safe to skip all validation.
> 
> what I was having in mind is enabling unsafe (without validation) by default but having a executor configuration switch which could revert this. It might be easy to cover case with arrow flight but a bit harder when ShuffleReader reads local shuffle file
> 
> > ### Addition
> > Should we consider the power down scene? It may cause Arrow ipc file broken if `ShuffleWriterExec` is writting. Maybe we can support Job recover and reuse the partition file in the future.
> 
> Im not sure this case can be achieved with this validation? how can we be sure that file is written fully? I guess we'd need some kind of checksum for this scenario.

All right, will move the validation config to Executor.

And will try to add a checksum validation with failed Job rerun after finish the wasm udf.

milenkovicm:
hey @westhide are you still interested to get this PR merged?

Dandandan:
As far as I can see, we don't have to validate the IPC files:

* Ballista has control over writing the output
* In a power down scenario where the file is being written but the stage is not yet completed, we will / should not read the resulting files. As the stage is never completed, the files may be ignored (or cleaned up).

> Maybe we can support Job recover and reuse the partition file in the future

I think it makes sense to support job recovery either by stage or by completed task rather than per IPC file.
Supporting it by partial IPC file seems it might be complicated, as it keep track of input and execution state, which is lost when powering down as well.

killzoner:
Is there anything left to do apart from the config based setup on the executor for this PR?
If that's all, I can push back this with the suggested changes

milenkovicm:
closing this as other PR addressed the issue

PR #1226, opened 2025-04-04
feat: make scheduler session context stateless 

### LINKED ISSUE #1220: `JobStateEventStream` does not emit events related to Session
**Describe the bug**

At the moment `JobStateEvent` does not emit: `JobStateEvent::SessionCreated`, `JobStateEvent::SessionUpdated`,  and missing `JobStateEvent::SessionRemoved` those events are not propagated to `JobStateEventStream`

**To Reproduce**

add custom listener  `cluster.job_state().job_state_events()` and listen for those events

**Expected behavior**

Expected events to be published, with `SessionConfig` rather than `BallistaConfig`

**Additional context**

- session context does not expire as it looks like, it will be kept in memory forever. it hast o be addressed as well

milenkovicm:
... as session context was not cleaned up.

Closes #1220.

 # Rationale for this change

- Scheduler was keeping HashMap of session_id -> SessionContext which was never cleaned up. 
- Scheduler will never emit events about accessed or removed SessionContexts 
- Client SessionContext is authoritative not the scheduler  

- client sets session context, no need for extra `create_context`
- get_context has been replaced with `create_or_update_context`

### DIFF
--- a/ballista/client/src/extension.rs
+++ b/ballista/client/src/extension.rs
@@ -15,16 +15,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use ballista_core::extension::SessionConfigHelperExt;
 pub use ballista_core::extension::{SessionConfigExt, SessionStateExt};
-use ballista_core::{
-    serde::protobuf::{scheduler_grpc_client::SchedulerGrpcClient, CreateSessionParams},
-    utils::create_grpc_client_connection,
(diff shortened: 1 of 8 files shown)


andygrove (approved):
I don't currently have time to review large PRs like this unfortunately, but I skimmed through it and it looks reasonable.

PR #1232, opened 2025-04-09
chore: Remove some arrow references

andygrove:

Closes #.

 # Rationale for this change


ballista/core/proto/datafusion.proto
```diff
@@ -21,7 +21,7 @@
  package datafusion;
  
  option java_multiple_files = true;
- option java_package = "org.apache.arrow.datafusion.protobuf";
```
milenkovicm, line 24: this file is coped from datafusion, I believe we should change it there 

https://github.com/apache/datafusion/blob/7ef5c4b291bdb13d5b53e7e344c7a6ec1cb2c8b4/datafusion/proto/proto/datafusion.proto#L24

LICENSE.txt
milenkovicm, line 1: do we need to have LICENSE.txt datafusion changed it license in https://github.com/apache/datafusion/pull/3154

ballista/core/proto/datafusion.proto
andygrove: Thanks. I also created https://github.com/apache/datafusion/pull/15658

LICENSE.txt
andygrove: I pushed another commit to update the LICENSE.txt to match the one from DataFusion

https://github.com/apache/datafusion/blob/main/LICENSE.txt

dev/build-ballista-docker.sh
milenkovicm, line 1: `apache/datafusion-ballista-*` is referenced in documentation but it was building `apache/arrow-ballista_*`,  sorry

Issue #1235, opened 2025-04-11
executor can't read s3 config in push-staged mode.


mmooyyii:
**Describe the bug**
run example custom-scheduler in push-staged mode, it returns 
```
Error: ArrowError(ExternalError(Execution("Job ScIzlxB failed: Job failed due to stage 1 failed: Task failed due to runtime execution error: DataFusionError(Execution(\"DataFusionError(Configuration(\\\"'s3.access_key_id' & 's3.secret_access_key' must be configured\\\"))\"))\n")), None)
```

**To Reproduce**
use custom-client.rs, custom-executor.rs,custom-scheduler.rs in example

1. change  the default TaskSchedulingPolicy to push staged in datafusion-ballista/ballista/core/src/config.rs
```rust
pub enum TaskSchedulingPolicy {
    /// push-based scheduling can result in lower latency.
    #[default]
    PushStaged,
    /// Pull-based scheduling works in a similar way to Apache Spark
    PullStaged,
}
```
2. run custom-scheduler, custom-executor and custom-client
```
yimo@YideMacBook-Pro examples % cargo run --example custom-scheduler
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `/Users/yimo/RustroverProjects/datafusion-ballista/target/debug/examples/custom-scheduler`
[2025-04-11T06:05:34Z INFO  ballista_scheduler::scheduler_process] Ballista v45.0.0 Scheduler listening on 127.0.0.1:50050
[2025-04-11T06:05:34Z INFO  ballista_scheduler::scheduler_process] Starting Scheduler grpc server with task scheduling policy of PushStaged
[2025-04-11T06:05:34Z INFO  ballista_scheduler::scheduler_server::query_stage_scheduler] Starting QueryStageScheduler
[2025-04-11T06:05:34Z INFO  ballista_core::event_loop] Starting the event loop query_stage
```

```
yimo@YideMacBook-Pro datafusion-ballista % cargo run --example custom-executor
   Compiling ballista-examples v45.0.0 (/Users/yimo/RustroverProjects/datafusion-ballista/examples)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.47s
     Running `target/debug/examples/custom-executor`
[2025-04-11T06:05:53Z INFO  ballista_executor::executor_process] Executor starting ... (Datafusion Ballista 45.0.0)
[...]

PR #1236, opened 2025-04-11
fix: executor can't read s3 config in push-staged mode

mmooyyii:

Closes https://github.com/apache/datafusion-ballista/issues/1235

 # Rationale for this change


milenkovicm:
thanks for patch @mmooyyii 
there is a test to test object store access but it does not cover all cases unfortunately, we definitely need to improve testing.

just a heads up, object store part will be moved around in #1230, default features are already added.

milenkovicm (approved):

PR #1256, opened 2025-04-24
feat: make execution_graph.stages()  public

milenkovicm:

Closes #.

 # Rationale for this change

this is follow up on #1243 where I missed to make `stages()` public. Making stages public we can react on scheduler emitted events and inspect all stages for given job

- make single method public 

### DIFF
--- a/ballista/scheduler/src/state/execution_graph.rs
+++ b/ballista/scheduler/src/state/execution_graph.rs
@@ -218,7 +218,7 @@ impl ExecutionGraph {
         new_tid
     }
 
-    pub(crate) fn stages(&self) -> &HashMap<usize, ExecutionStage> {
+    pub fn stages(&self) -> &HashMap<usize, ExecutionStage> {
         &self.stages
     }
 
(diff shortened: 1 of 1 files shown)


andygrove (approved):

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -218,7 +218,7 @@ impl ExecutionGraph {
         new_tid
     }
 
-    pub(crate) fn stages(&self) -> &HashMap<usize, ExecutionStage> {
+    pub fn stages(&self) -> &HashMap<usize, ExecutionStage> {
```
andygrove, line 221: nit: It would be good practice to start adding documentation on public methods

milenkovicm: fully agree with you, it'll be a good exercise for copilot one of the following weekends. will fix this one personally this time :)

milenkovicm: created  #1258

PR #1260, opened 2025-05-10
feat: expose submit and cancel job methods as public in scheduler

milenkovicm:

Closes #.

 # Rationale for this change

- provide ability to submit and submit and cancel job without going through grpc 

- make methods submit and cancel job public
- submit job will return job id 

no

### DIFF
--- a/ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
+++ b/ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
@@ -376,9 +376,9 @@ mod tests {
         )
         .await?;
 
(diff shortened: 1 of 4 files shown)


ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -376,9 +376,9 @@ mod tests {
         )
         .await?;
 
-        let job_id = "job-1";
+        //let job_id = "job-1";
```
andygrove, line 379: nit: can this comment be removed?

andygrove (approved):

PR #1267, opened 2025-05-21
feat: job id is incremental

milenkovicm:

Closes #.

 # Rationale for this change

previously `job id` is generated randomly, without any ordering guarantees, which make is rather hard to determine ordering of jobs, and returning totally random results on rest api. Spark is generating job id incrementing atomic int. 

- use atomic integer to generate job id, i do not see valid reason not to do it
- move job id generation from task manager to scheduler (if we later want to make it configurable it would be easier to do)

- no, they might get confused with new job ids 

# Open points

- [ ] `job_id` is used as directory name where job (shuffle) data is persisted, if we use incremental id, we may get into problems with leftover directory names in case of scheduler restart

### DIFF
--- a/ballista/scheduler/src/display.rs
+++ b/ballista/scheduler/src/display.rs
@@ -53,7 +53,7 @@ pub fn print_stage_metrics(
             DisplayableBallistaExecutionPlan::new(plan, &plan_metrics).indent()
         );
     } else {
-        error!("Fail to combine stage metrics to plan for stage [{}/{}],  plan metrics array size {} does not equal
+        error!("Fail to combine stage metrics to plan for job/stage [{}/{}],  plan metrics array size {} does not equal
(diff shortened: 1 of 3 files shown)


andygrove (approved):

Dandandan:
How would this work in a multi-scheduler setup? Maybe add the scheduler id as a first item?

milenkovicm:
when going through logs, for example, it makes it easier to reason about.

milenkovicm:
there is issue with having job id tied to physical directory, which may make mess when scheduler is restarted without restarting executors, making possibility to overlap job data directories (shuffle readers may read wrong spill in current implementation) thus just using incremental id in current implementation may make problems. 

I was looking at the `ULID` they generate random but sortable IDs, downside is that they are too big (IMHO), not easiest to reason about ordering either.

snowflake like ids, are sortable and unique; for them machine id should be exposed as a configuration parameter. 

Current approach is not sortable, now there is slight chance to get name collision (with very low probability).

PR #1272, opened 2025-06-15
feat: consistent hash scheduling implemented as `DistributionPolicy`

milenkovicm:

Closes #.

 # Rationale for this change

pluggable `DistributionPolicy` provides a way to extend scheduler task binding policies.
with consistent hash task binding policy being quite use case specific implementing it as `DistributionPolicy` would make a sense.  This would reduce api surface we need to maintain. 

follow up needed:

- improved tests
- example how to use it 
- documentation 

- custom `DistributionPolicy` for consistent hash scheduling 
- moved to its own submodule
- policy has been removed from out of the box policy, users should configure it programmatically

- consistent hash option removed from scheduler configuration


milenkovicm:
closed as stale, will reopen at some point

Issue #1128, opened 2024-11-23
Streamline github actions


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Github actions look very busy, and repeatable (multiple actions do same check), on top of that they need 25+ mins to finish. 
rust compilation is done in dedicated rust container.

There is no clean difference between  "Test Workspace ..." & "Test Ballista" ...

Also, some of the actions are complaining about deprecations.

**Describe the solution you'd like**

streamline and modernise github actions file, use github action for rust setup, build caching ... . remove options which are not used anymore. 

**Describe alternatives you've considered**

**Additional context**

milenkovicm:
<img width="1076" alt="image" src="https://github.com/user-attachments/assets/c23f8f23-327e-4f68-b587-3f73357335bd">

milenkovicm:
there is #703 which is asking same

Huy1Ng:
I can give thisa try, but do we have any idea of what an ideal workflow should be? For example. datafusion Rust workflow also takes 25-30min to complete:
https://github.com/apache/datafusion/actions/workflows/rust.yml

PR #1275, opened 2025-06-29
improve rust workflows without cache

Huy1Ng:

Part of #1128.

 # Rationale for this change
In the issue

Improve the rust workflows by following the practices in parent datafusion repo (with couple inspiration from popular rust project like uv or polars). Changes are:
- Parallelize all jobs. I don't see any reason to wait for linux-build-lib to complete first before running other jobs
- Move prep steps out into actions
- Use cargo check instead of cargo build
- Remove caches altogether. While caches help a lot for a particular job, the compiler couldn't reuse other jobs' artifacts, so dependent jobs will always re-download and recompile their own => take up lots of cache storage space. Also, cache is not commonly used in other projects that I researched. 

No


milenkovicm:
Have no access to computer at the moment. Can you please push change, it should trigger new job

Two questions 

- What's the reason to remove cancel job? It was helpful killing old runs
- We do push docker, that job is used, would it be possible to revert back that change

milenkovicm:
You don't have to if you think this is better solution. I don't know much about this topic. What do you think ?

I guess we need to publish docker, cancel job is your call

.github/actions/setup-builder/action.yaml
```diff
@@ -37,3 +37,5 @@ runs:
         rustup toolchain install ${{ inputs.rust-version }}
```
milenkovicm, line 37: I wonder should we use GitHub action such as `dtolnay/rust-toolchain@stable` to setup rather than use bash script for it?

.github/workflows/rust.yml
```diff
@@ -17,275 +17,151 @@
 
 name: Rust
 
+concurrency:
+  group: ${{ github.repository }}-${{ github.workflow }}
+  cancel-in-progress: true
```
milenkovicm, line 22: 👍🏻

.github/workflows/rust.yml
```diff
@@ -17,275 +17,151 @@
...
-          CARGO_HOME: "/github/home/.cargo"
-          CARGO_TARGET_DIR: "/github/home/target"
-      - name: Save artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: rust-artifacts
-          path: |
-            /github/home/target/release/ballista-scheduler
-            /github/home/target/release/ballista-executor
+          # Adding `--locked` here to assert that the `Cargo.lock` file is up to
```
milenkovicm, line 56: 👍🏻

.github/workflows/rust.yml
```diff
@@ -17,275 +17,151 @@
...
-        run: |
-          apt-get -qq update && apt-get -y -qq install protobuf-compiler
-          protoc --version
-      - name: Cache Cargo
-        uses: actions/cache@v4
-        with:
-          path: /github/home/.cargo
-          # this key equals the ones on `linux-build-lib` for re-use
-          key: cargo-cache-
-      - name: Cache Rust dependencies
```
milenkovicm, line 276: I'm fine to remove cache is it helps, there are another two locations where cache is used `cargo-toml-formatting-checks` and `Clippy`. Would it make sense to remove it there as well or we keep it?

.github/actions/setup-builder/action.yaml
Huy1Ng: I tried but the current repo settings doesn't allow `dtolnay/rust-toolchain@stable`. `datafusion` repo seems to have the same limitations: https://github.com/apache/datafusion/pull/15315. 

P/s: my bad. I was using different toolchain. Still, I think we should mirror datafusion CI workflows as much as possible (offload the thinking and decision to the bigger community).

.github/workflows/rust.yml
Huy1Ng: I will remove them. Those are fast to run so cache wouldn't do much anyway.

Huy1Ng: I also wonder if benchmark test is where we should leave the caching. This is the bottleneck of rust workflows, so any optimization would help.

Huy1Ng:
this should be ready for rust. I will tackle Python workflow next. This is going to be spitted into 3 small PRs.

milenkovicm:
thanks againn @Huy1Ng

Issue #1288, opened 2025-07-31
Improve Python CI workflow


Huy1Ng:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Python CI time is getting worse. It's now taking over 40 minutes to complete. The workflow is severely bottlenecked at [Mac x86_64 (3.10)](https://github.com/apache/datafusion-ballista/actions/runs/16625562819/job/47040965358#logs) and [Mac/Win (3.10, windows-latest)](https://github.com/apache/datafusion-ballista/actions/runs/16625562819/job/47040965361#logs) (41 and 34min respectively). All other jobs ran fine under 20 minutes. And if we remove the redundant `cargo build` step, it could shave off 8 minutes for each job.

**Describe the solution you'd like**
The build time should be under 20 minutes, in line with other CI workflows

**Describe alternatives you've considered**
We could: **1.** accept the slow build time for Python **2.** remove those 2 slow jobs **3.** use some optimization flags to speed up the build (doubt that we could bring them back under 20 minutes this way tho)

**Additional context**
I didn't find any way to speed up the build process other than removing the redundant `cargo build` before `maturin build`

Issue #1271, opened 2025-06-14
update rust edition to 2024


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

to keep up with latest rust edition we need to update rust to 2024

**Describe the solution you'd like**

update rust edition and code to support `edition = "2024`

**Describe alternatives you've considered**

**Additional context**

`configure_me` dependency looks as very outdated (and not maintained) from what i can tell code generated by it (`configure_me_codegen`) may not be valid according to `2024` specification

- [x] #1281
- [x] Fix issues with code

milenkovicm:
`configure_me` deps has clippy problems as well (https://github.com/apache/datafusion-ballista/pull/1270/commits/9131cc8395c0c86c89e38a94ba4b8f1ae91b8451)

milenkovicm:
We should track, Minimum Supported Rust Version (MSRV) of datafusion as well

```
# Minimum Supported Rust Version (MSRV)
rust-version = "1.85.1"
```

killzoner:
> Looks like pydf 51 has not been released yet. Maybe we should disable msrv check for pyballista for now

I just made sense of this. I thought current main was tracking datafusion 50, but it's 51

milenkovicm:
pydf 51 not released yet

PR #1289, opened 2025-08-10
Improve GitHub actions/python workflows

Huy1Ng:
#1288 
#1271

Closes #.

 # Rationale for this change
This is to improve the Python CI workflow, which is horribly slow at the moment. Like previous PR, I tried to reuse the CI workflows from the bigger datafusion-python repo. This also update rust to latest version and edition, which seems to help a lot with compilation time on rust and python workflows

- use uv for python build (in line with datafusion-python)
- update to edition 2024 and rust 1.89.0
- re-enable python formater
- add pytest job to pre-verify python build integrity

Yes. uv use pyproject.toml and uv.lock, so requirements.txt is no longer needed for build steps.


Huy1Ng:
I will re-add restriction to run python workflows on main branch once we are done with the reviews. Except for the workflows and Cargo.toml, all changes are from automatic formatting and clippy fixes.

milenkovicm (commented):
thanks @Huy1Ng i will have a look in next day or so.
One thing I would suggest is to keep MSRV and edition aligned with datafusion 49 (msrv 1.85.1) 

also would it be possible to have one PR for clippy and other for python related changes. I dont really understand where are all those clippy issues coming from

Huy1Ng:
@milenkovicm I separate the rust/clippy change here #1290. You can review there first.

milenkovicm:
@Huy1Ng would it to too big change if we have this merged without MSRV and rust edition changes ?

milenkovicm:
> I keep the Python portion only. The later rust update seem to improve compiling time so would be nice to have that as well.

Which part helps compile time rust edition or MSRV?

Will have look at the PR today, one small comment as we use uv now can python/README.md be updated to reflect change

.github/workflows/dev.yml
```diff
@@ -40,17 +40,16 @@ jobs:
     name: Use prettier to check formatting of documents
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
```
milenkovicm, line 43: Is there a reason we have this specific hash? If there is, can we have a comment to note that? Also, looks like there is version 5 now

milenkovicm (approved):
I think this makes sense, just two minor comments to address

.github/workflows/dev.yml
Huy1Ng: It's the code from datafusion-python, which also seems to follow this discussion here: https://github.com/apache/datafusion/issues/15298

milenkovicm:
Thanks again @Huy1Ng

Issue #1296, opened 2025-08-25
Failed to execute tpcds q47.sql in ballista remote mode


ZihuanLing:
**Describe the bug**
I tried to run tpcds sql test with ballista, when I tried to run q47.sql, it just failed in remote executor, which tell me: `Job zRGEAB9 failed: Job failed due to stage 10 failed: Task failed due to runtime execution error: DataFusionError(Internal("All partition by columns should have an ordering"))`

**To Reproduce**

step 1: build ballista v48.0.1
step 2: start ballista-scheduler and ballista-executor
  🌟note: please ensure the ballista-scheduler and executor run in the same directory with the demo. so they can load the test data correctly.

step 3: build and run the demo code below.

Here is the code to reproduce:
```rust
use datafusion::prelude::*;
use std::error::Error;
use std::fs;
use std::path::Path;
use ballista::prelude::SessionContextExt;

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    // create temp dir to store csv table.
    let temp_dir = "tpcds_test_data";
    fs::create_dir_all(temp_dir)?;

    // generate csv demo data.
    create_date_dim_csv(temp_dir)?;
    create_store_csv(temp_dir)?;
    create_item_csv(temp_dir)?;
    create_store_sales_csv(temp_dir)?;

    // let ctx = SessionContext::new();
    let ctx = SessionContext::remote("df://localhost:50050").await?;

    // Just register a few table which will be used.
    ctx.register_csv(
        "date_dim",
        &format!("{}/date_dim.csv", temp_dir),
        CsvReadOptions::new().has_header(true),
    )
        .await?;

    ctx.register_csv(
        "store",
        &format!("{}/store.csv", temp_dir),
        CsvReadOptions::new().has_header(true),
    )
        .await?;

    ctx.register_csv(
        "item",
        &format!("{}/item.csv", temp_dir),
        CsvReadOptions::new().has_header(true),
    )
        .await?;

    ctx.register_csv(
        "store_sales",
        &format!("{}/store_sales.csv", temp_dir),
        CsvReadOptions::new().has_header(true),
    )
        .await?;

    let query = r#"
--q47.sql--

with v1 as(
    select i_category,
           i_brand,
           s_store_name,
           s_company_name,
           d_year,
           d_moy,
           sum(ss_sales_price) sum_sales,
           avg(sum(ss_sales_price)) over
        (partition by i_category, i_brand,
[...]

PR #1306, opened 2025-08-31
Ensure stage-level sort requirements are enforced in distributed planning

metegenez:

Closes https://github.com/apache/datafusion-ballista/issues/1296.

 # Rationale for this change

The issue stems from the staging logic. When there’s a functional dependency on columns (e.g., column_a = 1), sorting may be unnecessary at that point, so SortExec won’t sort that column. However, once the plan is split into stages, the planner can lose this dependency information that was available in the single, unified plan, causing upstream operators that require ordering to miss it.

EnforceSorting for stages.

No.


milenkovicm (approved):
LGTM thanks @metegenez

PR #123, opened 2022-08-10
Using tokio tracing for log file

Ted-Jiang:
# Which issue does this PR close?

Closes #122 .

 # Rationale for this change
Enable thread info and cut log file by day.
Now logs look like:
```
arrow-ballista % cat logs/scheduler_ballista_0.0.0.0_50050.2022-08-10 
2022-08-10T07:28:45.457283Z  INFO main ThreadId(01) ballista_scheduler: Ballista v0.7.0 Scheduler listening on 0.0.0.0:50050    
2022-08-10T07:28:45.457489Z  INFO main ThreadId(01) ballista_scheduler: Starting Scheduler grpc server with task scheduling policy of PullStaged    
2022-08-10T07:28:45.458130Z  INFO main ThreadId(01) ballista_scheduler::scheduler_server::query_stage_scheduler: Starting QueryStageScheduler    
2022-08-10T07:28:45.458307Z  INFO tokio-runtime-worker ThreadId(17) ballista_core::event_loop: Starting the event loop query_stage    
2022-08-10T07:29:46.156777Z  INFO main ThreadId(01) ballista_scheduler: Ballista v0.7.0 Scheduler listening on 0.0.0.0:50050    
2022-08-10T07:29:46.157159Z  INFO main ThreadId(01) ballista_scheduler: Starting Scheduler grpc server with task scheduling policy of PullStaged    
2022-08-10T07:29:46.157729Z  INFO main ThreadId(01) ballista_scheduler::scheduler_server::query_stage_scheduler: Starting QueryStageScheduler    
2022-08-10T07:29:46.157922Z  INFO tokio-runtime-worker ThreadId(15) ballista_core::event_loop: Starting the event loop query_stage    
2022-08-10T07:30:09.766192Z  INFO main ThreadId(01) ballista_scheduler: Ballista v0.7.0 Scheduler listening on 0.0.0.0:50050    
2022-08-10T07:30:09.768707Z  INFO main ThreadId(01) ballista_scheduler: Starting Scheduler grpc server with task scheduling policy of PullStaged    
2022-08-10T07:30:09.769133Z  INFO main ThreadId(01) ballista_scheduler::scheduler_server::query_stage_scheduler: Starting QueryStageScheduler    
2022-08-10T07:30:09.769264Z  INFO tokio-runtime-worker ThreadId(09) ballista_core::event_loop: Starting the event loop query_stage    
```

# What changes are included in this PR?

# Are there any user-facing changes?

Issue #1293, opened 2025-08-17
Improve `EXPLAN` command support


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`EXPLAN` query provides valuable information to understand job behaviour, at the moment it will provide datafusion plan without shuffles, also it does not really provide information for `EXPLAN ANALYSE`

**Describe the solution you'd like**

Improve `EXPLAN` command support 

**Describe alternatives you've considered**

**Additional context**

danielhumanmod:
Hey team, I’m new to the DataFusion/Ballista community and would love to start contributing from this issue 🙂

milenkovicm:
Welcome, let us know if you need any help

milenkovicm:
Your plan makes sense @danielhumanmod,
If it is easier we could split this issue in two, explain and explain analyse and deliver it as two separate PRs.

Regarding options, B looks better to me but once you start implementing you'll have better idea.

danielhumanmod:
Hey @milenkovicm , I have finished the draft version for `EXPLAIN` implementation #1309 and would love to get some early feedbacks from you :)

Issue #1292, opened 2025-08-17
Expose  Logical and Physical plan details in the REST API


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Expand information REST API provides, making it the first point of contact to understand job behaviour. Logical and Physical plan are crucial for job understanding. 

**Describe the solution you'd like**

REST API to provide more job related information, like logical and physical plan, it can also provide all the relevant plan execution statistic.

**Describe alternatives you've considered**

**Additional context**

PR #1309, opened 2025-09-04
feat: Support distributed plan in `EXPLAIN` command

danielhumanmod:

<!--
We generally require a GitHub issue to be filed for all bug fixes and enhancements and this helps us generate change logs for our releases. You can link an issue to this PR using the GitHub syntax. For example `Closes #123` indicates that this PR will close issue #123.
-->

Closes [#1293](https://github.com/apache/datafusion-ballista/issues/1293)



 # Rationale for this change
`EXPLAN` query provides valuable information to understand job behaviour, at the moment it will provide datafusion plan without shuffles

- Core
  1. Add a new custom Exec `BallistaExplainExec` to include distributed plan info
  2. Add proto and codec support for new ballista exec
- Scheduler
  1. Distributed plan genereation
  2. Physical plan transformation, ExplainExec (DF) -> BallistaExplainExec

Yes, when user query with `EXPLAIN ...`, example output: 
```
            "| distributed_plan | =========ResolvedStage[stage_id=1.0, partitions=1]=========                                                                                                                                                                                                                                              |",
            "|                  | ShuffleWriterExec: partitions:Some(Hash([Column { name: \"id\", index: 0 }], 16))                                                                                                                                                                                                                        |",
            "|                  |   AggregateExec: mode=Partial, gby=[id@0 as id], aggr=[count(Int64(1))]                                                                                                                                                                                                                                  |",
            "|                  |     CoalesceBatchesExec: target_batch_size=8192                                                                                                                                                                                                                                                          |",
            "|                  |       FilterExec: id@0 > 4                                                                                                                                                                                                                                                                               |",
            "|                  |         DataSourceExec: file_groups={1 group: [[Users/ballista/git/datafusion-ballista/ballista/client/testdata/alltypes_plain.parquet]]}, projection=[id], file_type=parquet, predicate=id@0 > 4, pruning_predicate=id_null_count@1 != row_count@2 AND id_max@0 > 4, required_guarantees=[]       |",
            "|                  |                                                                                                                                                                                                                                                                                                          |",
            "|                  |                                                                                                                                                                                                                                                                                                          |",
            "|                  | =========UnResolvedStage[stage_id=2.0, children=1]=========                                                                                                                                                                                                                                              |",
            "|                  | Inputs{1: StageOutput { partition_locations: {}, complete: false }}                                                                                                                                                                                                                                      |",
            "|                  | ShuffleWriterExec: partitions:None                                                                                                                                                                                                                                                                       |",
            "|                  |   ProjectionExec: expr=[count(Int64(1))@1 as count(*), id@0 as id]                                                                                                                                                                                                                                       |",
            "|                  |     AggregateExec: mode=FinalPartitioned, gby=[id@0 as id], aggr=[count(Int64(1))]                                                                                                                                                                                                                       |",
            "|                  |       CoalesceBatchesExec: target_batch_size=8192                                                                                                                                                                                                                                                        |",
            "|                  |         UnresolvedShuffleExec: partitions=Hash([Column { name: \"id\", index: 0 }], 16)                                                                                                                                                                                                                  |",
            "|                  |                                                                                                                                                                                                                                                                                                          |",
            "|                  |                                                                                                                                                                                                                                                                                                          |",
            "+------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+",  
```


milenkovicm:
@danielhumanmod will try to review it in next few days,
thanks a lot

ballista/core/src/execution_plans/distributed_explain.rs
```diff
@@ -0,0 +1,236 @@
...
+
+#[derive(Debug, Clone)]
+pub struct BallistaExplainExec {
+    schema: SchemaRef,
+    stringified_plans: Vec<BallistaStringifiedPlan>,
+    verbose: bool,
+    cache: PlanProperties,
+}
+
+impl BallistaExplainExec {
```
milenkovicm: I wonder do we need new Exec? It prints string literals, can `PlaceholderRowExec` used instead of it? something like: 

```rust
    ctx.sql("select unnest (['logical_plan','physical_plan']) as plan_type, unnest(['Projection ...','ProjectionExec ...']) as plan ")
        .await
        .unwrap()
        .show()
        .await
        .unwrap();
```

or what we get in physical plan: 


```
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                                                                                                                                                                                                                                                                                          |
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| logical_plan  | Projection: __unnest_placeholder(make_array(Utf8("logical_plan"),Utf8("physical_plan")),depth=1) AS UNNEST(make_array(Utf8("logical_plan"),Utf8("physical_plan"))) AS plan_type, __unnest_placeholder(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ...")),depth=1) AS UNNEST(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ..."))) AS plan |
|               |   Unnest: lists[__unnest_placeholder(make_array(Utf8("logical_plan"),Utf8("physical_plan")))|depth=1, __unnest_placeholder(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ...")))|depth=1] structs[]                                                                                                                                                  |
|               |     Projection: List([logical_plan, physical_plan]) AS __unnest_placeholder(make_array(Utf8("logical_plan"),Utf8("physical_plan"))), List([Projection ..., ProjectionExec ...]) AS __unnest_placeholder(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ...")))                                                                                        |
|               |       EmptyRelation                                                                                                                                                                                                                                                                                                                                           |
| physical_plan | ProjectionExec: expr=[__unnest_placeholder(make_array(Utf8("logical_plan"),Utf8("physical_plan")),depth=1)@0 as plan_type, __unnest_placeholder(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ...")),depth=1)@1 as plan]                                                                                                                             |
|               |   UnnestExec                                                                                                                                                                                                                                                                                                                                                  |
|               |     ProjectionExec: expr=[[logical_plan, physical_plan] as __unnest_placeholder(make_array(Utf8("logical_plan"),Utf8("physical_plan"))), [Projection ..., ProjectionExec ...] as __unnest_placeholder(make_array(Utf8("Projection ..."),Utf8("ProjectionExec ...")))]                                                                                         |
|               |       PlaceholderRowExec                                                                                                                                                                                                                                                                                                                                      |
|               |                                                                                                                                                                                                                                                                                                                                                               |
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

ballista/client/tests/context_unsupported.rs
milenkovicm, line 1: I guess this test should not panic, test should move to `context_checks.rs`. Now it might be a bit of problem how do we handle different environments and file locations.

ballista/core/src/execution_plans/distributed_explain.rs
milenkovicm: On the second thoughts, would this design with separate ExplainExec node help to implement `EXPLAIN ANALYSE` ?

what do you think @danielhumanmod ?

danielhumanmod:
> hey @danielhumanmod any help needed from me to finalise this PR ?

Hey @milenkovicm thanks for checking in! Sorry for the delay on my side — a little busy with some personal stuff these days. The solution makes total sense to me: using an existing Exec is definitely a simpler way to solve this. I’ll submit the change in the next few days. Appreciate your support! 🙏

danielhumanmod:
> no worries @danielhumanmod whenever you get time, I just want to check as I believe this is a good PR and did not wan to loose it :)
> 
> please do not change PR just yet, lets think if current PR wold work better with `EXPLAIN ANALYSE`, not sure what do you think!

Hey @milenkovicm about `EXPLAIN ANALYSE` I do have a question,

I identified our stage metrics is in https://github.com/apache/datafusion-ballista/blob/b0fd0875e49eb9bdcd4825efde2bca562e88ab28/ballista/executor/src/metrics/mod.rs. It is a logging for now, we might need to find a way to use it for `EXPLAIN ANALYSE`.

 But not sure a whether a separate ExplainExec node can help `EXPLAIN ANALYSE`, as we need to collect the information across stages. Could you give me some hints on this, should we do it on scheduler layer or a exec node? Thanks you very much!

danielhumanmod:
> Sorry, for late reply @danielhumanmod I'm not quite sure, i guess all the metrics are collected at the scheduler side, so scheduler should have it all once job finishes.
> 
> It's a question how can we get it out of scheduler? Once the job is scheduled its question where/how do we wait for job competition. Do we lunch another job, which would poll rest api or we change scheduler grpc to wait there, i'm not sure.
> 
> #1292 may be related if we get it from rest api

Oh, that’s a really good point. I was also thinking about how to handle this outside of the scheduler, since adding special-case logic directly into the scheduler feels like a bad pattern to me. Let me look into how we could leverage the REST API for this. But I do have a feeling we may need a new Exec node if we go down this path. I’ll do some investigation, dig into it further, and get back to you soon. Thanks for the inspiration!

milenkovicm:
Sorry for late reply @danielhumanmod 
I think your breakdown makes sense, I just wonder what would be the easiest way to implement this.

Scheduler emits job completion events, so we could react on job completion, but this approach would be a bit messy with observers, as evens will be emitted for all jobs.

An alternative would be to add ability to register job completion callback which triggers change of job B once the job A finishes. Trigger can change UnresolvedShuffle to actual Exec (we could use same logic like we use for simple EXPLAIN). 

I believe job finish callback would be useful for other cases, I'm not sure at this point how complicated would it be to implement, may have a look this weekend.

danielhumanmod:
> Sorry for late reply @danielhumanmod I think your breakdown makes sense, I just wonder what would be the easiest way to implement this.
> 
> Scheduler emits job completion events, so we could react on job completion, but this approach would be a bit messy with observers, as evens will be emitted for all jobs.
> 
> An alternative would be to add ability to register job completion callback which triggers change of job B once the job A finishes. Trigger can change UnresolvedShuffle to actual Exec (we could use same logic like we use for simple EXPLAIN).
> 
> I believe job finish callback would be useful for other cases, I'm not sure at this point how complicated would it be to implement, may have a look this weekend.

Sorry for late response @milenkovicm !

The job completion callback sounds like a more reasonable approach for a dynamic scenario like this, and it also seems more generic for other future use cases.

I’d be happy to help investigate or implement the callback if I could get a bit more context how you envision the design.

Also, do you think it makes sense to finish EXPLAIN first, and then tackle EXPLAIN ANALYZE in a follow-up PR once the approach is finalized?

ballista/client/tests/context_unsupported.rs
danielhumanmod: Have migrated this test to the `context_checkes.rs` in latest commit, thanks!

danielhumanmod:
> I agree with you let's finish explain first.
> 
> I had a quick look at internals but I never finished my investigation how to add callback in a better way, will try to do it before you finish explain

Hey @milenkovicm I have updated our `EXPLAIN` implementation based on our prev discussion. Basically I follow this exec rewrite: 

```
Previous:
     ExplainExec
After:
     CoalescePartitionsExec
       └─ ProjectionExec: expr=[list_type -> plan_type, list_plan -> plan]
           └─ UnnestExec
                └─ ProjectionExec: expr=[list_type, list_plan]
                     └─ PlaceholderRowExec 
```
which produces a table like this:
```
| plan_type        | plan            |
|------------------|-----------------|
| logical_plan     | logical_txt     |
| physical_plan    | physical_txt    |
| distributed_plan | distributed_txt |
```

Appreciate any feedback on this! And I’m also happy to help with the callback part — just let me know the direction or any context I should be aware of, and I can dig into it as well.

ballista/client/tests/context_checks.rs
```diff
@@ -979,4 +979,83 @@ mod supported {
 
         Ok(())
     }
+
+    #[rstest]
```
milenkovicm, line 983: this test fails in my environment, looks like its prone to user name change

ballista/client/tests/context_checks.rs
```diff
@@ -979,4 +979,83 @@ mod supported {
...
+        ctx: SessionContext,
+        test_data: String,
+    ) {
+        let parquet_path = format!("{test_data}/alltypes_plain.parquet");
+        ctx.register_parquet("test", &parquet_path, Default::default())
+            .await
+            .unwrap();
+
+        let result = ctx
+            .sql("EXPLAIN select count(*), id from test where id > 4 group by id")
```
milenkovicm: can we do something like?

```rust
ctx.sql("EXPLAIN select count(*), id from (select unnest([1,2,3,4,5]) as id) group by id")
```

danielhumanmod: Thanks for the review! I’ve fixed the previous test failure and added this test as a case for constant tables.
Do you mind triggering the CI again?

ballista/client/tests/context_checks.rs
milenkovicm: I'd suggest to remove this test if we have the other one

danielhumanmod: > I'd suggest to remove this test if we have the other one

Removed, thanks!

milenkovicm:
follow up issue would be good for tracking, thanks @danielhumanmod

PR #1312, opened 2025-09-07
refactor: expand ClusterEventSender visibility to public

Th824:

This is merely a minor modification, so I haven't linked it to an issue. Please let me know if an issue is required, and I'd be happy to create one and link it accordingly.

 # Rationale for this change

When extending scheduler we noticed that cluster event sender would be useful primitive to public accessed.

expand ClusterEventSender visibility from `public crate` to `public`

no


milenkovicm:
Thanks @Th824

PR #1314, opened 2025-09-08
feat: implement job data cleanup in pull-staged strategy #1219

### LINKED ISSUE #602: Shuffle files should get deleted immediately after job finishes by default
I am trying to run automated benchmarks and keep running out of disk space

I would like shuffle files and other temp files to be deleted immediately after a job finishes by default.

None

**Additional context**
None

### LINKED ISSUE #1219: job data cleanup does not work if `pull-staged` strategy selected
**Describe the bug**

If the `pull-staged` strategy is selected job data won't be cleared after the job is finished, leaving data hanging on the executes. This can lead to piling old shuffle files on the executor. One way to prevent is to set executors to cleanup data more aggressively.

**To Reproduce**

just run default ballista cluster setup 

**Expected behavior**

Shuffle files to be removed when job finishes or when there is no need for them 

**Additional context**

- `push-based` strategy works as expected 
- It might be related to #1175

KR-bluejay:

Closes #1219 

Closes #602

 # Rationale for this change

In pull-based strategy, executors did not receive job data cleanup requests; only for the push-based (scheduler --> executors RemoveJobData RPC). As a result, completed jobs could leave their local working directories behind.

This PR adds a scheduler‑managed store‑and‑send path for cleanup in pull mode. The scheduler records per‑executor job IDs to clean and, on poll_work, returns them via PollWorkResult.jobs_to_clean. Executors then validate paths (is_dir + is_subdirectory) and remove those job directorie.

1. Proto: Add jobs_to_clean (repeated CleanJobDataParams, tag=2) to PollWorkResult
2. Scheduler: Maintain per-executor pending_cleanup_jobs and drain on poll_work
3. Executor: Consume jobs_to_clean; delete only if is_dir + is_subdirectory

### DIFF
--- a/ballista/executor/src/execution_loop.rs
+++ b/ballista/executor/src/execution_loop.rs
@@ -17,7 +17,7 @@
 
 use crate::cpu_bound_executor::DedicatedExecutor;
 use crate::executor::Executor;
-use crate::executor_server::is_subdirectory;
+use crate::executor_process::remove_job_dir;
 use crate::{as_task_status, TaskExecutionTimes};
 use ballista_core::error::BallistaError;
 use ballista_core::extension::SessionConfigHelperExt;
@@ -37,7 +37,6 @@ use std::any::Any;
 use std::convert::TryInto;
 use std::error::Error;
 use std::ops::Deref;
-use std::path::PathBuf;
 use std::sync::mpsc::{Receiver, Sender, TryRecvError};
 use std::time::{SystemTime, UNIX_EPOCH};
 use std::{sync::Arc, time::Duration};
@@ -95,23 +94,19 @@ pub async fn poll_loop<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                     jobs_to_clean,
                 } = result.into_inner();
                 active_job = !tasks.is_empty();
-                let work_dir = PathBuf::from(&executor.work_dir);
 
                 // Clean up any state related to the listed jobs
                 for cleanup in jobs_to_clean {
-                    let cleanup_job_id = cleanup.job_id;
-                    let path = work_dir.join(&cleanup_job_id);
-                    let is_valid_path = path.is_dir()
-                        && is_subdirectory(path.as_path(), work_dir.as_path());
+                    let job_id = cleanup.job_id.clone();
+                    let work_dir = executor.work_dir.clone();
 
-                    if !is_valid_path {
-                        warn!("Invalid job cleanup path: {:?}", path);
-
-                        continue;
-                    }
-
-                    info!("Remove data for job {:?}", cleanup_job_id);
-                    std::fs::remove_dir_all(&path)?;
+                    // In poll-based cleanup, removing job data is fire-and-forget.
+                    // Failures here do not affect task execution and are only logged.
+                    tokio::spawn(async move {
+                        if let Err(e) = remove_job_dir(&work_dir, &job_id).await {
+                            error!("failed to remove job dir {job_id}: {e}");
+                        }
+                    });
                 }
 
                 for task in tasks {
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -18,6 +18,7 @@
 //! Ballista Executor Process
 
 use std::net::SocketAddr;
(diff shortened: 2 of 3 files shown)


ballista/executor/src/execution_loop.rs
```diff
@@ -88,8 +90,29 @@ pub async fn poll_loop<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
             Ok(result) => {
-                let tasks = result.into_inner().tasks;
+                let PollWorkResult {
+                    tasks,
+                    jobs_to_clean,
+                } = result.into_inner();
                 active_job = !tasks.is_empty();
+                let work_dir = PathBuf::from(&executor.work_dir);
+
+                // Clean up any state related to the listed jobs
```
milenkovicm, line 98: would it make sense to remove files on a separate tread(s) rather than on a work polling thread? In case of exception while removing we wont break the thread ?

KR-bluejay: Good point — I think the main issue is with `std::fs::remove_dir_all`.  
Would it make sense to switch this to `tokio::fs::remove_dir_all` instead?

1. **Async**: Since `tokio::fs::remove_dir_all` is async, it won’t block the work polling thread.  
2. **Exception isolation**: In case of an error when removing the job directory, I don’t think it should affect the main execution flow.

What do you think about this approach?

milenkovicm: first point is that it should not block pull thread, we could use tokio async but also it would make sense to have a separate facility to do file removal. where we can just schedule request for removal and it would do it in its separate threads, or similar. 

we have another removal method at https://github.com/milenkovicm/datafusion-ballista/blob/4c9b360b51afd3f29614b43b4f3727c473e515a5/ballista/executor/src/executor_server.rs#L726 so it we can unify them at the same time it would be great.

Would something like that make sense?

KR-bluejay: ### 1. Thoughts on the existing function
I looked at the `remove_job_data` function when writing this.  
At that time, I considered splitting it out. Eventually, I decided it would be better to leave that refactoring for another issue and PR, because `remove_job_data` is a Protobuf response handler.

### 2. About unification
Yeah, I could work on unifying the two job data removal functions.  
Would it be okay if I take on this task?

milenkovicm: lets do 2 as a follow up task, for now lets focus on offloading deletion from a pull task, and if we can merge duplicated code as a single function. would that make sense ?

KR-bluejay: Hi, sorry for the delay — I’ve pushed the changes.

I think I should add a test before merge, but I’m not sure how.  
There is a `test_executor_clean_up`, but it seems to cover different functionality, not `remove_job_dir`.

Could you suggest how I should write such a test?

milenkovicm: thanks @KR-bluejay, no delays at all! will have a look later today hopefully

KR-bluejay: Thank you for reviewing this, @milenkovicm — I really appreciate it.

milenkovicm: Which test would you like to add, testing event loop or end to, end to end test or something else?

KR-bluejay: Cleanup logic is tied to both scheduler and executor, so I wasn’t sure how to add a focused test here. 
 
I did verify that job removal works through `integration-test.sh` (end-to-end).

Do you think that is enough for now, or should I also add an additional unit/integration test?

milenkovicm: I think it may be fine for now. I believe we need to redesign file handling, allowing more file types apart from shuffle files, at that point we could revisit whole file deletion process.

KR-bluejay: Got it, thanks! Since it’s fine for now, I’ll mark this PR ready for review.

milenkovicm: I'll believe its good to go, will have another look later. Thanks a lot @KR-bluejay

PR #1318, opened 2025-09-13
feat: Improve Remote Shuffle Read Speed and Resource Utilisation 

### LINKED ISSUE #1315: Improve Remote Shuffle Read Speed and Resource Utilisation
## Executive Summary

By optimizing the Ballista Arrow Flight Server and ballista client implementation, the time required to transfer a 5.5 GB shuffle file can be reduced from 61 seconds to 11 seconds. This represents nearly a sixfold improvement in transfer speed and approximately a threefold reduction in the amount of data transmitted.

## Is your feature request related to a problem or challenge? Please describe what you are trying to do

Ballista `ShuffleWriterExec` writes shuffle files as compressed arrow stream (`CompressionType::LZ4_FRAME`). Shuffle files are read by `ShuffleReaderExec` either locally using local file read or remotely using ballista flight server.

Local read will be performed using `StreamReader`.

Remote reads will be performed using `BallistaClient` which will use arrow flight stream to fetch data from a arrow flight server end point.

Image

`doGet` method implementation at, the flight server side ,  will read arrow stream file, uncompress it decode it (split it in batches) and, decode, compress and ship data over GRPC. This all will take some time, with compress/decompress taking most of the time, as illustrated in

flight server flamegraph
[flight server flamegraph](https://github.com/user-attachments/assets/065b0e1f-e2fa-4d89-aae9-3ce595409b76)

Actual ballista [flight server implementation](https://github.com/apache/datafusion-ballista/blob/4c9b360b51afd3f29614b43b4f3727c473e515a5/ballista/executor/src/flight_service.rs#L88:L120) details.

At the client side we see similar flamegraph where decompression takes most of the time

ballista flight client flamegraph
[ballista flight client flamegraph](https://github.com/user-attachments/assets/3a3f53bd-b4f2-4da9-badf-5888a895879b)

The question to be asked is why do we need to do all the work on the server side when source file is already arrow stream file? One of the possible answers could be: "to comply with `do_get` method semantics".

Modify the Flight server implementation to eliminate shuffle partition decoding and decompressing. Instead, the server should divide the file into smaller, fixed-size chunks (for example, 4 MB byte arrays) and transmit these chunks sequentially over the network. This approach avoids unnecessary decoding and leverages efficient bulk data compression.

Adopting this proposal will alter/replace the existing method, where each transmitted chunk is directly convertible to a `RecordBatch`. In the revised approach, each chunk may correspond to zero or more `RecordBatch` instances, depending on the data boundaries within the chunk. Transmitting file chunks rather than serialized record batches is expected to enhance compression efficiency, as the compression algorithm can operate on larger, contiguous data segments.

Image

On the client side, implement data fetching stream logic and decode it using custom `StreamDecoder`.

This would free arrow flight server of very heavy compression/decompression operation, making it more scalable.

block transfer flight server flamegraph
[block transfer flight server flamegraph](https://github.com/user-attachments/assets/554305f7-6087-4724-a03c-994649e64ed0)

As we can see, server just moves data from file to socket.

At the client side we see similar flamegraph to previous client implementation, decompression is operation taking the most time, but a bit less than in previous case.

block transfer flight client
[block transfer flight client](https://github.com/user-attachments/assets/8d96fe6e-fec4-4b66-a979-18b7f4744170)

This approach has one drawback: it can’t push down filters or column projections because the shuffle server doesn’t decompress and decode the files. Pushdown or projection isn’t supported in the current implementation.

### Experimental Results

A compressed streaming file has been created, taking 5945958784 bytes (~5.5G) to act as shuffle file (~60M rows, 59986051 to be precise).

Running old ballista setup takes 61.67s to transfer whole file, but total transferred bytes is 14802700720 (~14.5GB), looks like compression is not helping much.

On the other hand, new implementation takes 10.75s and transfers 5945958784 bytes (actual file size)!

|                      | Time (sec) | Bytes Transferred  |
| -------------------- |-----------:|-------------------:|
| Old Implementation   | 61         |  14802700720       |
| New Implementation   | 11         |  5945958784        |

- _NOTE 1: this approach uses `LZ4_FRAME` compression._
- _NOTE 2: all tests use local network interface._

### Implementation Details

We have two options to implement it, extend do `do_get(...)` or implement it as `do_action(...)`.

`do_get(...)` returns stream of `FlightData`, which is declared as:

```grpc
/*
 * A batch of Arrow data as part of a stream of batches.
 */
message FlightData {
  /* [removed] ... */
  FlightDescriptor flight_descriptor = 1;
  /* [removed] ... */
  bytes data_header = 2;
  /* [removed] ... */
  bytes app_metadata = 3;
  /*
   * The actual batch of Arrow data. [removed] ...
   */
  bytes data_body = 1000;
}
```

`data_body` property has to contain "The actual batch of Arrow data", our proposal clearly does not provide such guarantee, so extension of `do_get` may break semantic of the interface.

`do_action`, on the other hand, can be extended in any way needed:

```grpc
///
/// Flight services can support an arbitrary number of simple actions in
/// addition to the possible ListFlights, GetFlightInfo, DoGet, DoPut
/// operations that are potentially available. DoAction allows a flight client
/// to do a specific action against a flight service. An action includes
/// opaque request and response objects that are specific to the type action
/// being undertaken.
async fn do_action(
    &self,
    request: tonic::Request<super::Action>,
) -> std::result::Result<tonic::Response<Self::DoActionStream>, tonic::Status>;
/// Server streaming response type for the ListActions method.
type ListActionsStream: tonic::codegen::tokio_stream::Stream<
        Item = std::result::Result<super::ActionType, tonic::Status>,
    >
    + std::marker::Send
    + 'static;

```

which returns type which may be the perfect fit in our case:

```grpc
///
/// An opaque result returned after executing an action.
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Result {
    #[prost(bytes = "bytes", tag = "1")]
    pub body: ::prost::bytes::Bytes,
}
```

additionally to `do_action` we would need to implement `list_actions`

```grcp
///
/// A flight service exposes all of the available action types that it has
/// along with descriptions. This allows different flight consumers to
/// understand the capabilities of the flight service.
async fn list_actions(
    &self,
    request: tonic::Request<super::Empty>,
) -> std::result::Result<
    tonic::Response<Self::ListActionsStream>,
    tonic::Status,
>;
```

Avoiding GPRC protocol would probably be even more performant, but at the moment I haven’t considered it as it would slightly complicate design and implementation and introduce a bigger change. We can re-consider it at some point if needed.

## Additional context

The proposed "block" file transport mechanism should be introduced as a configurable option. It remains to be determined whether this configuration will be applied at compile time or runtime. Further evaluation is required to select the most appropriate approach, ensuring flexibility and compatibility with existing deployment workflows.

This issue may invalidate 

- #932
- #944
- #1189

Possible follow up tasks

- #660

milenkovicm:

Closes #1315

 # Rationale for this change

- changes to arrow flight server to support optimised transport
- changes to ballista client 
- configuration option to disable protocol `ballista.shuffle.force_remote_prefer_flight`

### DIFF
--- a/ballista/executor/src/flight_service.rs
+++ b/ballista/executor/src/flight_service.rs
@@ -21,6 +21,7 @@ use datafusion::arrow::ipc::reader::StreamReader;
 use std::convert::TryFrom;
 use std::fs::File;
 use std::pin::Pin;
+use tokio_util::io::ReaderStream;
 
 use arrow_flight::encode::FlightDataEncoderBuilder;
 use arrow_flight::error::FlightError;
@@ -64,6 +65,8 @@ impl Default for BallistaFlightService {
 
 type BoxedFlightStream<T> =
     Pin<Box<dyn Stream<Item = Result<T, Status>> + Send + 'static>>;
+
+const BUFFER_CAPACITY: usize = 8 * 1024 * 1024;
 
 #[tonic::async_trait]
 impl FlightService for BallistaFlightService {
@@ -185,16 +188,59 @@ impl FlightService for BallistaFlightService {
     ) -> Result<Response<Self::DoActionStream>, Status> {
         let action = request.into_inner();
 
-        let _action = decode_protobuf(&action.body).map_err(|e| from_ballista_err(&e))?;
+        match action.r#type.as_str() {
+            "IO_BLOCK_TRANSFER" => {
+                let action =
+                    decode_protobuf(&action.body).map_err(|e| from_ballista_err(&e))?;
 
-        Err(Status::unimplemented("do_action"))
+                match &action {
+                    BallistaAction::FetchPartition { path, .. } => {
+                        debug!("FetchPartition reading {path}");
+                        let file = tokio::fs::File::open(&path).await.map_err(|e| {
+                            Status::internal(format!("Failed to open file: {e}"))
+                        })?;
+
+                        debug!(
+                            "streaming file: {} with size: {}",
+                            path,
+                            file.metadata().await?.len()
+                        );
+                        let reader =
+                            tokio::io::BufReader::with_capacity(BUFFER_CAPACITY, file);
+                        let file_stream =
+                            ReaderStream::with_capacity(reader, BUFFER_CAPACITY);
+
+                        let flight_data_stream = file_stream.map(|result| {
+                            result
+                                .map(|bytes| arrow_flight::Result { body: bytes.into() })
+                                .map_err(|e| Status::internal(format!("I/O error: {e}")))
+                        });
+
+                        Ok(Response::new(
+                            Box::pin(flight_data_stream) as Self::DoActionStream
+                        ))
+                    }
+                }
+            }
+            action_type => Err(Status::unimplemented(format!(
+                "do_action does not implement: {}",
+                action_type
+            ))),
+        }
     }
 
     async fn list_actions(
         &self,
         _request: Request<Empty>,
     ) -> Result<Response<Self::ListActionsStream>, Status> {
-        Err(Status::unimplemented("list_actions"))
+        let actions = vec![Ok(ActionType {
+            r#type: "IO_BLOCK_TRANSFER".to_owned(),
+            description: "optimized shuffle data transfer".to_owned(),
+        })];
+
+        Ok(Response::new(
+            Box::pin(futures::stream::iter(actions)) as Self::ListActionsStream
+        ))
     }
 
     async fn do_exchange(
--- a/ballista/executor/src/flight_service.rs
+++ b/ballista/executor/src/flight_service.rs
@@ -189,7 +189,7 @@ impl FlightService for BallistaFlightService {
         let action = request.into_inner();
 
         match action.r#type.as_str() {
-            "IO_BLOCK_TRANSFER" => {
+            "IO_BLOCK_TRANSPORT" => {
                 let action =
                     decode_protobuf(&action.body).map_err(|e| from_ballista_err(&e))?;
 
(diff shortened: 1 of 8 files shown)


ballista/executor/src/flight_service.rs
```diff
@@ -185,16 +189,68 @@ impl FlightService for BallistaFlightService {
     ) -> Result<Response<Self::DoActionStream>, Status> {
         let action = request.into_inner();
 
-        let _action = decode_protobuf(&action.body).map_err(|e| from_ballista_err(&e))?;
-
-        Err(Status::unimplemented("do_action"))
+        match action.r#type.as_str() {
+            // Block transfer will transfer arrow ipc file block by block
+            // without decoding or decompressing, this will provide less resource utilization
+            // as file are not decoded nor decompressed/compressed. Usually this would transfer less data across
```
alamb, line 195: FWIW the idea of sending uncompressed streams over gRPC makes a lot of sense to me -- gRCP already has the ability to gzip communications at a higher so there is no need to always force compression. 

Not compressing will likely perform better in environments with fast networks (aka data centers) as you have pointed out. For environments with slower networks, you could still get the benefits of compression by enabling `gzip` compression in the grpc layer.

milenkovicm: just small note @alamb stream is compressed, shuffle file at rest (at local directory) is already compressed. point i was trying to make is, when we have compressed file, there is no need to uncompressed it just to be compressed again. we can just ship file across, block by block where it will be decompressed and decoded

alamb: Even better to remove an entirely redundant layer of compression 👍

PR #1324, opened 2025-10-03
infra: macos-13 is deprecated

kevinjqliu:
# Which issue does this PR close?

Closes #.

 # Rationale for this change

https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

# What changes are included in this PR?

# Are there any user-facing changes?

PR #1325, opened 2025-10-04
chore(infra): remove build-macos-x86_64

kevinjqliu:

Closes #.

 # Rationale for this change

~~Follow up to https://github.com/apache/datafusion-ballista/pull/1324, the correct replacement for `macos-13` should be `macos-15-intel` for Intel-based architecture~~
~~according to 
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories~~

Since both `build-python-mac-win` and `build-macos-x86_64` use `macos-latest`, we can consolidate and refactor the steps to run for individual os

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,7 +107,7 @@ jobs:
 
   build-windows:
     needs: [generate-license]
-    name: Windows
+    name: Windows x86_64
     runs-on: windows-latest
     strategy:
       fail-fast: false
@@ -154,7 +154,7 @@ jobs:
       - name: Archive wheels
         uses: actions/upload-artifact@v4
         with:
-          name: dist-windows
+          name: dist-windows-x86_64
           path: python/target/wheels/*
 
   build-macos-arm64:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,15 +105,14 @@ jobs:
           name: python-wheel-license
           path: LICENSE.txt
 
-  build-python-mac-win:
+  build-windows:
     needs: [generate-license]
-    name: Mac/Win
-    runs-on: ${{ matrix.os }}
+    name: Windows
+    runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.10"]
-        os: [macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v5
 
@@ -147,25 +146,20 @@ jobs:
           uv run --no-project maturin build --release --strip
 
       - name: List Windows wheels
-        if: matrix.os == 'windows-latest'
         run: dir python\target\wheels\
         # since the runner is dynamic shellcheck (from actionlint) can't infer this is powershell
         # so we specify it explicitly
         shell: powershell
 
-      - name: List Mac wheels
-        if: matrix.os != 'windows-latest'
-        run: find python/target/wheels/
-
       - name: Archive wheels
         uses: actions/upload-artifact@v4
         with:
-          name: dist-${{ matrix.os  }}
+          name: dist-windows
           path: python/target/wheels/*
 
-  build-macos-x86_64:
+  build-macos-arm64:
     needs: [generate-license]
-    name: Mac x86_64
+    name: Mac arm64
     runs-on: macos-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -166,7 +166,7 @@ jobs:
   build-macos-x86_64:
     needs: [generate-license]
     name: Mac x86_64
-    runs-on: macos-latest
+    runs-on: macos-15-intel
```
milenkovicm: I'm not sure, why would we test it on intel chipset, arm (latest) makes more sense to me

milenkovicm (commented):
Personally I don't think we should care about Intel Macs at this stage, latest makes more sense to me

.github/workflows/build.yml
kevinjqliu: making this change because this was [originally using `macos-13`](https://github.com/apache/datafusion-ballista/pull/1324/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L169)
macos-15-intel is the direct replacement

kevinjqliu: the name of the step "Mac x86_64" suggest that we should use an intel based processor. 

macos-latest is arm64
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

kevinjqliu: opened this pr to correct the mistake from #1324
we can follow up with removing this github action step is necessary

milenkovicm: I believe intel architecture for Mac is on its way out, let's keep arm, and just rename action, Wdyt?

kevinjqliu: i remove the `build-macos-x86_64` step entirely, the step above, `build-python-mac-win`, already covers mac-latest

milenkovicm: can we rename

```
  build-macos-x86_64:
    needs: [generate-license]
    name: Mac x86_64
    runs-on: macos-latest
```
`build-macos-x86_64` -> `build-macos-arm64`

kevinjqliu: so instead of removing `build-macos-x86_64`, we want to just rename to `build-macos-arm64`?

kevinjqliu:
@milenkovicm i updated the pr to split out windows and mac. lmk if this is what you were thinking about

.github/workflows/build.yml
```diff
@@ -147,25 +146,20 @@ jobs:
...
-
       - name: Archive wheels
         uses: actions/upload-artifact@v4
         with:
-          name: dist-${{ matrix.os  }}
+          name: dist-windows-x86_64
           path: python/target/wheels/*
 
-  build-macos-x86_64:
+  build-macos-arm64:
```
milenkovicm, line 160: this is what i was asking, as name changed after you push update to use arm64 👍🏻

.github/workflows/build.yml
```diff
@@ -105,15 +105,14 @@ jobs:
           name: python-wheel-license
           path: LICENSE.txt
 
-  build-python-mac-win:
+  build-windows:
```
milenkovicm, line 108: I'm not sure why you removed mac from here? is there other job which tests this python on mac

kevinjqliu: yes, `build-macos-arm64`. 

`build-python-mac-win` used to run for windows and macos. but since `build-macos-arm64` already runs for macos, im chaning this to just run windows.

milenkovicm: `build-python-mac-win` builds `ballista/python/` extension, on the other hand `build-macos-arm64` builds core `ballista/` or I'm missing something

kevinjqliu: Screenshot 2025-10-24 at 5 17 46 PM

heres a side by side diff of `build-python-mac-win` and `build-macos-arm64` taken from latest main 
https://github.com/apache/datafusion-ballista/blob/400f818ccd84126ae1733cf6457c592c542d0b63/.github/workflows/build.yml#L108-L213

the only difference is which os is running. the refactor here is to get rid of the `matrix.os` in `build-python-mac-win` since `build-macos-x86_64` is already running for `macos-latest`

milenkovicm:
superseded by #1612

Issue #1258, opened 2025-04-29
Fail the build if public method is missing rustdoc


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

As ballista moved towards library model having a good documentation and examples is becoming imperative. As the first step we should fail the build if public method does not have proper rustdoc.

**Describe the solution you'd like**

I believe there is similar functionality in datafusion repository, we can re-use it 

**Describe alternatives you've considered**

**Additional context**

martin-g:
What do you think about enabling `#![warn(missing_docs)]` and adding `/// TODO` to fix all warnings? 
Later whenever someone works on code with a TODO could improve it.
The benefit is that any new code will be checked and should be documented.

milenkovicm:
i was hoping somebody would "vibe" comment the code :)

PR #1327, opened 2025-10-13
document private items and update docs CI

killzoner:

Closes partially [https://github.com/apache/datafusion-ballista/issues/1258 #](https://github.com/apache/datafusion-ballista/issues/1258).

 # Rationale for this change

https://github.com/apache/datafusion-ballista/issues/1258 would need `#![warn(missing_docs)]`, but there is currently a lot of missing documentation.

This is a first step towards avoid functions not appearing in documentation.

Updating the CI to have (partially) same checks around docs as `datafusion`.

Will document private items, so they will now appear in docs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,9 +46,10 @@ jobs:
     container:
       image: amd64/rust
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: true
+          fetch-depth: 1
       - name: Setup Rust toolchain
         uses: ./.github/actions/setup-builder
       - name: Prepare cargo build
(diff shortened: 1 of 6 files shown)


.github/workflows/rust.yml
```diff
@@ -79,15 +81,53 @@ jobs:
...
 
+  # Run `cargo test doc` (test documentation examples)
+  linux-test-doc:
+    name: cargo test doc (amd64)
+    needs: linux-build-lib
+    runs-on: ubuntu-latest
+    container:
+      image: amd64/rust
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
```
milenkovicm: why do we use hash here and version in others cases ?

killzoner: Sorry that's a miss. Basically `datafusion` uses this hash (with the comment) everywhere in the CI, so I started replicating and figured out it might be cleaner to just point to v5

I will have a look at the failing doc test, I might just remove it for now as I copied it from upstream `datafusion` and it's not serving any specific purpose at least for this PR

killzoner: removed the section

PR #1328, opened 2025-10-13
add msrvcheck

killzoner:

It's an automated way to track msrv with upstream `datafusion`.
Currently cargo checks for msrv when adding a new dependency on latest resolver (https://rust-lang.github.io/rfcs/3537-msrv-resolver.html), not for existing ones. 

 # Rationale for this change

This is one of the items from https://github.com/apache/datafusion-ballista/issues/1271

Add a CI check to make sure MSRV is aligned with upstream `datafusion` projects.

No

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -1,0 +1,58 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Dependencies
+
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches-ignore:
+      - 'gh-readonly-queue/**'
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+  pull_request:
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+  merge_group:
+  # manual trigger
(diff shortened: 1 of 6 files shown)


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,58 @@
...
+  workflow_dispatch:
+
+jobs:
+  depcheck:
+    name: msrv dependency check
+    runs-on: ubuntu-latest
+    container:
+      image: amd64/rust
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
```
milenkovicm: can you please replace hashes with versions here as well? thanks

killzoner: doing it in https://github.com/apache/datafusion-ballista/pull/1327 (sorry 2 concurrent PRs :smile: )

dev/msrvcheck/src/main.rs
```diff
@@ -0,0 +1,69 @@
...
+    let (_, resolve) = cargo::ops::resolve_ws(&workspace, false)?;
+    let packages_with_rust_version: Vec<_> = resolve
+        .iter()
+        .filter(|id| id.name().starts_with("datafusion"))
+        .map(|e| resolve.summary(e))
+        .map(|e| (e.name(), e.rust_version()))
+        .collect();
+
+    println!("Current project MSRV: {}", project_msrv);
+
```
milenkovicm, line 56: can we add print statement to print datafusion MSRV as well? it would help with tracking datafsuion version on update

killzoner: will do

killzoner: done

.github/workflows/dependencies.yml
milenkovicm: are you sure this will be covered in that other PR? this is new job definition

killzoner: I'm gonna make sure to rebase / fix the thing once the other PR is merged.

Either ways, i'm expecting some kind of git conflict, and I'm keeping this comment as unresolved to make sure it's addressed if no conflict

killzoner: Rebased and cleaned

milenkovicm:
> Example of failing check with MSRV downgraded: 
> 
> https://github.com/killzoner/datafusion-ballista/commit/6d7022f9f196f605b0eb4e91c947ef903263ee40
> 
> https://github.com/killzoner/datafusion-ballista/actions/runs/18509960270/job/52748285312#step:5:636
> 
> 
> 
> ```bash
> 
> Current project MSRV: 1.85.0
> 
> thread 'main' panicked at src/main.rs:60:17:
> 
> package 'datafusion' has MSRV 1.86.0 not compatible with current project MSRV 1.85.0
> 
> ```

I see, should it be other way around I wonder. We need to support MSRV which datafusion supports or older

killzoner:
> thanks @killzoner

Thank you for the detailed reviews

PR #1333, opened 2025-10-24
[Client]: Sync scheduler's catalog metadata to client, to allow planning against remote tables

mach-kernel:
### Which issue does this PR close?
TBD

### Rationale for this change
- A Ballista scheduler may have tables in its catalog that the client is unaware of. Currently, `show tables` only produces what is available in the client's local session context. This change allows the client to plan queries against schemas available in the Ballista cluster but not the client.
- AFAICT there is no way to load e.g. a custom table or catalog provider at runtime. This approach could let any Ballista client query clusters with custom table providers (and eventually UDFs) without having to resolve them clientside.

###### ballista
- New protos to represent the hierarchy of catalog -> schema -> table
- Scheduler RPC to pack the catalog names into protos
- `SessionContextExt::remote*` calls scheduler and loads tables using `MemoryCatalogProvider`
- A custom `RemoteTableProvider` that encapsulates names (catalog, schema, table), and a concrete Arrow schema
  - `RemoteTableProvider::scan` errors -- we should never try to actually scan it.
- `BallistaLogicalExtensionCodec` treats `RemoteTableProvider` as a special case
  - Encode just packs into the table provider proto representation
  - Decode makes a concrete table reference and attempts to resolve it in the remote context

###### ballista-python
- Add to workspace to consume new client changes

### Are there any user-facing changes?

`ballista-cli` can now run queries against cluster tables!

```bash
Ballista CLI v49.0.0
❯ show tables;
+---------------+--------------------+---------------+------------+
| table_catalog | table_schema       | table_name    | table_type |
+---------------+--------------------+---------------+------------+
| spice         | runtime            | task_history  | BASE TABLE |
| spice         | public             | wiki_a_potion | BASE TABLE |
| spice         | information_schema | tables        | VIEW       |
| spice         | information_schema | views         | VIEW       |
| spice         | information_schema | columns       | VIEW       |
| spice         | information_schema | df_settings   | VIEW       |
| spice         | information_schema | schemata      | VIEW       |
| spice         | information_schema | routines      | VIEW       |
| spice         | information_schema | parameters    | VIEW       |
| datafusion    | information_schema | tables        | VIEW       |
| datafusion    | information_schema | views         | VIEW       |
| datafusion    | information_schema | columns       | VIEW       |
| datafusion    | information_schema | df_settings   | VIEW       |
| datafusion    | information_schema | schemata      | VIEW       |
| datafusion    | information_schema | routines      | VIEW       |
| datafusion    | information_schema | parameters    | VIEW       |
+---------------+--------------------+---------------+------------+
18446744073709551615 row(s) fetched.
Elapsed 0.016 seconds.

❯ select id, title from spice.public.wiki_a_potion limit 2;
+----------+----------+
| id       | title    |
+----------+----------+
| 4549720  | Al Salvi |
| 36963017 | Al Samha |
+----------+----------+
18446744073709551615 row(s) fetched.
Elapsed 0.777 seconds.
```

And with the Python client, in notebooks too!!

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,15 +51,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -79,7 +70,7 @@ checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
  "const-random",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -116,12 +107,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -132,9 +117,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.20"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -147,9 +132,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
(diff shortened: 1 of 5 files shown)


milenkovicm:
First of all thanks for contribution @mach-kernel! I'll try to support you as much as I can with your requirement. 

I'll start with a bit of history, it will help me to explain core direction. Last year we have started trimming down ballista to the shape it has today. Previously ballista was a one size fit all solution, it had a lot of code which was siting in the repo to support various very specific use cases. Instead of fit all solution we have decided to make it more generic, with ability to override or add required code to support specific cases which may not be seen as generic for which we can allocate effort to maintain it. This helps  to reduce burden on maintainers.

If there is use case specific behaviour needed, users can change and compile its own client, scheduler or/and executor. Main reason was, as you state it in discord discussion, we're unable just to drop a jar on the class path. This way user can rely on functionality provided by the core ballista library but extend it in a way to support its own use case. 

There are few examples of extensions of core functionalities, but I would say it's not documented as much as it needs to be. I have created few more show case projects [ballista python](https://github.com/milenkovicm/ballista_python), [ballista extensions](https://github.com/milenkovicm/ballista_extensions) & [ballista delta](https://github.com/milenkovicm/ballista_delta) to demonstrate how to extend ballista to fit specific use case, I'm not sure if they will help. 

You have mentioned UDF, at the moment there are few different approaches, not of them perfect, and from perspective of ballista, we hope we can support all of them, but we do not want to maintain them. 

Regarding your code, I will have a better look, but at the moment most of the things look like they can be implemented out of the core library. you could create your own extensions codecs to support your specific tables. Maybe the missing part which could be added is registering additional (GRPC) service(es) in addition to core scheduler service, which could support centralised schema location. 

Let me know what you think,

mach-kernel:
> If there is use case specific behaviour needed, users can change and compile its own client, scheduler or/and executor. Main reason was, as you state it in discord discussion, we're unable just to drop a jar on the class path. This way user can rely on functionality provided by the core ballista library but extend it in a way to support its own use case.

Ah, sorry, I had no idea this was how it was intended to be used. The example repos are helpful, thank you for linking them.

> I will have a better look, but at the moment most of the things look like they can be implemented out of the core library. you could create your own extensions codecs to support your specific tables. Maybe the missing part which could be added is registering additional (GRPC) service(es) in addition to core scheduler service, which could support centralised schema location.

My thoughts are: in general the client is useful (e.g., vs Flight SQL endpoint) for the ergonomics of the DataFrame API, or secondarily running a 'bad query' (like selecting a col that does not exist) can error immediately in the client vs having to make the request to find out. 

To make a logical plan on the client it feels like all we need is a schema for a given table ref to do this. And to run UDFs (outside of packing Python UDFs) that may be present in the scheduler/executor runtime but not the client, if we know its signature and return type, we can do the same. On encountering a stub, the scheduler looks up the table or function in its runtime. The additional RPC are around getting these 'bare minimum' shapes over to the client.

It would be a really nice out of box experience if the client could run queries against any remote cluster without needing to worry about all of its runtime customizations. It's convenient to get the default lib off of PyPI and/or not worry about shipping a client as you change the cluster runtime. 

I think this is how the client/Python lib would be most commonly used (with a persistent remote cluster). Maybe it could be a new 'remote mode'? The default client could still continue to not care about the concrete implementations of custom tables, fns, etc, outside of what is shipped with the default distribution.

Re the Python library: if the client worked like this plus a solution was figured out for Python UDFs, then that would (at least for me) feel very useful. 

Hope this makes sense. I am not sure of what the limitations are to this idea (one is that it assumes the scheduler/executor runtimes are the same). Thanks for getting back to me, I appreciate it!

milenkovicm:
> > I don't disagree with what you say, we only differ in opinion who should provide this. I believe you could extend current ballista to do this for you (or you could do it if we add ability to register additional gprc services as I have mentioned)
> 
> I have one tiny disagreement here: providing an RPC extension hook is fine, but it doesn't change the developer experience much. In other words I would still need to compile/vendor a scheduler server with my custom extension to be able to do this with the client.

I'm not sure I understand. What would be different than providing a GRPC service hook with your custom extension ? 
Your custom extension would register additional GRPC service with the current scheduler GRPC, you would have to compile scheduler even if you would add your extension to existing GPRC to existing `get_catalog`.

PR #1337, opened 2025-10-25
feat: make gRPC timeout configurations user-configurable

### LINKED ISSUE #117: Make gRPC timeouts user-configurable
PR https://github.com/apache/arrow-ballista/pull/115 specifies timeouts but they are hard-coded.

I would like these to be configurable using the extensible config mechanism that was recently added in DataFusion but has not yet been integrated with Ballista. Adding this would allow users to configure DataFusion-specific options in Ballista as well.

None

**Additional context**
None

CuteChuanChuan:
Closes #117

# Rationale for this change
PR #115 introduced gRPC timeout configurations, but these values were hard-coded. This PR extends #115 by making all gRPC timeout values user-configurable through DataFusion's extensible configuration mechanism.

- Added 9 new configuration options to `BallistaConfig` (5 client-side, 4 server-side)
- Modified `create_grpc_client_connection()` and `create_grpc_server()` to accept config parameter
- Updated all call sites across executor and scheduler

**Yes.** Users can now configure gRPC timeouts.

All existing tests pass.

### DIFF
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -54,7 +54,7 @@ use ballista_core::serde::{
 };
 use ballista_core::utils::{
     create_grpc_client_connection, create_grpc_server, default_config_producer,
-    get_time_before, GrpcClientConfig, GrpcServerConfig,
+    get_time_before, GrpcServerConfig,
 };
 use ballista_core::{ConfigProducer, RuntimeProducer, BALLISTA_VERSION};
 
--- a/ballista/executor/src/executor_server.rs
+++ b/ballista/executor/src/executor_server.rs
@@ -44,9 +44,7 @@ use ballista_core::serde::scheduler::from_proto::{
 use ballista_core::serde::scheduler::PartitionId;
 use ballista_core::serde::scheduler::TaskDefinition;
 use ballista_core::serde::BallistaCodec;
-use ballista_core::utils::{
-    create_grpc_client_connection, create_grpc_server, GrpcClientConfig,
-};
+use ballista_core::utils::{create_grpc_client_connection, create_grpc_server};
 use dashmap::DashMap;
 use datafusion::execution::TaskContext;
 use datafusion_proto::{logical_plan::AsLogicalPlan, physical_plan::AsExecutionPlan};


ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -238,6 +238,7 @@ impl<T: 'static + AsLogicalPlan> ExecutionPlan for DistributedQueryExec<T> {
                 self.session_id.clone(),
                 query,
                 self.config.default_grpc_client_max_message_size(),
+                self.config.clone(),
```
milenkovicm: we already have `gpc_max_message_size` as a parameter can we pass other needed values as parameters rather than passing config

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -274,10 +275,11 @@ async fn execute_query(
     session_id: String,
     query: ExecuteQueryParams,
     max_message_size: usize,
+    config: BallistaConfig,
```
milenkovicm: we already have gpc_max_message_size as a parameter can we pass other needed values as parameters rather than passing config

ballista/core/src/config.rs
```diff
@@ -39,6 +39,28 @@ pub const BALLISTA_SHUFFLE_READER_FORCE_REMOTE_READ: &str =
...
+pub const BALLISTA_GRPC_CLIENT_TIMEOUT_SECONDS: &str =
+    "ballista.grpc.client.timeout_seconds";
+pub const BALLISTA_GRPC_CLIENT_TCP_KEEPALIVE_SECONDS: &str =
+    "ballista.grpc.client.tcp_keepalive_seconds";
+pub const BALLISTA_GRPC_CLIENT_HTTP2_KEEPALIVE_INTERVAL_SECONDS: &str =
+    "ballista.grpc.client.http2_keepalive_interval_seconds";
+pub const BALLISTA_GRPC_CLIENT_KEEPALIVE_TIMEOUT_SECONDS: &str =
+    "ballista.grpc.client.keepalive_timeout_seconds";
+
+// gRPC server timeout configurations
```
milenkovicm: this config is relevant to `standalone`  version only can we prefix it somehow to identify it as standalone option?

ballista/executor/src/executor_server.rs
```diff
@@ -114,7 +117,7 @@ pub async fn startup<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>(
         let mut grpc_shutdown = shutdown_noti.subscribe_for_shutdown();
         tokio::spawn(async move {
             let shutdown_signal = grpc_shutdown.recv();
-            let grpc_server_future = create_grpc_server()
+            let grpc_server_future = create_grpc_server(&ballista_config)
```
milenkovicm: this configuration should be provided as part of executor binary configuration, not ballista condif

ballista/core/src/utils.rs
```diff
@@ -106,31 +107,50 @@ pub async fn collect_stream(
 
 pub async fn create_grpc_client_connection<D>(
     dst: D,
+    config: &BallistaConfig,
```
milenkovicm: as i mentioned before, IMHO it makes more sense to pass individual parameters, or we can derive some kind of `GprcConfig` if there are too many of them

ballista/core/src/utils.rs
```diff
@@ -106,31 +107,50 @@ pub async fn collect_stream(
...
+        ))
+        .keep_alive_timeout(Duration::from_secs(
+            config.default_grpc_client_keepalive_timeout_seconds() as u64,
+        ))
         .keep_alive_while_idle(true);
     endpoint.connect().await
 }
 
-pub fn create_grpc_server() -> Server {
+pub fn create_grpc_server(config: &BallistaConfig) -> Server {
```
milenkovicm: as i mentioned before, IMHO it makes more sense to pass individual parameters, or we can derive some kind of GprcConfig if there are too many of them

CuteChuanChuan:
Hi @milenkovicm ,
Thanks for your comments and pointing out better approach. PTAL when you have a chance.
Thanks.

ballista/core/src/utils.rs
CuteChuanChuan: I derive `GrpcServerConfig` for this.

ballista/core/src/utils.rs
CuteChuanChuan: I derive `GrpcClientConfig` for this.

ballista/client/tests/context_checks.rs
```diff
@@ -240,12 +240,12 @@ mod supported {
...
         let expected = [
-            "+---------------------------------------+----------+",
-            "| name                                  | value    |",
-            "+---------------------------------------+----------+",
-            "| ballista.grpc_client_max_message_size | 16777216 |",
-            "| ballista.job.name                     |          |",
-            "+---------------------------------------+----------+",
+            "+-------------------------------------------------------+-------+",
+            "| name                                                  | value |",
+            "+-------------------------------------------------------+-------+",
```
milenkovicm: can we please update query to return ` ballista.job.name` only. so we don't change the test every time we add new option.

ballista/core/src/config.rs
```diff
@@ -39,6 +39,18 @@ pub const BALLISTA_SHUFFLE_READER_FORCE_REMOTE_READ: &str =
...
+// gRPC client timeout configurations
+pub const BALLISTA_GRPC_CLIENT_CONNECT_TIMEOUT_SECONDS: &str =
+    "ballista.grpc.client.connect_timeout_seconds";
+pub const BALLISTA_GRPC_CLIENT_TIMEOUT_SECONDS: &str =
+    "ballista.grpc.client.timeout_seconds";
+pub const BALLISTA_GRPC_CLIENT_TCP_KEEPALIVE_SECONDS: &str =
+    "ballista.grpc.client.tcp_keepalive_seconds";
+pub const BALLISTA_GRPC_CLIENT_HTTP2_KEEPALIVE_INTERVAL_SECONDS: &str =
+    "ballista.grpc.client.http2_keepalive_interval_seconds";
+pub const BALLISTA_STANDALONE_GRPC_CLIENT_KEEPALIVE_TIMEOUT_SECONDS: &str =
```
milenkovicm: Do we need standalone config? Standalone is used just for local testing, maybe if we can live without it. It would make configuration options simpler. WDYT?

ballista/core/src/utils.rs
```diff
@@ -36,6 +37,64 @@ use std::{fs::File, pin::Pin};
...
+pub struct GrpcClientConfig {
+    pub connect_timeout_seconds: u64,
+    pub timeout_seconds: u64,
+    pub tcp_keepalive_seconds: u64,
+    pub http2_keepalive_interval_seconds: u64,
+    pub keepalive_timeout_seconds: u64,
+}
+
+impl GrpcClientConfig {
+    pub fn from_ballista_config(config: &BallistaConfig) -> Self {
```
milenkovicm: would it make sense to implement this as `From` or `Into` not sure which one would be better

ballista/core/src/utils.rs
```diff
@@ -36,6 +37,64 @@ use std::{fs::File, pin::Pin};
...
+            timeout_seconds: 20,
+            tcp_keepalive_seconds: 3600,
+            http2_keepalive_interval_seconds: 300,
+            keepalive_timeout_seconds: 20,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct GrpcServerConfig {
```
milenkovicm, line 107: can you please put some docs please, we will try to catch up with documentation at some point so it would help.

ballista/core/src/utils.rs
```diff
@@ -36,6 +37,64 @@ use std::{fs::File, pin::Pin};
 use tonic::codegen::StdError;
 use tonic::transport::{Channel, Error, Server};
 
+#[derive(Debug, Clone)]
+pub struct GrpcClientConfig {
```
milenkovicm, line 56: can you please put some docs please, we will try to catch up with documentation at some point so it would help.

ballista/executor/src/standalone.rs
```diff
@@ -123,8 +123,9 @@ pub async fn new_standalone_executor_from_builder(
         .max_decoding_message_size(max_message_size)
         .max_encoding_message_size(max_message_size);
 
+    let grpc_server_config = GrpcServerConfig::default();
```
milenkovicm: can you avoid variable here and create config at method call 

```rust
create_grpc_server(&GrpcServerConfig::default();)
```

ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -427,7 +429,9 @@ impl ExecutorManager {
                 "http://{}:{}",
                 executor_metadata.host, executor_metadata.grpc_port
             );
-            let connection = create_grpc_client_connection(executor_url).await?;
+            let grpc_config = GrpcClientConfig::default();
```
milenkovicm: can this be method parameter, so we can use `BallistaConfig` value instead of default values here?

ballista/scheduler/src/standalone.rs
```diff
@@ -92,8 +93,9 @@ pub async fn new_standalone_scheduler_with_builder(
     let listener = TcpListener::bind("localhost:0").await?;
     let addr = listener.local_addr()?;
     info!("Ballista v{BALLISTA_VERSION} Rust Scheduler listening on {addr:?}");
+    let grpc_server_config = GrpcServerConfig::default();
     tokio::spawn(
-        create_grpc_server()
+        create_grpc_server(&grpc_server_config)
```
milenkovicm: I would suggest

```rust
create_grpc_server(&GrpcServerConfig::default();)
```

if possible

milenkovicm (commented):
Firs of all sorry for late review, and thank you for your contribution ! 
I think changes make sense, I have just few minor comments, please have a look when you get chance

ballista/core/src/config.rs
CuteChuanChuan: Thanks for pointing this out. You're right. We can make configuration options simpler.

ballista/core/src/utils.rs
CuteChuanChuan: Considering that GrpcClientConfig is derived by extracting specific configs from BallistaConfig, I think `From` is more appropriate than `Into` here. WDYT?

ballista/core/src/utils.rs
CuteChuanChuan: Absolutely!

ballista/core/src/utils.rs
CuteChuanChuan: Absolutely!

ballista/core/src/utils.rs
milenkovicm: `From` make sense, it will implement into as well, if i'm not mistaken

ballista/executor/src/executor_process.rs
```diff
@@ -245,8 +249,11 @@ pub async fn start_executor_process(
     ));
 
     let connect_timeout = opt.scheduler_connect_timeout_seconds as u64;
+    let session_config = (executor.config_producer)();
+    let ballista_config = session_config.ballista_config();
+    let grpc_client_config = GrpcClientConfig::from(&ballista_config);
     let connection = if connect_timeout == 0 {
-        create_grpc_client_connection(scheduler_url)
+        create_grpc_client_connection(scheduler_url, &grpc_client_config)
```
milenkovicm: could we have 

```rust
create_grpc_client_connection(scheduler_url, &ballista_config.into())
```

it would remove 

```rust
let grpc_client_config = GrpcClientConfig::from(&ballista_config);
```

making it a bit easier to read?

ballista/executor/src/executor_server.rs
```diff
@@ -235,7 +239,11 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorServer<T,
             Ok(scheduler)
         } else {
             let scheduler_url = format!("http://{scheduler_id}");
-            let connection = create_grpc_client_connection(scheduler_url).await?;
+            let session_config = (self.executor.config_producer)();
+            let ballista_config = session_config.ballista_config();
+            let grpc_client_config = GrpcClientConfig::from(&ballista_config);
+            let connection =
+                create_grpc_client_connection(scheduler_url, &grpc_client_config).await?;
```
milenkovicm: could we have 

```rust
create_grpc_client_connection(scheduler_url, &ballista_config.into())
```

it would remove 

```rust
let grpc_client_config = GrpcClientConfig::from(&ballista_config);
```

making it a bit easier to read?

milenkovicm (approved):
i have just two small coments, if you agree

ballista/executor/src/executor_process.rs
CuteChuanChuan: Hi @milenkovicm ,

Good catch! However, since the `From` trait is implemented for `&BallistaConfig` (not `BallistaConfig`), I need `&(&ballista_config).into()` instead of `&ballista_config.into()`.

Would you prefer I change the implementation to `impl From<BallistaConfig>` so we can use the simpler `&ballista_config.into()` syntax?

CuteChuanChuan:
Hi @milenkovicm ,

I've addressed the feedback about simplifying the config usage and error found in Clippy. However, I noticed the CI is failing with rustdoc errors in `display.rs` and `execution_graph.rs`, which are files I haven't modified in this PR.

Would you like me to:
  1. Fix them in this PR as well, or
  2. Create a separate issue/PR for these rustdoc fixes?

milenkovicm:
thanks on your contribution @CuteChuanChuan

PR #1338, opened 2025-10-30
feat: New ballista python interface

### LINKED ISSUE #1142: Ballista Python Issue(s)
First of all, I'm not expert in rust-python (pyo3) integration, if I've done/said something stupid,
my apologies.

Current implementation of (py)ballista has limitation when it comes to `DataFrame` operations.

following code will result with an error:

```python
from pyballista import BallistaBuilder
from datafusion import SessionContext
from datafusion import functions as f

# %%
ctx: SessionContext = BallistaBuilder()\
    .config("ballista.job.name", "example ballista")\
    .config("ballista.shuffle.partitions", "16")\
    .standalone()
    
df = ctx.sql("SELECT 1 as r").aggregate(
    [f.col("r")], [f.count_star()]
)
df.show()
```

it will throw exception (similar to):

```text
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File /Users/user/git/arrow-ballista/python/examples/example.py:3
      1 # %% 
      2 # Select 1 to verify its working
----> 3 df = ctx.sql("SELECT 1 as r").aggregate(
      4     [f.col("r")], [f.count_star()]
      5 )
      6 df.show()

TypeError: argument 'group_by': 'Expr' object cannot be converted to 'Expr'
```

Actually previous implementation had the same problem, the same error will be thrown (`git checkout 2f223db21557c15080bf865ac692d276b8f0b770`)

```python
# %%
from pyballista import SessionContext
from datafusion import functions as f

ctx = SessionContext("localhost", 50050)

df = ctx.sql("SELECT 1 as r").aggregate(
    [f.col("r")], [f.count_star()]
)
df.show()
```

The similar issue is there if `SessionConfig` is used:

```python
from ballista import Ballista, RuntimeConfig, SessionConfig
from datafusion import RuntimeConfig, SessionConfig, SessionContext

runtime = RuntimeConfig().with_disk_manager_os().with_fair_spill_pool(10000000)
config = (
    SessionConfig()
    .with_create_default_catalog_and_schema(True)
    .with_default_catalog_and_schema("foo", "bar")
    .with_target_partitions(8)
    .with_information_schema(True)
    .with_repartition_joins(False)
    .with_repartition_aggregations(False)
    .with_repartition_windows(False)
    .with_parquet_pruning(False)
    .set("datafusion.execution.parquet.pushdown_filters", "true")
)

# %%
ctx: SessionContext = Ballista.builder\
    .with_runtime(runtime)\ # it will panic at this point, complaining that `RuntimeConfig` object cannot be converted to `RuntimeConfig`
    .with_config(config)\
    .standalone()

ctx.sql("SELECT 1").show()
```

problem with `RuntimeConfig`, `SessionConfig` could be solved if they are re-exported in ballista:

```python
from ballista import Ballista, RuntimeConfig, SessionConfig
from datafusion import SessionContext
```

but the first problem with `DataFrame` would still remain.

My guess is that there is FFI issue as ballista and datafusion is different package, I'm not sure what the problem is nor how to resolve this issue.

@timsaucer comment <https://github.com/apache/datafusion-ballista/issues/1091#issuecomment-2436167064> make more sense to me now.

### Possible Solution (I)

One obvious way would be to move ballista context creation to datafusion-python. We need one line context creation:

```rust
let ctx = datafusion::prelude::SessionContext::remote("df://localhost:50050").await?;
```

As ballista context is the `SessionContext` it would be trivial to integrate, and, I believe, it would avoid previous issues.

We could only provide "remote context" (no standalone), making it optional feature for which users python datafusion users could to opt in. This would somewhat limit number of libraries ballista would bring to datafusion-python (we could split core to core and client-core to further reduce deps)

This proposal would mean that we would have to bring optional dependency to datafusion-python, and additional complexity in (datafusion-python) release process.

(py)ballista would stay, it could expose scheduler and executor control as proposed in <https://github.com/apache/datafusion-ballista/issues/1107>

Big risk for of this proposal is that ballista could block datafusion python release in case it goes back to unmaintained mode.

### Possible Solution (II)

Another possible solution is to re-export all classes from datafusion-python in ballista. I'm not sure how complex or practical this is going to be.
I'm not sure if datafusion python applications would need some kind of re-writing to be able to run on ballista.

This would put additional responsibility to ballista maintainers (not too many of them).

### Any Other Solution?

I'm not sure, open to suggestions

## Proposal

Short term proposal:

- Remove test used to test (py)ballista code on ballista commit <https://github.com/apache/datafusion-ballista/blob/81cfa632f94ef794cc9f35c81676cf9a010c1dbe/.github/workflows/rust.yml#L121-L122> .
- update datafusion to latest <https://github.com/apache/datafusion-ballista/pull/1125> .
- focus on ballista (rust) release <https://github.com/apache/datafusion-ballista/issues/974> .

We should release (py)ballista once we figure out the best approach to fix it.

milenkovicm:

Closes #1142

 # Rationale for this change

For quite some time, we wanted to provide a Ballista Python interface and make it an extension of DataFusion Python. For the reasons mentioned in #1142, we haven't been able to do so. The main issue was that we could not use DataFrame as there was a class mismatch, something like 

```python
from pyballista import BallistaBuilder
from datafusion import SessionContext
from datafusion import functions as f

# %%
ctx: SessionContext = BallistaBuilder()\
    .config("ballista.job.name", "example ballista")\
    .config("ballista.shuffle.partitions", "16")\
    .standalone()
    
df = ctx.sql("SELECT 1 as r").aggregate(
    [f.col("r")], [f.count_star()]
)
df.show()
``` 

was not possible due to FFI boundary between datafusion and ballista python. 

We also wanted to reuse and rely on datafusion python as much as possible, as we do not have bandwidth to maintain duplicated effort. 

This PR relies on python duck typing, to "fake" `DataFrame` interface, with `DistributedDataFrame` extension which would execute query on ballista cluster.

```python
from ballista import BallistaSessionContext
from datafusion import col, lit, DataFrame
from datafusion import functions as f

# we replace 
# ctx = SessionContext()
# with
ctx = BallistaSessionContext(address="df://127.0.0.1:50050")

df : DataFrame = ctx.table("t")
df.filter(col("id") > lit(4)).show()

df0 = ctx.sql("SELECT 1 as r")

df0.aggregate(
    [f.col("r")], [f.count_star()]
)

df0.show()
```

There is slight inneficiency where original logical plan will be serialised in datafusion python and deserialised in ballista python in order to cross FFI boundary as well as re-creation of BallistaContext

Also, we would need to override few methods to make it work with ballista.

Things left to do / address

- [x] write method support
- [x] printing in jupiter
- [x] collect_partitioned
- [x] standalone support
- [ ] propagate SessionConfig across 
- [x] keep session_id between calls 
- [x] create test cluster
- [x] add test
- [x] deprecate/remove old interface
- [ ] ...

Note of caution: there are too many sharp edges, but so far it looks as most promising approach 

There will be change in interface, but too early to tell


python/python/ballista/extension.py
```diff
@@ -0,0 +1,114 @@
...
+        super().__init__(df.df)
+        self.address = address
+        self.session_id = session_id
+
+    def _to_internal_df(self):
+        blob_plan = self.logical_plan().to_proto()
+        df = PyBallistaRemoteExecutor.create_data_frame(
+            blob_plan, self.address, self.session_id
+        )
+        return df
```
timsaucer: This is the main idea, right? That you expose to the user a python class that uses the datafusion class as a Base and then extend on top of it. Did I get that right?

I think this is  good approach. The one concern I have is users who accidentally use a mixture of DataFusion objects and Ballista extensions. The way I could imaging this happening was that someone has some portion of their code base that they've copied over from their single node work.

The *other* approach, and the one that I think in the long term better would be to make it possible for Ballista to just replace the execution plan via a physical plan optimizer.

andygrove: +1 for replacing the physical planning/optimizer path to use Ballista instead of DataFusion

milenkovicm: physical planner would be the best option, not sure if with @timsaucer FFI effort we're closer to get there

andygrove (approved):
Thanks @milenkovicm

milenkovicm:
I believe the Ballista needs a query planner. Please let me know Tim, I’m a bit stuck with time but I’ll try to help.

PR #1351, opened 2025-12-18
feat: Add arrow flight proxy to scheduler

### LINKED ISSUE #1349: Clarify usage of advertise_flight_sql_endpoint
Hi,

I hope creating an issue for this is okay...
Is there some more documentation / an example available for the use of the `advertise_flight_sql_endpoint` scheduler config?

I was trying run ballista in docker-compose and connecting from the outside, with code based on https://github.com/milenkovicm/ballista_delta.
When fetching results, ballista seems to connect to the executor that produced the result.

My understanding of the docs on `advertise_flight_sql_endpoint` is that this would allow fetching the result from the scheduler?
Just setting `advertise_flight_sql_endpoint` to "localhost:12345" didn't show any effect.
Is there any example available that demonstrates how this is intended to be used/configured?

I also tried searching the codebase of usage of this config - but actually only found it used in the config.

sebbegg:

Closes #1349

 # Rationale for this change

See #1349. Having a proxy on the scheduler makes it easier to e.g. expose this in docker-compose or kubernetes.

Making this a draft first, let me know what you think.

This adds a "flight proxy" service to the scheduler that's optionally startet, when `advertise_flight_sql_endpoint` is set.
This implements only `do_get` and simply proxies the requests to the actual executor.
I reasoned that having this as a separate service (instead of another method in the scheduler grpc) makes this more flexible since the code on the client side remains almost unchanged - except for the logic to pick scheduler or executor host as the endpoint.

Kind of - using `advertise_flight_sql_endpoint` now actually has an effect.

### DIFF
--- a/ballista/scheduler/src/config.rs
+++ b/ballista/scheduler/src/config.rs
@@ -41,6 +41,8 @@ pub struct Config {
     /// Route for proxying flight results via scheduler (IP:PORT format).
     #[arg(
         long,
+        num_args = 0..=1,
+        default_missing_value = "",
         help = "Route for proxying flight results via scheduler. Use 'HOST:PORT' to let clients fetch results from the specified address. If empty a flight proxy will be started on the scheduler host and port."
     )]
     pub advertise_flight_sql_endpoint: Option<String>,


milenkovicm:
Perhaps @martin-g may be able to help with quick review, I'd be very thankful

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -360,9 +363,22 @@ async fn execute_query(
                 let duration = Duration::from_millis(duration);
 
                 info!("Job {job_id} finished executing in {duration:?} ");
+                let FlightEndpointInfo {
+                    address: flight_proxy_address,
+                } = scheduler
+                    .get_flight_endpoint_info(FlightEndpointInfoParams {})
+                    .await
+                    .map_err(|e| DataFusionError::Execution(format!("{e:?}")))?
```
martin-g: Thinking out loud: Does it need to fail here or it is OK to log it and fallback to no-proxy ?

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+async fn get_flight_client(
+    host: &String,
+    port: &u16,
+) -> Result<FlightServiceClient<tonic::transport::channel::Channel>, BallistaError> {
+    let addr = format!("http://{host}:{port}");
+    let grpc_config = GrpcClientConfig::default();
+    debug!("FlightProxyService connecting to {addr}");
+    let connection = create_grpc_client_connection(addr.clone(), &grpc_config)
```
martin-g, line 175: Since the host & port come from client provided ticket I think it would be good to check somehow that they point to a registered executor. Otherwise it might be used for Server-Side Request Forgery attack.

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,157 @@
...
+        .await
+        .map_err(|e| {
+            BallistaError::GrpcConnectionError(format!(
+                "Error connecting to Ballista scheduler or executor at {addr}: {e:?}"
+            ))
+        })?;
+    let flight_client = FlightServiceClient::new(connection)
+        .max_decoding_message_size(16 * 1024 * 1024)
```
martin-g: Can/should we use `config.grpc_server_max_encoding_message_size` or a new setting ?
Same for `min` below.

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,157 @@
...
+        Err(Status::unimplemented("list_actions"))
+    }
+}
+
+fn from_ballista_err(e: &ballista_core::error::BallistaError) -> Status {
+    Status::internal(format!("Ballista Error: {e:?}"))
+}
+
+async fn get_flight_client(
+    host: &String,
```
martin-g:
```suggestion
    host: &str,
```

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,157 @@
...
+    port: &u16,
+) -> Result<FlightServiceClient<tonic::transport::channel::Channel>, BallistaError> {
```
martin-g, line 172: Consider connection pooling. Currently a new client is created for each request.

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -131,7 +176,19 @@ pub async fn start_server(
...
 
-    start_grpc_service(address, scheduler).await
+    info!(
+        "advertise_flight_sql_endpoint: {:?}",
+        config.advertise_flight_sql_endpoint
+    );
+    match config.advertise_flight_sql_endpoint {
+        Some(_) => {
+            info!("Starting flight proxy");
+            let _flight_proxy = start_flight_proxy_server(config);
```
martin-g: This ignores any errors returned by the the proxy. They are logged, so this might be intentional - to continue working even without the proxy.

If both services should fail if either of them fail then you could use `select!()`. E.g.:
```diff
-            let _flight_proxy = start_flight_proxy_server(config);
-            start_grpc_service(address, scheduler).await
+            let flight_proxy = start_flight_proxy_server(config);
+            tokio::select! {
+                result = start_grpc_service(address, scheduler) => result,
+                result = flight_proxy => {
+                    result.map_err(|e| BallistaError::Internal(format!("Flight proxy task panicked: {e:?}")))?
+                }
+            }
```

Also there is a racing here - the main service may start faster and any requests won't be routed thru the proxy until it fully starts too.

ballista/core/proto/ballista.proto
```diff
@@ -707,6 +707,12 @@ message RunningTaskInfo {
   uint32 partition_id = 4;;
```
martin-g, line 711:
```suggestion
  uint32 partition_id = 4;
```

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+fn from_ballista_err(e: &ballista_core::error::BallistaError) -> Status {
+    Status::internal(format!("Ballista Error: {e:?}"))
+}
+
+async fn get_flight_client(
+    host: &String,
+    port: &u16,
```
martin-g:
```suggestion
    port: u16,
```

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -121,6 +123,49 @@ pub async fn start_grpc_service<
...
+        };
+
+        let max_encoding_message_size =
+            config.grpc_server_max_encoding_message_size as usize;
+        let max_decoding_message_size =
+            config.grpc_server_max_decoding_message_size as usize;
+        info!("Built-in arrow flight server proxy listening on: {address:?} max_encoding_size: {max_encoding_message_size} max_decoding_size: {max_decoding_message_size}");
+
+        let grpc_server_config = GrpcServerConfig::default();
+        let server_future = create_grpc_server(&grpc_server_config)
```
martin-g: There is no authentication layer.
But there is no authentication for the main service too, so this is not required at the moment.

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -360,9 +363,22 @@ async fn execute_query(
                 let duration = Duration::from_millis(duration);
 
                 info!("Job {job_id} finished executing in {duration:?} ");
+                let FlightEndpointInfo {
```
milenkovicm: I dont think we should do a round trip to fetch endpoint info. Could we add optional response parameter in `message SuccessfulJob` ? and if it is present do a proxy request?

milenkovicm (commented):
its a good start, i just think we could remove part additional request to check for proxy endpoint

ballista/core/src/serde/generated/ballista.rs
```diff
@@ -1449,6 +1456,36 @@ pub mod scheduler_grpc_client {
                 );
             self.inner.unary(req, path, codec).await
         }
+        /// Provide info on flight proxy address
```
milenkovicm: as previous comment, maybe we do not need this endpoint, so we save a request

ballista/core/src/execution_plans/distributed_query.rs
sebbegg: Changing `SuccessfulJob` would require a few changes in more modules… not sure it’s the right place?
An alternative would be to add it as a second field to `GetJobStatusResult?`
That would limit the impact to the scheduler grpc server only

milenkovicm: https://github.com/apache/datafusion-ballista/blob/118327ca424cf14e38e522b7f0af4ed9c6b3c8c3/ballista/core/src/execution_plans/distributed_query.rs#L356

gives you back partition locations which indicates where data is, so adding optional proxy parameter you could ask it "give me data from `partition_location`" otherwise you just fetch it as it is at the moment (if proxy not provided)

sebbegg: I get that, my worry was about where to fill the information.
The `SuccessfulJob` and `PartitionLocation` objects are all created in `execution_graph.rs` it seems. It appears weird to somehow forward the proxy information all the way into the execution graph in order to be able to fill a new field like `SuccessfulJob.flight_endpoint`

The alternative could be to clone & update the `SuccessfulJob` in the grpc endpoint:

https://github.com/apache/datafusion-ballista/blob/118327ca424cf14e38e522b7f0af4ed9c6b3c8c3/ballista/scheduler/src/scheduler_server/grpc.rs#L450

pseudo:
```
fn get_job_status(job_id) {
  job_status = task_manager.get_job_status(job_id)
  if job_status.status is SuccessfulJob:
      job_status.status.flight_endpoint = self.state.config.advertise_flight_sql_endpoint
  
  return job_status
}
```

ballista/scheduler/src/flight_proxy_service.rs
sebbegg: Certainly.

No idea on whether this should be a new setting 🤔 
I'd guess starting with the current one might be fine and if there should be need for a dedicated setting one could revisit?

ballista/core/src/execution_plans/distributed_query.rs
sebbegg: obsolete with sending the flight-endpoint as part of the job status

ballista/scheduler/src/flight_proxy_service.rs
sebbegg: Added a check that host+port match with a known (and alive?) executor.

ballista/scheduler/src/flight_proxy_service.rs
sebbegg: Looked into that. I mentioned in the issue - I'm new to rust so maybe I'm missing something:

My first attempt for a pool would have been a plain `HashMap<(String, u16), FlightClient>` that just keeps existing clients for executors alive for reuse, with `(host, port)` being the key.
Unfortunately the `FlightService` trait has all methods as non-mutable. So I can't make `do_get` use `&mut self`...

Is there a smart way around that?
It seems tonic also doesn't offer pools out of the box...

milenkovicm: I would ignore pooling for now, we can add it later if you agree

ballista/scheduler/src/flight_proxy_service.rs
milenkovicm: Same like pooling, do we need this check in the first version?

milenkovicm:
hey @sebbegg is there anything i can do to help you with this PR?

ballista/scheduler/src/flight_proxy_service.rs
sebbegg: This appeared to be straight forward, I've added a check that the requested `host:port` belongs to a registered & active executor.

milenkovicm:
thanks @sebbegg will have a look today/tomorrow

ballista/core/proto/ballista.proto
```diff
@@ -642,6 +642,7 @@ message JobStatus {
 
 message GetJobStatusResult {
   JobStatus status = 1;
+  optional string flight_endpoint = 2;
```
milenkovicm: would it make sense to name this as "flight proxy" or similar?

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -327,7 +328,10 @@ async fn execute_query(
     let mut prev_status: Option<job_status::Status> = None;
 
     loop {
-        let GetJobStatusResult { status } = scheduler
+        let GetJobStatusResult {
+            status,
+            flight_endpoint,
```
milenkovicm: would it make sense to support Some("") in which case client should use scheduler address and port ? This way scheduler should not relly care about its public port?

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -131,7 +180,24 @@ pub async fn start_server(
...
 
-    start_grpc_service(address, scheduler).await
+    info!(
+        "advertise_flight_sql_endpoint: {:?}",
+        config.advertise_flight_sql_endpoint
+    );
+    match config.advertise_flight_sql_endpoint {
+        Some(_) => {
+            info!("Starting flight proxy");
+            let flight_proxy = start_flight_proxy_server(config, scheduler.state.clone());
```
milenkovicm: Would it make sense to run proxy as a service on the same port with a scheduler service? It would simplify configuration.

ballista/core/src/execution_plans/distributed_query.rs
milenkovicm: not sure should we use Some("") or we have proto enums to represent proxy cases

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -131,7 +191,24 @@ pub async fn start_server(
...
+        create_scheduler::<LogicalPlanNode, PhysicalPlanNode>(cluster, config.clone())
+            .await?;
 
-    start_grpc_service(address, scheduler).await
+    info!(
+        "advertise_flight_sql_endpoint: {:?}",
+        config.advertise_flight_sql_endpoint
+    );
+    match config.advertise_flight_sql_endpoint.clone() {
+        Some(s) if s != "" => {
```
milenkovicm: Sorry I might be unclear. If we specify different port (or ip port) that would mean there is external process running proxy, not the scheduler process.

So we have three configuration options

- no proxy 
- in process (no need to specify ip/port client should use scheduler ip port)
- external process (ip / port specified) client should use given ip/port

ballista/core/proto/ballista.proto
```diff
@@ -642,6 +642,7 @@ message JobStatus {
 
 message GetJobStatusResult {
   JobStatus status = 1;
+  optional string flight_proxy = 2;
```
milenkovicm: can we make this one of to represent proxy statuses

- no proxy
- in process (no need to specify ip/port client should use scheduler ip port)
- external process (ip / port specified) client should use given ip/port

that would remove check if for empty string on client side

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -450,7 +450,22 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
-            Ok(status) => Ok(Response::new(GetJobStatusResult { status })),
+            Ok(status) => Ok(Response::new(GetJobStatusResult {
+                status,
+                flight_proxy: self
+                    .state
+                    .config
+                    .advertise_flight_sql_endpoint
+                    .clone()
+                    .map(|s| match s {
+                        s if s.is_empty() => format!(
```
milenkovicm: i guess same thing here, if configuration is empty client should dial back on scheduler address / port

ballista/scheduler/src/config.rs
```diff
@@ -41,7 +41,7 @@ pub struct Config {
     /// Route for proxying flight results via scheduler (IP:PORT format).
     #[arg(
         long,
-        help = "Route for proxying flight results via scheduler. Should be of the form 'IP:PORT"
+        help = "Route for proxying flight results via scheduler. Should be of the form 'IP:PORT'"
```
milenkovicm: same comment regarding empty address. if address/port not specified client need to dial back on scheduler address

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,209 @@
...
+use datafusion_proto::logical_plan::AsLogicalPlan;
+use datafusion_proto::physical_plan::AsExecutionPlan;
+use futures::{Stream, TryFutureExt};
+use log::debug;
+use std::collections::HashSet;
+use std::pin::Pin;
+use std::sync::Arc;
+use tonic::{Request, Response, Status, Streaming};
+
+/// Service implementing a proxy from scheduler to executor Apache Arrow Flight Protocol
```
milenkovicm, line 34: it would be great if we could add more comments here. describe how it can be configured

milenkovicm (commented):
thanks @sebbegg, 

just to clarify, we can have three configuration options:

- proxy not configured, client needs to fetc data from executors
- proxy configured, no ip address or port provided, scheduler needs to start proxy on the same port (withing process)
- proxy configured ip/port provided, scheduler considers this as extenal process running proxy, it just needs to put that value in the response, scheduler will not start proxy. client needs to use that ip/port combination to connect to process

ballista/scheduler/src/scheduler_server/grpc.rs
sebbegg: That's what happens - just felt that the "switch" was easier to implement on scheduler side.
This way there's a bit less logic on the client side. Can move this though.

ballista/scheduler/src/scheduler_process.rs
sebbegg: That’s what happens though?
This just puts the logic to use the scheduler host:port in the scheduler rather than the client.

milenkovicm: Yes, but you need to specify advertising address and configure it correctly, which may be tricky in docker containers. 

Suggestion would eliminate that as client already knows scheduler address.

ballista/core/proto/ballista.proto
sebbegg: E.g. like
```protobuf
oneof flight_proxy {
    bool no_proxy = 2;
    bool in_scheduler = 3;
    string external_address = 4;
}
```
?

milenkovicm: yes something like that, 
perhaps,
```
oneof flight_proxy {
    bool local = 1; 
    string external = 4;
}
```

something like that

milenkovicm:
> Ok, so for the scope of this PR, should we add the extra proxy as an additional executable? A minimalistic approach could be to only implement the embedded variant and leave an external flight-proxy executable up to users.

I agree

milenkovicm:
I apologise @sebbegg, I'm catching up with reviews

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -415,22 +426,75 @@ async fn execute_query(
...
+            .to_string();
+        let port: u16 = url.port().ok_or(DataFusionError::Execution(format!(
+            "No port in {address:?}"
+        )))?;
+        Ok((host, port))
+    }
+
+    match flight_proxy {
+        Some(FlightProxy::External(address)) => {
+            info!("Fetching results from external flight proxy: {}", address);
```
milenkovicm: please use debug

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -415,22 +426,75 @@ async fn execute_query(
...
+            split_host_port(format!("http://{address}").as_str())
+        }
+        Some(FlightProxy::Local(true)) => {
+            info!("Fetching results from scheduler: {}", scheduler_url);
```
milenkovicm: please use debug

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -415,22 +426,75 @@ async fn execute_query(
...
+            split_host_port(scheduler_url)
+        }
+        Some(FlightProxy::Local(false)) | None => {
+            info!(
```
milenkovicm: please use debug

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -96,6 +99,24 @@ pub async fn start_grpc_service<
     let mut tonic_builder = RoutesBuilder::default();
     tonic_builder.add_service(scheduler_grpc_server);
 
+    match &config.advertise_flight_sql_endpoint {
+        Some(proxy) if proxy.is_empty() => {
+            info!("Adding embeddded flight proxy service on scheduler");
```
milenkovicm: typo `embedded`

milenkovicm (commented):
thanks @sebbegg 
I think this can be merged, I just have a few minor comments and one case to be fixed

running scheduler with:

```
cargo run --bin ballista-scheduler -- --advertise-flight-sql-endpoint 
```

will return error 

```
error: a value is required for '--advertise-flight-sql-endpoint <ADVERTISE_FLIGHT_SQL_ENDPOINT>' but none was supplied
```
not sure how to configure local proxy to test this

ballista/scheduler/src/config.rs
```diff
@@ -41,7 +41,7 @@ pub struct Config {
     /// Route for proxying flight results via scheduler (IP:PORT format).
     #[arg(
```
milenkovicm, line 42:
```
#[arg(
        long,
        num_args = 0..=1,
        default_missing_value = "",
        help = "Route for proxying flight results via scheduler. Use 'HOST:PORT' to let clients fetch results from the specified address. If empty a flight proxy will be started on the scheduler host and port."
    )]
```

ballista/scheduler/src/flight_proxy_service.rs
```diff
@@ -0,0 +1,188 @@
...
+    Status::internal(format!("Ballista Error: {e:?}"))
+}
+
+async fn get_flight_client(
+    host: &str,
+    port: u16,
+    max_decoding_message_size: usize,
+    max_encoding_message_size: usize,
+) -> Result<FlightServiceClient<tonic::transport::channel::Channel>, BallistaError> {
+    let addr = format!("http://{host}:{port}");
```
milenkovicm, line 173: we should not assume `http` here

sebbegg: Hm, other usages of `create_grpc_client_connection` follow the same usage:

https://github.com/apache/datafusion-ballista/blob/caea92969f34a0c0899a6eb13713b9f7e377f80d/ballista/core/src/client.rs#L73-L76

https://github.com/apache/datafusion-ballista/blob/caea92969f34a0c0899a6eb13713b9f7e377f80d/ballista/executor/src/executor_process.rs#L201

It's somewhat inconsistent that some parts of the code use host+port while at other places require `Url`s or url-like strings.

milenkovicm: its great you catch those, maybe we should fix them with follow up or as part of #1400

ballista/scheduler/src/config.rs
sebbegg: Thanks 👍 
Looks good:

```console
% cargo run --bin ballista-scheduler -- --advertise-flight-sql-endpoint
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/ballista-scheduler --advertise-flight-sql-endpoint`
2026-01-23T15:44:14.805527Z  INFO main ThreadId(01) ballista_scheduler::scheduler_process: Ballista v51.0.0 Scheduler listening on 0.0.0.0:50050
2026-01-23T15:44:14.805652Z  INFO main ThreadId(01) ballista_scheduler::scheduler_process: Starting Scheduler grpc server with task scheduling policy of PullStaged
2026-01-23T15:44:14.808318Z  INFO main ThreadId(01) ballista_scheduler::scheduler_server::query_stage_scheduler: Starting QueryStageScheduler
2026-01-23T15:44:14.809134Z  INFO tokio-runtime-worker ThreadId(11) ballista_core::event_loop: Starting the event loop query_stage
2026-01-23T15:44:14.810022Z  INFO                 main ThreadId(01) ballista_scheduler::scheduler_process: Adding embedded flight proxy service on scheduler
```

PR #1353, opened 2025-12-28
feat: capture more metrics in distributed_query

### LINKED ISSUE #1252: Improve `DistributedQueryExec` adding job related metrics

With #1251 `DistributedQueryExec` exposed few metrics, `row count` and `transferred bytes` we could extend metrics adding job related metrics like run and scheduling time and time needed to transfer data from flight servers to the client. 

This would enable us to retrieve job related statistics without needed to contact REST API, and improve observability. 

Method `execute_query` already has times, we could utilise it to retrieve those metrics.

**Additional context**

PhVHoang:

Closes #1252 .

 # Rationale for this change

### DIFF
--- a/ballista/core/src/execution_plans/distributed_query.rs
+++ b/ballista/core/src/execution_plans/distributed_query.rs
@@ -75,8 +75,8 @@ pub struct DistributedQueryExec<T: 'static + AsLogicalPlan> {
     /// - transferred_bytes: Total bytes transferred from executors
     /// - job_execution_time_ms: Time spent executing on the cluster (server-side)
     /// - job_scheduling_in_ms: Time from query submission to job start (includes queue time)
-    /// - total_query_time_ms: Total end-to-end query latency from client perspective
-    /// - data_transfer_time_ms: Time spent fetching results (not yet implemented)
+    /// - job_execution_time_ms: Time spent executing on the cluster (ended_at - started_at)
+    /// - job_scheduling_in_ms: Time job waited in scheduler queue (started_at - queued_at)
     metrics: ExecutionPlanMetricsSet,
 }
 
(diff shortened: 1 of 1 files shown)


ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -289,6 +301,8 @@ async fn execute_query(
         .max_encoding_message_size(max_message_size)
         .max_decoding_message_size(max_message_size);
 
+    let query_submit_time = std::time::Instant::now();
```
milenkovicm: just wonder if `SuccessfulJob::queued_at` could be used instead of this one? I'm not quite sure, wdyt?

milenkovicm (commented):
i think this makes sense @PhVHoang thanks a lot!
One minor comment to investigate

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -278,7 +285,12 @@ async fn execute_query(
     query: ExecuteQueryParams,
     max_message_size: usize,
     grpc_config: GrpcClientConfig,
+    metrics: Arc<ExecutionPlanMetricsSet>,
+    partition: usize,
```
milenkovicm, line 289: looks like there is single partition as the result of this exec, so not sure if we need partition here or we could use 0 later?

ballista/core/src/execution_plans/distributed_query.rs
PhVHoang: Thanks @milenkovicm ,

I think this is a great catch. Using `queued_at` would be more accurate since it’s a server-side timestamp set by the scheduler when the job is queued. This avoids client-side clock skew and ensures the measurement doesn’t include any network latency.

Let me fix it.

ballista/core/src/execution_plans/distributed_query.rs
milenkovicm: Let's leave it as it is for now

Issue #32, opened 2022-05-21
Adopt substrait.io for serializing query plans


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Ballista (and DataFusion) has a proprietary protobuf-based format for serializing query plans. This really ties Ballista to DataFusion and does not allow other query engines and/or compute kernals to be used easily.

**Describe the solution you'd like**
There is now an emerging standard for query plan serialization at https://substrait.io/ and this is also protobuf-based. It would be good to move towards this over time.

**Describe alternatives you've considered**
None

**Additional context**
None

andygrove:
Substrait support is now in DataFusion, so I plan on working on this soon

milenkovicm:
I propose to close this EPIC if @andygrove agrees.

#1358 adds scheduler support for substrait, i'm not sure if there is much value in point 2. Datafusion substrait physisical is not really mature, and it will always lag behind datafusion proto, so not much benefit from point 3 in the description

andygrove:
Sound good to me

PR #1360, opened 2026-01-02
feat: Scheduler supports `substrait` logical plan and remove deprecated `sql` support

### LINKED ISSUE #1358: Remove query type `sql` from ballista protocol

at the moment ballista protocol supports sending raw SQL to be executed

https://github.com/apache/datafusion-ballista/blob/32f281a2a7cada087c7eebe6d3a4cbd355d45af0/ballista/core/proto/ballista.proto#L540

this has been deprecated for quite some time, maybe its time to be removed as it wont behave correctly

Remove query type sql

**Additional context**

mattcuento:

Closes #1358.
Part of #32.

 # Rationale for this change

As part of this [epic](https://github.com/apache/datafusion-ballista/issues/32), adding server-side support for Substrait plans (to the Scheduler `ExecuteQuery` API). Substrait support is gated by the `substrait` feature with conditional depdencies for `datafusion-substrait` and therefore `substrait`. This feature is on by default.

The `sql` field was marked as deprecated from the Ballista protocol two years back [here](https://github.com/apache/datafusion-ballista/commit/5bb69a77a91c2cf4f76e04933dc98347bef17a92). This is clean up of the field. The field ID and name have been marked as reserved to prevent corrupted requests if client/server versions drift across this change.

- Bumped base image to `rust:1.92-trixie` in order to pull an updated `protobuf-compiler` version that supports optional fields for proto3 by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,9 @@ rpath = false
 [profile.ci]
 inherits = "dev"
 incremental = false
+debug = false
+debug-assertions = false
+strip = "debuginfo"
 
 # ci turns off debug info, etc. for dependencies to allow for smaller binaries making caching more effective
 [profile.ci.package."*"]


mattcuento:
Hey @milenkovicm, I added support for Substrait on `ExecuteQuery`. My assumption from the epic is that this is strictly intended to be a server-side change for use with different front-ends, and that we won't be changing the client to serialize into Substrait (sounds like that would be an unnecessary extra serialization step on top of `LogicalPlan`).

Assuming that's true, it was a little tricky to test this well aside from calling the scheduler RPC service directly ([`test_substrait_compatibility`](https://github.com/apache/datafusion-ballista/pull/1360/changes#diff-7dacf191970a82fbfac6d832289fb555079fe3de0ada9036741996fdf53a84d9R877)). I'm brand new to the project, so if you have any opinions/advice on where or how to better test this do let me know!

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -873,4 +873,77 @@ mod test {
...
+        assert_eq!(stored_executor.host, "http://localhost:8080".to_owned());
+
+        // Context strictly used for values-based query serialization to avoid
+        // needing to register tables and keep them in sync with the scheduler instance.
+        // We only truly desire to test proper reception of a Substrait plan, not explicit
+        // SubstraitPlan -> LogicalPlan conversions.
+        let config = SessionConfig::new();
+        let ctx = SessionContext::new_with_config(config);
+        let serialized_substrait_plan =
+            serialize_bytes("SELECT a, b FROM (VALUES (1, 2), (3, 4)) AS t(a, b)", &ctx)
```
milenkovicm: maybe we could add another column in select like `abs(a) + abs(b)` or similar

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -873,4 +873,77 @@ mod test {
         assert!(active_executors.is_empty());
         Ok(())
     }
+    #[tokio::test]
+    async fn test_substrait_compatibility() -> Result<(), BallistaError> {
```
milenkovicm, line 901: test makes sense  👍🏻

milenkovicm:
Maybe as a follow up we should put a bit more documentation around this and example(s)

milenkovicm:
Also, could we gate substrait with config option, which could be on by default? 
Users not needing it could disable it at compile time.

ballista/scheduler/Cargo.toml
```diff
@@ -52,6 +53,7 @@ clap = { workspace = true, optional = true }
 dashmap = { workspace = true }
 datafusion = { workspace = true }
 datafusion-proto = { workspace = true }
+datafusion-substrait = { workspace = true, optional = true }
```
milenkovicm, line 56: will datafusion-substrait option `protoc` help with the proto issue `protoc = ["substrait/protoc"]` https://github.com/apache/datafusion/blob/418f62ae3bed96e8d2dc86d5085580eb1b8c5b3c/datafusion/substrait/Cargo.toml#L61

milenkovicm (commented):
thanks @mattcuento changes make sense, we just need to fix build issue and we can merge this

ballista/scheduler/Cargo.toml
mattcuento: I'd tried that and found that a lot of the build actions seemingly don't ship with cmake for executing `protoc` compilation as part of that feature. If we're good with adding cmake to the failing builds, that should solve it

milenkovicm: Whats puzzling me is that we build protoc in ballista and it does not complain about cmake, not sure whats the difference

mattcuento: Yeah fairly odd... I've noticed that alternatively, ballista core is able to pass a flag that makes the compilation work without a newer `protoc` version `--experimental_allow_proto3_optional`. However, for us here this is in the compilation of the proto files within the `substrait dependency (`algebra.proto`?) as opposed to source code.

I'm not sure if there's an equivalent way to pass that flag down.


```
tonic_prost_build::configure()
            .extern_path(".datafusion_common", "::datafusion_proto_common")
            .extern_path(".datafusion", "::datafusion_proto::protobuf")
            .protoc_arg("--experimental_allow_proto3_optional")
            .compile_protos(&["proto/ballista.proto"], &["proto"])
            .map_err(|e| format!("protobuf compilation failed: {e}"))?;
```

I'll keep messing around here for a bit to see what I can do!

milenkovicm: Build fails at `protobuf-src v2.1.1+27.1` which comes from https://github.com/substrait-io/substrait-rs/blob/f097b314fb9726866abd61a9935e48aeac68817e/Cargo.toml#L32 where proto feature should not be enabled. I wonder https://github.com/apache/datafusion/blob/fd35a09438a2b4841431f5e86ffef378cbbda7c9/datafusion/substrait/Cargo.toml#L68 is triggering issue. 

I wonder if `--no-default-features` for that specific dependency will make any difference ?

I had a look at datafusion, they do install `cmake `for that specific target (which should not be default) 

As the last resort we could add `cmake`  not sure if it will break windows build.

Cargo.toml
```diff
@@ -39,6 +39,7 @@ datafusion = "51.0.0"
 datafusion-cli = "51.0.0"
 datafusion-proto = "51.0.0"
 datafusion-proto-common = "51.0.0"
+datafusion-substrait = { version = "51.0.0", features = ["protoc"] }
```
milenkovicm: can you remove `features = ["protoc"]` from here? I've missed this file completely

mattcuento: Done, without this we should be back to a state where only the docker build fails. From local:
```
   Compiling substrait v0.62.2
error: failed to run custom build command for `substrait v0.62.2`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `/home/builder/workspace/target/release/build/substrait-2824a6349ada2508/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=FORCE_REBUILD
  cargo:rerun-if-changed=substrait
  cargo:rerun-if-changed=substrait/text/simple_extensions_schema.yaml
  cargo:rerun-if-changed=substrait/text/dialect_schema.yaml
  cargo:rerun-if-changed=substrait/proto/substrait/algebra.proto
  cargo:rerun-if-changed=substrait/proto/substrait/plan.proto
  cargo:rerun-if-changed=substrait/proto/substrait/type_expressions.proto
  cargo:rerun-if-changed=substrait/proto/substrait/parameterized_types.proto
  cargo:rerun-if-changed=substrait/proto/substrait/capabilities.proto
  cargo:rerun-if-changed=substrait/proto/substrait/type.proto
  cargo:rerun-if-changed=substrait/proto/substrait/function.proto
  cargo:rerun-if-changed=substrait/proto/substrait/extensions/extensions.proto
  cargo:rerun-if-changed=substrait/proto/substrait/extended_expression.proto

  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: substrait/algebra.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }
```

milenkovicm: I wonder if updating builder image will make any difference ? `FROM rust:1.85-bullseye`?

milenkovicm: probably nothing as they come with same protobuf compiler

mattcuento: Great call actually, googling around it looks like the necessary protoc upgrade happens in bookworm, i'll bump. https://packages.debian.org/bookworm/protobuf-compiler

milenkovicm: will updating builder to `rust:1.92-trixie` and others to latest ubuntu fix issue ?

mattcuento: Works from local on `rust:1.85-bookworm` 🥳, yeah happy to bump to `rust:1.92-trixie` if that's preferred! 

Thanks a lot for the help ❤️

milenkovicm: update to latest please, plus other ubuntu references. it would be great if you could open new PR for that please

mattcuento: I see datafusion has 1.92 in their toolchain, so bumped and committed `rust:1.92-trixie` here (required in this PR for protoc)

I'll open another PR for the CLI image, it does look like ubuntu references are already up to date with trixie via 24.04

mattcuento:
Will review the latest `test linux balista/crates` failures this evening

mattcuento:
Tried to initiate a re-run in hopes the linux test builds will pass the linking phase, GH runner usage metrics are pretty minimal.

milenkovicm:
https://github.com/milenkovicm/datafusion-ballista/commit/b60d2457550b79727bb968479738a21ba4eeed65 adding `cargo clean` may help

Huy1Ng:
I couldn't replicate the error on my fork, but add this to the CI section in `Cargo.toml` 
```
[profile.ci]
inherits = "dev"
incremental = false
debug = false
debug-assertions = false
strip = "debuginfo"
```
reduce the ./target directory size by ~10%. I'm not sure how `cargo clean` would help as we should start with empty ./target for each job (or do we?!)

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -392,25 +398,36 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
                         }
                     }
                 }
-                Query::Sql(sql) => {
-                    match session_ctx
-                        .sql(&sql)
-                        .await
-                        .and_then(|df| df.into_optimized_plan())
+                Query::SubstraitPlan(bytes) => {
```
milenkovicm, line 415:
```
warning: unused variable: `bytes`
   --> ballista/scheduler/src/scheduler_server/grpc.rs:401:38
    |
401 |                 Query::SubstraitPlan(bytes) => {
    |                                      ^^^^^ help: if this is intentional, prefix it with an underscore: `_bytes`
    |
    = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default
```

if substrait disabled

mattcuento: Thanks, fixed!

Issue #1359, opened 2026-01-02
[EPIC] Adaptive Query Execution (AQE)


milenkovicm:
## Executive Summary

- Create a distributed query planner which can change stages, add/or remove stages, or cancel stage (partitioning) to  reflect collected runtime statistics.
- An extendable distributed query planner, simple to extend, just like adding a new physical rule.

## Overview

Currently, jobs in Ballista are divided into stages at job submission time according to predefined rules.
Ballista can optimize stages based on runtime information using set of physical optimizers, changing the join order for example. This has been [implemented here](https://github.com/milenkovicm/datafusion-ballista/blob/26fcbf66233de2c6d12545277242e16b4ede386a/ballista/scheduler/src/state/execution_stage.rs#L372-L378) with only few physical optimizer rules used.

Also, few more rules are scattered around  [here](https://github.com/milenkovicm/datafusion-ballista/blob/26fcbf66233de2c6d12545277242e16b4ede386a/ballista/scheduler/src/planner.rs#L130) and [here](https://github.com/milenkovicm/datafusion-ballista/blob/26fcbf66233de2c6d12545277242e16b4ede386a/ballista/scheduler/src/planner.rs#L162)

Current approach may fail to generate valid plan in case there is [dependency between physical optimizers](https://github.com/apache/datafusion-ballista/issues/1296). Furthermore, some optimizations can’t be propagated across stage boundaries potentially impacting overall performance.

Physical optimizers used work well with a major downside, rules being scattered across code, making adding or removing rules non-trivial task. Scattered rules are hard to control, thus centralized rule "facility" is need. It would make it easier to change or alter behavior of physical planners or to experiment with new solutions.

Another major problem arises when stages are split; it becomes impossible to change their layout, which can be necessary in certain situations. For instance, if an exchange produces an empty result, we’d need to run a subsequent stage even though the system knows it will also produce an empty result. This can lead to significant overhead on the scheduler, especially with many tasks. Having the ability to remove a stage would be helpful in this scenario.

### Proposal

Implement adaptive query planner which would address some of the issues with the current approach. In the core of the design are pluggable physical optimizer rules which would re-optimize plan after each stage.

This proposal is strongly influenced by [Adaptive and Robust Query Execution for Lakehouses at Scale](https://www.cs.cmu.edu/~15721-f24/papers/AQP_in_Lakehouse.pdf) paper. 

Major difference from it is that ballista works on physical plan rather than logical plan.

### Impact on DataFusion

[...]

PR #1361, opened 2026-01-03
feat: Extract `execution_graph` to a trait

milenkovicm:

Closes #.

Relates to #1359.

 # Rationale for this change

`AdaptiveExecutionGraph` progress looks decent (#1372), with basic flow passing (most of), we can proceed merging this to main.

(as described in #1359)

Rename `ExecutionGraph` to `StaticExecutionGraph` and introduce `dyn ExecutionGraph` which would be implemented by `StaticExecutionGraph` and new `AdaptiveExecutionGraph`. 

Change implementation in `TaskManger` based on predefined configuration parameter.  It would enable us to have two implementation running in parallel until AQE implementation matures.

No

### DIFF
--- a/ballista/scheduler/src/cluster/mod.rs
+++ b/ballista/scheduler/src/cluster/mod.rs
@@ -43,7 +43,9 @@ use crate::cluster::memory::{InMemoryClusterState, InMemoryJobState};
 
 use crate::config::{SchedulerConfig, TaskDistributionPolicy};
 use crate::scheduler_server::SessionBuilder;
-use crate::state::execution_graph::{ExecutionGraph, TaskDescription, create_task_info};
+use crate::state::execution_graph::{
+    ExecutionGraphBox, TaskDescription, create_task_info,
+};
 use crate::state::task_manager::JobInfoCache;
 
 /// Event broadcasting and subscription for cluster state changes.
@@ -299,7 +301,7 @@ pub trait JobState: Send + Sync {
     /// Submits a new job to the job state.
     ///
     /// The submitter is assumed to own the job.
-    async fn submit_job(&self, job_id: String, graph: &ExecutionGraph) -> Result<()>;
+    async fn submit_job(&self, job_id: String, graph: &ExecutionGraphBox) -> Result<()>;
 
     /// Returns the set of all active job IDs.
     async fn get_jobs(&self) -> Result<HashSet<String>>;
@@ -311,12 +313,15 @@ pub trait JobState: Send + Sync {
     ///
     /// The job may not belong to the caller, and the graph may be updated
     /// by another scheduler after this call returns.
-    async fn get_execution_graph(&self, job_id: &str) -> Result<Option<ExecutionGraph>>;
+    async fn get_execution_graph(
+        &self,
+        job_id: &str,
+    ) -> Result<Option<ExecutionGraphBox>>;
 
     /// Persists the current state of an owned job.
     ///
     /// Returns an error if the job is not owned by the caller.
-    async fn save_job(&self, job_id: &str, graph: &ExecutionGraph) -> Result<()>;
+    async fn save_job(&self, job_id: &str, graph: &ExecutionGraphBox) -> Result<()>;
 
     /// Marks an unscheduled job as failed.
     ///
@@ -330,7 +335,7 @@ pub trait JobState: Send + Sync {
     ///
     /// Returns the execution graph if the job is still running and successfully acquired,
     /// otherwise returns None.
-    async fn try_acquire_job(&self, job_id: &str) -> Result<Option<ExecutionGraph>>;
+    async fn try_acquire_job(&self, job_id: &str) -> Result<Option<ExecutionGraphBox>>;
 
     /// Returns a stream of job state events.
     async fn job_state_events(&self) -> Result<JobStateEventStream>;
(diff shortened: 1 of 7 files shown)


milenkovicm:
if anybody have time for review plase cc: @danielhumanmod @killzoner @mattcuento @martin-g

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -193,109 +326,379 @@ impl ExecutionGraph {
...
+        completed_stages
+    }
+    /// An ExecutionGraph is successful if all its stages are successful
+    fn is_successful(&self) -> bool {
+        self.stages
+            .values()
+            .all(|s| matches!(s, ExecutionStage::Successful(_)))
+    }
+
+    // pub fn is_complete(&self) -> bool {
```
mattcuento, line 662: Can probably remove this

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -1704,8 +1821,6 @@ mod test {
 
         let outputs = agg_graph.output_locations();
 
-        assert_eq!(outputs.len(), agg_graph.output_partitions);
```
mattcuento: why remove this?

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -350,15 +350,15 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
         if let Some(cached) = self.get_active_execution_graph(job_id) {
             let guard = cached.read().await;
 
-            Ok(Some(Arc::new(guard.deref().clone())))
+            Ok(Some(guard.deref().cloned()))
         } else {
             let graph = self.state.get_execution_graph(job_id).await?;
 
-            Ok(graph.map(Arc::new))
+            Ok(graph)
```
mattcuento: What changed here such that we dont need an `Arc`? Not super familiar with the threading models in ballista yet. Is this since the execution graph trait is `Send + Sync`?

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -516,15 +516,13 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
     pub async fn executor_lost(&self, executor_id: &str) -> Result<Vec<RunningTaskInfo>> {
         // Collect all the running task need to cancel when there are running stages rolled back.
         let mut running_tasks_to_cancel: Vec<RunningTaskInfo> = vec![];
-        // Collect graphs we update so we can update them in storage
-        let updated_graphs: DashMap<String, ExecutionGraph> = DashMap::new();
```
mattcuento, line 520: hm, was this just doing nothing as is? Any issue we should log for this based on the comment?

mattcuento (commented):
One question and nit, otherwise looks good to me!

ballista/scheduler/src/cluster/mod.rs
```diff
@@ -311,12 +313,15 @@ pub trait JobState: Send + Sync {
     ///
     /// The job may not belong to the caller, and the graph may be updated
     /// by another scheduler after this call returns.
-    async fn get_execution_graph(&self, job_id: &str) -> Result<Option<ExecutionGraph>>;
+    async fn get_execution_graph(
+        &self,
+        job_id: &str,
+    ) -> Result<Option<ExecutionGraphBox>>;
 
     /// Persists the current state of an owned job.
```
danielhumanmod, line 321: At first glance, the doc comment mentioning "Persists..." made me worried about potential deserialization issues as we move to `Box`.

But after checking the codebase, I found `InMemoryJobState` is currently the only implementation, so we should be good here.

BTW just curious, do we plan to support persistent storage for job state in the future?

milenkovicm: There used to be persistent backends but I have removed them as part of refactor. Not sure what would be good case for them, and spark does not have similar functionality.

So at the moment I have no plans for it, also if there is need users can plug their own implementation

ballista/scheduler/src/state/task_manager.rs
milenkovicm: it was dead code, hard to tell if we need follow up issue

ballista/scheduler/src/state/execution_graph.rs
milenkovicm: will clean it up as part of next PR, i was not sure if this method is needed or not

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -52,6 +52,9 @@ pub(crate) use crate::state::execution_stage::{
 };
 use crate::state::task_manager::UpdatedStages;
 
+/// Boxed [ExecutionGraph]
+pub type ExecutionGraphBox = Box<dyn ExecutionGraph + Send + Sync>;
```
killzoner, line 56: Probably doesn't hurt, but wonder why we have both `Send` and `Sync` bounds, I've seen a lot in async librabries being overly restrictive. Can be changed later though

killzoner (approved):
Not much to say, this LGTM!
Personally I like the fact that the new trait is explicity mentionning Box so you don't have to figure it out by looking at it.
One (nit) question

ballista/scheduler/src/state/execution_graph.rs
milenkovicm: At the moment both are needed. Compiler complains a lot if any is removed. We could come back and revisit it later

milenkovicm:
thanks everybody for reviews, will merge this

PR #1345, opened 2025-11-21
chore: update datafusion to 51.0

danielhumanmod:
# Which issue does this PR close?

Closes #1343 .

 # Rationale for this change

DataFusion 51 introduces new changes in its physical plan serde APIs, metrics system, and tonic/prost/Arrow dependencies

Reference: https://datafusion.apache.org/library-user-guide/upgrading.html#datafusion-51-0-0

# What changes are included in this PR?

1. Update Rust MSRV
  - rust-version 1.86 → 1.88 (required by DF51)

2. Upgrade DataFusion ecosystem and protobuf/RPC dependencies
  - datafusion* → 51.0.0
  - arrow* → 57.0.0
  - tonic* / prost* → 0.14

3. Adapt to DF51 API breaking changes
  - SessionContext → TaskContext
  - LogicalPlan::schema().as_ref() → as_arrow()
  - UnnestExec::new()
  - Updated physical plan serde:
    - physical_plan_from_bytes(.., ctx.task_ctx())
    - try_into_physical_plan(ctx.task_ctx(), codec)
    - PhysicalExtensionCodec::try_decode(ctx: &TaskContext)
  - Updated related tests

4. Support new DF51 metrics in Ballista (proto / from_proto / to_proto)
 - OutputBytes
 - PruningMetrics
 - Ratio

(no semantic changes)

# Are there any user-facing changes?

No user-facing behavior changes.
This PR only updates internal dependencies and adapts Ballista to the breaking changes introduced by DataFusion 51.
All external APIs, CLI behavior, and user-visible features remain unchanged.

PR #1363, opened 2026-01-04
chore: update ballista version to 51.0.0 (from 50.0.0)

milenkovicm:

as part of datafusion upgrade we missed upgrading ballista versions as well

Closes #.

 # Rationale for this change


milenkovicm:
perhaps @martin-g or @danielhumanmod could help with review, we have missed to increment ballista version as part of #1345

PR #1372, opened 2026-01-07
feat: Adaptive query execution (AQE) planner fundamentals

milenkovicm:

Closes #.
Relates to #1359.

 # Rationale for this change

as described in #1359 

Implementation of `ExecutionGraph` which computes DAG in runtime, hence
it can add, remove or change stages based on collected statistics.

It has been implemented on top of `AdaptivePlanner` which runs set of predefined
optimizers.

Current implementation covers happy day scenarios, such as:
- join reordering
- elimination of stages which do not produce data (initial implementation)
- ...

with many limitations, such as:

- it does not cover executor failure
- dynamically coalescing shuffle partitions, not supported yet
- does not switch from hash join to sort merge join
- does not switch from streaming aggregation to hash aggregation
- planner traverses whole job graph every time, even for stages which are completed (can we trim the graph and keep traversal faster?) 
- does not support sort shuffle
- ...

Those functionalities are yet to be implemented.

Note, this [AdaptiveExecutionGraph] is a fork of [crate::state::execution_graph::StaticExecutionGraph],
with changes needed to support adaptive execution. Notable change that at the moment
there is not need for `ExecutionStage::UnResolved` as stages are `ExecutionStage::Resolved`
when they are returned from `AdaptivePlanner`.

Adaptive planner is disable by default at the moment, and can be enabled with 

```
SET ballista.planner.adaptive.enabled=true
```

current implementation passes `TPCH` queries, but it still has rough edges

No, functionality disabled by default

### DIFF
--- a/ballista/scheduler/src/state/aqe/optimizer_rule/datafusion_patch.rs
+++ b/ballista/scheduler/src/state/aqe/optimizer_rule/datafusion_patch.rs
@@ -29,6 +29,8 @@ use datafusion::physical_plan::{ExecutionPlan, displayable, execution_plan};
 use log::warn;
 use std::sync::Arc;
 
+// not required anymore
+#[allow(dead_code)]
 #[derive(Debug, Clone, Default)]
 pub struct EliminateRoundRobbinRule {}
 impl EliminateRoundRobbinRule {
@@ -60,7 +62,7 @@ impl PhysicalOptimizerRule for EliminateRoundRobbinRule {
     }
 
     fn name(&self) -> &str {
-        "DistributedEliminateRoundRobbinRule"
+        "EliminateRoundRobbinRule"
     }
 
     fn schema_check(&self) -> bool {
--- a/ballista/scheduler/src/state/aqe/planner.rs
+++ b/ballista/scheduler/src/state/aqe/planner.rs
@@ -18,7 +18,7 @@ use crate::state::aqe::adapter::BallistaAdapter;
 use crate::state::aqe::execution_plan::{AdaptiveDatafusionExec, ExchangeExec};
 use crate::state::aqe::optimizer_rule::{
     DistributedExchangeRule, EliminateCooperativeExecRule, EliminateEmptyExchangeRule,
-    EliminateRoundRobbinRule, PropagateEmptyExecRule, WarnOnDuplicateExecRule,
+    PropagateEmptyExecRule, WarnOnDuplicateExecRule,
 };
 
 use crate::state::execution_stage::StageOutput;
@@ -445,7 +445,8 @@ impl AdaptivePlanner {
         let default_optimizers = PhysicalOptimizer::new();
         physical_optimizers.extend(default_optimizers.rules.iter().cloned());
         physical_optimizers.push(Arc::new(PropagateEmptyExecRule::default()));
-        physical_optimizers.push(Arc::new(EliminateRoundRobbinRule::default()));
+        // this rule is not required anymore
+        // physical_optimizers.push(Arc::new(EliminateRoundRobbinRule::default()));
         physical_optimizers.push(Arc::new(EliminateCooperativeExecRule::default()));
         // we should remove it at the later stage
         // this is just temporary to detect possible duplicate
--- a/ballista/scheduler/src/state/aqe/test/mod.rs
+++ b/ballista/scheduler/src/state/aqe/test/mod.rs
@@ -108,7 +108,9 @@ pub(crate) fn mock_memory_table() -> Arc<dyn TableProvider> {
 }
 
 pub(crate) fn mock_context() -> SessionContext {
-    let config = SessionConfig::new().with_target_partitions(2);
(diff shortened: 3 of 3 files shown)


ballista/scheduler/src/state/aqe/optimizer_rule/datafusion_patch.rs
```diff
@@ -0,0 +1,163 @@
...
+use datafusion::config::ConfigOptions;
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+use datafusion::physical_plan::coop::CooperativeExec;
+use datafusion::physical_plan::repartition::RepartitionExec;
+use datafusion::physical_plan::{ExecutionPlan, displayable, execution_plan};
+use log::warn;
+use std::sync::Arc;
+
+#[derive(Debug, Clone, Default)]
+pub struct EliminateRoundRobbinRule {}
```
danielhumanmod, line 35: Should we investigate making this idempotent on the DataFusion side? (This could be handled in a separate task

milenkovicm: I'm not sure if this is required anymore as configuration disables round robin repartition, worth to check

ballista/scheduler/src/state/aqe/optimizer_rule/distributed_exchange.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn plan_invalid(
+        &self,
+        execution_plan: Arc<dyn ExecutionPlan>,
+    ) -> datafusion::error::Result<bool> {
+        execution_plan
+            .transform_up(|p| self.transform(p))
+            .transformed()
+    }
+
+    pub(crate) fn transform(
```
danielhumanmod, line 43: Question for my understanding, if we have a tree like
```
Filter
  HashJoin (Inner)
    Repartition (Hash) -> Input A
    Repartition (Hash) -> Input B
```

After transformation it will be:
```
AdaptiveDatafusionExec
  Filter
    HashJoin
      ExchangeExec -> Input A
      ExchangeExec -> Input B
```
so that `AdaptivePlanner` can replace these ExchangeExec later if needed?

milenkovicm: yes, you're right. `ExchangeExec` will be stage boundary and will be replaced by shuffle write and shuffle read

ballista/scheduler/src/state/aqe/optimizer_rule/datafusion_patch.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn optimize(
+        &self,
+        execution_plan: Arc<dyn ExecutionPlan>,
+        _config: &datafusion::config::ConfigOptions,
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        Ok(execution_plan.transform_up(Self::transform)?.data)
+    }
+
+    fn name(&self) -> &str {
+        "DistributedEliminateRoundRobbinRule"
```
killzoner: any reason the name is different from the struct (`EliminateRoundRobbinRule` vs `DistributedEliminateRoundRobbinRule`) ? a convention?

ballista/scheduler/src/state/aqe/optimizer_rule/distributed_exchange.rs
```diff
@@ -0,0 +1,140 @@
...
+            let exchange_exec = ExchangeExec::new(
+                sort_preserving_merge.input().clone(),
+                None,
+                self.plan_id_generator
+                    .fetch_add(1, std::sync::atomic::Ordering::Relaxed),
+            );
+            Ok(Transformed::yes(
+                execution_plan.with_new_children(vec![Arc::new(exchange_exec)])?,
+            ))
+        } else if let Some(repartition) =
```
killzoner, line 83: looks like there could be an opportunity to factor these, not sure how though

milenkovicm: There are many such opportunities, but we can push them off for a while, I would like if we could prove  few key points such as shuffle coalescing ...

ballista/scheduler/src/state/aqe/optimizer_rule/datafusion_patch.rs
milenkovicm: No good reason, I refactored struct name but did not change string 😀

ballista/scheduler/src/state/aqe/optimizer_rule/distributed_exchange.rs
killzoner: I will see if I can bridge some good old macro with the protobuf enums and a match when I get time

milenkovicm:
Thanks lads for reviews, will merge this later.

Feel free to take any bits you're interested in, just announce it so we do not overlap

PR #1376, opened 2026-01-10
feat: Creating SubstraitSchedulerClient and standalone Substrait examples

### LINKED ISSUE #1368: Add documentation/examples on how to use substrait plans with Ballista

As part of #32 and #1360, the Scheduler now accepts Substrait plans via gRPC. It is currently undocumented on how this can be used.

We should add examples in the repo on how to create a gRPC client to the scheduler and use datafusion-substrait to generate plans for examples.

Additionally, this should be added to the docs. In particular, I think two areas could be updated to mention this support:
- Overview
- A new page under "REFERENCE" called "Substrait with Ballista"

**Additional context**

Ideally, in addition to examples via datafusion-substrait, maybe we could add one with a front-end like https://ibis-project.org/ ?

mattcuento:

Closes #1368.

 # Rationale for this change

Documentation of how to pass substrait plans and examples of how to do so with existing extensions. Additionally, the interface for executing Substrait plans wasn't very intuitive, so a thin wrapper was created around the scheduler client in the example for easier execution.

No

### DIFF
--- a/examples/examples/standalone-substrait.rs
+++ b/examples/examples/standalone-substrait.rs
@@ -1,0 +1,588 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use ballista::datafusion::common::Result;
+use ballista_core::client::BallistaClient;
+use ballista_core::extension::SessionConfigExt;
+use ballista_core::serde::protobuf::execute_query_params::Query::SubstraitPlan;
+use ballista_core::serde::protobuf::scheduler_grpc_client::SchedulerGrpcClient;
+use ballista_core::serde::protobuf::{
+    ExecuteQueryParams, GetJobStatusParams, GetJobStatusResult, PartitionLocation,
+    SuccessfulJob, execute_query_result, job_status,
+};
+use ballista_core::utils::{GrpcClientConfig, create_grpc_client_connection};
+use ballista_examples::test_util;
+use datafusion::arrow::array::RecordBatch;
+use datafusion::arrow::error::ArrowError;
+use datafusion::catalog::MemoryCatalogProviderList;
+use datafusion::common::DataFusionError;
+use datafusion::execution::{
+    SendableRecordBatchStream, SessionState, SessionStateBuilder,
+};
+use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricBuilder};
+use datafusion::prelude::{SessionConfig, SessionContext};
+use datafusion_substrait::serializer::serialize_bytes;
+use futures::{Stream, StreamExt, TryFutureExt, TryStreamExt};
+use log::{error, info};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tonic::transport::Channel;
+
+/// This example demonstrates executing Substrait plans against external tables
+/// using datafusion-substrait as our frontend.
+///
+/// In a distributed setting, an external catalog would be required to resolve Substrait
+/// `NamedTable` references. Substrait DDL is currently not supported by
+/// [datafusion_substrait]; producing plans results in an [LogicalPlan::EmptyRelation] and
+/// consuming DDL Substrait plans will result in an error.
+#[tokio::main]
+async fn main() -> Result<()> {
+    let catalog_list = Arc::new(MemoryCatalogProviderList::new());
+
+    let config = SessionConfig::new_with_ballista();
+    let state = SessionStateBuilder::new()
+        .with_config(config)
+        .with_default_features()
+        .with_catalog_list(catalog_list.clone())
+        .build();
+    let ctx = SessionContext::new_with_state(state.clone());
+
+    // Use any frontend to serialize a Substrait plan
+    let test_data = test_util::examples_test_data();
+    let ddl_plan_bytes = serialize_bytes(
+        &format!(
+            "CREATE EXTERNAL TABLE IF NOT EXISTS another_data \
+         STORED AS PARQUET \
+         LOCATION '{}/alltypes_plain.parquet'",
+            test_data
+        ),
+        &ctx,
+    )
+    .await?;
+    let select_plan_bytes =
+        serialize_bytes("SELECT id, string_col FROM another_data", &ctx).await?;
+
+    let session_id = ctx.session_id();
+    let scheduler_url = setup_standalone(Some(&state)).await?;
+    let client =
+        SubstraitSchedulerClient::new(scheduler_url, session_id.to_string()).await?;
+
+    client.execute_query(ddl_plan_bytes).await?;
+    let mut stream = client.execute_query(select_plan_bytes).await?;
+
+    let mut batch_count = 0;
+    let mut total_rows = 0;
+    while let Some(batch_result) = stream.next().await {
+        let batch = batch_result?;
+        batch_count += 1;
+        total_rows += batch.num_rows();
+
+        println!("Batch {}: {} rows", batch_count, batch.num_rows());
+        println!("{:?}", batch);
+    }
+    println!("---------");
+    println!("Query executed successfully!");
+    println!("Total batches: {}, Total rows: {}", batch_count, total_rows);
+
+    Ok(())
+}
+
+/// Wrapper class to simplify execution of Substrait plans.
+pub struct SubstraitSchedulerClient {
+    scheduler_url: String,
+    session_id: String,
+    grpc_config: GrpcClientConfig,
+    max_message_size: usize,
+    use_flight_transport: bool,
+}
+
+impl SubstraitSchedulerClient {
+    /// Creates a new [SubstraitSchedulerClient] with default configuration.
+    ///
+    /// # Example
+    /// ```no_run
+    /// use ballista::extension::SubstraitSchedulerClient;
+    ///
+    /// # #[tokio::main]
+    /// # async fn main() -> datafusion::error::Result<()> {
+    /// let client = SubstraitSchedulerClient::new(
+    ///     "http://localhost:50050".to_string(),
+    ///     "session-123".to_string()
+    /// ).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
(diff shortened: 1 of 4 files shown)


examples/examples/standalone-substrait.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let execute_query_params = ExecuteQueryParams {
+        session_id: uuid::Uuid::new_v4().to_string(),
+        settings: vec![],
+        operation_id: uuid::Uuid::now_v7().to_string(),
+        query: Some(SubstraitPlan(plan_bytes)),
+    };
+
+    let response = scheduler.execute_query(execute_query_params).await.expect("Error executing query");
+    response.into_inner().result.expect("Failed query");
```
mattcuento: @milenkovicm, it looks like we end up with some leaky internals since we don't have an existing entry point on the client side for a Substrait plan. 

All interfaces inside `DataFrame` seem to use `LogicalPlan` which makes it difficult to extend. Would it make sense to add some sort of substrait pass through implementation to avoid the extra thrashing between Substrait and `LogicalPlan`?

I know there was some talk in #32 about using Substrait as a replacement for `LogicalPlan`, but that seems like quite the overhaul off the bat.

milenkovicm: > @milenkovicm, it looks like we end up with some leaky internals since we don't have an existing entry point on the client side for a Substrait plan.
> 
> All interfaces inside `DataFrame` seem to use `LogicalPlan` which makes it difficult to extend. Would it make sense to add some sort of substrait pass through implementation to avoid the extra thrashing between Substrait and `LogicalPlan`?

I think it makes sense if we want to "advertise" this feature

> I know there was some talk in #32 about using Substrait as a replacement for `LogicalPlan`, but that seems like quite the overhaul off the bat.

it looks to me that substrait support is always lagging behind datafusion proto, so it might be additional effort on us. I don't see big benefit for us. What do you think?

mattcuento: > it looks to me that substrait support is always lagging behind datafusion proto, so it might be additional effort on us. I don't see big benefit for us. What do you think?

Agreed I don't see big benefit for the effort. I think a way to pass through Substrait via clients would be the most valuable thing. The project seems wholly  centered around the `LogicalPlan` so I don't see much reason to go through effort to change that other than interop with other systems, which we get anyways with just accepting Substrait in the client.

I would be happy to file an issue on how we could implement a substrait pass through after exploring the codebase more

mattcuento:
@milenkovicm yep no problem, and no worries! I liked the idea

examples/examples/standalone-substrait.rs
milenkovicm: @mattcuento I'm not sure if i understand your last point

mattcuento: Sorry, a bit long winded... in sum I don't see much benefit to using Substrait as a replacement of `LogicalPlan`. I _do_ think we should provide an extension to `SessionContext` to pass a Substrait plan.

milenkovicm: What if we want to use it from a framework which does not have SC? Would it be better to have a method which will take a substait blog and return a stream of arrow c batches ?

mattcuento: Yeah you're right. 

I think I'm getting confused in figuring out how Substrait plans can share a catalog with the scheduler for resolving named tables. 

I just pushed up an example `standalone-subtrait.rs`, where we try to create a listing table that points to a parquet file, then read from it. When I run it, we try to consume the substrait select statement into a logical plan. As part of consuming the plan, the `DefaultSubstraitConsumer::resolve_table_ref` method is called, which tries to get a `TableProvider` from a catalog, but fails to do so stating the table doesn't exist. It appears that `CREATE EXTERNAL TABLE` doesn't seem to get registered anywhere on the scheduler side.

Do we have any documentation on how catalogs work in Ballista? I've been comparing this to the `standalone-sql` example as well as a variation that uses DDL for an external table and I see that table providers seem to only be kept locally. Do we have a source of truth catalog in the scheduler or would that need to be a required extension for using Substrait with Ballista?

Sorry for the novel, still learning a lot here, so I hope this is clear 🙂

mattcuento: Aside from that, virtual / values-based tables work just fine per `remote-substrait.rs` example!

milenkovicm: Looks like you found a discrepancy between datafusion proto and substrait. 

At the moment client session context is authoritative. `CREATE TABLE`  will create table locally in the client context, logical plan will encode all required information and send it to scheduler, which will just convert it to physical plan without any additional lookups at the scheduler side (which has empty catalog and stateless session context). 

In most of the cases this works fine, and avoids bringing some kind of shared catalog but in substrait case it might be a bit of a show stopper and external catalog might be required or session at the scheduler side should keep the state. 

The reason to have it implemented the way it was DDL statement handling in datafusion. From what I recall they will never hit query planner so they cant sent from client to scheduler. As substrait may not have this limitation we would need to implement session caching on scheduler side, which was disabled in #1226

examples/examples/standalone-substrait.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    match response.into_inner().result.unwrap() {
+        execute_query_result::Result::Failure(failure) => {
+            return Err(DataFusionError::Execution(format!(
+                "Failed to execute query: {:?}",
+                failure
+            )));
+        }
+        _ => {}
+    }
```
milenkovicm: would it be able to hide this behind a common method or common interface? 

for example if you create a struct `SustraitExec` (use some better name), something along 

```rust
struct SustraitExec {
 scheduler : SchedulerGrpcClient,
}

impl SustraitExec {

 async fn execute_sustrait(plan: &[u8]) -> Result<RecordBatchStream> {
   // this boiler plate code 
 }
```

then later we can extend it to be FFI safe https://docs.rs/arrow/latest/arrow/ffi_stream/index.html

examples/examples/standalone-substrait.rs
milenkovicm: btw, i would be open to bring session context caching back if there is interest

mattcuento: Ahhh thanks for the context, this makes a lot more sense now as to what's going on.

> The reason to have it implemented the way it is was DDL statement handling in datafusion. From what I recall they will never hit query planner so they cant sent from client to scheduler.

Finally have a clear mental model on this. `SessionContext::execute_logical_plan` registers tables in the catalog but then overrides the plan with `SessionContext::return_empty_dataframe` so it's an `EmptyRelation`. Basically, it does hit the logical planner in the client but we early exit in `BallistaQueryPlanner::create_physical_plan` since we have an empty relation.

> btw, i would be open to bring session context caching back if there is interest

Bringing back session context caching does seem like the most simple path forward here to support Substrait plans I think. Do you have any concerns in doing so? I'd be happy to help with this

milenkovicm: If you're willing to implement it please do. It should not be too problematic. 

I can recall a few concerns from the past:

- actual cache invalidation (maybe we can learn from spark connect, as they use grpc) 
- do we uses a library for cache 
- will session config update work if we cache session context 

once you have rough idea we can have a discussion

mattcuento: > once you have rough idea we can have a discussion

Great, can do, will file a new issue once I have a rough idea on caching.

In relation to your earlier point:
> external catalog might be required or session at the scheduler side should keep the state.

I believe a combination of both might be required for a full-fledged solution.

Session caching on the scheduler-side will help with synchronizing settings like `CatalogOptions`, but I still think a user would need to use an external catalog. The scheduler catalog won't get populated with schema/table providers on DDL statements as we clarified earlier that DDL doesn't reach the scheduler. 

As a result, any Substrait `ReadType::NamedTable` that we try to convert to a `TableScan` in the scheduler will fail to resolve table references into `TableProvider`s.

In essence, it sounds like to support Substrait fully, one would either need a persistent external catalog, or the scheduler would need to be authoritative and process DDL. 

Do you have any thoughts here?

milenkovicm: External catalog would solve all our issues I would say 😁

Can we make a "mock external catalog" to fix failing test, merge this PR and then think about next steps. 

DataFusion does not propagate DDLs at the moment, but if we  expose substrait we should not assume client is DataFusion, so if DDL does not come we can always push back with claim the client is broken 😁. Definitely we would need to provide cached session context or statefull catalog for substrait, with later being current workaround. 

Does it make sense ?

milenkovicm: If we document current limitation nobody could complain its a bug  😀

mattcuento: > Can we make a "mock external catalog" to fix failing test, merge this PR and then think about next steps.

Can do 🙂 

> DataFusion does not propagate DDLs at the moment, but if we expose substrait we should not assume client is DataFusion, so if DDL does not come we can always push back with claim the client is broken 😁. Definitely we would need to provide cached session context or statefull catalog for substrait, with later being current workaround.

Yep this makes sense to me!

ballista/client/src/extension.rs
```diff
@@ -159,6 +163,23 @@ impl SessionContextExt for SessionContext {
     }
 }
 
+#[cfg(feature = "standalone")]
```
milenkovicm: I don't think we should add this to ballista interface, if needed consider adding it as part of example.

ballista/client/src/extension.rs
```diff
@@ -234,3 +256,301 @@ impl Extension {
         Ok(scheduler_url)
     }
 }
+
```
milenkovicm: same with this one, if needed please add it as part of example

ballista/client/tests/substrait_client.rs
```diff
@@ -0,0 +1,160 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
milenkovicm: if needed should be moved to be part of example

ballista/client/Cargo.toml
```diff
@@ -33,24 +33,28 @@ ballista-core = { path = "../core", version = "51.0.0" }
...
 datafusion-proto = { workspace = true }
+datafusion-substrait = { workspace = true }
 env_logger = { workspace = true }
 rstest = { workspace = true }
 tempfile = { workspace = true }
 tonic = { workspace = true }
 
 [features]
-default = ["standalone"]
+default = ["standalone", "substrait"]
```
milenkovicm: client should not advertise substrait

examples/examples/standalone-substrait.rs
```diff
@@ -0,0 +1,82 @@
...
+use datafusion::execution::SessionStateBuilder;
+use datafusion::prelude::{SessionConfig, SessionContext};
+use datafusion_substrait::serializer::serialize_bytes;
+use futures::StreamExt;
+
+/// This example demonstrates executing Substrait plans against external tables
+/// using datafusion-substrait as our frontend.
+#[tokio::main]
+async fn main() -> Result<()> {
+    let catalog_list = Arc::new(MemoryCatalogProviderList::new());
```
milenkovicm, line 54: it would be good to add the comment here to describe whats the current limitation with the substrait

mattcuento: Done!

Issue #1367, opened 2026-01-06
Provide support for SSL/TSL connections


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Support optional TLS/SSL connections between clients schedulers and executors.  Currently plain HTTP is used.

**Describe the solution you'd like**

I’m not an expert in this area but sensible security would make sense.  Plain HTTP should remain the default.  Good documentation should also cover security configuration.

**Describe alternatives you've considered**

**Additional context**

mohit7705:
Hi @milenkovicm 👋  

I’d like to help with this issue.  
I’m planning to start by:
- Understanding the current HTTP communication flow between client, scheduler, and executors
- Researching how TLS is typically handled in similar distributed systems
- Proposing a minimal, optional TLS configuration (keeping plain HTTP as default)

Before proceeding further, I’d appreciate guidance on:
1. Whether TLS should be implemented at the transport layer only or also include mutual TLS (mTLS)
2. Any preferred libraries or existing patterns in the project for security-related configuration
3. Whether a design proposal document would be helpful before implementation

Thanks!

mohit7705:
## Design Proposal: Optional TLS via Channel and Server Factories

### Problem
Currently, communication between clients, schedulers, and executors uses plain HTTP, which does not provide transport-level encryption.

### Goal
Enable optional TLS support while keeping the default behavior unchanged (plain HTTP).

### Non-goals
- Enforcing TLS by default
- Implementing a full security or certificate management framework

### Proposed Solution
Introduce configurable abstractions for:
- `ChannelFactory` – responsible for creating outbound client channels
- `ServerFactory` – responsible for creating inbound server listeners

The existing HTTP-based implementation will remain the default.  
Users who require TLS can provide custom implementations (for example, using `rustls`) without modifying core logic.

### Configuration
Factories can be selected via existing configuration mechanisms (e.g., config files or environment variables), allowing users to opt into TLS explicitly.

### Backward Compatibility
This approach is fully backward-compatible. Existing deployments will continue to work without any changes.

### Next Steps
After feedback and approval:
1. Identify instantiation points for channels and servers
2. Introduce factory interfaces
3. Wire configuration to select the desired factory implementation

PR #1388, opened 2026-01-17
perf: add connection pooling for shuffle reader remote fetches

### LINKED ISSUE #736: [Shuffle] Support cache remote shuffle reader client in executor.

In our env, when executor grows to 100+, when running sql (shuffle 100G) with 10 parallel.  Got 
```
Caused by: java.lang.RuntimeException: Query with job id wXHGdKf failed due to Job failed due to stage 3 failed: Task failed due to runtime execution error: DataFusionError(ArrowError(ExternalError("Arrow error: External error: Shuffle fetch partition error from Executor 10.69.147.148 : 7115d88d-1212-4a4d-95b7-58f5a68600be, map_stage 2, map_partition 3, error desc: Error connecting to Ballista scheduler or executor at http://10.69.147.148:50051/: tonic::transport::Error(Transport, hyper::Error(Connect, Custom { kind: TimedOut, error: Elapsed(()) })) @ GrpcConnectionError")))
```
Seems cache the client is a solution, All cut-edge systems like `iox` and `tikv` did this.

**Additional context**

andygrove:
## Summary

Adds a `BallistaClientPool` that caches gRPC connections to executors by (host, port). This avoids the overhead of establishing new connections for each partition fetch during shuffle reads.

### Changes

- Add `BallistaClientPool` struct in `client.rs` with:
  - `get_or_connect(host, port, max_message_size)` - gets cached connection or creates new one
  - `remove(host, port)` - removes a connection (useful for error recovery)
  - `clear()` - clears all cached connections
- Add `global_client_pool()` function for shared pool access across all shuffle reads
- Update `fetch_partition_remote()` in shuffle_reader.rs to use the connection pool
- **Remove stale connections from pool on fetch errors**

### Implementation Details

The pool uses a read-write lock pattern with `parking_lot::RwLock`:
- **Fast path**: read lock to check for existing connection (no contention for cache hits)
- **Slow path**: create connection without holding lock, then write lock to cache

Race handling: if multiple tasks try to connect to the same host simultaneously, each creates a connection but only one gets cached. The others detect this and return the cached one, dropping their extra connection. This is a rare case and avoids holding locks during async connection establishment.

### Error Handling

On `FetchFailed` or `GrpcActionError`, the connection is removed from the pool so subsequent requests create fresh connections. This handles cases where:
- The remote server closes the connection
- Network issues cause connection problems
- The executor is restarted

### Motivation

This addresses the TODO comment in shuffle_reader.rs:
```rust
```

Connection establishment involves TCP handshake, TLS negotiation (if enabled), and HTTP/2 setup. Reusing connections eliminates this overhead for repeated fetches from the same executor.

## Note

This PR was created with AI assistance using [Claude Code](https://claude.ai/code). All changes were reviewed and approved by a human maintainer.

## Test plan

- [x] Existing shuffle_reader tests pass (`cargo test -p ballista-core shuffle_reader`)
- [x] Existing client tests pass (`cargo test -p ballista-core client`)
- [ ] Manual testing with distributed queries to verify connection reuse

Closes #736

### DIFF
--- a/ballista/core/src/client.rs
+++ b/ballista/core/src/client.rs
@@ -19,6 +19,7 @@
 
 use std::collections::HashMap;
 use std::sync::{Arc, OnceLock};
+use std::time::{Duration, Instant};
 
 use std::{
     convert::{TryFrom, TryInto},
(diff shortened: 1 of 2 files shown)


milenkovicm:
is there a chance we cover  #1367 somehow as well? (if it is obvious)

milenkovicm:
there is a test https://github.com/apache/datafusion-ballista/blob/fce60b73533195abfbaf73e3ee361f1ea4a398b6/ballista/client/tests/context_checks.rs#L918 
which forces remote read always, not sure if that would cover your last task in the list.

ballista/core/src/client.rs
```diff
@@ -287,6 +289,122 @@ impl BallistaClient {
...
+/// establishing new gRPC connections for each partition fetch during shuffle reads.
+/// Connections are stored indefinitely and reused across multiple fetch operations.
+///
+/// # Thread Safety
+///
+/// The pool uses a `RwLock` to allow concurrent reads while ensuring exclusive
+/// access during connection creation. The `BallistaClient` itself is `Clone`
+/// (wrapping an `Arc`), so cloned clients share the underlying connection.
+#[derive(Default)]
+pub struct BallistaClientPool {
```
milenkovicm, line 319: just wonder would it make sense to use some off-the-shelf like `deadpool` or `r2d2` instead of implementing our own ?

If i'm not mistaken, current implementation can, in theory, leak connections; connection will be removed from the pool only when it fails, but, in theory, we can have a executor removed/replaced and connection never get chance to fail

andygrove: I pushed a TTL improvement, but moved this to draft for now

milenkovicm:
closes #736 i believe

sqlbenchmark:
## @sqlbenchmark usage

This bot is whitelisted for DataFusion committers.

### Commands

```
@sqlbenchmark tpch [--iterations N] [--scale-factor N] [--baseline BRANCH] [--baseline-pr N] [-c key=value ...]
@sqlbenchmark criterion [--bench NAME] [--filter PATTERN] [--baseline BRANCH] [--baseline-pr N]
@sqlbenchmark help
```

### Examples

```
@sqlbenchmark tpch
@sqlbenchmark tpch --iterations 5 --scale-factor 10
@sqlbenchmark tpch -c datafusion.execution.batch_size=4096
@sqlbenchmark tpch --baseline my-branch
@sqlbenchmark criterion
@sqlbenchmark criterion --bench sql_planner
@sqlbenchmark criterion --baseline my-branch
@sqlbenchmark criterion --bench sort_shuffle --baseline-pr 1430
```

---
*Automated by dfbench*

milenkovicm:
@sqlbenchmark tpch --iterations 5 --scale-factor 10

milenkovicm:
I spent some time with this, the reason being with latest changes TPCH (SF10) starts braking on my machine, as the process is getting out of ephemeral  client ports (around 16K on macos) and macos has 1 min `TIME_WAIT` before port can be re-assigned. 

so the ininital approach was to cache channel per address

```rust
static CHANNEL_CACHE: LazyLock<DashMap<String, Channel>> =
    LazyLock::new(DashMap::new);
```

clone it and re-use it. Aparently this does not work well if there is a lot of data to be transferred from server to client (everything get stuck) `initial_connection_window_size` & `initial_stream_window_size` help in some cases, not in all.

Looks like single tcp connection can't be pushed to transfer as much data as shuffle files generate. 

Alternative approach is to keep multiple connections for same address and then load balance across them 

```rust
static CHANNEL_CACHE: LazyLock<DashMap<String, Vec<Channel>>> =
    LazyLock::new(DashMap::new);
```

apparently connection re-use is not as trivial as i initially thought

milenkovicm:
closing in favour of #1578

Issue #1319, opened 2025-09-14
[EPIC] Shuffle file execs improvement


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

shuffle files execs could be improved in few ways

- [ ] make better metrics, especially at reader side
- [x] offload I/O operations to dedicated executor
- [ ] offload arrow flight server to dedicated executor

**Describe the solution you'd like**

there is design document at in #660

**Describe alternatives you've considered**

**Additional context**

- [x] #1287
- [ ] #660

PhVHoang:
Yes, that’s exactly what I meant when I mentioned TaskGroup as well. I brought it up because it was discussed in the related issue (#660).

I agree that having a separate IoExecutor for I/O-heavy workloads makes sense, and the same rationale applies to the Arrow Flight server. Happy to work on the IoExecutor implementation and explore how best to integrate it for shuffle I/O (and potentially Flight as well), if that direction aligns with your expectations.

milenkovicm:
closing this as there were quite few changes to shuffle files since this was opened

Issue #660, opened 2023-02-08
Proposal for more efficient disk-based shuffle mechanism


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
The current shuffle mechanism is too basic and produces too many small shuffle files.

**Describe the solution you'd like**
See https://docs.google.com/document/d/16SIEoniAWKSFt8XKDLsOfRQ0sU--5E9_OIyE42Zj808/edit?usp=sharing

**Describe alternatives you've considered**

**Additional context**

andygrove:
@thinkharderdev @yahoNanJing @mingmwang @avantgardnerio Let me know what you think. If this seems like a good idea, I may have some time later this month to try and implement it. I am also happy for someone else to pick this up.

Dandandan:
Sounds like a great idea to me

andygrove:
I ran into some challenges with the proposed design, which I have documented in the Google doc (near the end). Feedback welcome.

Dandandan:
I read through it, sounds indeed a bit more simple.

A nice side effect btw of this optimization is that limit on the shufflewriter is also a bit more effective - allowing to quicker terminate the tasks.

yahoNanJing:
Thanks @andygrove for raising the discussion for this topic. 

For the second approach, the optimization change is also limited when tasks of a query stage are assigned to different executors, which is a common case when using the RoundRobin task scheduling policy for load balancing.

Actually, to reduce the shuffle write file, I recommend to use the sort-based shuffle writer used in Spark https://issues.apache.org/jira/browse/SPARK-2045. Then for each original `ShuffleWriterExec`, there will be only 2 output files rather than N files for its downside stage. One file for shuffling data with concatenating all of the output partition data, and the other one for the indexes of each partition's offset in the data file.

An intuitive graph can be find here, https://github.com/blaze-init/blaze/blob/master/dev/doc/architectural_overview.md.

Hi @yjshen, could you share your opinions?

andygrove:
This sounds great @yahoNanJing. I would also like to hear if @yjshen has an opinion on this.

Issue #320, opened 2022-10-05
Implement memory management in shuffle writer


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
There is no memory management in shuffle writer

**Describe the solution you'd like**

Blaze implements file-based shuffle with memory management in https://github.com/blaze-init/blaze/blob/master/native-engine/datafusion-ext/src/shuffle_writer_exec.rs

Perhaps we can learn from that?

**Describe alternatives you've considered**

**Additional context**
See discussion in https://github.com/apache/arrow-datafusion/pull/3493#issuecomment-1247587017

mingmwang:
I will take care of this ticket.

PR #1389, opened 2026-01-17
feat: Add sort-based shuffle implementation

### LINKED ISSUE #1287: Investigate use of comet shuffle

Can it be investigated if use of comet shuffle will benefit ballista? If it would can we share the implementation. More details in #1245

**Additional context**

andygrove:
## Summary

This PR implements a sort-based shuffle mechanism for Ballista, similar to Spark's approach. Instead of creating one file per output partition (N×M files for N input partitions and M output partitions), each input partition writes to a single consolidated file with an index file mapping partition IDs to batch ranges.

### Related Issues
- Closes #1287 - Investigate use of comet shuffle
- Contributes to #1319 - [EPIC] Shuffle file execs improvement
- Addresses #660 - Proposal for more efficient disk-based shuffle mechanism
- Addresses #320 - Implement memory management in shuffle writer

## Key Features

- **Consolidated output files**: Each input partition produces one data file + one index file (2×N files instead of N×M)
- **Arrow IPC File format**: Uses FileWriter/FileReader for efficient random access to specific record batches
- **Memory management**: Configurable buffer sizes with spill-to-disk support to prevent OOM
- **Backward compatible**: Sort-based shuffle is opt-in via configuration; default remains hash-based shuffle

## Architecture

```
ballista/core/src/execution_plans/sort_shuffle/
├── mod.rs          # Module exports
├── config.rs       # SortShuffleConfig with buffer/memory settings
├── buffer.rs       # PartitionBuffer for in-memory buffering
├── spill.rs        # SpillManager for disk spilling
├── index.rs        # ShuffleIndex for batch range tracking
├── writer.rs       # SortShuffleWriterExec execution plan
└── reader.rs       # Reading logic with FileReader random access
```

## Configuration Options

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `ballista.shuffle.sort_based.enabled` | bool | false | Enable sort-based shuffle |
| `ballista.shuffle.sort_based.buffer_size` | bytes | 1MB | Per-partition buffer size |
| `ballista.shuffle.sort_based.memory_limit` | bytes | 256MB | Total memory limit for buffers |
| `ballista.shuffle.sort_based.spill_threshold` | float | 0.8 | Spill when memory exceeds this fraction |

### Core
- Added `sort_shuffle` module with all components (buffer, spill, index, writer, reader)
- Added `ShuffleWriter` trait for polymorphic shuffle writer handling
- Updated `shuffle_reader.rs` to detect and read sort-based shuffle format
- Uses Arrow IPC File format (FileWriter/FileReader) for efficient random access

### Executor
- Updated `DefaultExecutionEngine` to handle both `ShuffleWriterExec` and `SortShuffleWriterExec`
- Added `ShuffleWriterVariant` enum for type-safe dispatch

### Scheduler/Planner
- Updated `DistributedPlanner` to create `SortShuffleWriterExec` when enabled
- Updated `ExecutionStageBuilder` to use `ShuffleWriter` trait

### Benchmarks
- Added `shuffle_bench` binary for comparing hash vs sort shuffle performance

## Benchmark Results

Configuration: 1M rows, 4 input partitions, 16 output partitions

```
=== Hash-Based Shuffle ===
  Average time: 57.94ms
  Files created: 44
  Total size: 13077 KB
  Throughput: 493.81 MB/s

=== Sort-Based Shuffle ===
  Average time: 44.14ms
  Files created: 8
  Total size: 13101 KB
  Throughput: 648.14 MB/s
```

**Performance Improvements:**
- **24% faster** execution time (57.94ms → 44.14ms)
- **82% fewer files** (44 files → 8 files)
- **31% higher throughput** (493.81 MB/s → 648.14 MB/s)

## Test Plan

- [x] Unit tests for all new components
- [x] All 43 ballista-core tests pass
- [x] All 15 ballista-executor tests pass
- [x] All 50 ballista-scheduler tests pass
- [x] 17 end-to-end integration tests (`ballista/client/tests/sort_shuffle.rs`)
- [x] Comparison tests verifying sort shuffle produces same results as hash shuffle

## Running the Benchmark

```bash
# Default settings (1M rows, 16 output partitions)
cargo run --release --bin shuffle_bench

# Custom settings
cargo run --release --bin shuffle_bench -- --rows 1000000 --partitions 32 --input-partitions 8

# Run only sort shuffle  
cargo run --release --bin shuffle_bench -- --sort-only
```

## References

This implementation is inspired by:
- **Apache DataFusion Comet's shuffle writer**: https://github.com/apache/datafusion-comet/blob/main/native/core/src/execution/shuffle/shuffle_writer.rs
- **Apache Spark's sort-based shuffle**: https://spark.apache.org/docs/latest/rdd-programming-guide.html#shuffle-operations

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/benchmarks/src/bin/shuffle_bench.rs
+++ b/benchmarks/src/bin/shuffle_bench.rs
@@ -128,7 +128,7 @@ fn create_test_data(
     partition_count: usize,
 ) -> Vec<Vec<RecordBatch>> {
     let rows_per_partition = rows / input_partitions;
-    let batches_per_partition = (rows_per_partition + batch_size - 1) / batch_size;
+    let batches_per_partition = rows_per_partition.div_ceil(batch_size);
 
     let mut partitions = Vec::with_capacity(input_partitions);
     for p in 0..input_partitions {


ballista/core/src/execution_plans/sort_shuffle/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Sort-based shuffle implementation for Ballista.
+//!
+//! This module provides an alternative to the hash-based shuffle that writes
+//! a single consolidated file per input partition (sorted by output partition ID)
+//! along with an index file mapping partition IDs to byte offsets.
+//!
```
milenkovicm, line 23: maybe we could add paragraph from referred spark documentation, I personally find it as good algorithm outline: "Internally, results from individual map tasks are kept in memory until they can’t fit. Then, these are sorted based on the target partition and written to a single file. On the reduce side, tasks read the relevant sorted blocks."

ballista/client/tests/sort_shuffle.rs
```diff
@@ -0,0 +1,450 @@
...
+    use datafusion::execution::SessionStateBuilder;
+    use datafusion::prelude::{ParquetReadOptions, SessionConfig, SessionContext};
+    use std::collections::HashSet;
+
+    /// Creates a standalone session context with sort-based shuffle enabled.
+    async fn create_sort_shuffle_context() -> SessionContext {
+        let config = SessionConfig::new_with_ballista()
+            .set_str(BALLISTA_SHUFFLE_SORT_BASED_ENABLED, "true")
+            .set_str(BALLISTA_SHUFFLE_SORT_BASED_BUFFER_SIZE, "1048576") // 1MB
+            .set_str(BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT, "268435456"); // 256MB
```
milenkovicm, line 59: adding `.set_str(BALLISTA_SHUFFLE_READER_FORCE_REMOTE_READ, "true")` will force remote read using flight. at the moment tests will fail if set. I would suggest using `rstest` to test local and remote read (contex with and without this option set)

milenkovicm, line 59: also, one note, #1318 introduces block transfer rather than flight transfer to override and fail back to flight transport   `.set_str(BALLISTA_SHUFFLE_READER_REMOTE_PREFER_FLIGHT, "true")` is needed. 

I believe we would need to patch flight server to support transport of sorted shuffles

ballista/core/src/execution_plans/sort_shuffle/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Sort-based shuffle implementation for Ballista.
+//!
+//! This module provides an alternative to the hash-based shuffle that writes
```
milenkovicm: I would suggest change from:

"This module provides an alternative to the hash-based shuffle that writes" to
"This module provides an alternative to the hash-based shuffle. It writes"

I personally find it easier to understand, but it's a personal option

ballista/core/src/execution_plans/sort_shuffle/reader.rs
```diff
@@ -0,0 +1,195 @@
...
+/// * `index_path` - Path to the index file
+/// * `partition_id` - The partition to read
+///
+/// # Returns
+/// Vector of record batches for the requested partition.
+pub fn read_sort_shuffle_partition(
+    data_path: &Path,
+    index_path: &Path,
+    partition_id: usize,
+) -> Result<Vec<RecordBatch>> {
```
milenkovicm, line 67: Would it make sense to return a stream instead of loading all vectors in a single vector?

ballista/core/src/execution_plans/sort_shuffle/reader.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/// Reads all batches from a sort shuffle data file.
+///
+/// # Arguments
+/// * `data_path` - Path to the consolidated data file (Arrow IPC File format)
+///
+/// # Returns
+/// Vector of all record batches in the file.
+pub fn read_all_batches(data_path: &Path) -> Result<Vec<RecordBatch>> {
```
milenkovicm, line 234: Would it make sense to return a stream instead of loading all vectors in a single vector

ballista/core/src/execution_plans/sort_shuffle/spill.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    /// Returns true if the partition has spill files.
+    pub fn has_spill_files(&self, partition_id: usize) -> bool {
+        self.spill_files
+            .get(&partition_id)
+            .is_some_and(|v| !v.is_empty())
+    }
+
+    /// Reads all spill files for a partition and returns the batches.
+    pub fn read_spill_files(&self, partition_id: usize) -> Result<Vec<RecordBatch>> {
```
milenkovicm, line 158: would it make sense to avoid vector and use stream?

ballista/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -525,7 +528,58 @@ async fn fetch_partition_local(
...
+                    metadata.id.clone(),
+                    partition_id.stage_id,
+                    partition_id.partition_id,
+                    e.to_string(),
+                )
+            })?;
+            reader.schema()
+        };
+
+        let stream = futures::stream::iter(batches.into_iter().map(Ok));
```
milenkovicm: IMHO, getting stream from reader instead of all batches may make sense

milenkovicm:
Thanks @andygrove will try to have a look later. 

One question, I'm not sure if it makes sense. 

- What if instead spilling to temporary file spill goes to output file directly, and index to keep more than one partition id -> offset mapping. 
- Read would need to do few more file seeks, as batches for same partition are scattered around, but should not be too bad as reads should be able to read many batches together (as batches are buffered before write). This would save spill batch reconciliation at the end.

ballista/executor/src/flight_service.rs
```diff
@@ -95,8 +99,43 @@ impl FlightService for BallistaFlightService {
...
+                    let flight_data_stream = FlightDataEncoderBuilder::new()
+                        .with_schema(schema)
+                        .with_options(write_options)
+                        .build(stream)
+                        .map_err(|err| Status::from_error(Box::new(err)));
+
+                    return Ok(Response::new(
+                        Box::pin(flight_data_stream) as Self::DoGetStream
+                    ));
+                }
```
milenkovicm, line 136: nitpicking, consider adding else branch

ballista/core/src/execution_plans/sort_shuffle/spill.rs
milenkovicm: Do we need to keep this as vector, perhaps just return `StreamReader`?

milenkovicm (approved):
it looks good to me @andygrove 
there is one vector which should be stream, and a minor nitpicking

andygrove:
Thanks for the review @milenkovicm. I'll go ahead and merge and will keep iterating on this to address the remaining feedback.

PR #1393, opened 2026-01-19
doc: add more details for protobuf extension

LouisBurke:
Closes #.

 # Rationale for this change
Doc clean up and minor addition to extensions example docs.

Markdown/Sphinx changes.

No

### DIFF
--- a/docs/source/user-guide/extensions-example.md
+++ b/docs/source/user-guide/extensions-example.md
@@ -255,7 +255,7 @@ message LSample {
 message PMessage {
     oneof Extension {
         // opaque message means that extension codec does not handle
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/extensions-example.md
```diff
@@ -214,16 +214,60 @@ scheduler.
 The logical extension codec typically consists of two components: Google Protocol Buffer definitions:
 
 ```proto
+syntax = "proto3";
+
+package extension.ballista;
+
+// we do not use this in this example 
+// i'm leaving it as it might be important 
+// later
```
milenkovicm: we do use it in codecs, maybe just clarify that it will be translated to rust structures at the build time

Issue #1398, opened 2026-01-22
Improve Jupyter notebook support with SQL magic commands and examples


andygrove:
## Summary

Improve the Jupyter notebook experience for Ballista by adding SQL magic commands, example notebooks, and notebook-specific features. While basic notebook support already works via `_repr_html_`, there's an opportunity to provide a richer, more integrated experience.

## Current State

PyBallista already supports basic Jupyter usage:

```python
from ballista import BallistaSessionContext

ctx = BallistaSessionContext("df://localhost:50050")
df = ctx.sql("SELECT * FROM my_table LIMIT 10")
df  # Renders as HTML table via _repr_html_()
```

**What works today:**
- `_repr_html_()` - DataFrames render as styled HTML tables
- `to_pandas()` / `to_arrow_table()` / `to_polars()` - Data conversion
- `show()` - Terminal-style output
- Example `.py` files with `# %%` cell markers

### Phase 1: Documentation & Examples (Low Effort)

1. **Add example Jupyter notebooks** to `python/examples/`:
   - `getting_started.ipynb` - Basic connection and queries
   - `dataframe_api.ipynb` - DataFrame transformations
   - `distributed_queries.ipynb` - Multi-stage query examples

2. **Document notebook support** in `python/README.md`

### Phase 2: SQL Magic Commands (Medium Effort)

Add IPython magic commands for a more interactive SQL experience:

```python
%load_ext ballista.jupyter

# Connect to cluster
%ballista connect df://localhost:50050

# Line magic for simple queries
%sql SELECT COUNT(*) FROM orders

# Cell magic for complex queries
%%sql
SELECT 
    customer_id,
    SUM(amount) as total
FROM orders
GROUP BY customer_id
ORDER BY total DESC
LIMIT 10
```

**Implementation sketch:**

```python
# ballista/jupyter.py
from IPython.core.magic import Magics, magics_class, line_magic, cell_magic

@magics_class
class BallistaMagics(Magics):
    def __init__(self, shell):
        super().__init__(shell)
        self.ctx = None
    
    @line_magic
    def ballista(self, line):
        """Ballista commands: connect, status, disconnect"""
        cmd, *args = line.split()
        if cmd == "connect":
            self.ctx = BallistaSessionContext(args[0])
            return f"Connected to {args[0]}"
        elif cmd == "status":
            # Show cluster status
            pass
    
    @cell_magic
    def sql(self, line, cell):
        """Execute SQL query"""
[...]

PR #1399, opened 2026-01-22
docs: add Jupyter notebook support documentation

andygrove:
## Summary

Add documentation for using Ballista in Jupyter notebooks to the Python user guide.

## What's Included

New "Jupyter Notebook Support" section covering:

- **Basic Usage**: How DataFrames automatically render as HTML tables via `_repr_html_()`
- **Converting Results**: Examples of `to_pandas()`, `to_arrow_table()`, `to_polars()`, and `collect()`
- **Example Notebook Workflow**: Complete example showing setup, exploration, analysis, and visualization with matplotlib
- **Running a Local Cluster**: Using `setup_test_cluster()` for development/testing in notebooks

## Why This Change

While Ballista already supports Jupyter notebooks (the `_repr_html_` method is proxied through to DataFusion), this wasn't documented. Users may not realize they can use Ballista effectively in notebooks without explicit documentation.

This addresses part of #1398.

## Testing

Documentation-only change. Verified the markdown renders correctly.

### DIFF
--- a/docs/source/user-guide/python.md
+++ b/docs/source/user-guide/python.md
@@ -144,6 +144,105 @@ assert result.column(0) == pyarrow.array([5, 7, 9])
 assert result.column(1) == pyarrow.array([-3, -3, -3])
 ```
 
+## Jupyter Notebook Support
+
+Ballista works well in Jupyter notebooks. DataFrames automatically render as formatted HTML tables when displayed
+in a notebook cell.
+
+### Basic Usage
+
+```python
+from ballista import BallistaSessionContext
+
+# Connect to a Ballista cluster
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/python.md
```diff
@@ -144,6 +144,105 @@ assert result.column(0) == pyarrow.array([5, 7, 9])
...
+```
+
+### Example Notebook Workflow
+
+A typical notebook workflow might look like:
+
+```python
+# Cell 1: Setup
+from ballista import BallistaSessionContext
+from datafusion import col, lit
```
milenkovicm, line 201: Not used in this example

PR #1416, opened 2026-01-27
feat: Add spark-compat mode to integrate datafusion-spark features au…

### LINKED ISSUE #1397: Add Spark compatibility mode using datafusion-spark expressions
## Summary

Add a configuration option to enable Spark-compatible expression behavior by registering functions from the [`datafusion-spark`](https://crates.io/crates/datafusion-spark) crate. This would help users migrating from Spark get more consistent behavior without requiring a full Spark Connect implementation.

## Motivation

Ballista aims to be a compelling alternative to Apache Spark. While full Spark Connect protocol support is being addressed by other projects like [LakeSail Sail](https://github.com/lakehq/sail), there's a simpler improvement that would help Spark users: ensuring expression/function behavior matches Spark semantics.

The `datafusion-spark` crate (version 51.0.0, maintained alongside DataFusion) provides:
- Spark-compatible scalar functions
- Spark-compatible aggregate functions  
- Spark-compatible window functions
- Spark-compatible table functions

These functions implement Spark's specific semantics which can differ from DataFusion's defaults (e.g., null handling, type coercion, edge cases).

### New Configuration Option

Add a new Ballista configuration key:

```rust
pub const BALLISTA_SPARK_COMPAT_MODE: &str = "ballista.spark_compat_mode";
```

With the config entry:
```rust
ConfigEntry::new(
    BALLISTA_SPARK_COMPAT_MODE.to_string(),
    "Enable Spark compatibility mode which registers Spark-compatible expressions from datafusion-spark".to_string(),
    DataType::Boolean,
    Some("false".to_string())
)
```

### Implementation

When `ballista.spark_compat_mode` is enabled:

1. **Scheduler side**: Register datafusion-spark functions when creating the SessionContext
2. **Executor side**: Ensure the same functions are available during plan execution

```rust
use datafusion_spark::register_all;

if config.spark_compat_mode() {
    register_all(&mut ctx)?;
}
```

### Feature Flag

Add an optional feature to ballista-core and ballista-scheduler:

```toml
[features]
spark-compat = ["datafusion-spark"]

[dependencies]
datafusion-spark = { version = "51", optional = true }
```

This keeps the dependency optional for users who don't need Spark compatibility.

### CLI
```bash
ballista-scheduler --spark-compat-mode
ballista-executor --spark-compat-mode
```

### Environment Variable
```bash
BALLISTA_SPARK_COMPAT_MODE=true ballista-scheduler
```

### Programmatic
```rust
let config = BallistaConfig::builder()
    .set(BALLISTA_SPARK_COMPAT_MODE, "true")
    .build()?;
```

## Benefits

1. **Low effort, high value**: Leverages existing datafusion-spark crate
2. **Incremental migration path**: Users can test Spark compatibility without full commitment
3. **Transparent**: Clear config flag makes behavior explicit
4. **Optional**: Feature-flagged to avoid bloating builds for users who don't need it

## Future Extensions

This could be extended to include:
- Spark SQL dialect parsing (when available in DataFusion)
- Additional Spark-specific behaviors (null ordering, case sensitivity)
- Integration with datafusion-comet-spark-expr for even more compatibility

## References

- [datafusion-spark crate](https://crates.io/crates/datafusion-spark)
- [DataFusion Spark Functions docs](https://datafusion.apache.org/library-user-guide/functions/spark.html)
- [datafusion-comet-spark-expr](https://crates.io/crates/datafusion-comet-spark-expr) (alternative/complementary)
- Ballista Spark Connect discussion: #964

mattcuento:
…tomatically

Closes #1397.

 # Rationale for this change

Exposing `datafusion-spark` functions as a 'bundled' feature with Ballista. This simplifies augmenting Ballista with spark features for quicker and more simple experimentation/adoption.

Documentation has been added in the user guide to describe how to enable these functions in builds.

- Introduce `spark-compat` feature to register all spark scalar/agg/window/table functions by default
- Created helper `datafusion_x_functions` to register default (and spark) functions if applicable in the `SessionState` for scheduler/client usage
- Extended most usages of session state builder to register functions using the helper
- Register spark functions if applicable in the executor via `BallistaFunctionRegistry::default()`

I've rendered the docs locally to ensure all looks well.

- New feature flag `spark-compat` to automatically register Spark-compatible scalar, aggregate, and window functions from `datafusion-spark`

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -42,6 +42,10 @@ required-features = ["tls"]
 name = "remote-spark-functions"
 required-features = ["ballista-core/spark-compat"]
 
+[[example]]
+name = "standalone-substrait"
+required-features = ["ballista-scheduler/substrait"]
+
 [dependencies]
 # Optional dependency for TLS support
 rustls = { version = "0.23", features = ["ring"], optional = true }
@@ -74,9 +78,8 @@ url = { workspace = true }
 uuid = { workspace = true }
 
 [features]
-default = ["substrait", "standalone"]
+default = ["standalone"]
 standalone = ["ballista/standalone"]
-substrait = ["ballista-scheduler/substrait"]
 testcontainers = []
 # Enable TLS support for the mtls-cluster example
 tls = ["tonic/tls-ring", "dep:rustls"]


ballista/core/src/extension.rs
```diff
@@ -177,6 +203,9 @@ impl SessionStateExt for SessionState {
             .with_query_planner(Arc::new(planner))
             .build();
 
+        #[cfg(feature = "spark-compat")]
+        let session_state = register_spark_functions(session_state);
```
mattcuento: Instead of needing to call `register_spark_functions` from wherever we build state, I wonder if it'd be more elegant to push the registration of these functions/the spark-compat feature upstream to the datafusion `SessionStateBuilder`.

ballista/core/src/registry.rs
```diff
@@ -34,29 +44,68 @@ pub struct BallistaFunctionRegistry {
     pub aggregate_functions: HashMap<String, Arc<AggregateUDF>>,
     /// Window user-defined functions.
     pub window_functions: HashMap<String, Arc<WindowUDF>>,
+    /// Table functions.
```
milenkovicm: do we need table functions here? BFR is used at scheduler and executor after logical plan has been ser/de so table function should be resolved to TableProvider (or whatever is exactly) so table function should be encoded between client and scheduler

ballista/executor/src/executor_server.rs
```diff
@@ -661,6 +661,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
                         self.executor.function_registry.scalar_functions.clone(),
                         self.executor.function_registry.aggregate_functions.clone(),
                         self.executor.function_registry.window_functions.clone(),
+                        self.executor.function_registry.table_functions.clone(),
```
milenkovicm: same as previous comment, do se need table functions here?

examples/Cargo.toml
```diff
@@ -71,6 +71,7 @@ uuid = { workspace = true }
 
 [features]
 default = ["substrait", "standalone"]
+spark-compat = ["ballista-core/spark-compat"]
```
milenkovicm: we need to add

```
[[example]]
name = "remote-spark-functions"
required-features = ["spark-compat"]
```

at the bottom of the file, so it triggers so it triggers build with spark-compact

please add same for `substrait` example and remove it from default

milenkovicm (approved):
thanks @mattcuento 
just one comment to address, otherwise its a good in my opinion

examples/Cargo.toml
mattcuento: ah good call, added for both, did substrait in a separate commit. Ran both locally. Thanks!

PR #1420, opened 2026-01-28
feat: add `Dataframe.cache()` factory (no planner handling)

### LINKED ISSUE #1395: bug: Disable `DataFrame.cache()` for ballista

Ballista at the moment does not support cache functionality, `df.cache()` does not produce expected behaviour, so until we come up with the caching strategy for ballista I would suggest disabling the functionality 

DataFusion 52 comes with [method to override cache factory `SessionState::with_cache_factory`](https://github.com/apache/datafusion/issues/17297) we should use it to override cache behaviour when `SessionContext` is created, and return error if `df.cache` is called.

EDIT: maybe we could introduce logical plan extension which represents distributed cache, and then users can plug-in their behaviour on scheduler side.

Alternative would be to implement cache behaviour, which should be done, but as it is non trivial task, having dependencies on few other things we would have incorrect behaviour.

**Additional context**

- This task depends on #1357 
- Also there is example how to create cache factory in https://github.com/apache/datafusion/pull/19139

killzoner:

Closes #1395 (partially if we want to expand on future configurable cache later).

 # Rationale for this change

Disable explicitly `cache()` method with good user level error rather than relying on current `panic` 

- Add cache factory to properly error the unsupported feature (links to https://github.com/apache/datafusion/issues/18893)
- Cache factory creates cache node extension, which is serialized to scheduler but not handled for now (can be through custom quer planner)

User level error instead of panic

### DIFF
--- a/ballista/core/src/serde/mod.rs
+++ b/ballista/core/src/serde/mod.rs
@@ -191,21 +191,15 @@ impl LogicalExtensionCodec for BallistaLogicalExtensionCodec {
         &self,
         buf: &[u8],
         inputs: &[datafusion::logical_expr::LogicalPlan],
-        _ctx: &TaskContext,
+        ctx: &TaskContext,
     ) -> Result<datafusion::logical_expr::Extension> {
-        let plan: protobuf::BallistaLogicalPlanNode =
-            protobuf::BallistaLogicalPlanNode::decode(buf).map_err(|e| {
-                DataFusionError::Internal(format!(
-                    "Could not deserialize BallistaLogicalPlanNode: {e}"
-                ))
-            })?;
+        let plan = protobuf::BallistaLogicalPlanNode::decode(buf)
+            .ok()
+            .and_then(|node| node.logical_plan_type);
 
-        let plan: protobuf::ballista_logical_plan_node::LogicalPlanType =
-            plan.logical_plan_type.ok_or_else(|| {
-                DataFusionError::Internal(
-                    "Could not deserialize BallistaLogicalPlanNode because it's logical_plan_type is none".to_string()
-                )
-            })?;
+        let Some(plan) = plan else {
+            return self.default_codec.try_decode(buf, inputs, ctx);
+        };
 
         match plan {
             LogicalPlanType::CacheNode(plan_cache) => Ok(Extension {
@@ -246,7 +240,7 @@ impl LogicalExtensionCodec for BallistaLogicalExtensionCodec {
 
             Ok(())
         } else {
-            Ok(())
+            self.default_codec.try_encode(node, buf)
         }
     }
 


ballista/client/tests/context_unsupported.rs
```diff
@@ -96,4 +96,26 @@ mod unsupported {
 
         assert_batches_eq!(expected, &result);
     }
+
+    #[rstest]
+    #[case::standalone(standalone_context())]
+    #[case::remote(remote_context())]
+    #[tokio::test]
+    async fn should_support_caching_data_frame_without_error(
```
killzoner: This one is a bit duplicated with https://github.com/killzoner/datafusion-ballista/blob/8d0f8421df3391479153b018da6d9e879cd60124/ballista/client/tests/context_unsupported.rs#L67, however the previous one will still be useful when we want to test the user level caching strategy later

ballista/core/src/extension.rs
```diff
@@ -528,6 +531,27 @@ impl BallistaQueryPlannerExtension {
...
+#[derive(Debug)]
+struct BallistaCacheFactory;
+
+impl BallistaCacheFactory {
+    fn new() -> Self {
+        Self {}
+    }
+}
+
+impl CacheFactory for BallistaCacheFactory {
```
milenkovicm, line 738: can we make cache factory return a logical plan extension, something like https://github.com/apache/datafusion/blob/20d5f06a554f39407e52dd1d4848267dbc1467ce/datafusion-examples/examples/dataframe/cache_factory.rs#L103

but with two additional properties, session_id and cache_id (populated from random id), we can then serialize it and send it to scheduler. Scheduler will then fail as there is no implementation. So when users want to implement cache they just need to add new query planner

killzoner: Ah i see the link clearer with https://github.com/apache/datafusion-ballista/issues/1395#issuecomment-3798590795 now.
Didn't know about this behaviour with the scheduler, will look into this, thanks !

milenkovicm:
no need to implement anything on scheduler side at the moment

ballista/client/tests/context_unsupported.rs
```diff
@@ -96,4 +96,28 @@ mod unsupported {
...
+    #[case::remote(remote_context())]
+    #[tokio::test]
+    async fn should_support_on_cache_collect(
+        #[future(awt)]
+        #[case]
+        ctx: SessionContext,
+    ) -> datafusion::error::Result<()> {
+        let cached_df = ctx.sql("SELECT 1").await?.cache().await?;
+
+        // Collect fails because extension node can't be serialized
```
killzoner: Need to update this comment

ballista/core/proto/ballista.proto
```diff
@@ -27,6 +27,15 @@ option java_outer_classname = "BallistaProto";
 import "datafusion.proto";
 import "datafusion_common.proto";
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// Ballista Logical Plan
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
+message BallistaLogicalPlanCacheNode {
```
killzoner: Is there any specific place we want to put this?

milenkovicm: ballista proto is ok for now

ballista/client/tests/context_unsupported.rs
killzoner: done

ballista/core/proto/ballista.proto
```diff
@@ -27,6 +27,15 @@ option java_outer_classname = "BallistaProto";
 import "datafusion.proto";
 import "datafusion_common.proto";
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// Ballista Logical Plan
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
```
milenkovicm, line 38: I would suggest to add additional node, `BallistaLogicalPlaneNode` which would have `oneof` where `BallistaLogicalPlanCacheNode` would be part of it, similar to `BallistaPhysicalPlanNode`

maybe `CacheLogicalPlanExtension` would be better name for `BallistaLogicalPlanCacheNode`

ballista/core/src/extension.rs
```diff
@@ -665,6 +671,100 @@ impl BallistaConfigGrpcEndpoint {
...
+    fn inputs(&self) -> Vec<&LogicalPlan> {
+        vec![&self.input]
+    }
+
+    fn schema(&self) -> &DFSchemaRef {
+        self.input.schema()
+    }
+
+    fn expressions(&self) -> Vec<Expr> {
+        self.exprs.clone()
```
milenkovicm, line 803: can we just return empty vector instead of having property exprs, will they ever be non empty?

ballista/core/src/extension.rs
```diff
@@ -665,6 +671,100 @@ impl BallistaConfigGrpcEndpoint {
...
+    }
+}
+
+impl CacheFactory for BallistaCacheFactory {
+    fn create(
+        &self,
+        plan: LogicalPlan,
+        session_state: &SessionState,
+    ) -> datafusion::error::Result<LogicalPlan> {
+        Ok(LogicalPlan::Extension(Extension {
```
milenkovicm: i wonder if we should have a ballista config option to "shortcut" cache. At the moment cache() will fail as there is no cache functionality implemented. But if we flip the switch, instead of returning LogicalPlan extension we can return `plan`. There will be no cache but pipelines relying on cache will not fail (it will be noop), wdyt?

milenkovicm:
```
if config.ballista_cache_noop {
 plan
} else {
  Ok(LogicalPlan::Extension ...
  ..
  }
}

```

milenkovicm (commented):
thansk @killzoner  few comments and one idea. 

what do you think if we add configuration option which would make cache operation a no op instead of throwing exception?

ballista/core/proto/ballista.proto
killzoner: I was also wondering for the oneof, thanks will do

ballista/core/src/extension.rs
killzoner: from what I get from the trait, it could be non-empty if you use `with_exprs_and_inputs` ? maybe it's not intended this way, I'm not very familiar with this

https://github.com/apache/datafusion-ballista/pull/1420/changes#diff-77305d1314ffee1d338be417750116a3501c19274c426a3380e9efd967f25e6dR751-R765

ballista/core/src/extension.rs
killzoner: Would it not defeat the purpose of the initial issue https://github.com/apache/datafusion-ballista/issues/1395 of informing explicitely the user that the feature is not supported?

Will add this in a dedicated commit

ballista/client/tests/context_unsupported.rs
milenkovicm, line 68: i guess we should move this to other test, as it is not expected to fail

ballista/client/tests/context_unsupported.rs
```diff
@@ -143,4 +149,52 @@ mod unsupported {
...
+        assert!(result.is_ok());
+        let expected = [
+            "+----------+",
+            "| Int64(1) |",
+            "+----------+",
+            "| 1        |",
+            "+----------+",
+        ];
+        assert_batches_eq!(expected, &result?);
+
```
milenkovicm: we have other test to test this

killzoner: i wanted to group test with both `ballista.cache_noop` values, will move it to specific cases then

ballista/client/tests/context_unsupported.rs
```diff
@@ -143,4 +149,52 @@ mod unsupported {
...
+        assert!(result.is_err());
+        let err_msg = result.unwrap_err().to_string();
+        assert!(
+            err_msg.contains("No installed planner was able to convert the custom node to an execution plan: BallistaCacheNode"),
+            "Expected planner error, got: {err_msg}"
+        );
+
+        // opt in case, should succeed transparently without effective cache
+        ctx.sql("SET ballista.cache_noop = true")
```
milenkovicm: can we rename config to `ballista.cache.noop` or similar  ?

killzoner: done

ballista/client/tests/context_unsupported.rs
killzoner: removed some parts and moved the remainder next to `should_support_caching_data_frame` test

ballista/core/src/serde/mod.rs
```diff
@@ -186,17 +191,63 @@ impl LogicalExtensionCodec for BallistaLogicalExtensionCodec {
...
+
+            proto.encode(buf).map_err(|e| {
+                DataFusionError::Internal(format!(
+                    "failed to encode cache node logical plan: {e:?}"
+                ))
+            })?;
+
+            Ok(())
+        } else {
+            Ok(())
```
killzoner: this should be `self.default_codec.try_encode(node, buf)`

ballista/core/src/serde/mod.rs
```diff
@@ -186,17 +191,63 @@ impl LogicalExtensionCodec for BallistaLogicalExtensionCodec {
...
-        ctx: &TaskContext,
+        _ctx: &TaskContext,
     ) -> Result<datafusion::logical_expr::Extension> {
-        self.default_codec.try_decode(buf, inputs, ctx)
+        let plan: protobuf::BallistaLogicalPlanNode =
+            protobuf::BallistaLogicalPlanNode::decode(buf).map_err(|e| {
+                DataFusionError::Internal(format!(
+                    "Could not deserialize BallistaLogicalPlanNode: {e}"
+                ))
+            })?;
```
killzoner: there should probably be a fallback to `self.default_codec.try_decode(buf, inputs, ctx)` here

ballista/core/src/serde/mod.rs
milenkovicm: you might be right

ballista/core/src/serde/mod.rs
milenkovicm: yes

ballista/core/src/serde/mod.rs
killzoner: done in 38cd250c9016fb0cf4f73cd9cb6ea5e92260a9de

Issue #1396, opened 2026-01-22
Add interactive TUI (Text User Interface) for cluster monitoring


andygrove:
## Summary

Add an interactive terminal-based user interface (TUI) for monitoring Ballista clusters, viewing query plans, and inspecting job statistics. This would provide a lightweight alternative to a web UI that can be used directly from the command line.

## Motivation

Currently, monitoring a Ballista cluster requires either:
- Manually calling REST API endpoints with curl/httpie
- Building custom tooling to consume the API
- No built-in visualization of query plans or job progress

A TUI would make Ballista more accessible for operators and developers who prefer terminal workflows, and would provide immediate visibility into cluster health and job execution.

### Dashboard View
- Cluster overview: scheduler uptime, version, connected executors
- Real-time executor list with host, port, and last-seen status
- Job summary: running, queued, completed, failed counts

### Jobs View
- List all jobs with status, progress percentage, and stage completion
- Filter/search jobs by ID or name
- Sort by status, start time, or progress
- Cancel jobs interactively

### Job Detail View
- Stage breakdown with input/output rows and elapsed compute time
- Query plan visualization (render DOT graph as ASCII art or tree view)
- Per-stage drill-down

### Metrics View
- Display Prometheus metrics in a formatted table
- Highlight key metrics: job execution times, planning times, failure counts

## Available REST Endpoints

The scheduler already exposes all necessary data via REST API (requires `rest-api` feature):

| Endpoint | Data Available |
|----------|----------------|
| `GET /api/state` | Scheduler start time, version |
| `GET /api/executors` | Executor id, host, port, last_seen |
| `GET /api/jobs` | Job id, name, status, num_stages, completed_stages, percent_complete |
| `PATCH /api/job/{job_id}` | Cancel a job |
| `GET /api/job/{job_id}/stages` | Stage id, status, input_rows, output_rows, elapsed_compute |
| `GET /api/job/{job_id}/dot` | Graphviz DOT representation of execution graph |
| `GET /api/job/{job_id}/stage/{stage_id}/dot` | DOT for specific stage |
| `GET /api/metrics` | Prometheus metrics (requires `prometheus-metrics` feature) |

### Crate Structure
- New crate: `ballista-tui` (similar to `ballista-cli`)
- Or extend `ballista-cli` with a `--monitor` or `--tui` flag

### Recommended Libraries
- [ratatui](https://github.com/ratatui/ratatui) - Modern TUI framework (successor to tui-rs)
[...]

PR #1433, opened 2026-01-31
WIP: Ballista Text User Interface app

martin-g:

Part of #1396.

# Rationale for this change

See #1396 

Initial version of the TUI app that includes:
* new Cargo feature `tui` to include it as part of the CLI app
* New CLI option `--tui` to start the TUI app
* Header navigation (Dashboard, Jobs, Metrics)
* Help popup (activate with `?` or `h`)
* Dashboard view (scheduler state and executors info)
* Placeholder for the Jobs and Metrics views 
* Yaml configuration
  * hardcoded defaults
  * OS specific config (e.g. ~/.config/ballista-tui/config.yaml)
  * env variables, e.g. `BALLISTA_TUI_HTTP_TIMEOUT=3000`

New option for the CLI application, `--tui` that starts the TUI application

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,6 +71,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -547,7 +553,7 @@ dependencies = [
  "aws-sdk-ssooidc",
  "aws-sdk-sts",
  "aws-smithy-async",
- "aws-smithy-http",
+ "aws-smithy-http 0.62.6",
  "aws-smithy-json",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
@@ -608,7 +614,7 @@ dependencies = [
  "aws-credential-types",
  "aws-sigv4",
  "aws-smithy-async",
- "aws-smithy-http",
+ "aws-smithy-http 0.62.6",
  "aws-smithy-runtime",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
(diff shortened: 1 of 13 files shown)


martin-g:
The Dashboard view:
ballista-tui

The help popup:
ballista-tui-with-help

milenkovicm:
would it make sense to switch from CLI to TUI?  would it be possible to have  `\tui` command which would switch for SQL to cluster management?

ballista-cli/src/tui/infrastructure/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+http:
+  timeout: 2000
+"#;
+
+impl Settings {
+    pub(crate) fn new() -> Result<Self, ConfigError> {
+        let config_dir = dirs::config_dir()
+            .unwrap_or_else(|| dirs::home_dir().unwrap().join(".config"))
+            .join("ballista-tui");
```
milenkovicm: can we keep it as `ballista-cli`?

martin-g: How about just `ballista` ?
And then use `cli` and/or `tui` as a parent node in the config.

This way the same config file could be used to provide the defaults for the whole cli+tui app and maybe even for the scheduler and executors later. AFAIS the cli app starts a scheduler and executor if host+port are not provided.

ballista-cli/src/tui/ui/header.rs
```diff
@@ -0,0 +1,92 @@
...
+
+const MENU_ITEMS: [&str; 3] = ["Dashboard", "Jobs", "Metrics"];
+const PERCENTAGE: u16 = 100 / MENU_ITEMS.len() as u16;
+const MENU_CONSTRAINTS: [Constraint; MENU_ITEMS.len()] =
+    [Constraint::Percentage(PERCENTAGE); MENU_ITEMS.len()];
+const BANNER: &'static str = r#"
+ _               _                    _
+| \  _. _|_  _. |_     _ o  _  ._    |_)  _. | | o  _ _|_  _.
+|_/ (_|  |_ (_| | |_| _> | (_) | |   |_) (_| | | | _>  |_ (_|
+"#;
```
andygrove: Could we have capital 'F' in DataFusion

martin-g: I think it is capital. Let me experiment with some other text banner

martin-g:
> would it make sense to switch from CLI to TUI? would it be possible to have `\tui` command which would switch for SQL to cluster management?

implemented with 3b289d94

PR #1435, opened 2026-02-01
feat: add -c config override flag to tpch benchmark

andygrove:

Closes #.

 # Rationale for this change

Allow arbitrary ballista and datafusion config overrides when running the tpch benchmark via `-c key=value` flags, removing the need to add a dedicated CLI flag for each config option.

### DIFF
--- a/benchmarks/src/bin/tpch.rs
+++ b/benchmarks/src/bin/tpch.rs
@@ -417,7 +417,9 @@ async fn benchmark_ballista(opt: BallistaBenchmarkOpt) -> Result<()> {
 
         for kv in &opt.config_overrides {
             if let Some((key, value)) = kv.split_once('=') {
-                config = config.set_str(key.trim(), value.trim());
+                if let Err(e) = config.options_mut().set(key.trim(), value.trim()) {
+                    println!("Warning: could not set config '{}': {}", kv, e);
+                }
             } else {
                 println!(
                     "Warning: ignoring invalid config override '{}'. \


benchmarks/src/bin/tpch.rs
```diff
@@ -417,6 +424,17 @@ async fn benchmark_ballista(opt: BallistaBenchmarkOpt) -> Result<()> {
             config = config.set_str(BALLISTA_SHUFFLE_SORT_BASED_ENABLED, "true");
```
milenkovicm: Maybe it can be removed now?

benchmarks/src/bin/tpch.rs
```diff
@@ -417,6 +424,17 @@ async fn benchmark_ballista(opt: BallistaBenchmarkOpt) -> Result<()> {
...
         }
 
+        for kv in &opt.config_overrides {
+            if let Some((key, value)) = kv.split_once('=') {
+                config = config.set_str(key.trim(), value.trim());
+            } else {
+                return Err(DataFusionError::Configuration(format!(
```
milenkovicm: Not sure if we need to throw error or just print warning. When comparing two branches one may not have (new) config. Wdyt?

milenkovicm: I was wrong, I apologise. I thought `Config::set_str` will return an error, unfortunately it will panic  internally, and previous change makes sense. It your call @andygrove if you wish to return previous change

andygrove: The check is only for malformed key-value pairs - so if the the user calls `-c foo`, which is not a valid key-value pair then we now print a message instead of failing. I think the current version is fine?

milenkovicm: we probably need 

```rust
config.options_mut().set("","")

```

which will not panic, will return result which could be ignored

milenkovicm: > The check is only for malformed key-value pairs - so if the the user calls `-c foo`, which is not a valid key-value pair then we now print a message instead of failing. I think the current version is fine?

your call @andygrove, 

i believe its more important to use 

```
config.options_mut().set("","")
```

andygrove: Thanks. Updated.

PR #1450, opened 2026-02-10
minor: add test to cover IPC arrow file read

milenkovicm:

Closes #.

 # Rationale for this change

looks like arrow read is not supported, opened https://github.com/apache/datafusion/issues/20280 to add proto support for it.

- created "unsupported" test 
- removed unsupported tests
- clean up few minor things

### DIFF
--- a/ballista/client/tests/context_unsupported.rs
+++ b/ballista/client/tests/context_unsupported.rs
@@ -102,7 +102,7 @@ mod unsupported {
     #[case::remote(remote_context())]
     #[tokio::test]
 
-    async fn should_execute_sql_collect_from_arrow(
+    async fn should_execute_sql_collect_from_arrow_file(
         #[future(awt)]
         #[case]
         ctx: SessionContext,


ballista/client/tests/context_checks.rs
```diff
@@ -481,48 +483,6 @@ mod supported {
         Ok(())
     }
 
-    #[rstest]
-    #[case::standalone(standalone_context())]
-    #[case::remote(remote_context())]
-    #[tokio::test]
-    async fn should_execute_sql_app_name_show(
```
mattcuento, line 488: what's the reasoning on removing this one? Wondering since the table registered is in parquet

mattcuento: Nvm I see the sql is the exact same as `should_execute_sql_collect`. Was this case intended to assert on the job name/test session properties?

mattcuento (approved):
lgtm

ballista/client/tests/context_unsupported.rs
```diff
@@ -96,4 +96,51 @@ mod unsupported {
 
         assert_batches_eq!(expected, &result);
     }
+
+    #[rstest]
+    #[case::standalone(standalone_context())]
+    #[case::remote(remote_context())]
+    #[tokio::test]
+
+    async fn should_execute_sql_collect_from_arrow(
```
killzoner:
```suggestion
    async fn should_execute_sql_collect_from_arrow_file(
```
nit : to me arrow is memory, so it's not 100% clear at first sight that we're talking about serialized arrow file

ballista/client/tests/context_checks.rs
milenkovicm: there is another test which test app name i believe

PR #1478, opened 2026-02-27
feat: job scheduling with push based job status updates

### LINKED ISSUE #1475: `ExecuteQuery` to support push based job status notifications

At the moment, `ExecuteQuery` submits job, then it goes and pulls data from `GetJobStatusParams` every predefined number of milliseconds

https://github.com/milenkovicm/datafusion-ballista/blob/debc627353587f523ef2707101d40864699d1a72/ballista/core/src/execution_plans/distributed_query.rs#L369-L370

this does work but puts unnecessary pressure on scheduler, on top of this, minimum query latency is 50ms which is at the moment predefined sleep timeout, 

Provide additional method, something like `ExecuteQueryAndWaitJobCompletion`, which would return a stream of  `GetJobStatusResult` which would be push based. We would need to change scheduler task manager to send updates to observers. 

An alternative approach would be to change `rpc GetJobStatus (GetJobStatusParams) returns (GetJobStatusResult) {}` to return stream of `GetJobStatusResult`. 

**Additional context**

milenkovicm:

Closes #1475

# Rationale for this change

Initial implementation will submit job and then pull scheduler for status updates, this comes with the price of 50ms wait time between pulls, which will affect latency of short queries. checking job status also involves lock of few structures in scheduler. 

- created new grpc endpoint, `rpc ExecuteQueryPush (ExecuteQueryParams) returns (stream GetJobStatusResult)` which streams back `JobStatus` (which is default implementation for job submission now)
- changed `JobState::async fn submit_job(&self, job_id: String, graph: &ExecutionGraphBox) -> Result<()>` adding ability to register job status subscriber 
- add ability to fall back to previous implementation with `SET ballista.client.pull = true`

there are change with errors returned from `execute_query` aligning them across proto and substrait, and newly added `execute_query_push`

### DIFF
--- a/ballista/scheduler/src/cluster/memory.rs
+++ b/ballista/scheduler/src/cluster/memory.rs
@@ -390,15 +390,15 @@ struct ExtendedJobStatus {
 impl ExtendedJobStatus {
     fn update_subscribers(&self, status: JobStatus) {
         let job_id = status.job_id.clone();
-        if let Some(subscriber) = &self.subscriber {
-            if matches!(subscriber.try_send(status), Err(TrySendError::Full(_))) {
-                // to be considered if we need another task to try to push this notification
-                // at the moment, it does not look as necessary as, buffer should be big enough
-                error!(
-                    "jobs notification subscriber for job {} is blocked, can't deliver status update, job notification will be missed",
-                    job_id
-                )
-            }
+        if let Some(subscriber) = &self.subscriber
+            && matches!(subscriber.try_send(status), Err(TrySendError::Full(_)))
+        {
+            // to be considered if we need another task to try to push this notification
+            // at the moment, it does not look as necessary as, buffer should be big enough for all cases
+            error!(
+                "jobs notification subscriber for job {} is blocked, can't deliver status update, job notification will be missed",
+                job_id
+            )
         }
     }
 }


milenkovicm:
@sqlbenchmark tpch --iterations 5 --scale-factor 1

milenkovicm:
@danielhumanmod @mattcuento @killzoner would you have time for review ?

sqlbenchmark:
## Ballista TPC-H Benchmark Results

**PR:** #1478 - add additional test
**PR Commit:** `bbabe9e5`
**Base Commit:** `97452516` (main)
**Scale Factor:** SF10
**Iterations:** 3

### Query Comparison

| Query | main (ms) | PR (ms) | Change |
|-------|-----------|---------|--------|
| Q1 | 2092.20 | 2009.30 | :white_circle: -4.0% |
| Q2 | 932.50 | 804.20 | :green_circle: -13.8% |
| Q3 | 1319.40 | 1299.90 | :white_circle: -1.5% |
| Q4 | 619.70 | 609.50 | :white_circle: -1.6% |
| Q5 | 2834.90 | 2779.60 | :white_circle: -2.0% |
| Q6 | 564.90 | 532.50 | :green_circle: -5.7% |
| Q7 | 4658.00 | 4174.50 | :green_circle: -10.4% |
| Q8 | 3346.70 | 3252.70 | :white_circle: -2.8% |
| Q9 | 4915.50 | 4937.90 | :white_circle: +0.5% |
| Q10 | 1721.40 | 1638.10 | :white_circle: -4.8% |
| Q11 | 588.20 | 571.40 | :white_circle: -2.9% |
| Q12 | 1062.70 | 994.30 | :green_circle: -6.4% |
| Q13 | 1496.20 | 1494.80 | :white_circle: -0.1% |
| Q14 | 623.00 | 587.10 | :green_circle: -5.8% |
| Q15 | 623.70 | 593.20 | :white_circle: -4.9% |
| Q16 | 367.10 | 356.80 | :white_circle: -2.8% |
| Q17 | 3775.80 | 3719.20 | :white_circle: -1.5% |
| Q18 | 6599.40 | 6151.80 | :green_circle: -6.8% |
| Q19 | 1359.00 | 1281.40 | :green_circle: -5.7% |
| Q20 | 1063.50 | 1021.00 | :white_circle: -4.0% |
| Q21 | 5039.30 | 4944.10 | :white_circle: -1.9% |
| Q22 | 343.40 | 308.20 | :green_circle: -10.3% |

**Total:** Main=45946.50ms, PR=44061.50ms (-4.1%)

---
*Automated benchmark run by dfbench*

milenkovicm:
@sqlbenchmark tpch --iterations 3 --scale-factor 1

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -333,6 +343,82 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
+            settings,
+        } = query_params
+        {
+            let job_name = settings
+                .iter()
+                .find(|s| s.key == BALLISTA_JOB_NAME)
+                .and_then(|s| s.value.clone())
+                .unwrap_or_default();
+
+            let job_id = self.state.task_manager.generate_job_id();
```
danielhumanmod: This `job_id` is created, logged but no further usage. In addition, `self.submit_job` (L402) will generate a `job_id` again. 

We might want to use only one job_id here, maybe the one from `submit_job`?

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -453,6 +485,160 @@ async fn execute_query(
...
+        .into_inner();
+
+    let mut prev_status: Option<job_status::Status> = None;
+
+    loop {
+        let item = query_status_stream
+            .next()
+            .await
+            .ok_or(DataFusionError::Execution(
+                "Stream closed without job completing".to_string(),
```
danielhumanmod, line 546: Based on my understanding, the current user experience is that users receive an error when the stream disconnects.
For better UX, one option we could consider adding a fallback to polling if an error occurs? (totally optional, maybe in a follow-up PR)

ballista/scheduler/src/cluster/memory.rs
```diff
@@ -450,12 +472,23 @@ impl JobState for InMemoryJobState {
...
+            }
+
             self.completed_jobs
                 .insert(job_id.to_string(), (status.clone(), Some(graph.cloned())));
-            self.running_jobs.remove(job_id);
         } else {
             // otherwise update running job
-            self.running_jobs.insert(job_id.to_string(), status.clone());
+            if let Some(mut job_info) = self.running_jobs.get_mut(job_id) {
+                job_info.update_subscribers(status.clone()).await;
```
danielhumanmod: This code awaits while holding a DashMap shard lock (get_mut() guard). Since await may suspend the future, this might potentially block other accesses to the same shard. Should we extract the subscriber, release the lock, and then call update_subscribers.await?

ballista/scheduler/src/scheduler_server/grpc.rs
milenkovicm: this is a good catch

ballista/core/src/execution_plans/distributed_query.rs
milenkovicm: this error means that subscriber has been dropped (end of stream has been received) but, no job completion or failure status received. its a corner case, not sure if the reconnect would help in this case, but definitely we can investigate further

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -351,83 +362,107 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
                 .and_then(|s| s.value.clone())
                 .unwrap_or_default();
 
-            let job_id = self.state.task_manager.generate_job_id();
+            info!(
+                "execution query (PUSH) job received - session_id: {session_id}, operation_id: {operation_id}, job_name: {job_name}"
+            );
+
+            let (_session_id, session_ctx) = self
```
killzoner: not sure why we create another `session_id` here instead of getting it like pull version did

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -569,6 +600,67 @@ fn extract_connect_info<T>(request: &Request<T>) -> Option<ConnectInfo<SocketAdd
         .cloned()
 }
 
+impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T, U> {
```
killzoner, line 602: checked locally, the `'static` lifetime didn't seem necessary, it's probably enough from the signature of `SchedulerServer`

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -108,10 +110,33 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
+                        if let Some(subscriber) = subscriber {
+                            let job_status = JobStatus {
+                                job_id: job_id.clone(),
+                                job_name,
+                                status: Some(ballista_core::serde::protobuf::job_status::Status::Failed(
+                                    FailedJob { error, queued_at, started_at: timestamp_millis(), ended_at: timestamp_millis() }
+                                ))
+                            };
+
+                            let _ = subscriber.send(job_status).await;
```
killzoner: shouldn't we at least `error!` ? this seems corner case in corner case but always feels like worth logging

killzoner (approved):
It's pretty clean! 
One thing I always wonder when seeing such refactoring, since stream can always do one off response, should we simplify the gRPC endpoints to have only 1 `ExecuteQuery` endpoint that does streaming with 1 or multiple responses?

It comes at its own cost though because this involves refactoring the existing, and a breaking change (and the one off case is more tedious to handle with stream API)

ballista/scheduler/src/scheduler_server/grpc.rs
milenkovicm: we can remove it as follow up

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
milenkovicm: i agree with you, but it will log errors even if client disconnects so thats why i did not add it. will add warn logs

ballista/scheduler/src/scheduler_server/grpc.rs
milenkovicm: it should be exactly the same behavior, apart we do not use session_id after, am i missing something ?

killzoner: nvm, i got lost with the different `execute_query_push`, `execute_query_pull` and `execute_query`

ballista/scheduler/src/scheduler_server/grpc.rs
```diff
@@ -351,83 +362,107 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerGrpc
...
+
+            let (subscriber, rx) = tokio::sync::mpsc::channel::<JobStatus>(16);
+            let stream = ReceiverStream::new(rx).map(move |status| {
+                Ok::<_, tonic::Status>(GetJobStatusResult {
+                    status: Some(status),
+                    flight_proxy: flight_proxy.clone(),
+                })
+            });
+
+            log::trace!("setting job name: {job_name}");
```
martin-g: This trace duplicates the earlier one at line 387

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -453,6 +485,160 @@ async fn execute_query(
...
+        .execute_query_push(query)
+        .await
+        .map_err(|e| DataFusionError::Execution(format!("{e:?}")))?
+        .into_inner();
+
+    let mut prev_status: Option<job_status::Status> = None;
+
+    loop {
+        let item = query_status_stream
+            .next()
```
martin-g, line 543: Should this wait for a new message for a particular time? I.e. does it need a  timeout ?

ballista/core/src/config.rs
```diff
@@ -156,7 +158,11 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
         ConfigEntry::new(BALLISTA_SHUFFLE_SORT_BASED_BATCH_SIZE.to_string(),
                          "Target batch size in rows for coalescing small batches in sort shuffle".to_string(),
                          DataType::UInt64,
-                         Some((8192).to_string()))
+                         Some((8192).to_string())),
+        ConfigEntry::new(BALLISTA_CLIENT_PULL.to_string(),
+                         "Should client employ pull or push job tracking. In pull mode client will make a request to server in the loop, until job finishes. Pull mode is keep for legacy clients.".to_string(),
```
```suggestion
                         "Should client employ pull or push job tracking. In pull mode client will make a request to server in the loop, until job finishes. Pull mode is kept for legacy clients.".to_string(),
```

ballista/scheduler/src/scheduler_server/grpc.rs
martin-g: AFAIS the session_id is used later in an info!() call.
So, here you can remove the `.clone()` at line 370 and _move_ the earlier variable, and remove the leading `_` for the returned variable that will be used for the logging/tracing.

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -247,33 +247,62 @@ impl<T: 'static + AsLogicalPlan> ExecutionPlan for DistributedQueryExec<T> {
...
+                        .map(|b| b.get_array_memory_size())
+                        .unwrap_or(0),
+                );
+
+                metric_row_count.add(batch.as_ref().map(|b| b.num_rows()).unwrap_or(0));
+            });
+
+            let schema = self.schema();
+            Ok(Box::pin(RecordBatchStreamAdapter::new(schema, stream)))
+        } else {
```
martin-g, line 278: The bodies of `if` and `else` look very similar. Could the logic be shared ?

ballista/scheduler/src/cluster/memory.rs
```diff
@@ -380,12 +380,34 @@ impl InMemoryJobState {
...
+#[derive(Clone)]
+struct ExtendedJobStatus {
+    status: JobStatus,
+    subscriber: Option<JobStatusSubscriber>,
+}
+
+impl ExtendedJobStatus {
+    async fn update_subscribers(&self, status: JobStatus) {
+        if let Some(subscriber) = &self.subscriber {
+            let _ = subscriber.send(status).await;
```
martin-g: Consider spawning a new task for the send because this may block the sender if the receiver is slow for some reason.

ballista/scheduler/src/cluster/memory.rs
```diff
@@ -450,12 +472,29 @@ impl JobState for InMemoryJobState {
...
+                job_info.subscriber.clone()
+            } else {
+                Err(BallistaError::Internal(format!(
+                    "scheduler state can't find job: {}",
+                    job_id
+                )))?
+            };
+
+            if let Some(subscriber) = subscriber {
+                let _ = subscriber.send(status.clone()).await;
```
martin-g: Same here: consider spawning a new task for the send because this may block the sender if the receiver is slow for some reason. Or use `try_send()` if stopping early would be better

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -108,10 +110,33 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
                         let fail_message = format!("Error planning job {job_id}: {e:?}");
+
+                        // this is a corner case, as most of job status changes are handled in
+                        // job state, after job is submitted to job state
+                        if let Some(subscriber) = subscriber {
+                            let job_status = JobStatus {
+                                job_id: job_id.clone(),
+                                job_name,
+                                status: Some(ballista_core::serde::protobuf::job_status::Status::Failed(
+                                    FailedJob { error, queued_at, started_at: timestamp_millis(), ended_at: timestamp_millis() }
```
martin-g: nit: Extract `timestamp_millis()` into a local variable and reuse it for both `started_at` and `ended_at`.

ballista/core/src/execution_plans/distributed_query.rs
milenkovicm: we do not know for how long job is going to run, hard to limit

ballista/scheduler/src/cluster/memory.rs
milenkovicm: you're right, but there is hand full of states job can go through, the underlying channel is setup with buffer of 16 ( i would even increase it) rather than complicating code with new task. wdyt ?

ballista/scheduler/src/cluster/memory.rs
milenkovicm: try send may make more sense, will add it

ballista/scheduler/src/cluster/memory.rs
```diff
@@ -464,7 +464,7 @@ impl JobState for InMemoryJobState {
 
     async fn save_job(&self, job_id: &str, graph: &ExecutionGraphBox) -> Result<()> {
         let status = graph.status().clone();
-
+        println!("--------> {job_id} -> {:?}", status);
```
martin-g: debug leftover

milenkovicm: Removed

ballista/core/src/execution_plans/distributed_query.rs
milenkovicm: if part will be removed once else part get stabilized. not too many duplicated lines, would suggest to keep it as it is for, now if you agree

PR #1436, opened 2026-02-01
feat: Ballista Text User Interface

martin-g:
# Which issue does this PR close?

Closes #1396

# Rationale for this change

See #1396

# What changes are included in this PR?

Initial version of the TUI app that includes:
* new Cargo feature `tui` to include it as part of the CLI app
* New CLI option `--tui` to start the TUI app
* Header navigation (Dashboard, Jobs, Metrics)
* Help popup (activate with `?` or `h`)
* Dashboard view (scheduler state and executors info)
* Placeholder for the Jobs and Metrics views
* Yaml configuration
  * hardcoded defaults
  * OS specific config (e.g. ~/.config/ballista/config.yaml)
  * env variables, e.g. `BALLISTA_HTTP_TIMEOUT=3000`

# Are there any user-facing changes?

New option for the CLI application, --tui that starts the TUI application

PR #1480, opened 2026-02-27
fix: Don't use `maxrows` as a "fetched rows" but calculate it from the batches

martin-g:

Closes N/A.

https://github.com/apache/datafusion-ballista/pull/1436#issuecomment-3975057288

# Rationale for this change

The Ballista CLI app was wrongly using the `MaxRows` config setting to render the `"N row(s) fetched"`.
It should use the sum of rows of each RecordBatch

Use the slice of `RecordBatch`es to calculate the fetched rows

No public API changes!
An output text in the CLI is corrected.

### DIFF
--- a/ballista-cli/src/exec.rs
+++ b/ballista-cli/src/exec.rs
@@ -105,11 +105,6 @@ pub async fn exec_from_repl(ctx: &SessionContext, print_options: &mut PrintOptio
 
     let mut print_options = print_options.clone();
 
-    let max_rows = match print_options.maxrows {
-        datafusion_cli::print_options::MaxRows::Unlimited => usize::MAX,
-        datafusion_cli::print_options::MaxRows::Limited(max_rows) => max_rows,
-    };
-
     loop {
         match rl.readline("❯ ") {
             Ok(line) if line.starts_with('\\') => {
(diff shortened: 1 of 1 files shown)


ballista-cli/src/exec.rs
```diff
@@ -179,13 +175,13 @@ async fn exec_and_print(
     ctx: &SessionContext,
     print_options: &PrintOptions,
     sql: String,
-    row_count: usize,
 ) -> Result<()> {
     let now = Instant::now();
     let df = ctx.sql(&sql).await?;
     let schema = Arc::new(df.schema().as_arrow().clone());
     let results = df.collect().await?;
```
milenkovicm, line 183: should we keep max_rows, add limit here, if limited ?

ballista-cli/src/exec.rs
```diff
@@ -179,13 +175,13 @@ async fn exec_and_print(
...
-    print_options.print_batches(schema, &results, now, row_count, &Default::default())?;
+    let rows = &results.iter().map(|b| b.num_rows()).sum::<usize>();
```
milenkovicm, line 184: this makes sense, we calculate total number here

PR #1506, opened 2026-03-14
feat: enable scheduler rest api by default

milenkovicm:

Closes #.

 # Rationale for this change

- user experience improvement

- rest api enabled by default 
- provided configuration option to disable rest api at scheduler start
- `404` handler will return `content-type: application/json` and `{"status_code":404,"reason":"Not Found"}` response body
- new get endpoint `/api/version` which returns same data as `/api/state`
- will specific error saying `/api/*` is disabled
- json message returned with all errors 

- rest endpoint is enabled by default, with compile time and runtime option to disable it

### DIFF
--- a/ballista/scheduler/src/api/handlers.rs
+++ b/ballista/scheduler/src/api/handlers.rs
@@ -44,6 +44,8 @@ struct SchedulerStateResponse {
     graphviz_support: bool,
     spark_support: bool,
     scheduling_policy: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    advertise_flight_sql_endpoint: Option<String>,
 }
 
 #[derive(Debug, serde::Serialize)]
@@ -106,6 +108,11 @@ pub async fn get_scheduler_state<
         graphviz_support: cfg!(feature = "graphviz-support"),
         spark_support: cfg!(feature = "spark-compat"),
         scheduling_policy: data_server.state.config.scheduling_policy.to_string(),
+        advertise_flight_sql_endpoint: data_server
+            .state
+            .config
+            .advertise_flight_sql_endpoint
+            .clone(),
     };
     Json(response)
 }


ballista/scheduler/src/api/handlers.rs
```diff
@@ -122,15 +123,13 @@ pub async fn get_jobs<
...
+) -> Result<impl IntoResponse, SchedulerErrorResponse> {
     let state = &data_server.state;
 
-    let jobs = state
-        .task_manager
-        .get_jobs()
-        .await
-        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    let jobs =
+        state.task_manager.get_jobs().await.map_err(|_| {
```
martin-g: The error is also not logged by TaskManager::get_jobs().
IMO it should be logger either there or here.

ballista/scheduler/src/api/handlers.rs
```diff
@@ -216,9 +215,12 @@ pub async fn cancel_job<
         .await
         .map_err(|err| {
             tracing::error!("Error getting job status: {err:?}");
-            StatusCode::INTERNAL_SERVER_ERROR
+            SchedulerErrorResponse::with_error(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("Error getting job status: {}", err),
```
martin-g:
```suggestion
                format!("Error getting job status: {err}"),
```

ballista/scheduler/src/api/handlers.rs
```diff
@@ -274,13 +278,18 @@ pub async fn get_query_stages<
...
     if let Some(graph) = data_server
         .state
         .task_manager
         .get_job_execution_graph(&job_id)
         .await
-        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?
+        .map_err(|e| {
+            SchedulerErrorResponse::with_error(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                e.to_string(),
```
martin-g: Usually the cause is logged and some generic message is sent back to the client, to prevent showing any sensitive data to users who don't have the respective permissions.

ballista/scheduler/src/api/handlers.rs
```diff
@@ -460,7 +469,9 @@ pub async fn get_job_svg_graph<
                 &mut PrinterContext::default(),
                 vec![CommandArg::Format(Format::Svg)],
             )
-            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+            .map_err(|_| {
```
martin-g: The cause should probably be logger here.

ballista/scheduler/src/api/mod.rs
```diff
@@ -27,6 +27,7 @@ pub fn get_routes<
 ) -> Router {
     let router = Router::new()
         .route("/api/state", get(handlers::get_scheduler_state::<T, U>))
```
martin-g: Is /api/state going to be deprecated in favour of /api/version ?

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -130,17 +132,42 @@ pub async fn start_grpc_service<
...
-        .merge(tonic)
-        .into_make_service_with_connect_info::<SocketAddr>();
+    let final_route = if config.disable_rest {
+        tonic
+            .route(
+                "/api/{*path}",
+                axum::routing::any(|| async {
+                    SchedulerErrorResponse::with_error(
+                        StatusCode::NOT_FOUND,
+                        "Rest api has been disabled at startup".to_string(),
```
martin-g:
```suggestion
                        "REST API has been disabled at startup".to_string(),
```

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -130,17 +132,42 @@ pub async fn start_grpc_service<
...
 
     #[cfg(not(feature = "rest-api"))]
-    let final_route = tonic.into_make_service_with_connect_info::<SocketAddr>();
+    let final_route = tonic
+        .route(
+            "/api/{*path}",
+            axum::routing::any(|| async {
+                SchedulerErrorResponse::with_error(
+                    StatusCode::NOT_FOUND,
+                    "Rest api has been disabled at compile time".to_string(),
```
martin-g:
```suggestion
                    "REST API has been disabled at compile time".to_string(),
```

ballista/scheduler/src/api/handlers.rs
```diff
@@ -409,16 +418,16 @@ pub async fn get_job_dot_graph<
...
         .get_job_execution_graph(&job_id)
         .await
-        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?
+        .map_err(|_| SchedulerErrorResponse::new(StatusCode::INTERNAL_SERVER_ERROR))?
     {
         ExecutionGraphDot::generate(graph.as_ref())
-            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)
+            .map_err(|_| SchedulerErrorResponse::new(StatusCode::INTERNAL_SERVER_ERROR))
     } else {
         Ok("Not Found".to_string())
```
martin-g: This is not changed in this PR but it should probably be 404 ?!

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -130,17 +132,42 @@ pub async fn start_grpc_service<
...
     #[cfg(feature = "rest-api")]
-    let axum = get_routes(Arc::new(scheduler));
-    #[cfg(feature = "rest-api")]
-    let final_route = axum
-        .merge(tonic)
-        .into_make_service_with_connect_info::<SocketAddr>();
+    let final_route = if config.disable_rest {
+        tonic
+            .route(
+                "/api/{*path}",
```
martin-g: I think the routing related code and SchedulerErrorResponse should be in api/mod.rs.

ballista/scheduler/src/config.rs
```diff
@@ -190,6 +190,11 @@ pub struct Config {
         help = "The interval to check expired or dead executors"
     )]
     pub expire_dead_executor_interval_seconds: u64,
+
+    #[cfg(feature = "rest-api")]
+    /// Should the rest api be disabled
+    #[arg(long, default_value_t = false, help = "Should the rest api be disable")]
```
martin-g:
```suggestion
    #[arg(long, default_value_t = false, help = "Should the REST API be disabled")]
```

ballista/scheduler/src/config.rs
```diff
@@ -190,6 +190,11 @@ pub struct Config {
...
+    pub disable_rest: bool,
```
martin-g: 1) nit: `disable_xyz = false` is a double negation and harder to reason about than `enable_xyz = true`

2) also consider appending `_api`, i.e. `enable_rest_api = true`. Without `_api` it may confuse someone that it disables/enables everything else (the rest).

ballista/scheduler/src/api/handlers.rs
```diff
@@ -166,17 +165,17 @@ pub async fn get_job<
...
     let graph = data_server
         .state
         .task_manager
         .get_job_execution_graph(&job_id)
         .await
         .map_err(|err| {
-            tracing::error!("Error occurred while getting the execution graph for job '{job_id}': {err:?}");
-            StatusCode::INTERNAL_SERVER_ERROR
+            tracing::error!("Error occurred while getting the execution graph for job '{job_id}' reason: {err:?}");
+            SchedulerErrorResponse::with_error(StatusCode::INTERNAL_SERVER_ERROR, format!("Error occurred while getting the execution graph for job '{job_id}' reason: {}", err))
```
martin-g: The cause may show too much (e.g. sensitive) information to the end user. Consider using some generic message in the JSON response. 
Only people with access to the logs should be able to see the complete error.

ballista/scheduler/src/api/handlers.rs
martin-g: This should probably return 404

ballista/scheduler/src/api/handlers.rs
```diff
@@ -430,16 +439,16 @@ pub async fn get_query_stage_dot_graph<
...
         .get_job_execution_graph(&job_id)
         .await
-        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?
+        .map_err(|_| SchedulerErrorResponse::new(StatusCode::INTERNAL_SERVER_ERROR))?
     {
         ExecutionGraphDot::generate_for_query_stage(graph.as_ref(), stage_id)
-            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)
+            .map_err(|_| SchedulerErrorResponse::new(StatusCode::INTERNAL_SERVER_ERROR))
     } else {
         Ok("Not Found".to_string())
```
martin-g: 404 ?!

ballista/scheduler/src/api/mod.rs
milenkovicm: we could, i have no strong opinion about it

ballista/scheduler/src/scheduler_process.rs
milenkovicm: api module not loaded if rest api disabled, but i agree with you otherwise

ballista/scheduler/src/api/mod.rs
martin-g: It is a bit strange both endpoints returning the same response.
But if /app/version was returning only the version and /api/state just the start time then every client will need to make two separate REST requests ...

Here is a third option: rename `/api/state` to `/api/info`
Spring Boot uses /info for such "server info" data

ballista/scheduler/src/scheduler_process.rs
martin-g: How about `#[cfg(not(feature="rest-api"))]` in api/mod.rs and move the logic there ?

ballista/scheduler/src/api/mod.rs
milenkovicm:
```
$ curl localhost:50050/api/version
{"version":"52.0.0"}%
```

```
$curl localhost:50050/api/state
{"started":1773582812453,"version":"52.0.0","substrait_support":true,"keda_support":false,"prometheus_support":false,"graphviz_support":false,"spark_support":false}%
```

wdyt?

martin-g: Lovely !

ballista/scheduler/src/config.rs
milenkovicm: I agree about double negation but in this case it makes sene in my opinion

ballista/scheduler/src/scheduler_process.rs
milenkovicm: moved, to api as well

PR #1536, opened 2026-03-31
feat: Refactor DistributedQueryExec to support job callback mechanisim

### LINKED ISSUE #1344: Support `EXPLAIN ANALYZE` in Ballista

danielhumanmod:

Closes #1344 , as the Step 1 - Building the job-level dependency logic.

For the next step, we will apply `EXPLAIN ANALYZE` job as the first use case.

 # Rationale for this change

Ballista currently treats a single job as the atomic unit of scheduling, with no mechanism to run multiple jobs sequentially.

To support features like `EXPLAIN ANALYZE`, we need job callback mechanism. This PR refactor client side (`DistributedQueryExec`) to introduce a semantic like this
```
let job_a = job_a.and_then(|job_info| {
 job_b_definition(job_info).and_then(|...| {...}
}

submit (job_a)
```

image

No

### DIFF
--- a/ballista/core/src/execution_plans/distributed_query.rs
+++ b/ballista/core/src/execution_plans/distributed_query.rs
@@ -15,27 +15,25 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::client::BallistaClient;
 use crate::config::BallistaConfig;
-use crate::extension::{BallistaConfigGrpcEndpoint, SessionConfigExt};
+use crate::execution_plans::job_callback_handler::{
+    FetchResultsHandler, JobCompletionAction, JobCompletionHandler,
+};
+use crate::extension::SessionConfigExt;
+use crate::serde::protobuf::SuccessfulJob;
 use crate::serde::protobuf::get_job_status_result::FlightProxy;
 use crate::serde::protobuf::{
     ExecuteQueryParams, GetJobStatusParams, GetJobStatusResult, KeyValuePair,
     PartitionLocation, execute_query_params::Query, execute_query_result, job_status,
     scheduler_grpc_client::SchedulerGrpcClient,
 };
-use crate::serde::protobuf::{ExecutorMetadata, SuccessfulJob};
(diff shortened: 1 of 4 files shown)


ballista/core/src/execution_plans/job_callback_handler/fetch_results.rs
danielhumanmod, line 1: This file is mostly migrated from the old code in `distributed_query.rs`. In my initial plan we will add more handlers in this crate (eg. explain analyze impl)

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -434,74 +466,33 @@ async fn execute_query_pull(
                 partition_location,
                 ..
             })) => {
-                // Calculate job execution time (server-side execution)
```
danielhumanmod, line 432: This part is same for pull/push, so extract the common code to the handler

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -588,204 +579,16 @@ async fn execute_query_push(
...
```
danielhumanmod, line 586: This part is same for pull/push, so extract the common code to the handler

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -588,204 +579,16 @@ async fn execute_query_push(
...
+                    query_start_time,
                 });
-
-                break Ok(futures::stream::iter(streams).flatten());
             }
         };
     }
 }
-
-fn get_client_host_port(
```
danielhumanmod, line 638: migrate to the fetch result handler too

PR #1527, opened 2026-03-29
feat: remove full path from partition locations

milenkovicm:
# Which issue does this PR close?

Closes #.

# Rationale for this change

- at the moment full path is used to access shuffle files, having full path in each partition location brings quite overhead.
- absolute path is a bio of security issue, probably low but still

# What changes are included in this PR?

- remove full path from partition locations, generate it using available information 
- shuffle reader will have executor path injected in the same way like shuffle writer

there are going to be follow up changes after this one, adding subdiroctory to store shuffle files and session id between work dir and job

# Are there any user-facing changes?

- no

PR #1537, opened 2026-04-02
feat: move shuffle writer disk I/O off tokio worker threads

### LINKED ISSUE #1387: perf: Consider using async I/O (tokio::fs) in shuffle writer
## Description

The shuffle writer currently uses synchronous file I/O operations (`std::fs`) inside async functions, which blocks the tokio runtime. This could impact performance, especially when multiple tasks are executing concurrently on the same executor.

## Current Implementation

In `ballista/core/src/execution_plans/shuffle_writer.rs` and `ballista/core/src/utils.rs`, the following synchronous operations are used within async contexts:

- `std::fs::create_dir_all()` - creating output directories
- `std::fs::File::create()` - creating shuffle output files  
- `std::fs::metadata()` - reading file size after write
- `StreamWriter::write()` / `StreamWriter::finish()` - writing data (synchronous write to file)

## Proposed Optimization

Consider using `tokio::fs` equivalents or wrapping synchronous I/O in `spawn_blocking`:

### Option 1: Use tokio::fs directly
```rust
use tokio::fs::{self, File};
use tokio::io::BufWriter;

// Directory creation
tokio::fs::create_dir_all(&path).await?;

// File creation
let file = BufWriter::new(File::create(path).await?);

// File metadata
let num_bytes = tokio::fs::metadata(&w.path).await?.len();
```

### Option 2: Use spawn_blocking for StreamWriter
Since Arrow's `StreamWriter` is synchronous, the actual writes would need `spawn_blocking`:
```rust
let batch_clone = batch.clone();
tokio::task::spawn_blocking(move || {
    writer.write(&batch_clone)
}).await??;
```

## Trade-offs to Consider

1. **Complexity**: Async file I/O adds complexity, especially for the `StreamWriter` which is inherently synchronous
2. **Performance**: The benefit depends on workload - may be more significant with many concurrent tasks and small batches
3. **Memory**: `spawn_blocking` requires cloning data or careful lifetime management
4. **Compatibility**: Arrow's IPC writer is synchronous; full async would require buffering or a different approach

## Benchmarking Needed

Before implementing, it would be valuable to benchmark:
- Current performance with synchronous I/O
- Impact of blocking on concurrent task execution
- Whether `spawn_blocking` overhead is worth the async benefit

## Related

This was identified during a code review of the shuffle writer. See also PR #1386 which adds `BufWriter` for buffered I/O.

hcrosse:

Closes #1387.

## Rationale

The shuffle writer performs synchronous `std::fs` I/O inside async task contexts. Under concurrent shuffle workloads, this blocks tokio worker threads and degrades runtime responsiveness for co-located async tasks.

Both the hash repartition path in `shuffle_writer.rs` and the no-repartition path in `utils::write_stream_to_disk` now use a bounded `mpsc::channel(2)` to bridge the async record-batch stream to a `spawn_blocking` task that owns all file I/O. This is the same pattern already used in `flight_service.rs`.

- `shuffle_writer.rs`: hash repartition path rewritten with bounded-channel + `spawn_blocking`, `create_dir_all` converted to `tokio::fs`, removed double-counting of `write_time` in the no-repartition path, two write-failure regression tests added
- `utils.rs`: `write_stream_to_disk` rewritten with bounded-channel + `spawn_blocking`
- `ballista/core/Cargo.toml`: added `"fs"` feature to tokio dependency
- `shuffle_bench.rs`: added `--concurrency` flag for measuring concurrent partition execution, modeling how the Ballista executor dispatches tasks via `Semaphore(task_slots)`

## Benchmark results

All numbers from `shuffle_bench` with 3 iterations averaged, targeting local tmpfs. Format is before -> after (delta).

Default configuration (8 input partitions, 100 output partitions):

| Rows | Concurrency 1 | Concurrency 4 | Concurrency 8 |
|------|---------------|---------------|---------------|
| 1,000,000 | - | 114 -> 97 (-14.4%) | 98 -> 93 (-4.7%) |
| 5,000,000 | 648 -> 650 (+0.4%) | 284 -> 272 (-4.4%) | 291 -> 293 (+0.9%) |
| 10,000,000 | - | 498 -> 487 (-2.4%) | - |

Output partition sensitivity (5,000,000 rows, 8 input partitions):

| Output Part. | Concurrency 4 | Concurrency 8 |
|--------------|---------------|---------------|
| 50 | 155 -> 164 (+6.3%) | - |
| 100 | 284 -> 272 (-4.4%) | 291 -> 293 (+0.9%) |
| 200 | 486 -> 491 (+1.0%) | 475 -> 473 (-0.4%) |

Input partition sensitivity (5,000,000 rows, concurrency 4, 100 output partitions):

| Input Part. | Before -> After |
|-------------|-----------------|
| 4 | 238 -> 241 (+1.2%) |
| 8 | 284 -> 272 (-4.4%) |

Throughput results are mixed, which is expected when the I/O target is local tmpfs with near-zero latency. In this scenario the channel indirection adds overhead that roughly offsets the benefit of freeing tokio workers.

The benefit shows most clearly at moderate concurrency (c=4) with 8 input partitions, where the tokio threadpool has enough in-flight tasks to take advantage of freed workers. At c=1 there is no contention so no benefit. At c=8, I/O bandwidth saturates and the channel overhead roughly cancels out the worker-freeing benefit. The slight regression at p=50 reflects the smaller per-batch partition cost making the channel overhead proportionally larger.

The primary benefit is runtime health rather than raw throughput: keeping tokio workers unblocked prevents latency spikes in co-located async tasks. A separate runtime-probe microbenchmark confirmed that inline sync I/O causes 3-8 second probe latencies vs 2-8 ms with `spawn_blocking`, even when aggregate throughput is similar.

## Future work

Both partitioning and I/O currently run together in `spawn_blocking` because `BatchPartitioner::partition` takes a sync closure. Making the private `partition_iter` method public upstream would let us partition on the async side and only push I/O to the blocking pool. See apache/datafusion#21311.

## Are these changes tested?

All existing `shuffle_writer` tests pass. Two new regression tests verify that errors propagate correctly through the bounded-channel for both the no-repartition and hash repartition paths. The `shuffle_bench` binary now accepts a `--concurrency` flag for validating concurrent execution behavior.

No. This is an internal change to the shuffle writer execution path.


ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -255,96 +252,114 @@ impl ShuffleWriterExec {
...
-                                        });
+                    let schema = stream.schema();
+                    let (tx, mut rx) = tokio::sync::mpsc::channel::<RecordBatch>(2);
+                    let write_time = write_metrics.write_time.clone();
+                    let repart_time = write_metrics.repart_time.clone();
+                    let output_rows = write_metrics.output_rows.clone();
+
+                    let handle = tokio::task::spawn_blocking(move || {
+                        let mut writers: Vec<Option<WriteTracker>> =
+                            (0..num_output_partitions).map(|_| None).collect();
```
martin-g, line 283: nit:

```suggestion
                        let mut writers: Vec<Option<WriteTracker>> = vec![None; num_output_partitions];
```

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -255,96 +252,114 @@ impl ShuffleWriterExec {
...
+                    };
+                    drop(tx);
+
+                    let write_result = handle.await.map_err(|e| {
+                        DataFusionError::Execution(format!(
+                            "Shuffle writer task failed: {e}"
+                        ))
+                    })?;
+                    if let Some(e) = stream_err {
+                        return Err(e);
```
martin-g, line 391: If `write_result` is also an Err then it is silently dropped.
Log it as a warning, at least ?

benchmarks/src/bin/shuffle_bench.rs
```diff
@@ -240,19 +275,53 @@ async fn benchmark_sort_shuffle(
...
+    if concurrency <= 1 {
+        for partition in 0..input_partition_count {
+            let mut stream = shuffle_writer.execute(partition, task_ctx.clone())?;
+            let batches = utils::collect_stream(&mut stream).await?;
+            if let Some(batch) = batches.first() {
+                total_files += batch.num_rows();
+            }
+        }
+    } else {
+        let semaphore = Arc::new(tokio::sync::Semaphore::new(concurrency));
```
martin-g: This code looks very similar (same?!) to https://github.com/apache/datafusion-ballista/pull/1537/changes?w=1#diff-bbacf10874172011fe967d766f8642a5b84583e33215b128f606ad52ef994329R200-R231
Maybe extract it to a helper function ?

ballista/core/src/utils.rs
```diff
@@ -159,42 +159,74 @@ pub fn default_config_producer() -> SessionConfig {
...
-        StreamWriter::try_new_with_options(file, stream.schema().as_ref(), options)?;
-
-    while let Some(result) = stream.next().await {
-        let batch = result?;
+    let stream_err = loop {
+        match stream.next().await {
+            Some(Ok(batch)) => {
+                num_batches += 1;
+                num_rows += batch.num_rows();
+                num_bytes += batch.get_array_memory_size();
```
martin-g: This is from before this PR:
Here `num_bytes` is the number of bytes in memory.
https://github.com/hcrosse/datafusion-ballista/blob/fdaf1b764bd4ab3320e7d0edbe70a911a9e954b9/ballista/core/src/execution_plans/shuffle_writer.rs#L323 is the number of bytes on disk, which is **compressed**.

ballista/core/src/utils.rs
```diff
@@ -159,42 +159,74 @@ pub fn default_config_producer() -> SessionConfig {
...
-        num_bytes += batch_size_bytes;
+    let write_result = handle
+        .await
+        .map_err(|e| BallistaError::General(format!("Disk writer task failed: {e}")))?;
 
-        let timer = disk_write_metric.timer();
-        writer.write(&batch)?;
-        timer.done();
+    if let Some(e) = stream_err {
+        return Err(e.into());
```
martin-g, line 237: Same issue - if `write_result` is Err then it is silently dropped.

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -214,14 +214,12 @@ impl ShuffleWriterExec {
 
             match output_partitioning {
                 None => {
-                    let timer = write_metrics.write_time.timer();
                     path.push(format!("{input_partition}"));
-                    std::fs::create_dir_all(&path)?;
+                    tokio::fs::create_dir_all(&path).await?;
                     path.push("data.arrow");
                     let path = path.to_str().unwrap();
```
martin-g: This is old (before the PR):
The `.unwrap()` may cause panic on non-UTF8 paths.
Looking at the code I think there is no need to make it a `&str`. Passing it as `&Path` should be OK too.

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -255,96 +252,114 @@ impl ShuffleWriterExec {
...
-
-                                        writer.write(&output_batch)?;
-                                        writers[output_partition] = Some(WriteTracker {
-                                            num_batches: 1,
-                                            num_rows: output_batch.num_rows(),
-                                            writer,
-                                            path,
-                                        });
+                    let schema = stream.schema();
+                    let (tx, mut rx) = tokio::sync::mpsc::channel::<RecordBatch>(2);
```
martin-g: Why `2` ?
Isn't it too low ? This may lead to less throughput due to more context switches.

ballista/core/src/utils.rs
```diff
@@ -159,42 +159,74 @@ pub fn default_config_producer() -> SessionConfig {
...
 ) -> Result<PartitionStats> {
-    let file = BufWriter::new(File::create(path).map_err(|e| {
-        error!("Failed to create partition file at {path}: {e:?}");
-        BallistaError::IoError(e)
-    })?);
+    let schema = stream.schema();
+    let path_owned = path.to_owned();
+    let write_metric = disk_write_metric.clone();
+
+    let (tx, mut rx) = tokio::sync::mpsc::channel::<RecordBatch>(2);
```
martin-g: Why `2` ?
Isn't it too low ? This may lead to less throughput due to more context switches.

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Thanks for flagging - logging an `error` now

ballista/core/src/utils.rs
hcrosse: Also logging an `error` for this

benchmarks/src/bin/shuffle_bench.rs
hcrosse: My bad 😅! Extracted to the helper function `run_concurrent`

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Switched to a Path 👍

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Applied this but had to revert when I actually compiled locally. `vec![None; N]` needs `Clone` on the element type, and `WriteTracker` holds a `StreamWriter<BufWriter<File>>` which isn't cloneable. Kept the map/collect form instead

ballista/core/src/utils.rs
hcrosse: Good catch! Fixed to use the bytes on disk.

sqlbenchmark:
## Ballista TPC-H Benchmark Results

**PR:** #1537 - Fix formatting in shuffle_bench and sort imports
**PR Commit:** `bd676818`
**Base Commit:** `d0b51dd3` (main)
**Scale Factor:** SF1
**Iterations:** 3

### Query Comparison

| Query | main (ms) | PR (ms) | Change |
|-------|-----------|---------|--------|
| Q1 | 208.00 | 220.50 | :red_circle: +6.0% |
| Q2 | 102.30 | 87.50 | :green_circle: -14.5% |
| Q3 | 137.00 | 116.10 | :green_circle: -15.3% |
| Q4 | 75.30 | 64.00 | :green_circle: -15.0% |
| Q5 | 275.70 | 226.10 | :green_circle: -18.0% |
| Q6 | 60.00 | 61.70 | :white_circle: +2.8% |
| Q7 | 368.40 | 267.30 | :green_circle: -27.4% |
| Q8 | 322.80 | 289.40 | :green_circle: -10.3% |
| Q9 | 464.60 | 383.10 | :green_circle: -17.5% |
| Q10 | 166.70 | 133.40 | :green_circle: -20.0% |
| Q11 | 70.70 | 63.60 | :green_circle: -10.0% |
| Q12 | 108.90 | 111.10 | :white_circle: +2.0% |
| Q13 | 131.60 | 116.60 | :green_circle: -11.4% |
| Q14 | 63.60 | 63.00 | :white_circle: -0.9% |
| Q15 | 61.70 | 62.00 | :white_circle: +0.5% |
| Q16 | 54.50 | 52.30 | :white_circle: -4.0% |
| Q17 | 214.60 | 199.20 | :green_circle: -7.2% |
| Q18 | 310.90 | 326.30 | :white_circle: +5.0% |
| Q19 | 141.60 | 141.50 | :white_circle: -0.1% |
| Q20 | 96.10 | 90.10 | :green_circle: -6.2% |
| Q21 | 450.00 | 365.50 | :green_circle: -18.8% |
| Q22 | 48.20 | 47.50 | :white_circle: -1.5% |

**Total:** Main=3933.20ms, PR=3487.80ms (-11.3%)

---
*Automated benchmark run by dfbench*

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: I originally went with `2` just to be conservative, but ran some benchmarks and I bumped the channel capacity from 2 to 8, and made it configurable via `ShuffleWriterExec::with_channel_capacity()`. I also added `--channel-capacity` to the benchmark so this can be checked more easily. Happy to remove the configurability if you feel like that's overkill.

For benchmarking, I ran capacity sweeps from 1 to 32 channels and compared against the baseline at a few concurrency levels (without much variance):

| Cap | Median | Inv CS |
|-----|--------|--------|
| 1 | 217.0ms | 883 |
| 2 | 203.3ms | 845 |
| 4 | 202.6ms | 834 |
| 8 | 199.2ms | 817 |
| 16 | 203.0ms | 787 |
| 32 | 200.2ms | 724 |

| Mode | Median | Inv CS |
|------|--------|--------|
| Channel (cap=8) | 199.2ms | 817 |
| Inline (no channel) | 199.6ms | 183 |

There wasn't a huge difference at any level beyond going from 1 -> 2, except for larger capacity modestly reducing involuntary context switches (from 883 to 724 from cap 1 to 32), which is consistent with fewer producer/consumer sync points.

I also ran a load test with 8 concurrent shuffles on 4 tokio workers, measuring `tokio::time::sleep` jitter as a proxy for worker health. Inline I/O had p99 jitter of 120-160ms vs ~3-11ms with the channel.

milenkovicm:
@sqlbenchmark tpch --iterations 3 --scale-factor 10

milenkovicm:
@sqlbenchmark tpch --iterations 3 --scale-factor 1

sqlbenchmark:
## Ballista TPC-H Benchmark Results

**PR:** #1537 - configurable channel capacity for shuffle writer, defaults to 8
**PR Commit:** `7ed27529`
**Base Commit:** `d0b51dd3` (main)
**Scale Factor:** SF1
**Iterations:** 3

### Query Comparison

| Query | main (ms) | PR (ms) | Change |
|-------|-----------|---------|--------|
| Q1 | 208.10 | 221.90 | :red_circle: +6.6% |
| Q2 | 104.50 | 89.20 | :green_circle: -14.6% |
| Q3 | 136.70 | 99.90 | :green_circle: -26.9% |
| Q4 | 75.80 | 58.80 | :green_circle: -22.4% |
| Q5 | 275.60 | 236.90 | :green_circle: -14.0% |
| Q6 | 59.60 | 61.20 | :white_circle: +2.7% |
| Q7 | 369.50 | 258.60 | :green_circle: -30.0% |
| Q8 | 321.50 | 273.90 | :green_circle: -14.8% |
| Q9 | 459.40 | 370.40 | :green_circle: -19.4% |
| Q10 | 158.00 | 136.80 | :green_circle: -13.4% |
| Q11 | 71.90 | 63.00 | :green_circle: -12.4% |
| Q12 | 109.30 | 111.40 | :white_circle: +1.9% |
| Q13 | 129.20 | 112.00 | :green_circle: -13.3% |
| Q14 | 63.90 | 63.20 | :white_circle: -1.1% |
| Q15 | 76.00 | 77.00 | :white_circle: +1.3% |
| Q16 | 55.10 | 66.00 | :red_circle: +19.8% |
| Q17 | 210.70 | 198.60 | :green_circle: -5.7% |
| Q18 | 316.40 | 314.60 | :white_circle: -0.6% |
| Q19 | 141.10 | 139.80 | :white_circle: -0.9% |
| Q20 | 100.50 | 89.50 | :green_circle: -10.9% |
| Q21 | 431.80 | 362.00 | :green_circle: -16.2% |
| Q22 | 48.00 | 48.00 | :white_circle: +0.0% |

**Total:** Main=3922.60ms, PR=3452.70ms (-12.0%)

---
*Automated benchmark run by dfbench*

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -168,6 +173,7 @@ impl ShuffleWriterExec {
             shuffle_output_partitioning,
             metrics: ExecutionPlanMetricsSet::new(),
             properties,
+            channel_capacity: DEFAULT_SHUFFLE_CHANNEL_CAPACITY,
```
milenkovicm: if we want to make this configurable, perhaps we could add it to `BallistaConfig` ?

milenkovicm (commented):
just a heads up #1527 will remove full path from shuffles

ballista/core/src/utils.rs
```diff
@@ -159,46 +160,83 @@ pub fn default_config_producer() -> SessionConfig {
...
-    while let Some(result) = stream.next().await {
-        let batch = result?;
+        let options = IpcWriteOptions::default()
+            .try_with_compression(Some(CompressionType::LZ4_FRAME))?;
 
-        let batch_size_bytes: usize = batch.get_array_memory_size();
-        num_batches += 1;
-        num_rows += batch.num_rows();
-        num_bytes += batch_size_bytes;
+        let mut writer =
```
Dandandan, line 197: Instead of doing more IO - I think this _mainly_ does more CPU in parallel.
(Not necessarily bad, but it will have an effect that each partitions work will no longer be bound to a thread).

hcrosse: Yeah good point! [apache/datafusion#21311](https://github.com/apache/datafusion/issues/21311) would let us partition on the async side and only move serialization + IO to the blocking pool.

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Agreed! Added in 1f7f67f

ballista/core/src/utils.rs
Dandandan: I mean something slightly differently.
This adds extra parallelism by adding extra worker threads.
This helps when there is some skew, i.e. one partition is finishing faster than the other (and this is likely the source of the speedup) but also increases memory usage and decreases cache-friendliness.

With e.g. morsel-driven parallelism: https://github.com/apache/datafusion/issues/20529 this would be not less needed as each thread is able to pull enough work from the scan.

You could implement a similar approach for ShuffleReader as well, eliminating / reducing the need to do it at the write side.

ballista/core/src/serde/generated/ballista.rs
```diff
@@ -65,6 +65,8 @@ pub struct ShuffleWriterExecNode {
     pub output_partitioning: ::core::option::Option<
         ::datafusion_proto::protobuf::PhysicalHashRepartition,
     >,
+    #[prost(uint32, tag = "5")]
+    pub channel_capacity: u32,
```
milenkovicm: this is not needed if it is part of config

hcrosse: My bad, still getting a feel for the project structure. Thanks for flagging, I removed this.

ballista/core/src/utils.rs
hcrosse: Cool proposal! I agree, the channel does add extra parallelism beyond just moving I/O off the runtime, and morsel-driven execution would likely eat into the throughput gains we're seeing from that in the TCP-H benchmarks.

The main motivation for the `spawn_blocking` work is keeping blocking writes off tokio workers so concurrent tasks don't stall, and I think that need stays even with morsel-driven execution upstream.

On the `ShuffleReader` side, could you say more about what you have in mind? Local reads already use `spawn_blocking` but process files sequentially. Are you suggesting we pipeline reads through a channel the same way, or something else?

Dandandan: For local disk IO stalling the runtime is usually not a big factor, as there is not so much it can do at the same time (in the same worker thread). I bet if you would *only* use spawn_blocking or tokio::fs for writes you would see smaller gains?
That in itself is interesting to learn.

Maybe I am wrong and mostly projecting my experience with Parquet reads :)

For Shufflereader, i think its a bit tricky given that the current format is entirely stream based.
It probably would need to write the shuffle files in smaller blocks (e.g. of 50MB or something each) so it can parallelize / morselize reads better.

Dandandan: Another approach currently could be with IPC format: flush / finish each file after x MB (so you end up with multiple files per partition).

hcrosse: Thanks for the continued discussion! Going to try and take a deeper look at this in a couple days

hcrosse: Ran this experiment with three modes on the hash-repartition path: channel (production), serial-offload (spawn_blocking + ack, no overlap), and inline (everything on tokio worker).

Isolated microbenchmark (MemorySourceConfig, 100 iterations): all three modes within ~1%. No difference with a cheap producer.

TPC-H SF10 through a real Ballista cluster (22 queries, 5 iterations, 16 partitions):

| Mode | Total time |
|---|---|
| channel | 154.5s |
| serial-offload | 195.5s |
| inline (main) | 181.4s |

Channel is 14.8% faster than main. Serial-offload is roughly the same as main. The heavy shuffle queries (q5, q7, q9, q10, q21) showed 15-28% improvement with channel.

So you were right that the benefit comes from pipelining, not from the offload itself. Real queries have enough upstream work that the overlap pays off. The DedicatedExecutor runtime probe showed only a small jitter difference under concurrent load, so the worker health argument is weaker than I originally framed it.

hcrosse: > Another approach currently could be with IPC format: flush / finish each file after x MB (so you end up with multiple files per partition).

Interesting idea. Chunked output files would give the read side independent work units for parallel/morsel-driven reads. Probably out of scope for this PR but worth a follow-up. I think that even with chunked files you'd still want the write path off tokio workers.

Dandandan: Indeed, the write path (but not the compress path).

milenkovicm: @andygrove is experimenting with shuffle chunking in https://github.com/apache/datafusion-comet/pull/3845

I wonder should we focus more on improving sort shuffle implementation as this one won't scale with huge number of partitions

hcrosse: Looked at the comet PR and batch coalescing per partition seems like the right fix for large partition counts, so I agree we should do that as a follow-up (happy to take an attempt myself!)

IMO they're complementary, but open to bundling it here or closing this and starting fresh if that makes more sense.

milenkovicm: It's your call @hcrosse and @Dandandan how to follow up. 

@Dandandan wdyt?

Dandandan: I am not too opinionated how to move forward.

This PR shows great perf improvements, probable the cost (some extra memory usage / bit of extra tokio overhead) is probably relatively minor for the improvement.

In the future we can probably benefit from queu-based scan/ workstealing for file sources and implement it for shuffle scans, so partition skew is handled better.

At that point we probably can disable it in most cases (only keep IO off the path instead of compress + write)

milenkovicm:
@sqlbenchmark tpch --iterations 3 --scale-factor 1

sqlbenchmark:
## Ballista TPC-H Benchmark Results

**PR:** #1537 - Read channel capacity from BallistaConfig in create_query_stage_exec
**PR Commit:** `004c5f96`
**Base Commit:** `aba8c6c0` (main)
**Scale Factor:** SF1
**Iterations:** 3

### Query Comparison

| Query | main (ms) | PR (ms) | Change |
|-------|-----------|---------|--------|
| Q1 | 215.30 | 205.10 | :white_circle: -4.7% |
| Q2 | 116.50 | 141.70 | :red_circle: +21.6% |
| Q3 | 138.30 | 122.90 | :green_circle: -11.1% |
| Q4 | 83.40 | 77.30 | :green_circle: -7.3% |
| Q5 | 283.20 | 235.10 | :green_circle: -17.0% |
| Q6 | 73.50 | 59.50 | :green_circle: -19.0% |
| Q7 | 366.80 | 267.80 | :green_circle: -27.0% |
| Q8 | 340.30 | 274.80 | :green_circle: -19.2% |
| Q9 | 470.60 | 417.50 | :green_circle: -11.3% |
| Q10 | 218.40 | 139.80 | :green_circle: -36.0% |
| Q11 | 107.90 | 72.30 | :green_circle: -33.0% |
| Q12 | 105.90 | 107.40 | :white_circle: +1.4% |
| Q13 | 128.30 | 126.10 | :white_circle: -1.7% |
| Q14 | 64.40 | 89.90 | :red_circle: +39.6% |
| Q15 | 101.40 | 62.80 | :green_circle: -38.1% |
| Q16 | 79.70 | 76.40 | :white_circle: -4.1% |
| Q17 | 211.70 | 223.90 | :red_circle: +5.8% |
| Q18 | 338.60 | 334.70 | :white_circle: -1.2% |
| Q19 | 137.90 | 138.50 | :white_circle: +0.4% |
| Q20 | 128.70 | 128.60 | :white_circle: -0.1% |
| Q21 | 445.90 | 372.50 | :green_circle: -16.5% |
| Q22 | 58.00 | 77.70 | :red_circle: +34.0% |

**Total:** Main=4214.70ms, PR=3752.30ms (-11.0%)

---
*Automated benchmark run by dfbench*

milenkovicm:
@sqlbenchmark tpch --iterations 3 --scale-factor 10

sqlbenchmark:
## Ballista TPC-H Benchmark Results

**PR:** #1537 - Read channel capacity from BallistaConfig in create_query_stage_exec
**PR Commit:** `004c5f96`
**Base Commit:** `aba8c6c0` (main)
**Scale Factor:** SF10
**Iterations:** 3

### Query Comparison

| Query | main (ms) | PR (ms) | Change |
|-------|-----------|---------|--------|
| Q1 | 1880.30 | 1967.00 | :white_circle: +4.6% |
| Q2 | 761.00 | 709.90 | :green_circle: -6.7% |
| Q3 | 1212.00 | 1031.90 | :green_circle: -14.9% |
| Q4 | 567.00 | 469.40 | :green_circle: -17.2% |
| Q5 | 2622.60 | 2218.40 | :green_circle: -15.4% |
| Q6 | 501.00 | 508.10 | :white_circle: +1.4% |
| Q7 | 3797.00 | 2777.10 | :green_circle: -26.9% |
| Q8 | 3241.20 | 2714.10 | :green_circle: -16.3% |
| Q9 | 4757.30 | 4169.60 | :green_circle: -12.4% |
| Q10 | 1536.00 | 1432.90 | :green_circle: -6.7% |
| Q11 | 553.70 | 575.30 | :white_circle: +3.9% |
| Q12 | 972.70 | 1009.50 | :white_circle: +3.8% |
| Q13 | 1430.40 | 1262.50 | :green_circle: -11.7% |
| Q14 | 527.90 | 541.90 | :white_circle: +2.7% |
| Q15 | 585.80 | 594.50 | :white_circle: +1.5% |
| Q16 | 341.30 | 277.20 | :green_circle: -18.8% |
| Q17 | 3682.00 | 3832.50 | :white_circle: +4.1% |
| Q18 | 3677.10 | 3991.30 | :red_circle: +8.5% |
| Q19 | 1255.90 | 1246.50 | :white_circle: -0.7% |
| Q20 | 974.10 | 886.30 | :green_circle: -9.0% |
| Q21 | 4995.60 | 4118.60 | :green_circle: -17.6% |
| Q22 | 290.80 | 274.30 | :green_circle: -5.7% |

**Total:** Main=40162.70ms, PR=36608.80ms (-8.8%)

---
*Automated benchmark run by dfbench*

milenkovicm:
looks like there is issue with some of the tests, they never finish (probably there is some unhandled panic)

hcrosse:
I wasn't able to reproduce the failure locally, so I'm adding some temporary debug logs to try and see where it fails.

milenkovicm:
Will try later as well

milenkovicm:
i could not reproduce it locally as well

milenkovicm:
image

milenkovicm:
main compiles and runs all the rests no problem #1560

hcrosse:
Huh, not sure why they passed this time. I guess I'll strip my debug printlns and 🤞 it doesn't flake. I still was never able to reproduce locally regardless of doing it in/outside of a container or with a variety of core-counts to mimic the action runner.

milenkovicm:
GH actions have only 4 cpu cores, not sure if that could cause something to deadlock

milenkovicm:
Will try to review this weekend. I apologise for slow reviews; a bit stuck with time.

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -87,6 +90,8 @@ pub struct ShuffleWriterExec {
     metrics: ExecutionPlanMetricsSet,
     /// Plan properties
     properties: Arc<PlanProperties>,
+    /// Bounded channel capacity for the async-to-blocking I/O bridge
```
milenkovicm: do we need channel capacity here if we can retrieve it from config?

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -228,11 +243,11 @@ impl ShuffleWriterExec {
 
                     debug!("Writing results to {path:?}");
 
-                    // stream results to disk
                     let stats = utils::write_stream_to_disk(
                         &mut stream,
                         path.as_path(),
                         &write_metrics.write_time,
+                        self.channel_capacity,
```
milenkovicm: i believe we could retrieve it from 

```rust
let channel_capacity = context.session_config().ballista_config().shuffle_writer_channel_capacity();
```

ballista/executor/src/execution_engine.rs
```diff
@@ -106,6 +109,14 @@ impl ExecutionEngine for DefaultExecutionEngine {
             })?
             .data;
 
+        let channel_capacity = config
```
milenkovicm: i dont think we need this here, see previous comment

milenkovicm (commented):
thanks @hcrosse for your contribution, and sorry for slow review. 
I have few comments mostly regarding channel_capacity configuration, i believe they are easy to be resolved.

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Nope, dropped it!

ballista/core/src/execution_plans/shuffle_writer.rs
hcrosse: Agreed, thanks for the catch

hcrosse:
@milenkovicm haha no worries about the delays, I'm responding fairly slowly myself 😅 

Thank you for all of the reviews on this PR!

milenkovicm:
can you please have a look @andygrove as you're looking into shuffles at the moment

andygrove:
Thanks for the contribution @hcrosse! I'd like to help review and test this, but there have been substantial changes to shuffle so this PR likely needs rebasing and updating to work with sort-based shuffle, which is now the default.

andygrove:
> Thanks for the contribution @hcrosse! I'd like to help review and test this, but there have been substantial changes to shuffle so this PR likely needs rebasing and updating to work with sort-based shuffle, which is now the default.

Actually, these changes are specific to the hash-based shuffle implementation. Could you fix the current conflict and I can help test with the standalone shuffle benchmark binary that was recently added. We can then consider a second PR to apply the same improvement to sort-based shuffle. WDYT?

hcrosse:
> Actually, these changes are specific to the hash-based shuffle implementation. Could you fix the current conflict and I can help test with the standalone shuffle benchmark binary that was recently added. We can then consider a second PR to apply the same improvement to sort-based shuffle. WDYT?

Hey @andygrove , apologies for the slow reply! Fixed the conflict. I think a second PR sounds good if we still see perf improvements here :)

andygrove (approved):
Thanks @hcrosse. I ran locally and confirmed no performance regressions. I did not expect speed up for local benchmark against fast NVMe drives.

Issue #1426, opened 2026-01-29
[EPIC] Improve Cluster Observability


milenkovicm:
Two major directions (as of now): 

## Collect Executor Statistics

As starting point we could start by collecting and aggregating executor statistic, something similar to [spark UI executor tab](https://spark.apache.org/docs/3.5.7/web-ui.html#executors-tab)

![](https://spark.apache.org/docs/3.5.7/img/webui-exe-tab.png)

some statistics, such as memory utilisation, disk usage, would be collected on the executor, other like shuffle read and write may be collected on scheduler side. 

We would need to expose additional rest interface to expose collected metrics 

## Per Stage Flame Graph

Similar to [Nvidia RAPIDS Per Stage Flame Graph](https://nvidia.github.io/spark-rapids/docs/additional-functionality/per-stage-flamegraph.html) collect stats and produce flame graphs. 

We would need to further investigate what should be done 

## TODO: 

- [x]  https://github.com/apache/datafusion-ballista/pull/1547
- [ ] per stage flame graphs

mohit7705:
Hi, @milenkovicm 
I’m interested in contributing to this EPIC. I’d like to start with executor memory usage metrics. Could you please guide me on where to begin?
Thanks!

sandugood:
Any updates on this epic?

milenkovicm:
i guess you've covered first part @sandugood 
maybe we can leave second part for later, but its up to you. If you wish to leave it for later we you could have a look at AQE epic, that epic is may bring huge impact (up to you)

PR #1547, opened 2026-04-09
feat: Executor system & process metrics reporting

sandugood:

Targets #1426  in terms of introducing system-wide metrics observing mechanism for executor process.

 # Rationale for this change
Current implementation of executor's metrics includes the as_task_status() module-wise function, which reports the current status of the task being executed on the executor. It would be nice to also include the system-wide metrics for each executor, which will be sent to the scheduler on each heartbeat.

1. Added additional enum fields for Metric enum (with .proto implementation)
2. Implemented get_executor_metrics() function, that is getting called on each hearbeat to produce metrics
3. Refactored minor parts of executor implementation and added docstrings.

Yes. In the scheduler's endpoints. Another endpoint is added: **/api/executor/{executor_id}**


ballista/executor/src/executor_server.rs
```diff
@@ -476,12 +497,44 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorServer<T,
...
+            if executor_system.last_refresh.elapsed() >= Duration::from_millis(100) {
+                executor_system.last_refresh = Instant::now();
+                true
+            } else {
+                false
+            }
+        };
+
+        if should_refresh {
+            executor_system.system.refresh_all();
```
milenkovicm: should we use refresh_memory_specifics ?

ballista/executor/src/executor_server.rs
```diff
@@ -476,12 +497,44 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorServer<T,
...
+        };
+        executor_metrics.push(available_memory);
+
+        let used_memory = ExecutorMetric {
+            metric: Some(executor_metric::Metric::UsedMemory(
+                executor_system.system.used_memory(),
+            )),
         };
-        let executor_metrics = vec![available_memory];
+        executor_metrics.push(used_memory);
```
milenkovicm: can we cache all this and replace it only if memory has been refreshed

ballista/executor/src/executor_server.rs
```diff
@@ -476,12 +497,44 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorServer<T,
         };
     }
 
-    // TODO populate with real metrics
+    // Getting system-wide executor metrics
     fn get_executor_metrics(&self) -> Vec<ExecutorMetric> {
+        let mut executor_system = self.executor_system.lock().unwrap();
+        let should_refresh = {
+            if executor_system.last_refresh.elapsed() >= Duration::from_millis(100) {
```
milenkovicm: is there a way to configure this ?

milenkovicm (commented):
thanks @sandugood 
few minor comments, also its not clear if we do anything with the collected statistics at the executor side? where can we keep it?

milenkovicm:
I also wonder, should we get current process memory rather than system memory?

ballista/executor/src/executor_server.rs
sandugood: Yes, thanks. Implemented that

ballista/executor/src/executor_server.rs
sandugood: With newest commit it doesnt matter. This function is being called when we send hearbeat to the scheduler, so as a "metric collector" tool we shouldnt cache it I think and let it be recalculated at each heartbeat

ballista/executor/src/executor_server.rs
sandugood: It is already configured (missed that) with ExecutorProcessConfig.executor_heartbeat_interval_seconds

sandugood:
> I also wonder, should we get current process memory rather than system memory?

I had some thoughts about that. Current implementation tackles the overall cluster observability (something similar to Hadoop's cluster WEB-UI monitoring, coupled with YARN) .


Right now we are sending task metrics using OperatorMetricsSet with the TaskStatus proto specification. I wonder if that is enought data for constructing spark-like web ui.

sandugood:
Or we could possibly make it this way, let me know if these are viable:
- send the system-level memory specification during executor registration in the **register_executor** function. Right now we are sending only the **available_parallelism** specification. Maybe the memory would also be viable for displaying in the UI, something like: **used_by_user_task_mem** / **total_executor_available_memory**
- send hearbeats to the scheduler with the info of currently running tasks? Something like a HashMap with job_id (if I'm not mistaken and this is the most top-level id that represents **app** that users created, similar to Spark).

milenkovicm:
> Or we could possibly make it this way, let me know if these are viable:
> 
> * send the system-level memory specification during executor registration in the **register_executor** function. Right now we are sending only the **available_parallelism** specification. Maybe the memory would also be viable for displaying in the UI, something like: **used_by_user_task_mem** / **total_executor_available_memory**

we can continue sending what we send now, it would make sense to get overall system utilisation (also we can make this configurable ) 

perhaps we could use https://lib.rs/crates/memory-stats to capture process memory as well (not sure if this is expensive to do)

> * send hearbeats to the scheduler with the info of currently running tasks? Something like a HashMap with job_id (if I'm not mistaken and this is the most top-level id that represents **app** that users created, similar to Spark).

i think scheduler already knows list of running tasks/jobs not sure what would this do

sandugood:
Added the configurable logic (at the start of the Executor) - defaulting to **ExecutorMetricCollectionPolicy::Off** (no metrics collected/sent to scheduler). 

Also added the process memory usage as well (memory-stats with **features = ["always_use_statm"]** which uses /proc/self/statm on linux, which is very lightweight)

milenkovicm:
image

just for reference, what spark exposes on executor page

sandugood:
> image
> just for reference, what spark exposes on executor page

Thank you for the response

I think it seems somewhat close to that of Spark, because we got:
1. Executor Cores - we get it when we are registering the executor in the scheduler
2. Active/Failed/Complete - through the TaskStatus specification, as well as ShuffleWrite stats
3. Memory (RAM) stats (from this PR)

Is adding disk info to that a good idea?

milenkovicm:
at the time of registration we can pull as much information as we can (we should not collect it on subsequent calls) 

- cpu info (cpu cores available) 
- user info 
- architecture 
- kernel version 
- physical cores 
- distribution 
- hostname 
- ...

we already have `sysinfo` added, lets use it.

sandugood:
Whole bunch of new LOC is because of handling the test cases
Now system-wide information (kernel ver, disk space, etc) is sent during startup time and memory (system-wide/per process) info is being sent during heartbeats

milenkovicm:
Maybe a method to create those test instances could be created?

sandugood:
> Maybe a method to create those test instances could be created?

Greate suggestion, thanks. Implemented separate methods to create desirable ExecutorSpecification for tests

milenkovicm:
Do we store data at the scheduler side? Can it be retrieved at the rest interface?

sandugood:
> Do we store data at the scheduler side? Can it be retrieved at the rest interface?

As far as I see - we don't have a persistance storage for executors' info (on the scheduler side). ExecutorManager stores everything in memory. 

We retreive the info using the /api/executors, which I rewrote to add system specs and metrics (that we receive during heartbeats)

sandugood:
I think that storing everything in memory (as it is implemented now) is ok.
Potential users would like to receive the most up-to-date info about executors when the cluster is running IMHO.

milenkovicm:
Memory persistence should be ok

sandugood:
Should we introduce something like a db for executor metrics? If so - I'm open to suggestions

milenkovicm:
I don't think db is needed, if it is we could export it to prometheus but we can do it later. 

Maybe we could capture max memory consumed by executor as well, that would be useful

sandugood:
Added logic for storing memory peaks

milenkovicm (commented):
thanks for your contribution @sandugood, i think it makes a lot of sense. 

I have one question for you, at the moment 

```bash
datafusion git:(...) ✗ curl -s http://localhost:50050/api/executors | jq
[
  {
    "id": "85bb0012-f387-409a-bd33-4582c3e12c24",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1775998022000,
    "specification": {
      "task_slots": 14,
      "physical_cores": 14,
      "num_disks": 2,
      "total_disk_space": 1989325168640,
      "total_available_disk_space": 1438860362222,
      "open_files_limit": 10240,
      "system_name": "Darwin",
      "kernel_ver": "Darwin 25.4.0",
      "os_ver": "26.4.1",
      "os_ver_long": "macOS 26.4.1"
    },
    "metrics": [],
    "peaks": {
      "physical_memory": 0,
      "virtual_memory": 0
    }
  }
]
```
specification previously captured user configured task slots, not it captures os information as well, would it make sense to move os specific properties to its own property

something like: 

```bash
datafusion git:(...) ✗ curl -s http://localhost:50050/api/executors | jq
[
  {
    "id": "85bb0012-f387-409a-bd33-4582c3e12c24",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1775998022000,
    "specification": {
      "task_slots": 14,
    }
    "os": {
      "physical_cores": 14,
      "num_disks": 2,
      "total_disk_space": 1989325168640,
      "total_available_disk_space": 1438860362222,
      "open_files_limit": 10240,
      "system_name": "Darwin",
      "kernel_ver": "Darwin 25.4.0",
      "os_ver": "26.4.1",
      "os_ver_long": "macOS 26.4.1"
    },
    "metrics": [],
    "peaks": {
      "physical_memory": 0,
      "virtual_memory": 0
    }
  }
]
```

also, i would it make sense to to merge `peaks` and `metrics`? at the moment it prints 0 which may be a bit confusing. we could rename them `peaks/physical_memory` -> `peak_physical_memory` ...

wdyt ?

sandugood:
All of your suggestions make perfect sense. Going to implement them (possibly tomorrow)
Thanks for your review!

martin-g:
How much of this extra info should be delivered from `/api/executors` ?
Maybe some of it should be delivered only when requested, e.g. with `/api/executor/<id>` ?!

milenkovicm:
We probably should

sandugood:
> How much of this extra info should be delivered from `/api/executors` ? Maybe some of it should be delivered only when requested, e.g. with `/api/executor/<id>` ?!

Seems like a good thing. I think the memory specs (as well as the peaks metrics) should be accessible via the /api/executors (so the user could see an "overview" of the current cluster resources)

And other stuff (like OS, kernel info etc.) should be accessible via /api/executor/{id} (like a separate page for each executor if thinking in terms of web ui)

martin-g:
> if thinking in terms of web ui

I am thinking in terms of the TUI app (`ballista-cli --tui` or `ballista-cli > \tui` command)

sandugood:
Tackled this problem by splitting information about the executors between two endpoints: /api/executors and /api/executor/{executor_id} 
More information is available through the latter

sandugood:
Ran the tests locally (on ubuntu) and they passed. Was this a github issue or PR's?

milenkovicm:
running tests i see:

```
[2026-04-14T13:47:03Z WARN  ballista_executor::execution_loop] Executor poll work loop failed. If this continues to happen the Scheduler might be marked as dead. Error: code: 'The operation was cancelled', message: "h2 protocol error: http2 error", source: tonic::transport::Error(Transport, hyper::Error(Http2, Error { kind: Reset(StreamId(103), CANCEL, Remote) }))

thread 'basic::test_aggregate_stddev' (1898262) panicked at ballista/scheduler/src/scheduler_server/grpc.rs:102:47:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/e408947bfd200af42db322daf0fadfe7e26d3bd1/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
       
```

hence test never finish, scheduler gets killed

sandugood:
Converting to WIP. Going to tackle it next week and finish. Thanks for the feedback

sandugood:
Fixed the tests

ballista/executor/src/metrics/mod.rs
```diff
@@ -53,3 +53,38 @@ impl ExecutorMetricsCollector for LoggingMetricsCollector {
...
 }
+
+/// Configures which executor's metrics should be collected
+#[derive(Clone, Copy, Debug, serde::Deserialize, Default)]
+#[cfg_attr(feature = "build-binary", derive(clap::ValueEnum))]
+pub enum ExecutorMetricCollectionPolicy {
+    /// Collect only system-wide metrics
+    SystemOnly,
+    /// Collect only current process metrics
+    ProcessOnly,
```
milenkovicm, line 67: Can we make process only default please?

milenkovicm (commented):
thanks a lot @sandugood 

Just one little suggestion

ballista/executor/src/metrics/mod.rs
sandugood: Done, thanks for the suggestion

ballista/executor/src/config.rs
```diff
@@ -142,6 +143,14 @@ pub struct Config {
         help = "The heartbeat interval in seconds to the scheduler for push-based task scheduling."
     )]
     pub executor_heartbeat_interval_seconds: u64,
+    /// Specifying which metrics should be collected and sent to scheduler
+    #[arg(
+        short = 'm',
+        long = "metrics",
+        default_value_t = ExecutorMetricCollectionPolicy::ProcessOnly,
```
milenkovicm: I believe you could use ::default() so we do not need to keep changing on two locations

sandugood:
Should we maybe rename the arguments to their short variants? Like: **sys** for **system-only** and **all** for **system-and-process**?

milenkovicm (approved):
Will have another look when I get back to my computer, but overall I think this is great contribution

ballista/executor/src/config.rs
sandugood: Yes, changed that

milenkovicm:
> Should we maybe rename the arguments to their short variants? Like: **sys** for **system-only** and **all** for **system-and-process**?

Makes sense

milenkovicm:
@martin-g any comments ?

milenkovicm:
```
➜  ~ curl -s http://localhost:50050/api/executors | jq
[
  {
    "id": "aa68ce95-4175-404e-a036-db12d7399e79",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1777154502000,
    "specification": {
      "task_slots": 14
    },
    "metrics": [
      {
        "type": "proc_physical_memory",
        "value": 34521088
      },
      {
        "type": "proc_virtual_memory",
        "value": 454171721728
      }
    ]
  }
]
```

```
➜  ~ curl -s http://localhost:50050/api/executor/aa68ce95-4175-404e-a036-db12d7399e79 | jq
{
  "executor_info": {
    "id": "aa68ce95-4175-404e-a036-db12d7399e79",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1777154562000,
    "specification": {
      "task_slots": 14
    },
    "metrics": [
      {
        "type": "proc_physical_memory",
        "value": 36339712
      },
      {
        "type": "proc_virtual_memory",
        "value": 454207553536
      }
    ]
  },
  "os_info": {
    "system_name": "Darwin",
    "kernel_ver": "Darwin 25.4.0",
    "os_ver": "26.4.1",
    "os_ver_long": "macOS 26.4.1",
    "physical_cores": 14,
    "num_disks": 2,
    "total_disk_space": 1989325168640,
    "total_available_disk_space": 1342051923950,
    "open_files_limit": 10240
  }
}
```

one note, @sandugood peak memory disappeared from this version, not sure if i'm doing something wrong ?

we had a

```
    "peaks": {
      "physical_memory": 0,
      "virtual_memory": 0
    }
```

previously and we said we're going to merge it with metrics adding `peak_physical_memory` and `peak_virtual_memory` ?

sandugood:
You're doing everything right. Missed that part, fixing it

sandugood:
Done

ballista/executor/src/executor_process.rs
```diff
@@ -229,8 +235,30 @@ pub async fn start_executor_process(
...
+    let open_files_limit = System::open_files_limit().unwrap_or(0) as u64;
+
+    let disks = Disks::new_with_refreshed_list();
+    let num_disks = disks.list().len() as u32;
+    let mut total_disk_space: u64 = 0;
+    let mut total_available_disk_space: u64 = 0;
+    for disk in &disks {
+        total_disk_space += disk.total_space();
+        total_available_disk_space += disk.available_space();
+    }
```
milenkovicm: nit pick, can we make this as a function please ?

milenkovicm (approved):
the last minor comment and we can merge this, if not problem

ballista/executor/src/executor_process.rs
sandugood: Great suggestion, thanks
moved to **structure_executor_metadata** function

milenkovicm:
Great feature, thanks a lot @sandugood

PR #1568, opened 2026-04-22
feat: add SchedulerEventListener for executor lifecycle events

akshaychitneni:
Add SchedulerConfig::event_listeners to observe executor join/leave without modifying core. Notifies on both pull-based and push-based registration, and on executor loss.

For example, [Kubeflow Data Cache](https://www.kubeflow.org/docs/components/trainer/user-guides/data-cache/) when implemented as a ballista extension, it can use this to redistribute cached dataset shards when executors join or leave the cluster on kubernetes

Closes #.

 # Rationale for this change

There is currently no way for external plugins to observe executor join/leave events. Plugins that manage distributed state (e.g., in-memory data caches) need to react when executors are added or removed to rebalance data. This change adds a minimal, non-breaking extension point for executor lifecycle observation.

New SchedulerEventListener trait and event_listeners field on SchedulerConfig with builder method. Listeners are notified on executor registration (pull + push), removal, and loss.
  Updated custom-scheduler example with usage.

New optional field on SchedulerConfig. Default is empty — no behavior change for existing users. No breaking changes to public APIs.

### DIFF
--- a/ballista/scheduler/src/state/executor_manager.rs
+++ b/ballista/scheduler/src/state/executor_manager.rs
@@ -271,7 +271,26 @@ impl ExecutorManager {
             "save executor metadata {} with {} task slots (pull-based registration)",
             metadata.id, metadata.specification.task_slots
         );
-        self.cluster_state.save_executor_metadata(metadata).await
+
+        // Check if this is a new executor (not seen before)
+        let is_new = self
+            .cluster_state
+            .get_executor_metadata(&metadata.id)
+            .await
+            .is_err();
+
+        self.cluster_state
+            .save_executor_metadata(metadata.clone())
+            .await?;
+
+        // Notify event listeners only for newly registered executors
+        if is_new {
+            for listener in &self.config.event_listeners {
+                listener.on_executor_registered(&metadata);
+            }
+        }
+
+        Ok(())
     }
 
     /// Registers the executor with the scheduler for push-based task scheduling.
@@ -291,8 +310,13 @@ impl ExecutorManager {
         ExecutorManager::test_connectivity(&metadata).await?;
 
         self.cluster_state
-            .register_executor(metadata, specification)
+            .register_executor(metadata.clone(), specification)
             .await?;
+
+        // Notify event listeners about the new executor
+        for listener in &self.config.event_listeners {
+            listener.on_executor_registered(&metadata);
+        }
 
         Ok(())
     }
--- a/ballista/scheduler/src/config.rs
+++ b/ballista/scheduler/src/config.rs
@@ -28,11 +28,23 @@
 use crate::SessionBuilder;
 use crate::cluster::DistributionPolicy;
 use ballista_core::extension::EndpointOverrideFn;
+use ballista_core::serde::scheduler::ExecutorMetadata;
 use ballista_core::{ConfigProducer, config::TaskSchedulingPolicy};
 use datafusion_proto::logical_plan::LogicalExtensionCodec;
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
 use std::fmt::Display;
 use std::sync::Arc;
+
+/// Listener for scheduler-level executor lifecycle events.
+///
+/// Implement this trait to react to executor join/leave events.
+/// Used by external plugins to observe executor topology changes.
+pub trait SchedulerEventListener: Send + Sync {
+    /// Called when a new executor registers with the scheduler.
+    fn on_executor_registered(&self, metadata: &ExecutorMetadata);
+    /// Called when an executor is removed or lost.
+    fn on_executor_lost(&self, executor_id: &str);
+}
 
 /// Command-line configuration for the scheduler binary.
 #[cfg(feature = "build-binary")]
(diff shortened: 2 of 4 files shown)


milenkovicm:
thanks for your contribution @akshaychitneni

FWIW, I believe its possible to get cluster state notifications in current implementation:

```rust
let cluster = BallistaCluster::new_from_config(&config).await?;

// this gets you a stream of `ClusterStateEvent` 
cluster.cluster_state().cluster_state_events()
```

where `ClusterStateEvent` is defined as 

```rust
#[derive(Debug, Clone, PartialEq)]
pub enum ClusterStateEvent {
    /// An executor has been registered with the cluster.
    RegisteredExecutor {
        /// ID of the registered executor.
        executor_id: String,
    },
    /// An executor has been removed from the cluster.
    RemovedExecutor {
        /// ID of the removed executor.
        executor_id: String,
    },
}
```
Having it as external stream gives possibility to run on different thread than scheduler,  having a listener on configuration might have better API.

Let me know what you think

milenkovicm:
btw @akshaychitneni can't wait to see your next talk "Speed up Your ML Workloads With Kubernetes & Ballista Powered In-memory Data"  😀

akshaychitneni:
> thanks for your contribution @akshaychitneni
> 
> FWIW, I believe its possible to get cluster state notifications in current implementation:
> 
> ```rust
> let cluster = BallistaCluster::new_from_config(&config).await?;
> 
> // this gets you a stream of `ClusterStateEvent` 
> cluster.cluster_state().cluster_state_events()
> ```
> 
> where `ClusterStateEvent` is defined as
> 
> ```rust
> #[derive(Debug, Clone, PartialEq)]
> pub enum ClusterStateEvent {
>     /// An executor has been registered with the cluster.
>     RegisteredExecutor {
>         /// ID of the registered executor.
>         executor_id: String,
>     },
>     /// An executor has been removed from the cluster.
>     RemovedExecutor {
>         /// ID of the removed executor.
>         executor_id: String,
>     },
> }
> ```
> 
> Having it as external stream gives possibility to run on different thread than scheduler, having a listener on configuration might have better API.
> 
> Let me know what you think

Thanks for pointing out. I think the listener approach might be helpful when running in Kubernetes-like environments, as callbacks run inline in the scheduler's event loop, avoiding silent event loss scenarios that can occur with a separate consumer thread.

milenkovicm:
> 
> 
> Thanks for pointing out. I think the listener approach might be helpful when running in Kubernetes-like environments, as callbacks run inline in the scheduler's event loop, avoiding silent event loss scenarios that can occur with a separate consumer thread.  

Having listeners on scheduler loop may keep it busy from handling scheduler related tasks, like heart bits.

If listeners are slow to process they may impacts the scheduler. In case of separate consumer thread it's up to consumer to keep up with the stream

I'm using this approach already and no problems so far 

I'd suggest we start with current implementation and change if we find it problematic 

Wdyt?

ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -271,7 +271,26 @@ impl ExecutorManager {
             "save executor metadata {} with {} task slots (pull-based registration)",
             metadata.id, metadata.specification.task_slots
         );
-        self.cluster_state.save_executor_metadata(metadata).await
```
milenkovicm, line 274: removing this may break pull (or push) executor notifications

milenkovicm (commented):
I'd suggest to start with current implementation, we could add helper methods to stream line user experience

milenkovicm:
Btw we use this functionality for exactly same use case

ballista/scheduler/src/state/executor_manager.rs
akshaychitneni: this was not removed but added below after checking if the executor already exists in the cluster state

akshaychitneni:
> > Thanks for pointing out. I think the listener approach might be helpful when running in Kubernetes-like environments, as callbacks run inline in the scheduler's event loop, avoiding silent event loss scenarios that can occur with a separate consumer thread.
> 
> Having listeners on scheduler loop may keep it busy from handling scheduler related tasks, like heart bits.
> 
> If listeners are slow to process they may impacts the scheduler. In case of separate consumer thread it's up to consumer to keep up with the stream
> 
> I'm using this approach already and no problems so far
> 
> I'd suggest we start with current implementation and change if we find it problematic
> 
> Wdyt?

Makes sense. thanks for sharing. I will try out.

ballista/scheduler/src/state/executor_manager.rs
milenkovicm: Sorry my bad

milenkovicm:
you may need job state notifications:

```rust
cluster.job_state().job_state_events()
```
and / or custom task distribution policy 

```rust
let config: SchedulerConfig = SchedulerConfig {
  // custom task distribution policy which can decide task allocations 
  task_distribution: ballista_scheduler::config::TaskDistributionPolicy::Custom(Arc::new(LocationAwarePolicy::new()),
  ..Default::default()
};
```

both of API are not perfect

akshaychitneni:
cluster_state_events() worked for me. thanks for pointing out @milenkovicm

milenkovicm:
excellent @akshaychitneni , 

perhaps we should create shortcut  methods for event streams on `BallistaCluster`, to make it easier to find

```rust
// from 
let cluster = BallistaCluster::new_from_config(&config).await?;
cluster.cluster_state().cluster_state_events()
// to
let cluster = BallistaCluster::new_from_config(&config).await?;
cluster.cluster_state_events();
```

wdyt?

milenkovicm:
Closing this for now thanks @akshaychitneni

PR #1577, opened 2026-04-25
feat: make ballista client retry policy configurable 

eachsaj:
# Which issue does this PR close?

changes to support configurable IO settings.

Closes #.

 # Rationale for this change

IO Configurations are static; it would be better to make them more configurable.

# What changes are included in this PR?

Modified certain constants to configurable parameters and updated the code to support the configurations

# Are there any user-facing changes?
No

PR #1578, opened 2026-04-26
feat: Cache ballista clients on executor

### LINKED ISSUE #736: [Shuffle] Support cache remote shuffle reader client in executor.

milenkovicm:

Closes #736.

# Rationale for this change

As mentioned in https://github.com/apache/datafusion-ballista/pull/1388#issuecomment-3980564362, we might run out of available ports to establish new remote connections.  Furthermore, connection establishment can sometimes be costly, such as during a TLS handshake. Therefore, we need a mechanism to cache established connections.

- A new pluggable trait `BallistaClientPool` allows for different client caching strategies. 
- `DefaultBallistaClientPool` implementation 

`DefaultBallistaClientPool` does not use GRPC channel "clone and use" as it might be harder to configure property, hence pool may have one or more cached connections for same endpoint.

Change is NOT backward compatible for implementors of custom executors. 

At the moment client pool is disabled, executors behave same  like before. To enable it executors should be started with `--connection-cache` option

```
cargo run --bin ballista-executor -- --connection-cache 30
```

After trail period we will make caching default.

### TODO:

- [x] provide a way to disable purging thread
- [x] wire up with executor (expose pool configuration on executor side)

> Note: this change supersedes #1388

### DIFF
--- a/ballista/core/Cargo.toml
+++ b/ballista/core/Cargo.toml
@@ -49,6 +49,7 @@ async-trait = { workspace = true }
 aws-config = { version = "1.8.16", optional = true }
 aws-credential-types = { version = "1.2.0", optional = true }
 chrono = { version = "0.4", default-features = false }
+dashmap = { workspace = true }
 clap = { workspace = true, optional = true }
 datafusion = { workspace = true }
 datafusion-proto = { workspace = true }
--- a/ballista/core/Cargo.toml
+++ b/ballista/core/Cargo.toml
@@ -49,7 +49,6 @@ async-trait = { workspace = true }
 aws-config = { version = "1.8.16", optional = true }
 aws-credential-types = { version = "1.2.0", optional = true }
 chrono = { version = "0.4", default-features = false }
-dashmap = { workspace = true }
 clap = { workspace = true, optional = true }
 datafusion = { workspace = true }
 datafusion-proto = { workspace = true }
--- a/ballista/core/src/client.rs
+++ b/ballista/core/src/client.rs
@@ -114,6 +114,24 @@ impl BallistaClient {
         })
     }
 
+    /// Build a client with a lazy channel — no real server required.
+    /// Intended only for unit tests that exercise pool logic without I/O.
+    #[cfg(test)]
+    pub(crate) fn new_for_test(host: &str, port: u16) -> Self {
+        use tonic::transport::Endpoint;
+        let addr = format!("http://{host}:{port}");
+        let channel = Endpoint::from_shared(addr)
+            .expect("valid address")
+            .connect_lazy();
+        Self {
+            io_retries_times: 3,
+            io_retry_wait_time_ms: 250,
+            host: host.to_string(),
+            port,
+            flight_client: FlightServiceClient::new(channel),
+        }
+    }
+
     /// Retrieves a partition from an executor.
     ///
     /// Depending on the value of the `flight_transport` parameter, this method will utilize either
--- a/ballista/core/src/client.rs
+++ b/ballista/core/src/client.rs
@@ -114,10 +114,10 @@ impl BallistaClient {
         })
     }
 
-    /// Build a client with a lazy channel — no real server required.
-    /// Intended only for unit tests that exercise pool logic without I/O.
-    #[cfg(test)]
-    pub(crate) fn new_for_test(host: &str, port: u16) -> Self {
+    /// creates a ballista client to be used for testing
+    /// it connects lazily and which can not really
+    /// be reconfigured.
+    pub fn new_for_test(host: &str, port: u16) -> Self {
         use tonic::transport::Endpoint;
         let addr = format!("http://{host}:{port}");
         let channel = Endpoint::from_shared(addr)
--- a/ballista/core/src/client_pool.rs
+++ b/ballista/core/src/client_pool.rs
@@ -1,0 +1,402 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Connection pool for `BallistaClient`` instances.
+//!
+//! `DefaultBallistaClientPool` maintains a `VecDeque`` of idle clients per
+//! `(host, port)` key backed by a `DashMap`. Callers `BallistaClientPool::acquire` a
+//! `PooledClient` guard; when the guard is dropped the underlying client is
+//! returned to the idle deque automatically.
+//!
+//! If the connection errored, call`PooledClient::discard` before dropping so
+//! the pool closes the channel rather than reusing it.
+//!
+//! A background tokio task evicts idle connections that have not been returned
+//! within the configured `idle_timeout`.
+
+use std::collections::VecDeque;
+use std::fmt::Debug;
+use std::ops::{Deref, DerefMut};
+use std::sync::{Arc, Weak};
+use std::time::{Duration, Instant};
+
+use async_trait::async_trait;
+use dashmap::DashMap;
+
+use crate::client::BallistaClient;
+use crate::error::Result;
+use crate::extension::BallistaConfigGrpcEndpoint;
+use crate::utils::GrpcClientConfig;
+
+// ---------------------------------------------------------------------------
+// PooledClient guard
+// ---------------------------------------------------------------------------
+
+/// A [BallistaClient] checked out from a pool.
+///
+/// Implements [Deref] / [DerefMut] so it can be used exactly like a
+/// [BallistaClient]. On drop, the inner client is returned to the pool
+/// automatically. Call [PooledClient::discard] before dropping if the
+/// connection should **not** be reused (e.g. after a transport error).
+pub struct PooledClient {
+    client: BallistaClient,
+    /// Invoked in `Drop::drop` to push the client back into the idle deque.
+    /// `None` after `discard()` is called.
+    return_fn: Option<Box<dyn FnOnce(BallistaClient) + Send>>,
+}
+
+impl PooledClient {
+    pub(crate) fn new(
+        client: BallistaClient,
+        return_fn: Box<dyn FnOnce(BallistaClient) + Send>,
+    ) -> Self {
+        Self {
+            client,
+            return_fn: Some(return_fn),
+        }
+    }
+
+    /// Close the connection instead of returning it to the pool.
+    pub fn discard(mut self) {
+        self.return_fn = None;
+    }
+}
+
+impl Deref for PooledClient {
+    type Target = BallistaClient;
+    fn deref(&self) -> &Self::Target {
+        &self.client
+    }
+}
+
+impl DerefMut for PooledClient {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.client
+    }
+}
+
+impl Drop for PooledClient {
+    fn drop(&mut self) {
+        if let Some(f) = self.return_fn.take() {
+            let client = self.client.clone();
+            f(client);
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Trait
+// ---------------------------------------------------------------------------
+
+/// Manages a pool of reusable [BallistaClient] connections.
+#[async_trait]
+pub trait BallistaClientPool: Send + Sync + Debug {
+    /// Acquire an idle client for `(host, port)`, or create a new one if the
+    /// pool is empty for that key. The returned [PooledClient] returns itself
+    /// to the pool on drop.
+    async fn acquire(
+        &self,
+        host: &str,
+        port: u16,
+        config: &GrpcClientConfig,
+        customize_endpoint: Option<Arc<BallistaConfigGrpcEndpoint>>,
+    ) -> Result<PooledClient>;
+
+    /// Remove all idle clients that have been sitting unused longer than the
+    /// configured idle timeout. Called automatically by the background task in
+    /// [DefaultBallistaClientPool]; can also be invoked on demand.
+    async fn evict_idle(&self);
+}
+
+// ---------------------------------------------------------------------------
+// DefaultBallistaClientPool
+// ---------------------------------------------------------------------------
+
+struct IdleEntry {
+    client: BallistaClient,
+    idle_since: Instant,
+}
+
+type IdleMap = DashMap<(String, u16), VecDeque<IdleEntry>>;
+
+struct Inner {
+    idle: IdleMap,
+    idle_timeout: Duration,
+}
+
+/// Default pool implementation.
+///
+/// Keeps a `VecDeque<BallistaClient>` per `(host, port)`. Idle clients are
+/// evicted by a background tokio task that runs at `idle_timeout / 5`
+/// intervals (minimum 10 s). The task exits automatically when the pool `Arc`
+/// is dropped.
+#[derive(Clone)]
+pub struct DefaultBallistaClientPool {
+    inner: Arc<Inner>,
+}
+
+impl Debug for DefaultBallistaClientPool {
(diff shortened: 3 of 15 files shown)


ballista/executor/src/executor.rs
```diff
@@ -127,7 +127,30 @@ impl Executor {
         function_registry: Arc<BallistaFunctionRegistry>,
         metrics_collector: Arc<dyn ExecutorMetricsCollector>,
         concurrent_tasks: usize,
-        execution_engine: Option<Arc<dyn ExecutionEngine>>,
+        execution_engine: Arc<dyn ExecutionEngine>,
```
andygrove, line 130: Nice work splitting out `with_default_execution_engine`. One thing worth checking. This changes `Executor::new` from `Option<Arc<dyn ExecutionEngine>>` to `Arc<dyn ExecutionEngine>`. Anyone downstream calling `Executor::new(..., None)` will hit a compile error once this lands. Could we keep the old signature and have `with_default_execution_engine` be purely additive? If we do want the breaking change, it might be worth flagging the PR with the `api change` label.

ballista/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -95,18 +97,48 @@ impl ShuffleReaderExec {
             metrics: ExecutionPlanMetricsSet::new(),
             properties,
             work_dir: None, // to be updated at the executor side
+            client_pool: None,
         })
     }
 
     /// changes work dir where shuffle files are located
-    pub fn change_work_dir(&self, work_dir: String) -> Self {
+    pub fn with_work_dir(&self, work_dir: String) -> Self {
```
andygrove, line 105: Small thought on the rename from `change_work_dir`. Since this is `pub` on a public struct, custom `ExecutionEngine` impls in downstream projects may be calling it. Would it be worth keeping a `#[deprecated]` shim under the old name that forwards to `with_work_dir`? That way folks get a warning instead of a hard compile error.

ballista/core/src/client_pool.rs
```diff
@@ -0,0 +1,426 @@
...
+        // BallistaClient::try_new call below.
+        let maybe_idle = self
+            .inner
+            .idle
+            .get_mut(&key)
+            .and_then(|mut deque| deque.pop_back())
+            .map(|e| e.client);
+
+        let client = match maybe_idle {
+            Some(c) => {
```
andygrove: Question about cached entries. The pool keys on `(host, port)`, but `GrpcClientConfig` carries values like `io_retries_times` and `io_retry_wait_time_ms` that became per-session-config in #1577. If those settings change between task runs, a recycled idle client will still be using whatever was in effect when it was first created. Is that something we should worry about in practice, or do those values rarely vary at runtime?

ballista/core/src/client_pool.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+fn evict(idle: &IdleMap, timeout: Duration) {
+    let deadline = Instant::now()
+        .checked_sub(timeout)
+        .unwrap_or_else(Instant::now);
+
+    // Drain expired entries from the front of each deque (oldest = front).
+    idle.retain(|_, deque| {
+        while deque.front().is_some_and(|e| e.idle_since <= deadline) {
```
andygrove: Tiny boundary thing. The trait doc above (`evict_idle`) says 'sitting unused longer than the configured idle timeout', but `idle_since <= deadline` will also evict entries that are exactly at the timeout. Should this be a strict `<` to match the wording?

ballista/core/src/client_pool.rs
```diff
@@ -0,0 +1,426 @@
...
+//!
+//! `DefaultBallistaClientPool` maintains a `VecDeque`` of idle clients per
+//! `(host, port)` key backed by a `DashMap`. Callers `BallistaClientPool::acquire` a
+//! `PooledClient` guard; when the guard is dropped the underlying client is
+//! returned to the idle deque automatically.
+//!
+//! If the connection errored, call`PooledClient::discard` before dropping so
+//! the pool closes the channel rather than reusing it.
+//!
+//! A background tokio task evicts idle connections that have not been returned
```
andygrove: I think this sentence may be inverted. The eviction task only sees connections that have already been returned to the idle deque. So it removes ones that have been sitting idle (returned but not re-acquired) for longer than `idle_timeout`. Maybe reword to something like 'evicts connections that have sat idle longer than `idle_timeout`'?

ballista/executor/src/config.rs
```diff
@@ -151,6 +151,13 @@ pub struct Config {
         help = "Metric collection policy of this executor instance"
     )]
     pub metric_collection_policy: ExecutorMetricCollectionPolicy,
+    /// Number of second established client connection should be cached (0 means no cache)
+    #[arg(
+        long,
+        default_value_t = 0,
+        help = "Number of second established client connection should be cached (0 means no cache) "
```
andygrove: Tiny typo: `Number of second` should be `Number of seconds`. The same wording also appears in the rustdoc just above on line 154, and in `ballista/executor/src/executor_process.rs` around line 141. Worth fixing all three since the help string shows up in `--help` output.

ballista/core/src/client_pool.rs
```diff
@@ -0,0 +1,426 @@
...
+
+impl Debug for DefaultBallistaClientPool {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DefaultBallistaClientPool").finish()
+    }
+}
+
+impl DefaultBallistaClientPool {
+    /// Create a pool that evicts connections idle longer than `idle_timeout`,
+    pub fn with_eviction_thread(idle_timeout: Duration) -> Self {
```
andygrove: Looks like this doc got truncated. It ends on a comma with no predicate. Looks like maybe an in-progress copy from `new()` below? Could be filled in with something like 'using a background eviction thread' to round out the sentence.

ballista/core/Cargo.toml
```diff
@@ -49,6 +49,7 @@ async-trait = { workspace = true }
 aws-config = { version = "1.8.16", optional = true }
 aws-credential-types = { version = "1.2.0", optional = true }
 chrono = { version = "0.4", default-features = false }
+dashmap = { workspace = true }
```
andygrove: Since pooling is opt-in at runtime (off by default via `--connection-cache 0`), would it make sense to gate `dashmap` and the new `client_pool` module behind a cargo feature? That way `--no-default-features` builds stay lean and folks who never enable pooling do not pull in dashmap.

andygrove (commented):
Really like the direction here. Caching shuffle clients should make a big difference for clusters that hit a lot of executors. Left a few inline thoughts.

ballista/core/src/execution_plans/shuffle_reader.rs
milenkovicm: this method has been introduced in #1527 which has not been released yet, so nobody should be impacted

ballista/core/src/client_pool.rs
milenkovicm: yes, this is boundary case, we could probably use all configuration values as key, but at the moment it looks like an overkill. would suggest to keep it as it is and just document behaviour

ballista/core/Cargo.toml
milenkovicm: DefaultBallistaPool can be moved to executor, where dashmap is already used

ballista/executor/src/executor.rs
milenkovicm: its a minimal change, not many users use this api, I suggest to keep the change as it is, if you agree

milenkovicm:
@martin-g @danielhumanmod @sandugood  it would be great if you could review this change. thanks a lot

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,351 @@
...
+type IdleMap = DashMap<(String, u16, GrpcClientConfig), VecDeque<IdleEntry>>;
+
+struct Inner {
+    idle: IdleMap,
+    idle_timeout: Duration,
+}
+
+/// Default pool implementation.
+///
+/// Keeps a `VecDeque<BallistaClient>` per `(host, port, config)`. Idle clients are
```
sandugood:
```suggestion
/// Keeps a `VecDeque<IdleEntry>` per `(host, port, config)`, which is itself a specification of a BallistaClient and duration of its idle status.
```

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,351 @@
...
+    /// Create a pool that evicts connections idle longer than `idle_timeout`,
+    /// if `enable_eviction_thread` is enabled
+    pub fn new(idle_timeout: Duration, enable_eviction_thread: bool) -> Self {
+        let inner = Arc::new(Inner {
+            idle: DashMap::new(),
+            idle_timeout,
+        });
+
+        let weak: Weak<Inner> = Arc::downgrade(&inner);
+        // TODO: do we limit minimum interval here?
```
sandugood: Since we add this in the user-facing config:
If the user sets the `client_ttl` in the (1, 45) range it will be transformed into 15. Maybe add a warning when user sets this value in the config and it is in the (1, 45) range?

ballista/core/src/client.rs
```diff
@@ -114,6 +114,24 @@ impl BallistaClient {
         })
     }
 
+    /// creates a ballista client to be used for testing
+    /// it connects lazily and which can not really
+    /// be reconfigured.
+    pub fn new_for_test(host: &str, port: u16) -> Self {
```
martin-g, line 120: There is no `#[cfg(test)]`.
Does it need to be here ? It could be a helper method in `mod tests` too

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,351 @@
...
+            .get_mut(&key)
+            .and_then(|mut deque| deque.pop_back()) // acquire from back of the queue
+            .map(|e| e.client);
+
+        let client = match maybe_idle_client {
+            Some(client) => {
+                log::trace!(
+                    "client connection pool - returning cached connection - host:{host}, port:{port}"
+                );
+                client
```
martin-g, line 175: does it need to check whether it is still connected ?
if the remote party has been restarted in the meantime then the client will probably fail without reconnecting.

ballista/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -94,19 +96,49 @@ impl ShuffleReaderExec {
...
+            schema: self.schema.clone(),
+            partition: self.partition.clone(),
+            metrics: self.metrics.clone(),
+            properties: self.properties.clone(),
+            work_dir: self.work_dir.clone(),
+            client_pool: Some(client_pool),
+        }
+    }
+    /// creates new shuffle reader with client pool and work dir
+    pub fn with_client_pool_and_work_dir(
```
martin-g: Is this method really needed ?
The user could do it in two steps: `with_client_pool(client_pool).with_work_dir(work_dir)`
At the moment `fn with_client_pool()` is not used anywhere.

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,351 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Connection pool for `BallistaClient` instances.
+//!
+//! `DefaultBallistaClientPool` maintains a `VecDeque`` of idle clients per
```
martin-g:
```suggestion
//! `DefaultBallistaClientPool` maintains a `VecDeque` of idle clients per
```

ballista/core/src/client.rs
milenkovicm: It should but apparently it does not work across different crates, or I'm doing something wrong

ballista/executor/src/client_pool.rs
milenkovicm: My understanding tonic should handle reconnection automatically

ballista/core/src/execution_plans/shuffle_reader.rs
milenkovicm: You're right, I was trying to save few clones

ballista/executor/src/client_pool.rs
milenkovicm: maybe it would make sense to have warning for ttl < 15 but then there would be a bit of explaining needed.
i was thinking to add it to cli help, but this is implementation specific as well. 
I'd propose to keep it as it is and add it later if needed

andygrove:
Thanks @milenkovicm. I will try and start reviewing this one tonight.

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,352 @@
...
+// ---------------------------------------------------------------------------
+// DefaultBallistaClientPool
+// ---------------------------------------------------------------------------
+
+struct IdleEntry {
+    client: BallistaClient,
+    idle_since: Instant,
+}
+
+type IdleMap = DashMap<(String, u16, GrpcClientConfig), VecDeque<IdleEntry>>;
```
danielhumanmod, line 53: Based on my understanding, the current pool is TTL-based and only evicts clients once they have been idle longer than the configured TTL.

One of my two-cent thought: should we also consider a per-endpoint or global max idle limit, and evict/drop clients when that limit is reached? That would help prevent the pool from retaining too many idle clients after a burst of high-concurrency shuffle fetches.

Not a blocker tho, we can also consider adding that if we meet this kind of issue

milenkovicm: i agree, we can do that as a follow up

ballista/core/src/client_pool.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::sync::Arc;
+
+// ---------------------------------------------------------------------------
+// Trait
+// ---------------------------------------------------------------------------
+
+/// Manages a pool of reusable [BallistaClient] connections.
+#[async_trait]
+pub trait BallistaClientPool: Send + Sync + Debug {
+    /// Acquire an idle client for `(host, port)`, or create a new one if the
```
andygrove: doc comment says key is `(host, port)` but should be `(host, port, config, customize_endpoint)` ?

ballista/executor/src/client_pool.rs
```diff
@@ -0,0 +1,352 @@
...
+#[async_trait]
+impl BallistaClientPool for DefaultBallistaClientPool {
+    async fn acquire(
+        &self,
+        host: &str,
+        port: u16,
+        config: &GrpcClientConfig,
+        customize_endpoint: Option<Arc<BallistaConfigGrpcEndpoint>>,
+    ) -> Result<PooledClient> {
+        let key = (host.to_string(), port, config.clone());
```
andygrove, line 158: key is missing `customize_endpoint`?

ballista/executor/src/execution_engine.rs
```diff
@@ -98,8 +115,18 @@ impl ExecutionEngine for DefaultExecutionEngine {
         let plan = plan
             .transform(|p| {
                 if let Some(reader) = p.as_any().downcast_ref::<ShuffleReaderExec>() {
-                    let reader = Arc::new(reader.change_work_dir(work_dir.to_string()));
-                    Ok(Transformed::yes(reader))
+                    // if client_pool is not None we need to get clone of it and
+                    // invoke reader.with_client_pool_and_work_dir
```
andygrove: I do not see method `with_client_pool_and_work_dir` being called

ballista/core/src/lib.rs
```diff
@@ -33,6 +33,8 @@ pub fn print_version() {
 
 /// Client utilities for connecting to Ballista schedulers.
 pub mod client;
+/// Connection pool for reusing `BallistaClient`` instances across requests.
```
andygrove:
```suggestion
/// Connection pool for reusing `BallistaClient` instances across requests.
```

ballista/executor/src/client_pool.rs
milenkovicm: Customise endpoint is compile time, it should not change. I propose it ignore it for now

PR #1513, opened 2026-03-16
feat: jupyter notebook support

sandugood:
# Which issue does this PR close?

Refactoring of PR #1430 

 # Rationale for this change

Latest commit to the mentioned PR was in January. Refactoring felt like a nice thing.

# What changes are included in this PR?

1. Fix of both @line_magic and @cell_magic decorators. Combined them under the @line_cell_magic decorator
2. Nicely HTML-formatted output of cells. Using display() func and also printing in a loop where applicable

# Are there any user-facing changes?

No.

PR #1585, opened 2026-04-26
fix: route collect/show/to_pandas through Ballista cluster

### LINKED ISSUE #1582: collect() runs query in client not in cluster
**Describe the bug**
I am running in k8s and see that queries run entirely in the client pod. There is no activity in executors.

I can work around this with ` df._to_internal_df().collect()`.

Claude claims that this is a recent regression caused by the Jupyter PR.

```
 What commit 6091b525 (feat: jupyter notebook support, Apr 9) broke:                                                                                                                                           
                                                                                                                                                                                                                
  Before that commit, DistributedDataFrame used a metaclass (RedefiningDataFrameMeta) that automatically wrapped execution methods to route through _to_internal_df():                                          
                                                                                                                                                                                                                
  OVERRIDDEN_EXECUTION_METHODS = [                                                                                                                                                                              
      "show", "count", "collect", "collect_partitioned",                                                                                                                                                        
      "write_json", "to_arrow_table", "to_pandas", ...                                                                                                                                                          
  ]                                                                                                                                                                                                             
                                                                                                                                                                                                                
  The Jupyter PR changed the metaclass to plain metaclass=type, silently removing all that wrapping. So collect(), show(), to_pandas() etc. now run locally. Only write_csv() and write_parquet() were manually 
  re-added as overrides.                                                                                                                                                                                        
                                                                                                                                                                                                                
  The tests in test_context.py still call df.collect() directly — they were written expecting the old behavior and presumably now pass only because they're running against a local/standalone cluster where    
  "local" and "cluster" are the same process.
```

**To Reproduce**

**Expected behavior**

**Additional context**

andygrove:
## Summary

`DistributedDataFrame.collect()`, `show()`, `count()`, `to_pandas()` and other execution methods were running locally in the client process instead of on the Ballista cluster.

**Root cause:** PR #1513 (jupyter notebook support) changed `DistributedDataFrame` from `metaclass=RedefiningDataFrameMeta` to `metaclass=type`. `RedefiningDataFrameMeta` was responsible for wrapping execution methods to route through `_to_internal_df()` before executing. With plain `metaclass=type`, those methods were inherited directly from DataFusion's `DataFrame` and ran locally. Only `write_csv` and `write_parquet` were manually re-added as overrides, so writes worked but reads did not.

**Fix:**
- Add `EXECUTION_METHODS` list to `RedefiningDataFrameMeta` covering `collect`, `collect_partitioned`, `show`, `count`, `to_arrow_table`, `to_pandas`, `to_polars`, `write_json`
- In `__new__`, wrap each of those methods to call `_to_internal_df()` first — skipping any method already explicitly overridden on the subclass (e.g. `write_parquet`)
- Switch `DistributedDataFrame` back to `metaclass=RedefiningDataFrameMeta`

Fixes #1582

## Test plan

- [ ] `df.collect()` executes on the cluster (verify executor logs show activity)
- [ ] `df.show()` executes on the cluster
- [ ] `df.to_pandas()` executes on the cluster
- [ ] `df.write_parquet()` still works (explicit override not clobbered by metaclass)
- [ ] Existing Python tests pass

### DIFF
--- a/python/python/ballista/extension.py
+++ b/python/python/ballista/extension.py
@@ -35,6 +35,18 @@ import pathlib
 # class used to redefine DataFrame object
 # intercepting execution methods and methods
 # which returns `DataFrame`
+EXECUTION_METHODS = [
+    "collect",
+    "collect_partitioned",
+    "show",
+    "count",
+    "to_arrow_table",
+    "to_pandas",
+    "to_polars",
+    "write_json",
+]
+
+
 class RedefiningDataFrameMeta(type):
     def __new__(cls, name, bases, attrs):
         # wrapper function intercept all execution functions
@@ -74,6 +86,13 @@ class RedefiningDataFrameMeta(type):
                 #
                 attrs[base_name] = __wrap_dataframe_result(base_value)
 
+        # Wrap execution methods to route through _to_internal_df() so they
+        # execute on the Ballista cluster rather than locally. Skip any method
+        # already explicitly defined on the subclass.
+        for func in EXECUTION_METHODS:
+            if func not in attrs:
+                attrs[func] = __wrap_dataframe_execution(func)
+
         return super().__new__(cls, name, bases, attrs)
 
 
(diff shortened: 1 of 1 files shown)


milenkovicm:
My bad, should have spotted that issue

milenkovicm (approved):

milenkovicm:
Please update ballista python to df and ballista 52 if you can

PR #1589, opened 2026-04-26
fix: warn and document scheduler --external-host default in cluster deploys

### LINKED ISSUE #1587: Executors try to connect with local scheduler instead of remote scheduler
**Describe the bug**

I am running queries in a k8s cluster. Exectuors are receiving tasks to run so can connect to the scheduler but then I see errors such as:

```
executor 2026-04-26T21:20:33.377969Z ERROR tokio-rt-worker ThreadId(03) ballista_executor::executor_server: Fail to connect to scheduler localhost:50050 due to TonicError(tonic::transport::Error(Transport │
```

There is no scheduler on localhost within the executor pod, so this is confusing.

**To Reproduce**

**Expected behavior**

**Additional context**

andygrove:

Closes #1587.

# Rationale for this change

When the scheduler launches tasks it embeds its own advertised address in `LaunchTaskParams.scheduler_id`, and executors use that string to dial back for task status and heartbeats (`executor_server.rs::get_scheduler_client` literally does `format!(\"http://{scheduler_id}\")`). The advertised address is built from `SchedulerConfig::scheduler_name()` = `format!(\"{external_host}:{bind_port}\")`, and `external_host` defaults to `\"localhost\"` with no env-var binding.

The Kubernetes deployment example in the user guide and the in-tree `docker-compose.yml` both omit `--external-host`, so any cluster deployed from those templates inherits the default. The executor's outgoing connection works because it is configured separately on the executor side via `--scheduler-host`, but every status report and heartbeat tries to dial `localhost:50050` inside the executor's own pod and fails with `Fail to connect to scheduler localhost:50050`. The failure mode is undocumented today.

- **Kubernetes deployment guide**: scheduler args now include `--external-host=ballista-scheduler`, with a paragraph explaining why and what the symptom looks like if it is omitted.
- **`docker-compose.yml`**: scheduler `command:` now sets `--external-host ballista-scheduler` to match the Compose service name.
- **Docker Compose deployment guide**: short note added describing why the scheduler advertises its Compose service name to executors.
- **`ballista/scheduler/src/scheduler_process.rs`**: emit a `WARN` at scheduler startup when `bind_host` is non-loopback but `external_host` is still the default `\"localhost\"`. This fires only on misconfigured cluster deploys (single-machine runs with the all-defaults `bind_host=127.0.0.1` are unaffected) and gives operators a clear diagnostic instead of waiting for the first task-status callback to fail.

No API changes. Operators of misconfigured clusters will see a new `WARN` log at scheduler startup; deployments using the updated example manifests will work without status-callback failures.

### DIFF
--- a/ballista/scheduler/src/scheduler_process.rs
+++ b/ballista/scheduler/src/scheduler_process.rs
@@ -40,7 +40,7 @@ use datafusion_proto::logical_plan::AsLogicalPlan;
 use datafusion_proto::physical_plan::AsExecutionPlan;
 use datafusion_proto::protobuf::{LogicalPlanNode, PhysicalPlanNode};
 use http::StatusCode;
-use log::info;
+use log::{info, warn};
 use std::{net::SocketAddr, sync::Arc};
 use tonic::service::RoutesBuilder;
 /// Creates as initialized scheduler service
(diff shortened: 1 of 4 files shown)


milenkovicm:
@andygrove is this PR still needed ?

PR #1590, opened 2026-04-26
chore: update python deps to ballista and datafusion 52

andygrove:

Closes #.

# Rationale for this change

The python client crate was last updated to ballista/datafusion 51 in #1335. Ballista 52 and datafusion 52 are now published on crates.io, so bring the python client up to that release.

- Bump `pyballista` to 52.0.0 and the ballista/datafusion crate dependencies (`ballista`, `ballista-core`, `ballista-executor`, `ballista-scheduler`, `datafusion`, `datafusion-proto`, `datafusion-python`) from 51 to 52.
- Bump the `datafusion` pip requirement to 52 in `pyproject.toml` and `requirements.txt`.
- The `pyarrow` feature was removed from the `datafusion` crate in 52, so drop it from the feature list. The `?` operator no longer converts `DataFusionError` to `PyErr` automatically as a result, so call `datafusion_python::errors::from_datafusion_error` explicitly via `map_err` in `python/src/lib.rs`.
- Refresh `python/Cargo.lock` and `python/uv.lock`.

No public Rust API changes; the python client now pulls in datafusion/datafusion-python 52 instead of 51.

### DIFF
--- a/python/Cargo.lock
+++ b/python/Cargo.lock
@@ -167,15 +167,16 @@ checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "apache-avro"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3a033b4ced7c585199fb78ef50fca7fe2f444369ec48080c5fd072efa1a03cc7"
+checksum = "36fa98bc79671c7981272d91a8753a928ff6a1cd8e4f20a44c45bd5d313840bf"
 dependencies = [
  "bigdecimal",
  "bon",
- "bzip2 0.6.1",
+ "bzip2",
  "crc32fast",
- "digest",
+ "digest 0.10.7",
+ "liblzma",
  "log",
  "miniz_oxide",
  "num-bigint",
@@ -190,7 +191,6 @@ dependencies = [
  "strum_macros",
  "thiserror",
  "uuid",
- "xz2",
  "zstd",
 ]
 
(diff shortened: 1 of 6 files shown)


milenkovicm:
If i recall correctly, there was few change of the interfaces/methods which may impact `DistributedDataFrame` and others

andygrove:
> If i recall correctly, there was few change of the interfaces/methods which may impact `DistributedDataFrame` and others

Everything is building in CI. Are there additional steps needed in this PR?

milenkovicm:
Tests are not great, unfortunately 

If i remember correctly some of datafusion-python interfaces changed between versions and `DistributedDataFrame` may not reflect the change (this is biggest issue with this approach, keeping an eye on changes).

andygrove:
> Tests are not great, unfortunately
> 
> If i remember correctly some of datafusion-python interfaces changed between versions and `DistributedDataFrame` may not reflect the change (this is biggest issue with this approach, keeping an eye on changes).

I see. Ok, I will pivot and start looking at the current testing approach and see what can be improved.

milenkovicm:
looks like `DataFrameWriteOptions` has been added between version so many of the writer methods are missing

```python
    def write_csv(
        self,
        path: str | pathlib.Path,
        with_header: bool = False,
        write_options: DataFrameWriteOptions | None = None, # <- added
    ) -> None:
```

```python
def write_parquet_with_options(
    path: str | Path,
    options: ParquetWriterOptions,
    write_options: DataFrameWriteOptions | None = None # <-  added
) -> None
```
I remember now, that was the missing link

milenkovicm:
we need to make mapping between `DataFrameWriteOptions` and `DataFrameWriteOptionsInternal` similar to

```python
        options_internal = ParquetWriterOptionsInternal(
            options.data_pagesize_limit,
            options.write_batch_size,
            options.writer_version,
            options.skip_arrow_metadata,
            options.compression,
            options.dictionary_enabled,
            options.dictionary_page_size_limit,
            options.statistics_enabled,
            options.max_row_group_size,
           # trimmed ...
        )
```
`DataFrameWriteOptions` is defined as 

```python
class DataFrameWriteOptions:
    """Writer options for DataFrame.

    There is no guarantee the table provider supports all writer options.
    See the individual implementation and documentation for details.
    """

    def __init__(
        self,
        insert_operation: InsertOp | None = None,
        single_file_output: bool = False,
        partition_by: str | Sequence[str] | None = None,
        sort_by: Expr | SortExpr | Sequence[Expr] | Sequence[SortExpr] | None = None,
    ) -> None:
 # trimmed ...
```

the only problem might be `sort_by: Expr | SortExpr | Sequence[Expr] | Sequence[SortExpr] | None = None,`

milenkovicm (approved):
I believe `df.write_` is broken, lets merge this, i'll follow up with the fix in next few days

PR #1592, opened 2026-04-26
fix: propagate session config from Python client to Ballista cluster

### LINKED ISSUE #1591: Cannot set target partitions from Python client
**Describe the bug**
I cannot find a way to set the number of partitions from the Python client. My cluster is configured with 32 executors with 8 cores each, so I want to process 256 partitions in parallel, but it looks like Ballista is using 8 (likely the number of cores in the pod where the client is running).

**To Reproduce**

**Expected behavior**

**Additional context**

andygrove:
Closes #1591.

# Rationale

The Python `BallistaSessionContext` had no way to override DataFusion / Ballista session settings (e.g. `target_partitions`). Worse, the Rust entry point built the remote session from `SessionStateBuilder::new_with_default_features()`, whose `target_partitions` defaults to the client pod's CPU count and was then shipped wholesale to the scheduler — overriding the scheduler's own defaults.

# Changes

- `python/src/lib.rs`: `create_ballista_data_frame` now starts the remote session from `SessionConfig::new_with_ballista()` and accepts an optional dict of overrides applied via `update_from_key_value_pair_mut`.
- `python/python/ballista/extension.py`: `BallistaSessionContext` accepts `cluster_config: dict[str, str]`. The dict is applied both to the local `SessionConfig` (needed for settings consulted during table registration / planning, e.g. `listing_table_factory_infer_partitions`) and forwarded through every `DistributedDataFrame._to_internal_df()` to the remote session.

# User-facing change

```python
ctx = BallistaSessionContext(
    "df://localhost:50050",
    cluster_config={"datafusion.execution.target_partitions": "256"},
)
```

New optional kwarg; existing call sites are unaffected.

### DIFF
--- a/python/python/ballista/extension.py
+++ b/python/python/ballista/extension.py
@@ -19,6 +19,7 @@ from datafusion import SessionContext, DataFrame, ParquetWriterOptions
 from datafusion.dataframe import Compression
 
 from typing import (
+    Dict,
     List,
     Union,
     Optional,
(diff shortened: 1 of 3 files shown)


milenkovicm (approved):
this was a know issue, its good to have a way to configure it now. 
i guess outstanding issue now is to change config use sql `SET ballista ...=value` but we can take that later

PR #1600, opened 2026-04-27
feat: Add standalone shuffle writer benchmark that shuffles real Parquet input

andygrove:
# Which issue does this PR close?

N/A

# Rationale for this change

This PR replaces `shuffle_bench` with a Parquet-driven runner modeled after [DataFusion Comet's `shuffle_bench`](https://github.com/apache/datafusion-comet/blob/main/native/shuffle/src/bin/shuffle_bench.rs). 
The CLI surface is intentionally kept close to Comet's so the same `--input` / `--partitions` / `--hash-columns` / `--memory-limit` arguments work against both binaries.

# What changes are included in this PR?

New benchmark.

# How to reproduce

Example:

```sh
cargo build --release --bin shuffle_bench

# Sort writer
./target/release/shuffle_bench \
  --input /path/to/tpch/sf100/lineitem \
  --writer sort --partitioning hash \
  --partitions 200 --hash-columns 0,3 \
  --memory-limit 8589934592 \
  --limit 1000000 --warmup 1 --iterations 3

# Hash writer
./target/release/shuffle_bench \
  --input /path/to/tpch/sf100/lineitem \
  --writer hash --partitioning hash \
  --partitions 200 --hash-columns 0,3 \
  --memory-limit 8589934592 \
  --limit 1000000 --warmup 1 --iterations 3
```

# Are there any user-facing changes?

The CLI surface of `shuffle_bench` changes (different flags, switched from `structopt` to `clap`). This binary is a developer profiling tool, not a public API. Anyone scripting against the previous synthetic flags (`--rows`, `--input-partitions`, `--hash-only`, `--sort-only`, etc.) will need to migrate to the new `--input` / `--writer` form.

No library or wire-format changes.

PR #1595, opened 2026-04-27
docs: document hash-based and sort-based shuffle implementations

andygrove:
# Which issue does this PR close?

Closes #.

# Rationale for this change

The user guide does not currently describe how Ballista's shuffle works or that two distinct writer implementations exist. The sort-based shuffle writer and its four `ballista.shuffle.sort_based.*` session-level config keys are only discoverable by reading the source. Anyone trying to tune a shuffle-heavy workload — particularly to reduce the `N × M` file count produced by the default hash writer — has no documentation to point them at the alternative.

# What changes are included in this PR?

- Add a "Shuffle Implementation" section to the tuning guide that explains the default hash-based writer (one file per `(input_partition, output_partition)` pair, no buffering) and the opt-in sort-based writer (per-output-partition in-memory buffers with spilling, consolidated data + index file per input partition), and lists the four `ballista.shuffle.sort_based.*` keys that tune it.
- Add a "Shuffle Settings" table to `configs.md` covering the same sort-shuffle keys plus the existing `ballista.shuffle.max_concurrent_read_requests`, `ballista.shuffle.force_remote_read`, and `ballista.shuffle.remote_read_prefer_flight` keys, which were also previously undocumented in the user guide.

# Are there any user-facing changes?

Documentation only. No code, behavior, or public API changes.

PR #1604, opened 2026-04-28
feat: default to sort-based shuffle writer [WIP]

andygrove:

N/A

# Rationale for this change

The sort-based shuffle writer (added recently and documented in #1595) is ~50x faster than the current default hash-based shuffle for real-world TPC workloads, as shown in https://github.com/apache/datafusion-ballista/pull/1600.

- Flip the default of `ballista.shuffle.sort_based.enabled` from `false` to `true` in `ballista/core/src/config.rs`.
- Update `docs/source/user-guide/configs.md` to show the new default and a note about falling back to the hash writer.
- Restructure the "Shuffle Implementation" section of `docs/source/user-guide/tuning-guide.md` so sort-based is presented as the default and hash-based as the opt-in fallback.

Yes. New sessions will use the sort-based shuffle writer by default. Users who want the previous behavior can set `ballista.shuffle.sort_based.enabled=false` per session.

### DIFF
--- a/ballista/client/tests/context_checks.rs
+++ b/ballista/client/tests/context_checks.rs
@@ -847,6 +847,13 @@ mod supported {
             .show()
             .await?;
 
+        // The block-transport remote reader does not support sort-based
+        // shuffle output, so this test exercises the hash-based writer.
+        ctx.sql("SET ballista.shuffle.sort_based.enabled = false")
+            .await?
+            .show()
+            .await?;
+
         let result = ctx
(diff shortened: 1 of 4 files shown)


andygrove:
I need to learn more about block vs flight before we make this change

milenkovicm:
U suggest going back to flight if it's required for sort shuttle, which obviously has benefits

andygrove:
I closed this and will make sure that we can support the more efficient
block transport for sort based shuffle - this is what we do in comet
already

I learned a lot from from the experiments today and will approach this a
bit more methodically in the next few days

Issue #1608, opened 2026-04-28
Python release wheels: macOS x86_64 wheel is missing and arm64 wheel is built twice


andygrove:
The Python release builds in `.github/workflows/build.yml` produce two macOS arm64 wheels and no macOS x86_64 wheel.

## Current behavior

Two jobs both target `macos-latest`, which has resolved to arm64 since GitHub's runner change in 2024:

- `build-python-mac-win` (matrix `os: [macos-latest, windows-latest]`) — uploads as `dist-macos-latest`
- `build-macos-x86_64` (job name "Mac x86_64") — uploads as `dist-macos-aarch64` despite the job name

Both jobs build the same `*_arm64.whl` artifact. Nothing in the workflow currently produces a wheel with a `*_x86_64.whl` macOS tag.

## Impact

Users installing `pip install ballista` on an Intel Mac fall back to the source distribution and have to compile the Rust extension locally (which requires `protoc` and a Rust toolchain). They likely cannot install a prebuilt wheel.

## Expected behavior

The release artifact set should contain exactly one wheel per supported platform:

- `manylinux_2_17_x86_64.manylinux2014_x86_64`
- `manylinux_2_17_aarch64.manylinux2014_aarch64`
- `macosx_*_arm64`
- `macosx_*_x86_64`
- `win_amd64`
- plus the `tar.gz` sdist

## Suggested fix

Either pin one of the macOS jobs to `macos-13` (the last x86_64 runner) so it actually produces the x86_64 wheel, or use cross-compilation via `--target x86_64-apple-darwin` from an arm64 runner. The job name `build-macos-x86_64` suggests x86_64 was the original intent for that job.

## References

- `.github/workflows/build.yml` (`build-python-mac-win`, `build-macos-x86_64`)
- GitHub runner image change: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

PR #1610, opened 2026-04-28
docs(release): document PyPI publish process for python client

andygrove:

Closes #.

# Rationale for this change

The Python client `ballista` is published to [PyPI](https://pypi.org/project/ballista/) as part of each release, but the procedure was undocumented. The bottom of `.github/workflows/build.yml` has a commented-out publish job with the note "PyPI publish needs to be done manually for now after release passed the vote", and `dev/release/README.md` documents the crates.io publish step but says nothing about PyPI. This PR fills that gap so any committee member added as a maintainer of the PyPI project can run the publish confidently after a successful release vote.

A new `### Publish Python Wheels to PyPI` section in `dev/release/README.md`, inserted between the existing "Publish on Crates.io" and "Publish Docker Images" sections. The new section covers:

- Prerequisites (PyPI / TestPyPI accounts, maintainer access, `~/.pypirc`, `twine`).
- Downloading the voted-on wheels from the RC's CI run via `dev/release/download-python-wheels.py`.
- Validating the artifacts with `twine check`.
- A TestPyPI dry-run with a fresh-venv smoke test, since PyPI uploads are immutable.
- The actual `twine upload` to PyPI.
- Verification.
- Recovery for the realistic failure modes (`twine check` fails, smoke test fails, partial upload).

One row is also added to the Post-Release task table near the top of the file.

The new section calls out a known CI bug — the macOS arm64 wheel is built twice and no macOS x86_64 wheel is produced — and links to #1608 which tracks the fix.

No code changes. Documentation only.

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -332,6 +332,14 @@ download the same wheels that were voted on from the RC's CI run. Retagging the
 RC commit does not trigger a fresh build, so the RC artifacts remain the
 canonical source.
 
+> **Artifact retention warning:** GitHub Actions artifacts default to 90-day
+> retention. If the elapsed time from cutting the RC to publishing PyPI wheels
+> exceeds that window, the wheels will have been deleted and are unrecoverable
+> — you cannot publish the voted-on artifacts and must cut a new RC and revote.
+> Plan the vote and the post-vote publish so the publish step happens
+> comfortably inside the 90-day window. Check the run's `expires_at` on
+> `https://github.com/apache/datafusion-ballista/actions` if in doubt.
+
 ```bash
 export GH_TOKEN=...     # GitHub PAT with read access to actions
 mkdir ballista-pypi-<version> && cd ballista-pypi-<version>
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -63,6 +63,7 @@ Although some tasks can only be performed by a PMC member, many tasks can be per
 | Task                                                         | Role Required |
 |--------------------------------------------------------------| ------------- |
 | Create PR against datafusion-site with updated documentation | None          |
+| Publish Python wheels to PyPI                                | PMC           |
 
 ## Detailed Guide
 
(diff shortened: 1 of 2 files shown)


dev/release/README.md
```diff
@@ -281,6 +282,158 @@ dot -Tsvg dev/release/crate-deps.dot > dev/release/crate-deps.svg
...
+### Publish Python Wheels to PyPI
+
+Only approved releases of the tarball should be published to PyPI, in order to
+conform to Apache Software Foundation governance standards. The Python wheels
+that get uploaded must be the same artifacts that the community voted on — they
+are downloaded from the release candidate's CI run, not rebuilt.
+
+#### Prerequisites
+
+A DataFusion committer can publish the [`ballista` package on
```
martin-g:
```suggestion
A DataFusion PMC member can publish the [`ballista` package on
```

dev/release/README.md
```diff
@@ -281,6 +282,158 @@ dot -Tsvg dev/release/crate-deps.dot > dev/release/crate-deps.svg
...
+  [testpypi]
+  repository = https://test.pypi.org/legacy/
+  username = __token__
+  password = pypi-...
+  ```
+
+- Install `twine`:
+
+  ```bash
+  pip install twine
```
martin-g: https://github.com/apache/datafusion-ballista/blob/68dfeddeeedec3382cfcae0a4139af7005dc3eac/dev/release/download-python-wheels.py#L23 also says that `requests` is needed

dev/release/README.md
```diff
@@ -281,6 +282,158 @@ dot -Tsvg dev/release/crate-deps.dot > dev/release/crate-deps.svg
...
+- `ballista-<version>-cp310-abi3-manylinux_2_17_aarch64.manylinux2014_aarch64.whl`
+- `ballista-<version>-cp310-abi3-macosx_*_arm64.whl`
+- `ballista-<version>-cp310-abi3-win_amd64.whl`
+- `ballista-<version>.tar.gz` (sdist)
+
+> **Known CI caveat:** the merged artifact currently contains the macOS arm64
+> wheel twice (jobs `build-python-mac-win`'s macOS leg and `build-macos-x86_64`
+> both run on `macos-latest`, which is now arm64) and **no** macOS x86_64 wheel.
+> Keep one copy of the arm64 wheel and delete any duplicate before upload.
+> Tracked in [#1608](https://github.com/apache/datafusion-ballista/issues/1608).
```
martin-g: Not sure this note is needed here.
The issue will be resolved soon (before next release).

dev/release/README.md
```diff
@@ -281,6 +282,158 @@ dot -Tsvg dev/release/crate-deps.dot > dev/release/crate-deps.svg
...
+#### Verify
+
+Confirm the new version appears at
+`https://pypi.org/project/ballista/<version>/`. Then in another fresh
+virtual environment:
+
+```bash
+python -m venv /tmp/ballista-pypi-verify
+source /tmp/ballista-pypi-verify/bin/activate
+pip install ballista==<version>
```
martin-g: Idea: Instead of using `<version>` I'd suggest to use `${BALLISTA_VERSION}` that is defined earlier. This way the release manager could just copy/paste/run the code without any manual edits.

dev/release/README.md
```diff
@@ -281,6 +282,166 @@ dot -Tsvg dev/release/crate-deps.dot > dev/release/crate-deps.svg
...
+
+  [pypi]
+  username = __token__
+  password = pypi-...
+
+  [testpypi]
+  repository = https://test.pypi.org/legacy/
+  username = __token__
+  password = pypi-...
+  ```
```
martin-g, line 320: - Secure ~/.pypirc
```bash
chmod 600 ~/.pypirc
```

dev/release/download-python-wheels.py
martin-g, line 93:
```suggestion
        download_url = artifact["archive_download_url"]
        break
```
The newest artifact is first. Like the workflow runs above

andygrove:
Thanks for feedback @martin-g - the PR is draft until I have actually tested this, so it is likely to change

dev/release/README.md
andygrove: Added, the install step now does `pip install twine requests`.

dev/release/README.md
andygrove: You were right that #1608 is closed (by #1612 dropping Intel Mac), so the specific arm64-duplicate note is gone. While testing today though I hit a different problem with the same `merge-build-artifacts` job: the merged `dist` artifact was missing the manylinux aarch64 wheel and the sdist, even though the per-platform jobs all succeeded. I replaced the note with a broader caveat that calls out checking every expected file and falling back to the per-platform `dist-*` artifacts if any are missing. Also added a side note that `build-sdist` currently has no upload-artifact step at all, so the sdist may be absent regardless.

dev/release/README.md
andygrove: Done. Introduced `BALLISTA_VERSION` and `BALLISTA_RC_NUM` at the top of the section and replaced every `<version>` in the PyPI publish section with `${BALLISTA_VERSION}`. Pre-existing `<version>` placeholders in earlier sections of this doc are left as-is so the convention there stays consistent.

dev/release/README.md
andygrove: Added as a step right after the `~/.pypirc` configuration.

andygrove:
Thanks for feedback so far @martin-g. I have addressed it and also made updated based on learnings so far from trying to follow the process.

andygrove:
I'm going to keep iterating in main as I work through this process

PR #1614, opened 2026-04-28
test(python): add datafusion-python compatibility tests

andygrove:

Closes #.

# Rationale for this change

Ballista's Python bindings extend `datafusion-python` heavily through subclassing and metaclass introspection (see `python/python/ballista/extension.py`):

- `RedefiningDataFrameMeta` walks the parent `DataFrame.__dict__` and re-wraps every method whose return annotation is the literal string `"DataFrame"` so it returns `DistributedDataFrame` instead.
- `RedefiningSessionContextMeta` does the same for `SessionContext`.
- A hardcoded `EXECUTION_METHODS = ["collect", "collect_partitioned", "show", "count", "to_arrow_table", "to_pandas", "to_polars", "write_json"]` is wrapped to route execution through the Ballista cluster.
- `DistributedDataFrame.write_csv`, `write_parquet`, and `write_parquet_with_options` are explicitly defined and call into `_internal_ballista` Rust bindings, bypassing the metaclass.

If a future `datafusion-python` release changes annotation style (e.g. switches from forward-reference strings to real class objects), renames methods, or alters signatures, the wrapping silently stops happening or breaks at runtime. Today only `collect()` was exercised under Ballista in `test_context.py`. The other seven `EXECUTION_METHODS` plus all three explicit write methods were entirely uncovered.

While adding the new tests, two real signature-drift bugs surfaced in `extension.py`: `write_csv` and `write_parquet_with_options` were not passing the `write_options` argument that the underlying PyO3-bound `DataFrame` requires, so any caller of `ctx.sql(...).write_csv(...)` got a `TypeError`. These are also fixed in this PR.

**`python/python/ballista/extension.py` bug fixes** — pass `None` for `write_options` in the underlying `DataFrame.write_csv` and `DataFrame.write_parquet_with_options` calls, matching the existing pattern used elsewhere.

**New file `python/python/tests/test_datafusion_compat.py`** with 14 tests in four groups:

1. **Metaclass smoke tests (3)** — fail loudly if introspection no longer matches:
   - `test_distributed_dataframe_wraps_dataframe_returning_methods` — confirms representative `DataFrame` methods (`select`, `filter`, `with_column`, `aggregate`) carry the string `"DataFrame"` return annotation and are re-wrapped on `DistributedDataFrame`.
   - `test_ballista_session_context_wraps_dataframe_returning_methods` — same check for `sql` / `read_csv` / `read_parquet` on `BallistaSessionContext`.
   - `test_execution_methods_are_present_on_dataframe` — every name in `EXECUTION_METHODS` still exists on `datafusion.DataFrame`.

2. **Per-`EXECUTION_METHODS` round-trip (8)** — one test per name in `EXECUTION_METHODS` (`collect`, `collect_partitioned`, `show`, `count`, `to_arrow_table`, `to_pandas`, `to_polars`, `write_json`). Builds a small `DistributedDataFrame` and calls each, asserting return shape and content.

3. **Write-method round-trip (3)** — covers the explicit, non-metaclass write methods on `DistributedDataFrame`. Each writes to `tmp_path` and reads back with `pyarrow` to verify row count and column values:
   - `test_write_csv_round_trip`
   - `test_write_parquet_round_trip`
   - `test_write_parquet_with_options_round_trip` — uses non-default `ParquetWriterOptions` (snappy compression, custom batch / row-group sizes, `statistics_enabled='chunk'`) so the ~20 attributes shovelled through `extension.py:173-194` are actually exercised. Asserts the compression attribute propagated to the written file.

**Dev dependency additions** — `pandas>=2.0.0` and `polars>=1.0.0` in `[dependency-groups].dev` so the `to_pandas` / `to_polars` tests run unconditionally in CI rather than skipping when those libraries are absent. `uv.lock` is regenerated accordingly.

Yes — `BallistaSessionContext.sql(...).write_csv(...)` and `.write_parquet_with_options(...)` no longer raise `TypeError` from a missing `write_options` argument. No API shape changes.

### DIFF
--- a/python/python/ballista/extension.py
+++ b/python/python/ballista/extension.py
@@ -163,7 +163,9 @@ class DistributedDataFrame(DataFrame, metaclass=RedefiningDataFrameMeta):
 
     def write_csv(self, path, with_header=False):
         df = self._to_internal_df()
-        df.write_csv(path, with_header)
+        # The PyO3-bound DataFrame.write_csv requires write_options to be
+        # passed even though its Python signature shows a None default.
+        df.write_csv(path, with_header, None)
 
     def write_parquet_with_options(
         self,
@@ -205,17 +207,16 @@ class DistributedDataFrame(DataFrame, metaclass=RedefiningDataFrameMeta):
                 bloom_filter_ndv=opts.bloom_filter_ndv,
             )
 
-        # raw_write_options = (
-        #     write_options._raw_write_options if write_options is not None else None
-        # )
-
         df = self._to_internal_df()
 
+        # The PyO3-bound DataFrame.write_parquet_with_options requires
+        # write_options to be passed even though its Python signature shows
+        # a None default.
         df.write_parquet_with_options(
             str(path),
             options_internal,
             column_specific_options_internal,
-            # raw_write_options,
+            None,
         )
 
     def write_parquet(
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -72,6 +72,8 @@ locked = true
 [dependency-groups]
 dev = [
     "maturin>=1.8.1",
+    "pandas>=2.0.0",
+    "polars>=1.0.0",
     "pytest>=8.4.1",
     "pytest-asyncio>=1.1.0",
     "ruff>=0.9.1",
--- a/python/python/tests/test_datafusion_compat.py
+++ b/python/python/tests/test_datafusion_compat.py
@@ -33,8 +33,10 @@ that quietly fall back to local execution.
 """
 
 import pyarrow as pa
+import pyarrow.csv as pa_csv
+import pyarrow.parquet as pa_parquet
 import pytest
-from datafusion import DataFrame, SessionContext
+from datafusion import DataFrame, ParquetWriterOptions, SessionContext
 
 from ballista import BallistaSessionContext, setup_test_cluster
 from ballista.extension import EXECUTION_METHODS, DistributedDataFrame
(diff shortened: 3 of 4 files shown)


milenkovicm:
df.write_* may have been changed a lot if i remember

andygrove:
> df.write_* may have been changed a lot if i remember

Thanks, I added tests for all the write methods and fixed one bug that was discovered 🎉

milenkovicm:
Should we merge this with #1590?

python/python/ballista/extension.py
```diff
@@ -163,7 +163,9 @@ def _to_internal_df(self):
 
     def write_csv(self, path, with_header=False):
         df = self._to_internal_df()
-        df.write_csv(path, with_header)
+        # The PyO3-bound DataFrame.write_csv requires write_options to be
+        # passed even though its Python signature shows a None default.
+        df.write_csv(path, with_header, None)
```
milenkovicm, line 168: as with #1590 we're missing `write_options: DataFrameWriteOptions`

andygrove:
> Should we merge this with #1590?

My plan was to get tests passing on main branch first, then rebase the DF 52 upgrade and make sure there were no regressions

milenkovicm:
makes sense. 

we could add `write_options: DataFrameWriteOptions` to signatures but internally ignore it for now. i left few comments  in #1590

python/python/ballista/extension.py
andygrove: I looked into this and this looks complex and beyond my current knowledge of how these things work 😞

milenkovicm:
lets merge both of them, i can follow up to fix them over weekend

milenkovicm:
closing as open for too long without progress, please reopen if still needed and updated (it might not be once we update to dfpy v.55)

PR #1615, opened 2026-04-28
feat(sort-shuffle): byte-copy spill files and enable block-IO transport

andygrove:

Closes #.

# Rationale for this change

Sort-shuffle finalize previously decoded every spilled batch and re-emitted it through an IPC `FileWriter`, paying decompress + Arrow allocation + recompress for every spilled byte. This PR replaces that round-trip with a `std::io::copy` of the spill file straight into the consolidated output. On Linux this engages `copy_file_range` / `sendfile`, so spilled bytes never re-enter user space.

The same byte-offset index also lifts the restriction that sort-shuffle remote reads must use Arrow Flight: the executor can now serve sort-shuffle partitions through `IO_BLOCK_TRANSPORT` (the existing block-IO action), bypassing decode/re-encode on the server.

The on-disk format for sort-shuffle output changes:

- **Data file**: was a single IPC File with a footer of batch-block offsets. Now it is a leading schema-header IPC stream followed by per-partition byte ranges, each holding zero or more concatenated self-contained IPC streams.
- **Index file**: was little-endian i64 cumulative batch indices (despite the docstring already promising byte offsets). Now it stores actual little-endian i64 byte offsets, matching what the docstring always claimed.
- **Reader**: `stream_sort_shuffle_partition` recovers the schema from the leading header stream and uses a new bounded multi-stream reader that crosses concatenated stream EOS markers within a partition's byte range.
- **Block-IO transport**: `do_action`'s `IO_BLOCK_TRANSPORT` arm now serves sort-shuffle by streaming `[schema-header bytes ++ partition byte range]`. The client's `BlockDataStream` resets its `StreamDecoder` on EOS so it walks the concatenated payload.

Hash-based shuffle is intentionally untouched. Public API of `is_sort_shuffle_output`, `get_index_path`, and `stream_sort_shuffle_partition` is unchanged, so `ShuffleReaderExec` and the executor's Arrow Flight service work without modification.

New tests cover multi-spill, in-memory-only, and empty-partition round-trips on the writer/reader side, plus concatenated-stream and schema-only-leading-stream round-trips on the block-IO client.

No public-API changes. The sort-shuffle on-disk format changes — it is executor-internal, but in-flight files written by older binaries are not readable by this version.

### DIFF
--- a/ballista/client/tests/sort_shuffle.rs
+++ b/ballista/client/tests/sort_shuffle.rs
@@ -46,8 +46,12 @@ mod sort_shuffle_tests {
     enum ReadMode {
         /// Read shuffle data locally (default)
         Local,
-        /// Read shuffle data via flight service (remote read)
+        /// Read shuffle data via the flight service over gRPC (`do_get`)
         RemoteFlight,
+        /// Read shuffle data via the block-IO transport over gRPC (`do_action`
+        /// with `IO_BLOCK_TRANSPORT`). The server streams the partition's
+        /// byte range directly without decode/re-encode.
+        RemoteBlockIo,
     }
 
     /// Creates a standalone session context with sort-based shuffle enabled.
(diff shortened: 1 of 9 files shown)


milenkovicm:
this is cool @andygrove i wonder do you have some numbers? i wonder if improvement is it similar to #1315 ?

andygrove:
> this is cool @andygrove i wonder do you have some numbers? i wonder if improvement is it similar to #1315 ?

I'm running benchmarks now ....

andygrove:
The performance benefits in this PR are:

1. make it cheaper to combine the spill files into the final file - file copy rather than re-encode
2. allow the use of `IO_BLOCK_TRANSPORT` with sort-based shuffle

The standalone benchmark only measures the first item. I did not see a significant performance improvement with local tests. The numbers were about the same.

milenkovicm (approved):
same, lets merge this to unblock sort shuffle work, we can revisit later

PR #1616, opened 2026-04-28
feat: SchedulerConfig improvement

sandugood:

Closes #.

 # Rationale for this change
Improving Scheduler's configuration when setting up the Ballista cluster

Moving the buffer capacity of the cluster's event sender to the config rather than a separate constant

Yes, another CLA added: `--event-sender-buffer-capacity`


sandugood:
I also mentioned with a comment the `TaskDistribution` part. Seems that it can be replaced fully by the `TaskDistributionPolicy`.
lmk what you think

milenkovicm:
Do we need to expose this config parameter to end user? In which case will they change it?

sandugood:
After evaluating it for a bit, seems that this parameter might remain a hardcoded value.

sandugood:
Closing for now

PR #1624, opened 2026-04-29
feat(executor): bound executor memory via --memory-pool-size

### LINKED ISSUE #1563: How to limit the maximum memory usage of the ballista executor?
How should I limit the maximum memory usage of the ballista executor?

andygrove:

Closes #1563

# Rationale for this change

The executor today builds a `RuntimeEnv` with no `MemoryPool`, so DataFusion uses its unbounded default and spillable operators (sort, hash join, hash agg) grow until the host OOMs. There is no way to bound executor memory or trigger spilling. The `tuning-guide.md` already flagged this as future work.

Adds an opt-in `--memory-pool-size <SIZE>` flag (e.g. `8GB`, `512MiB`, plain bytes) on the executor binary. When set, every task receives an isolated `FairSpillPool` of size `total / concurrent_tasks`. Wrapping is applied after the base `RuntimeProducer` is resolved, so it composes with embedder-supplied producers (including the existing S3 helper) and preserves their `DiskManager`, `CacheManager`, and `ObjectStoreRegistry`. Hard error at startup if the per-task share would round to zero.

Adds `bytesize` as a dep on `ballista-executor` for size parsing. No public API changes outside the executor crate.

Yes. New optional CLI flag and a new section in `docs/source/user-guide/tuning-guide.md`. Default behavior is unchanged when the flag is omitted.

### DIFF
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -74,7 +74,8 @@ use crate::shutdown::ShutdownNotifier;
 use crate::{ArrowFlightServerProvider, terminate};
 use crate::{execution_loop, executor_server};
 
-/// Wrap a [`RuntimeProducer`] so that every produced [`RuntimeEnv`] carries a
+/// Wrap a [`RuntimeProducer`] so that every produced
+/// [`RuntimeEnv`](datafusion::execution::runtime_env::RuntimeEnv) carries a
 /// fresh [`FairSpillPool`] of size `total_bytes / concurrent_tasks`.
 ///
 /// Returns an error if the per-task share would be zero (i.e. `total_bytes <
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -41,7 +41,9 @@ use tokio::task::JoinHandle;
 use tokio::{fs, time};
 use uuid::Uuid;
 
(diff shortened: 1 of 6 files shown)


ballista/executor/src/executor_process.rs
```diff
@@ -261,6 +296,21 @@ pub async fn start_executor_process(
             })
         });
 
+    let runtime_producer = if let Some(total) = opt.memory_pool_size {
+        let producer = wrap_runtime_producer_with_memory_pool(
+            runtime_producer,
+            total,
+            concurrent_tasks,
+        )?;
+        let per_task = total / concurrent_tasks as u64;
```
milenkovicm, line 305: i wonder should we divide usable memory across tasks or use shared memory pool across all of them? is spark doing this ?

milenkovicm: i guess this is consistent with spark executor, is it ?

milenkovicm (approved):

ballista/executor/src/executor_process.rs
andygrove: This is the approach that we use in Comet - each task gets its own fair pool and within a task, the operators share that pool. This keeps things predictable and stable.

We can certainly explore adding other config options in the future.

andygrove:
Thanks for the review @milenkovicm

PR #1625, opened 2026-04-29
fix: `df.write_` fix as it was broken after update

milenkovicm:
`write_options` parameter is not respected at the
moment

Closes #.

 # Rationale for this change

fix `df.write_` as it was broken after update. 

updated `DistributedDataFrame` adding missing write methods or missing `write_options` parameter in current method.

`write_options` is not respected at the moment 


andygrove (approved):
Thanks @milenkovicm

PR #1626, opened 2026-04-30
perf(sort-shuffle): fix performance regression caused by datafusion upgrade

andygrove:

Closes #.

# Rationale for this change

The sort-based shuffle writer is dramatically slower than expected on string-heavy inputs. On TPC-H lineitem (16 cols, 5 string cols, 200 output partitions, 1M rows, LZ4 compression) the standalone `shuffle_bench` was running at ~22K rows/s (~46s) — orders of magnitude slower than the equivalent benchmark in DataFusion Comet.

Root cause: DataFusion 52's parquet reader produces `Utf8View` (string-view) columns by default. `arrow::compute::interleave_record_batch` keeps every source data buffer alive when interleaving view arrays, even when the output emits only a few rows from each. With 123 input batches and 200 output partitions (~5K rows each), each materialized partition batch ended up referencing 130+ source data buffers totalling ~132 MB of underlying data — for ~5K rows of actual content. The downstream IPC `StreamWriter` then serialized all of that, hence the 46s finalize.

After running `gc()` on view columns, each materialized partition batch holds ~1 MB of compact data instead of 132 MB. The same workload now runs in ~0.5s (~2M rows/s), a ~90x speedup on this benchmark. Verified with `cargo test -p ballista-core --lib execution_plans::sort_shuffle` (28 round-trip / spill / multi-spill tests pass).

- **Compact view columns inside `PartitionedBatchIterator::next`** (`partitioned_batch_iterator.rs`): after `interleave_record_batch`, downcast each output column and call `gc()` on `StringViewArray` / `BinaryViewArray` so both the spill path and the finalize path emit self-contained batches. Non-view columns are passed through unchanged.
- **`spill_count` metric now counts spill events** (one per memory-pressure flush) in `SortShuffleWriterExec`, instead of counting per-partition batches written. The per-batch counter is preserved on `SpillManager` as `total_spilled_batches`, included in the existing completion log.
- **INFO log per spill event** with per-event and cumulative batches/bytes.
- **`shuffle_bench` initializes `env_logger`** so the new spill INFO logs surface during benchmarks without needing `RUST_LOG`.

- `spill_count` semantics changed from \"batches written to spill files\" to \"spill events\". The new value is strictly less than or equal to the old one; plan output that prints this metric will show smaller numbers on jobs that spill.
- No public API changes.

### DIFF
--- a/ballista/core/src/execution_plans/sort_shuffle/partitioned_batch_iterator.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/partitioned_batch_iterator.rs
@@ -18,10 +18,44 @@
 //! Iterator that materializes per-partition rows into well-sized
 //! `RecordBatch`es using `arrow::compute::interleave_record_batch`.
 
+use datafusion::arrow::array::{Array, ArrayRef, BinaryViewArray, StringViewArray};
 use datafusion::arrow::compute::interleave_record_batch;
 use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::common::DataFusionError;
 use datafusion::error::Result;
+use std::sync::Arc;
+
+/// Compacts `Utf8View` / `BinaryView` columns by running `gc()` on them.
+///
+/// `interleave_record_batch` preserves the original data buffers for view
+/// arrays, so the output references every source data buffer the inputs
+/// touched even if it only emits a few rows. Without compaction, downstream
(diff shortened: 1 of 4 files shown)


andygrove:
@milenkovicm I am going to go ahead and merge this because it fixes a ~100x performance regression

PR #1628, opened 2026-04-30
fix(rest): remove unwrap and return 404 if executor does not exist

milenkovicm:

Closes #.

 # Rationale for this change

rest api has an is unwrap which will panic if executor does not exist 

remove unwrap and return 404


milenkovicm:
@sandugood would you have time for review

martin-g (approved):

milenkovicm:
Thank you @martin-g

PR #1631, opened 2026-04-30
feat(benchmarks): add --config to tpch datafusion + contributor guide

andygrove:

Closes #.

# Rationale for this change

Two related gaps for contributors who want to benchmark Ballista:

1. The `tpch benchmark datafusion` subcommand has no way to override DataFusion session config keys. The `tpch benchmark ballista` subcommand has had a `-c key=value` flag for this since #1486 era, but the in-process DataFusion path requires editing source to change a single config. That makes apples-to-apples comparisons between the two modes hard.
2. The contributor development guide currently has a one-paragraph `Benchmarking` section that links out to `benchmarks/README.md`. The README covers cluster setup but not the day-to-day contributor workflow: which subcommand to use, how to override configs, how to read the metrics, and how to capture a flame graph when something is unexpectedly slow.

- **Code:** `benchmarks/src/bin/tpch.rs` adds a repeatable `-c`/`--config key=value` flag to the `datafusion` subcommand, mirroring the one already on `ballista`. Overrides are applied to the `SessionConfig` before constructing the `SessionContext`. Invalid keys log a warning and are skipped, matching the existing behavior on the ballista path.
- **Docs:** New page `docs/source/contributors-guide/benchmarking.md` covering:
  - Generating TPC-H input with `tpchgen-rs`
  - Running the `tpch` binary in `benchmark datafusion` (in-process) and `benchmark ballista` (against a cluster) modes, with notes on picking a representative single query while iterating
  - Setting session configs via `-c key=value`, with a table of commonly tuned `datafusion.*` and `ballista.*` keys
  - Reading metrics and a pointer to the existing Metrics user guide
  - Profiling with `cargo flamegraph` and `samply`
- The `Benchmarking` section in `contributors-guide/development.md` now points at the new page and keeps the link to `benchmarks/README.md`
- The new page is registered in the `index.rst` toctree under Contributors Guide

`tpch benchmark datafusion` accepts a new `-c`/`--config` flag. No existing flags or defaults change, so prior invocations behave the same.

### DIFF
--- a/benchmarks/src/bin/tpch.rs
+++ b/benchmarks/src/bin/tpch.rs
@@ -164,6 +164,13 @@ struct DataFusionBenchmarkOpt {
     /// Path to output directory where JSON summary file should be written to
     #[structopt(parse(from_os_str), short = "o", long = "output")]
     output_path: Option<PathBuf>,
+
+    /// Configuration overrides in key=value format.
+    /// Can be specified multiple times, e.g.
+    /// -c datafusion.execution.target_partitions=16
+    /// -c datafusion.execution.batch_size=8192
+    #[structopt(short = "c", long = "config", number_of_values = 1)]
+    config_overrides: Vec<String>,
 }
 
 #[derive(Debug, StructOpt, Clone)]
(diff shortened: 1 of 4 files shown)


milenkovicm:
@andygrove is this PR still needed?

Issue #1632, opened 2026-04-30
tracking tpc-h sf=100 benchmark results


andygrove:
I'm running TPC-H benchmarks and wanted somewhere to keep track of the the results. No action is needed for this issue.

Setup:

```shell
RUST_LOG=info ./target/release/ballista-scheduler
RUST_LOG=info ./target/release/ballista-executor -c 16
```

Benchmark command:

```shell
./target/release/tpch benchmark ballista --host localhost --port 50050 --path /mnt/bigdata/tpch/sf100 --format parquet --iterations 1
```

andygrove:
After fixing the recent regression in sort-based shuffle, running with default configs:

```
Query 1 took 19147.2 ms and returned 4 rows
Query 2 took 10234.8 ms and returned 100 rows
Query 3 took 73433.7 ms and returned 10 rows
Query 4 took 15675.1 ms and returned 5 rows
Query 5 took 190789.9 ms and returned 5 rows
Query 6 took 4440.7 ms and returned 1 rows
Query 7 took 77920.9 ms and returned 4 rows
Query 8 took 262859.9 ms and returned 2 rows
Query 9 took 341142.4 ms and returned 175 rows
Query 10 took 20336.1 ms and returned 20 rows
Query 11 took 8026.6 ms and returned 0 rows
Query 12 took 10911.1 ms and returned 2 rows
Query 13 took 18763.3 ms and returned 45 rows
Query 14 took 5046.3 ms and returned 1 rows
Query 15 took 6356.9 ms and returned 0 rows
Query 16 took 4254.7 ms and returned 27840 rows
Query 17 took 205500.7 ms and returned 1 rows
Query 18 took 106379.5 ms and returned 100 rows
Query 19 took 13369.0 ms and returned 1 rows
Query 20 took 12163.2 ms and returned 17971 rows
Query 21 took 188630.3 ms and returned 100 rows
Query 22 took 4723.8 ms and returned 7 rows
```

andygrove:
With sort-based shuffle disabled, it is much faster, confusingly.

```
Query 1 took 19427.7 ms and returned 4 rows
Query 2 took 6971.9 ms and returned 100 rows
Query 3 took 12378.4 ms and returned 10 rows
Query 4 took 6086.6 ms and returned 5 rows
Query 5 took 29025.9 ms and returned 5 rows
Query 6 took 4478.6 ms and returned 1 rows
Query 7 took 44554.4 ms and returned 4 rows
[...]

PR #1623, opened 2026-04-29
feat(sort-shuffle): enable sort-based shuffle by default

andygrove:
# Which issue does this PR close?

Closes #.

# Rationale for this change

The sort-based shuffle writer is faster and more scalable than hash-based shuffle. It now supports the block io transport so is ready for use.  

# What changes are included in this PR?

- Flip the default of `ballista.shuffle.sort_based.enabled` from `false` to `true` in `ballista/core/src/config.rs`.
- Update test expectations:
  - `context_checks.rs` EXPLAIN and EXPLAIN ANALYZE plans now show `SortShuffleWriterExec` for hash-partitioned stages (and the additional `spill_*` metrics).
  - `execution_graph_dot.rs` dot-graph fixtures show `SortShuffleWriter` for hash-partitioned stages; terminal stages with `partitioning: None` continue to use `ShuffleWriter`, since the planner only switches to sort-based when partitioning is `Hash(...)`.
  - `planner.rs` `distributed_window_plan` downcasts to `SortShuffleWriterExec`.
  - `sort_shuffle.rs` `create_hash_shuffle_context` now sets the flag explicitly to `false` since it can no longer rely on the default.
- Update `docs/source/user-guide/tuning-guide.md` and `docs/source/user-guide/configs.md` to mark sort-based shuffle as the default and document how to opt in to the hash-based writer.

# Are there any user-facing changes?

Yes, query plans will now show sort-based shuffle in many cases

PR #1633, opened 2026-04-30
feat(sort-shuffle): disable sort-based shuffle by default

andygrove:

Closes #.

# Rationale for this change

Although the standalone shuffle benchmark shows that sort-based shuffle is orders of magnitude faster than hash-based shuffle, TPC-H @ sf100 performance is worse, so we should revert making it the default until we can understand this better.

I have been tracking benchmarks in https://github.com/apache/datafusion-ballista/issues/1632

Reverts #1623. The default value of `ballista.shuffle.sort_based.enabled` returns to `false`, and the test golden plans, dot-graph labels, and user-guide docs are restored to describe hash-based shuffle as the default.

Yes. Users who were relying on sort-based shuffle being on by default will need to opt in explicitly:

```rust
let session_config = SessionConfig::new_with_ballista()
    .set_bool("ballista.shuffle.sort_based.enabled", true);
```

No public API signatures change, so no `api change` label is needed.


andygrove:
this is no longer - I found the root cause of the performance issues

PR #1636, opened 2026-05-01
fix(sort-shuffle): bound writer memory with per-task spill threshold

andygrove:

Closes #.

# Rationale for this change

This fixes performance of sort-based shuffle when running the TPC-H benchmarks.

Previously, memory was unbounded so all input data was getting accumulated up to the memory pool limit, and the executor was not imposing a limit.

This PR introduces a configurable memory limit per task (default 256 MB) so that spilling gets triggered and keeps memory under control.

# Performance

Benchmark command:

```
$ ./target/release/tpch benchmark ballista --host localhost --port 50050 --path /mnt/bigdata/tpch/sf100 --format parquet --iterations 1 --config ballista.shuffle.sort_based.enabled=true
```

Performance on main branch with sort-based shuffle with 2 output partitions (which is unrealistic, IMO, but is the default in the current benchmark):

```
Query 1 took 19147.2 ms and returned 4 rows
Query 2 took 10234.8 ms and returned 100 rows
Query 3 took 73433.7 ms and returned 10 rows
Query 4 took 15675.1 ms and returned 5 rows
Query 5 took 190789.9 ms and returned 5 rows
...
```

This was clearly no good.

With this PR :

```
Query 1 took 19227.0 ms and returned 4 rows
Query 2 took 9125.7 ms and returned 100 rows
Query 3 took 15806.4 ms and returned 10 rows
Query 4 took 7513.4 ms and returned 5 rows
Query 5 took 36166.8 ms and returned 5 rows
Query 6 took 4570.0 ms and returned 1 rows
Query 7 took 51075.8 ms and returned 4 rows
Query 8 took 37977.7 ms and returned 2 rows
Query 9 took 62603.1 ms and returned 175 rows
Query 10 took 17951.8 ms and returned 20 rows
Query 11 took 6616.2 ms and returned 0 rows
Query 12 took 10176.2 ms and returned 2 rows
Query 13 took 18517.7 ms and returned 45 rows
Query 14 took 5043.6 ms and returned 1 rows
Query 15 took 6152.1 ms and returned 0 rows
Query 16 took 4255.7 ms and returned 27840 rows
Query 17 took 52557.4 ms and returned 1 rows
Query 18 took 52189.7 ms and returned 100 rows
Query 19 took 13621.3 ms and returned 1 rows
Query 20 took 11598.2 ms and returned 17971 rows
Query 21 took 63261.7 ms and returned 100 rows
Query 22 took 4083.3 ms and returned 7 rows
```

This is now closer hash-based shuffle, which does better with small number of partitions :

```
$ ./target/release/tpch benchmark ballista --host localhost --port 50050 --path /mnt/bigdata/tpch/sf100 --format parquet --iterations 1 --config ballista.shuffle.sort_based.enabled=false

Query 1 took 19134.7 ms and returned 4 rows
Query 2 took 7295.9 ms and returned 100 rows
Query 3 took 12959.0 ms and returned 10 rows
Query 4 took 6147.2 ms and returned 5 rows
Query 5 took 28714.4 ms and returned 5 rows
Query 6 took 4544.4 ms and returned 1 rows
Query 7 took 40825.0 ms and returned 4 rows

```

With 200 partitions (Spark's default), performance is much better for many queries with sort-based shuffle:

sort-based:

```
$ ./target/release/tpch benchmark ballista --host localhost --port 50050 --path /mnt/bigdata/tpch/sf100 --format parquet --iterations 1 --config ballista.shuffle.sort_based.enabled=true --partitions 200

Query 1 took 3607.3 ms and returned 4 rows
Query 2 took 10533.4 ms and returned 100 rows
Query 3 took 9143.9 ms and returned 10 rows
Query 4 took 5031.2 ms and returned 5 rows
Query 5 took 16945.4 ms and returned 5 rows
...
```

With hash-based and 200 output partitions, I hit the file handle limit due to separate files per output partition. I cannot increase file handles beyond 8192 using `ulimit`.

```
$ ulimit -n 8192
$ ./target/release/tpch benchmark ballista --host localhost --port 50050 --path /mnt/bigdata/tpch/sf100 --format parquet --iterations 1 --config ballista.shuffle.sort_based.enabled=false --partitions 200
Running benchmarks with the following options: BallistaBenchmarkOpt { query: None, debug: false, expected_results: None, iterations: 1, batch_size: 8192, path: "/mnt/bigdata/tpch/sf100", file_format: "parquet", partitions: 200, host: Some("localhost"), port: Some(50050), output_path: None, config_overrides: ["ballista.shuffle.sort_based.enabled=false"] }
Query 1 took 3446.2 ms and returned 4 rows
[2026-05-01T03:31:48Z ERROR ballista_core::execution_plans::distributed_query] Job DRCZ6hQ failed: Job failed due to stage 5 failed: Task failed due to runtime execution error: DataFusionError(ParquetError(General("Failed to fetch metadata for file mnt/bigdata/tpch/sf100/partsupp.parquet/part10.parquet: Parquet error: External: Generic LocalFileSystem error: Unable to open file /mnt/bigdata/tpch/sf100/partsupp.parquet/part10.parquet: Too many open files (os error 24)")))
```

- New config `ballista.shuffle.sort_based.memory_limit_per_task_bytes` (default 256 MB).
- `SortShuffleConfig` carries the limit, plumbed from `BallistaConfig` through the planner.
- `SortShuffleWriterExec` tracks its own buffered-bytes counter and spills when the counter crosses the configured threshold, independently of the runtime `MemoryPool`. The `MemoryReservation` registration is kept (best-effort `try_grow`) so the pool still sees this writer as a memory citizen for visibility purposes.
- Existing spill round-trip tests rewired to drive spilling via the new per-task budget rather than via a tight `FairSpillPool` size; pool size in tests is now generous so the post-test `pool.reserved() == 0` assertion still validates that the writer releases its reservation.
- Total worst-case sort shuffle memory per executor is approximately `concurrent_tasks * memory_limit_per_task_bytes`. At the default 256 MB × 8 slots that's ~2 GB, which is safe on a laptop running multiple executors and still leaves the runtime pool free for other operators.

A new configuration key `ballista.shuffle.sort_based.memory_limit_per_task_bytes` is added with a default of 256 MB. Users running queries that previously fit comfortably in unbounded memory may now see spilling — set this higher if avoiding spills matters more than bounding memory.

### DIFF
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -75,6 +75,10 @@ pub const BALLISTA_SHUFFLE_SORT_BASED_ENABLED: &str =
 /// Configuration key for sort shuffle target batch size in rows.
 pub const BALLISTA_SHUFFLE_SORT_BASED_BATCH_SIZE: &str =
     "ballista.shuffle.sort_based.batch_size";
+/// Configuration key for the per-task buffered-bytes budget at which the
+/// sort shuffle writer spills its in-memory batches to disk.
+pub const BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT_PER_TASK_BYTES: &str =
+    "ballista.shuffle.sort_based.memory_limit_per_task_bytes";
 
 /// Result type for configuration parsing operations.
 pub type ParseResult<T> = result::Result<T, String>;
@@ -140,6 +144,13 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
                          "Target batch size in rows for coalescing small batches in sort shuffle".to_string(),
                          DataType::UInt64,
                          Some((8192).to_string())),
+        ConfigEntry::new(BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT_PER_TASK_BYTES.to_string(),
+                         "Per-task buffered-bytes budget at which the sort shuffle writer spills its \
+                         in-memory batches to disk. Counted independently of the runtime memory pool, so \
+                         spilling kicks in even when the pool is unbounded. Total worst-case sort shuffle \
+                         memory per executor is approximately concurrent_tasks * this value.".to_string(),
+                         DataType::UInt64,
+                         Some((256 * 1024 * 1024).to_string())),
         ConfigEntry::new(BALLISTA_CLIENT_PULL.to_string(),
                          "Should client employ pull or push job tracking. In pull mode client will make a request to server in the loop, until job finishes. Pull mode is kept for legacy clients.".to_string(),
                          DataType::Boolean,
@@ -344,6 +355,12 @@ impl BallistaConfig {
     /// Returns the target batch size for sort-based shuffle.
     pub fn shuffle_sort_based_batch_size(&self) -> usize {
         self.get_usize_setting(BALLISTA_SHUFFLE_SORT_BASED_BATCH_SIZE)
+    }
+
+    /// Returns the per-task buffered-bytes budget at which the sort shuffle
+    /// writer spills its in-memory batches to disk.
+    pub fn shuffle_sort_based_memory_limit_per_task_bytes(&self) -> usize {
+        self.get_usize_setting(BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT_PER_TASK_BYTES)
     }
 
     /// Should client employ pull or push job tracking strategy
--- a/ballista/core/src/execution_plans/sort_shuffle/config.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/config.rs
@@ -29,6 +29,10 @@ pub struct SortShuffleConfig {
     /// Target batch size in rows when materializing buffered indices via
     /// `interleave_record_batch` (default: 8192).
     pub batch_size: usize,
+    /// Per-task buffered-bytes budget at which the writer spills its in-memory
+    /// batches to disk. Counted independently of the runtime `MemoryPool`, so
+    /// spilling kicks in even when the pool is unbounded.
+    pub memory_limit_per_task_bytes: usize,
 }
 
 impl Default for SortShuffleConfig {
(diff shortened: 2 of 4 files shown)


andygrove:
@milenkovicm I'd like to merge this PR next and then have another go at running benchmarks at scale.

ballista/core/src/config.rs
```diff
@@ -140,6 +144,13 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
                          "Target batch size in rows for coalescing small batches in sort shuffle".to_string(),
                          DataType::UInt64,
                          Some((8192).to_string())),
+        ConfigEntry::new(BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT_PER_TASK_BYTES.to_string(),
+                         "Per-task buffered-bytes budget at which the sort shuffle writer spills its \
+                         in-memory batches to disk. Counted independently of the runtime memory pool, so \
+                         spilling kicks in even when the pool is unbounded. Total worst-case sort shuffle \
+                         memory per executor is approximately concurrent_tasks * this value.".to_string(),
+                         DataType::UInt64,
+                         Some((256 * 1024 * 1024).to_string())),
```
martin-g, line 153: `256 * 1024 * 1024` is used both here and at sort_shuffle/config.rs. They may drift.
Maybe declare a constant and reuse it ?

ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -262,7 +267,14 @@ impl SortShuffleWriterExec {
                 let after = buffered.indices_allocated_size();
                 growth += after.saturating_sub(before);
 
-                if reservation.try_grow(growth).is_err() {
+                // Mirror the growth in the runtime pool reservation so the pool
+                // sees this writer's memory usage. Best-effort: if the pool is
+                // bounded and rejects the grow, that's fine — the absolute
+                // counter below still triggers a spill.
+                let _ = reservation.try_grow(growth);
```
martin-g, line 274: What bothers me here is that a user may expect that setting up a memory pool limit will keep the process RSS memory close to that limit, but here it is ignored and depending on the value of the new config the process may consume much more memory than the pool limit and this will raise questions.

ballista/core/src/execution_plans/sort_shuffle/config.rs
```diff
@@ -37,19 +41,27 @@ impl Default for SortShuffleConfig {
...
 
 impl SortShuffleConfig {
-    /// Creates a new configuration.
+    /// Creates a new configuration with the default per-task memory limit.
     pub fn new(enabled: bool, compression: CompressionType, batch_size: usize) -> Self {
         Self {
             enabled,
             compression,
             batch_size,
+            memory_limit_per_task_bytes: Self::default().memory_limit_per_task_bytes,
```
martin-g, line 56: This is API preserving but error-prone.
https://github.com/apache/datafusion-ballista/blob/0e352d10175e8a5ac9e51d0adf30e82563460a56/ballista/core/src/serde/mod.rs#L384-L388 silently uses the default value instead of any custom value set with `.with_memory_limit_per_task_bytes()` on the sender side.

milenkovicm:
will have a look later or over weekend, @andygrove  please merge if you think it will unblock you, we can do one big review once sort shuffle settles 

also, please note, `view_types` are disabled in ballista due to arrow ipc issues, https://github.com/milenkovicm/datafusion-ballista/blob/b7c5e0975d303e79435d65262638b128a2f26956/ballista/core/src/extension.rs#L636

perhaps its worth enabling with the fix you've done yesterday

PR #1641, opened 2026-05-01
feat(bench): show TPC-H query timings in seconds and add total time

andygrove:

Closes #.

# Rationale for this change

The TPC-H benchmark output is hard to scan when comparing runs. Times
print as milliseconds with one decimal place, the default of three
iterations slows down ad-hoc runs, and there is no overall total at the
end. Aligning the per-query format and JSON summary durations with the
datafusion-comet benchmark makes it easier to compare results between
the two engines.

- Print per-query and average elapsed times in seconds with three
  decimal places (microsecond resolution).
- Print a `Total time` line at the end of the run, covering all
  queries and iterations.
- Change the default `--iterations` value from 3 to 1 for both the
  `ballista` and `datafusion` benchmark subcommands.
- Store JSON summary `elapsed` values in seconds so the file format
  lines up with datafusion-comet's `durations` array.

Yes:

- The default of `--iterations` changes from 3 to 1. Pass
  `--iterations 3` explicitly to keep the previous behavior.
- The console output now uses seconds (`s`) with three decimal places
  instead of milliseconds (`ms`) with one decimal place.
- `elapsed` values in the JSON summary are now in seconds rather than
  milliseconds.


benchmarks/src/bin/tpch.rs
```diff
@@ -354,34 +355,37 @@ async fn benchmark_datafusion(opt: DataFusionBenchmarkOpt) -> Result<Vec<RecordB
...
-            let avg = millis.iter().sum::<f64>() / millis.len() as f64;
-            println!("Query {} avg time: {:.1} ms", query, avg);
+            let avg = secs.iter().sum::<f64>() / secs.len() as f64;
+            println!("Query {} avg time: {:.3} s", query, avg);
         }
 
         benchmark_run.add_query_run(query_run);
     }
 
+    let total_elapsed = total_start.elapsed().as_secs_f64();
```
milenkovicm: Should this print total benchmark running time or total query running time.
If later is the case then it should be sum of all query times or sum of all average query times if there is more than one iteration per query

andygrove: Good question. Switched to summing per-query times: sum of elapsed when iterations is 1, sum of per-query averages when iterations > 1.

milenkovicm (approved):
Thanks @andygrove

PR #1644, opened 2026-05-01
chore: remove NYC Taxi benchmark

andygrove:

None.

# Rationale for this change

The NYC Taxi benchmark runs a single aggregation query against an unpartitioned dataset, which doesn't meaningfully exercise a distributed query engine. TPC-H and TPC-DS provide standardized, scalable workloads with joins, subqueries, and well-defined scale factors, so they are better suited for benchmarking Ballista.

- Delete `benchmarks/src/bin/nyctaxi.rs`
- Remove the NYC Taxi section from `benchmarks/README.md`
- Replace `nyctaxi` example paths in the Python user guide with generic ones

The `nyctaxi` benchmark binary is removed.


Dandandan (approved):

andygrove:
Thanks @Dandandan

Issue #1627, opened 2026-04-30
feat: make it user configurable to display plans as tree renderer


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

job REST endpoint returns plans as single line per node:

```bash
curl -sv http://localhost:50050/api/job/8Am4heu |jq

{
  "job_id": "8Am4heu",
  "job_name": "",
  "job_status": "Completed. Produced 1 partition containing 1 row. Elapsed time: 13125 ms.",
  "status": "Completed",
  "num_stages": 1,
  "completed_stages": 1,
  "percent_complete": 100,
  "start_time": 1777541048696,
  "end_time": 1777541061821,
  "logical_plan": "Projection: Int64(1)\n  EmptyRelation: rows=1",
  "physical_plan": "ProjectionExec: expr=[1 as Int64(1)]\n  PlaceholderRowExec\n",
  "stage_plan": "ExecutionGraph[job_id=8Am4heu, session_id=32ab0721-562a-47fa-a420-7dcc24e59f91, available_tasks=0, is_successful=true]\n=========SuccessfulStage[stage_id=1.0, partitions=1]=========\nShuffleWriterExec: partitioning: None, metrics=[output_rows=1, input_rows=1, repart_time=1ns, write_time=1.67ms]\n  ProjectionExec: expr=[1 as Int64(1)], metrics=[output_rows=1, elapsed_compute=296.83µs, output_bytes=8.0 B, output_batches=1, expr_0_eval_time=168.71µs]\n    PlaceholderRowExec, metrics=[]\n"
}
```

**Describe the solution you'd like**

make it user configurable to return plans as tree renderer

`displayable(running_stage.plan.as_ref()).indent(false).to_string()) -> displayable(running_stage.plan.as_ref()).tree_render() .to_string())`

**Describe alternatives you've considered**

**Additional context**

milenkovicm:
cc @sandugood if you're interested

sandugood:
take

milenkovicm:
Cc @martin-g i guess it should not make problems on TUI

martin-g:
> make it user configurable to return plans as tree renderer

with a query parameter ?!

sandugood:
> > make it user configurable to return plans as tree renderer
> 
> with a query parameter ?!

I think query parameter seems good for this
something like:
`api/job/8Am4heu?plan_format=tree`

milenkovicm:
[...]

PR #1650, opened 2026-05-03
feat: rest api supports plan tree rendering

sandugood:

Tackles #1627.

 # Rationale for this change
Add additional optional query parameters to the Scheduler's REST API to get data about the jobs. 

1. Added JobQueryParams struct for optional query parameters construction with two fields: `include_logical` and `tree_rendered`
2. Based on optional parameters a JSON response is formed:
- `include_logical` is true by default, but can be configured to be excluded from the response. Extracted by the name it is responsible for returning `logical_plan` to the user
- `tree_rendered` is the query parameter that controls how `physical_plan` is rendered. If `tree_rendered` is set to `false` then it is rendered as `DisplayableExecutionPlan::new(job.physical_plan().as_ref()).indent(false)`, otherwise `displayable(job.physical_plan().as_ref()).tree_render().to_string()`, which can be easily rendered on the user side
```
curl -s 'localhost:50050/api/job/JcYQYNY?tree_rendered=true' | jq -r ".physical_plan"
┌───────────────────────────┐
│       AggregateExec       │
│    --------------------   │
│           aggr:           │
│  min(test.c12), max(test  │
│           .c12)           │
│                           │
│        group_by: c1       │
│                           │
│           mode:           │
│      FinalPartitioned     │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      RepartitionExec      │
│    --------------------   │
│ partition_count(in->out): │
│           1 -> 4          │
│                           │
│    partitioning_scheme:   │
│      Hash([c1@0], 4)      │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       AggregateExec       │
│    --------------------   │
│           aggr:           │
│  min(test.c12), max(test  │
│           .c12)           │
│                           │
│        group_by: c1       │
│       mode: Partial       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         FilterExec        │
│    --------------------   │
│         predicate:        │
│  c11 > 0.1 AND c11 < 0.9  │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       DataSourceExec      │
│    --------------------   │
│          files: 1         │
│        format: csv        │
└───────────────────────────┘

```

Additional optional query parameters added to the Scheduler's REST API, so no breaking user-facing changes.


sandugood:
@milenkovicm would be glad if you could review this change
LogicalPlan can't be used directly in the displayable() or so it is in the current DF implementation

ballista/scheduler/src/state/mod.rs
```diff
@@ -444,17 +444,15 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
         } else {
             None
         };
-
+        // let logical_plan_tree = DisplayableExecutionPlan::new(plan.as_ref());
```
milenkovicm: nit, can we remove it

ballista/scheduler/src/state/aqe/mod.rs
```diff
@@ -130,7 +130,7 @@ impl AdaptiveExecutionGraph {
         queued_at: u64,
         session_config: Arc<SessionConfig>,
         logical_plan: Option<String>,
-        physical_plan: Option<String>,
+        physical_plan: Arc<dyn ExecutionPlan>,
```
milenkovicm: is `phisical_plan` duplication of `plan`?

ballista/scheduler/src/test_utils.rs
```diff
@@ -973,7 +974,7 @@ pub async fn test_two_aggregations_plan(partition: usize) -> StaticExecutionGrap
         Arc::new(SessionConfig::new_with_ballista()),
         &mut planner,
         None,
-        None,
+        cloned_plan,
```
milenkovicm: i guess we do not need this paramter

milenkovicm (commented):
i think this make sense, few minor comments

ballista/scheduler/src/state/aqe/mod.rs
sandugood: Actually, yeah. It was the planned before the `transform_down` in the `SchedulerState`'s `submit_job` function
Now using the physical plan we get after transformations.

ballista/scheduler/src/api/handlers.rs
```diff
@@ -203,6 +205,12 @@ pub struct QueryStageSummary {
     pub tasks: Vec<Option<TaskSummary>>,
 }
 
+#[derive(Debug, serde::Deserialize, Default)]
+pub struct JobQueryParams {
+    pub include_logical: Option<bool>,
```
milenkovicm: In which case we would not include logical plan? What's the idea behind this flag?

sandugood: Considering the fact, that LogicalPlan can't be rendered using the `displayable()` func, users might only be interested in the tree-rendered style of the physical plan. However, getting logical plan info is toggled by default, so users do not face any breaking changes in the API

sandugood: But we can also keep the old API, where logical_plan is just sent to the end user by default (no changes). Wdyt @milenkovicm?

milenkovicm: I'm struggling to find a case where we would disable logical plan. Let's not give that option for now and revert to previous behaviour if you agree @sandugood

sandugood: Yeah, I think that it makes perfect sense for now. Thanks @milenkovicm

milenkovicm (commented):
i believe this is the last comment: 

get_query_stages returns `QueryStageSummary` which includes plan string as well https://github.com/apache/datafusion-ballista/blob/9263e40916000b0508c263235a4da8b50625b5cc/ballista/scheduler/src/api/handlers.rs#L503

we need to handle this case as well

sandugood:
I thought that `render_tree` seemed clearer (I'm always unsure about the namings) as parameter name, so I switched to it.

Added this query param to the `/api/job/{job_id}/stages` endpoint. Looks like this:

```
~ curl -s localhost:50050/api/job/Za9mMBI/stages?render_tree=true | jq -r ".stages.[0].stage_plan"
┌───────────────────────────┐
│   SortShuffleWriterExec   │
│    --------------------   │
│       partitioning:       │
│      Hash([c1@0], 4)      │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       AggregateExec       │
│    --------------------   │
│           aggr:           │
│  min(test.c12), max(test  │
│           .c12)           │
│                           │
│        group_by: c1       │
│       mode: Partial       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         FilterExec        │
│    --------------------   │
│         predicate:        │
│  c11 > 0.1 AND c11 < 0.9  │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       DataSourceExec      │
│    --------------------   │
│          files: 1         │
│        format: csv        │
└───────────────────────────┘
```

milenkovicm (approved):
thanks @sandugood

milenkovicm:
@martin-g any suggestions ?

martin-g:
While adding support for tree rendering to the TUI app I noticed that the width of the ASCII chart is rather narrow.
I didn't find an easy way to make it wider. 

image

PR #1651, opened 2026-05-03
feat: default to sort-merge join

### LINKED ISSUE #1648: Should Ballista use sort-merge join rather than hash join by default?
DataFusion's hash join implementation does not have spilling yet, so can easily OOM. This is even more likely in Ballista because each executor is processing multiple tasks in parallel.

For example, I cannot get the integration tests to pass (`./dev/integration-test.sh`) due to OOM, unless I disable hash joins.

Personally, I think it would be better for Ballista to default to sort-merge join, and let the user override that if they want hash join. This would be the opposite of DataFusion's config, which enables hash join by default.

It would be nice if DataFusion had a spilling hash join implementation. I am working on one in Comet in https://github.com/apache/datafusion-comet/pull/4184

**Additional context**

andygrove:

Closes #1648.

## Rationale for this change

DataFusion's hash join has no spill support, so the build side must fit in memory per task. Ballista executors run multiple tasks per host, so per-task build sides aggregate and OOM the executor under realistic load — for example, the integration test suite (`./dev/integration-tests.sh`) does not complete on a typical machine without disabling hash joins.

DataFusion sets `datafusion.optimizer.prefer_hash_join = true`, which is fine for a single-node engine but a poor fit for a distributed multi-task one. This PR flips the Ballista default to sort-merge join, which spills, and leaves users a session-level knob to opt back into hash join when they know the build side fits.

- Set `datafusion.optimizer.prefer_hash_join = false` inside `ballista_restricted_configuration` (`ballista/core/src/extension.rs`), alongside the other soft DataFusion-default overrides. This is a soft default — users can override per session.
- Update the existing `should_support_sort_merge_join` integration test to assert the default picks `SortMergeJoinExec` via EXPLAIN, instead of relying on a result-row check that would pass under either join algorithm.
- Add `should_support_hash_join_when_opted_in` to verify users can still get `HashJoinExec` after `SET datafusion.optimizer.prefer_hash_join = true`.
- Document the new default and the opt-in path in `docs/source/user-guide/tuning-guide.md` under a new "Join Strategy" subsection.

Yes — behaviour change. Queries that previously planned a `HashJoinExec` will now plan a `SortMergeJoinExec` by default. The old behaviour is one `SET` away. Documented in the tuning guide. No public Rust API change, so no `api change` label.


milenkovicm (commented):
LGTM, sorry @andygrove I've missed this one

milenkovicm (approved):

martin-g (approved):

ballista/client/tests/context_checks.rs
```diff
@@ -39,6 +39,28 @@ mod supported {
         crate::common::example_test_data()
     }
 
+    /// Concatenate the `plan` column of an EXPLAIN result into a single
+    /// string for substring assertions.
+    fn plan_text(batches: &[RecordBatch]) -> String {
```
milenkovicm: Can we use standard batch assertion instead of bringing this function?

milenkovicm (approved):

ballista/client/tests/context_checks.rs
andygrove: Good call, switched to pretty_format_batches in 627e1521. The substring checks for SortMergeJoinExec/HashJoinExec still work against the formatted output, and we drop the custom helper.

I also tried walking the physical plan tree directly to avoid string matching entirely, but that meant bypassing BallistaQueryPlanner and planning locally with DefaultPhysicalPlanner. That tests DataFusion respects prefer_hash_join rather than that Ballista does, so I reverted. EXPLAIN actually round-trips through the scheduler in the remote case, which is what we want for this test.

PR #1652, opened 2026-05-03
fix(metrics): avoid stage metrics inflation by tracking partition snapshots

### LINKED ISSUE #1635: EXPLAIN ANALYZE row counts and elapsed_compute inflated by partition count
## Describe the bug

`EXPLAIN ANALYZE` reports operator metrics that are inflated by approximately the stage's partition count. The actual data flow and query results are correct — only the displayed metrics are wrong, which makes performance analysis misleading.

## To Reproduce

Run any TPC-H query with `EXPLAIN ANALYZE`, e.g. Q3 against SF100 with `--partitions 8`:

```sql
EXPLAIN ANALYZE
select l_orderkey, sum(l_extendedprice * (1 - l_discount)) as revenue, o_orderdate, o_shippriority
from customer, orders, lineitem
where c_mktsegment = 'BUILDING' and c_custkey = o_custkey and l_orderkey = o_orderkey
  and o_orderdate < date '1995-03-15' and l_shipdate > date '1995-03-15'
group by l_orderkey, o_orderdate, o_shippriority
order by revenue desc, o_orderdate
limit 10;
```

Look at stage 5's lineitem `ShuffleReaderExec`:

```
ShuffleReaderExec: partitioning: Hash([l_orderkey@0], 8), metrics=[output_rows=2.58 B, elapsed_compute=69.33s, ...]
```

The lineitem shuffle (stage 4) actually wrote 323M rows. The reader cannot have emitted 2.58 B rows — that is `~8 × 323M`, exactly the partition count multiplier. Same pattern shows up on every stage: each `ShuffleReaderExec` and the operators above it report `partition_count ×` the true row count.

I verified the actual row counts by adding a counter inside `CoalescedShuffleReaderStream::poll_next` that sums every batch returned to downstream operators. With `--partitions 8` the counter sums to ~320M (correct), while EXPLAIN ANALYZE reports 2.58 B (inflated by 8).

The query results themselves are correct — the inflation is purely a display artifact of how stage metrics are aggregated.

## Expected behavior

`output_rows` displayed in EXPLAIN ANALYZE should equal the actual rows the operator emitted, summed across that stage's tasks (so for an N-task stage where each task emits R rows, the metric should be N×R, not N²×R).

## Suspected location

The aggregation in `ballista/scheduler/src/state/execution_stage.rs::combine_metrics_set` repeatedly pushes each new task's metric values into the existing aggregated `MetricsSet` and then calls `aggregate_by_name()`. I suspect the previously-aggregated value is being summed back in on every task update, producing the `partition_count ×` inflation. Other Count and Time metrics on operators in the same stage (e.g. `elapsed_compute`, `build_time`, `join_time`) appear similarly inflated.

## Additional context

- Reproduced against current `main` with single-executor standalone cluster
- Inflation factor = stage's task count, consistent across `--partitions 2 / 4 / 8 / 16 / 32`
- Final query results are unaffected — actual data flow through the operators is correct

danielhumanmod:

Closes  #1635

 # Rationale for this change

There are two cause for this issue

1. Executor side
In `reset_metrics_for_execution_plan`, the function invokes `with_new_children` on ShuffleReadExec. However, in its current implementation, the metrics are not actually reset:
```
if children.is_empty() {
   Ok(self)
}
```
Since the metrics are stored as `Arc<Mutex<MetricsSet>>` and shared across multiple jobs, leading to inflated values at the data source level.

2. Scheduler side
Scheduler stage metrics were aggregated by name each time a task/partition reported metrics. This could cause already-aggregated stage metrics to be calculated multiple times when metrics from the same task/partition reported more than once.

for example:
```
partition 0 reports output_rows=100
stage_metrics stores output_rows=100, partition=None

partition 1 reports output_rows=100
stage_metrics stores output_rows=200, partition=None

partition 0 is processed again with output_rows=100
stage_metrics stores output_rows=300, partition=None  # should still be 200
```

1. Executor side
Reset metrics in `with_new_children` for shuffle reader exec

2. Scheduler side
After:
```
partition 0 reports output_rows=100
  output_rows=100, partition=0

partition 1 reports output_rows=100
  output_rows=100, partition=0
  output_rows=100, partition=1
```
Metrics from the same partition replace the previous snapshot, and aggregate_by_name() is only called in display / formatting paths.

No

### DIFF
--- a/ballista/scheduler/src/api/handlers.rs
+++ b/ballista/scheduler/src/api/handlers.rs
@@ -522,23 +522,13 @@ pub async fn get_query_stages<
                             .enumerate()
                             .map(|(partition_id, task_info)| {
                                 task_info.as_ref().map(|info| {
-                                    let partition_metrics = running_stage
+                                    let (input_rows, output_rows) = running_stage
                                         .stage_metrics
                                         .as_deref()
-                                        .and_then(|m| m.get(partition_id));
-                                    let (input_rows, output_rows) = partition_metrics
-                                    .map(|m| {
-                                        let input = get_combined_count(
-                                            std::slice::from_ref(m),
-                                            "input_rows",
-                                        );
-                                        let output = get_combined_count(
-                                            std::slice::from_ref(m),
-                                            "output_rows",
-                                        );
-                                        (input,output)
-                                    })
-                                    .unwrap_or((0,0));
+                                        .map(|metrics| {
+                                            get_partition_counts(metrics, partition_id)
+                                        })
+                                        .unwrap_or((0, 0));
 
                                     let start_exec_time = info.start_exec_time as u64;
                                     let end_exec_time = info.end_exec_time as u64;
@@ -580,21 +570,10 @@ pub async fn get_query_stages<
                             .iter()
                             .enumerate()
                             .map(|(partition_id, task_info)| {
-                                let partition_metrics =
-                                    completed_stage.stage_metrics.get(partition_id);
-                                let (input_rows, output_rows) = partition_metrics
-                                    .map(|m| {
-                                        let input = get_combined_count(
-                                            std::slice::from_ref(m),
-                                            "input_rows",
-                                        );
-                                        let output = get_combined_count(
-                                            std::slice::from_ref(m),
-                                            "output_rows",
-                                        );
-                                        (input,output)
-                                    })
-                                    .unwrap_or((0,0));
+                                let (input_rows, output_rows) = get_partition_counts(
+                                    &completed_stage.stage_metrics,
+                                    partition_id,
+                                );
 
                                 let start_exec_time = task_info.start_exec_time as u64;
                                 let end_exec_time = task_info.end_exec_time as u64;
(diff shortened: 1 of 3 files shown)


ballista/scheduler/src/api/handlers.rs
```diff
@@ -522,23 +522,13 @@ pub async fn get_query_stages<
                             .enumerate()
                             .map(|(partition_id, task_info)| {
                                 task_info.as_ref().map(|info| {
-                                    let partition_metrics = running_stage
+                                    let (input_rows, output_rows) = running_stage
                                         .stage_metrics
                                         .as_deref()
-                                        .and_then(|m| m.get(partition_id));
```
danielhumanmod, line 528: This change is optional, but I think it better aligns with the current metrics format. `stage_metrics` is indexed by operator, so the index here does not necessarily correspond to the partition id.

danielhumanmod:
cc @andygrove @milenkovicm

danielhumanmod:
At the same time I am validating this change via TPC-H query

danielhumanmod:
> At the same time I am validating this change via TPC-H query

Run TPC-H Q3 against SF1 with target_partitions=8 on a single-executor local cluster:
```
SET datafusion.execution.target_partitions = 8;

EXPLAIN ANALYZE
select l_orderkey, sum(l_extendedprice * (1 - l_discount)) as revenue, o_orderdate, o_shippriority
from customer, orders, lineitem
where c_mktsegment = 'BUILDING' and c_custkey = o_custkey and l_orderkey = o_orderkey
  and o_orderdate < date '1995-03-15' and l_shipdate > date '1995-03-15'
group by l_orderkey, o_orderdate, o_shippriority
order by revenue desc, o_orderdate
limit 10;
```

Before
```
 ShuffleReaderExec: partitioning: Hash([l_orderkey@0], 8), metrics=[output_rows=24.08 M, elapsed_compute=478.20ms, output_bytes=0.0 B, output_batches=0]        
```

After:
```
 ShuffleReaderExec: partitioning: Hash([l_orderkey@0], 8), metrics=[output_rows=3.24 M, elapsed_compute=75.88ms, output_bytes=0.0 B, output_batches=0]        
```

andygrove (approved):
Thanks @danielhumanmod!

milenkovicm:
thanks @danielhumanmod and @andygrove

PR #1670, opened 2026-05-08
[TUI] Show executor's details in a popup

### LINKED ISSUE #1642: [TUI] Render Executor's system & process metrics
```
➜  ~ curl -s http://localhost:50050/api/executors | jq
[
  {
    "id": "aa68ce95-4175-404e-a036-db12d7399e79",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1777154502000,
    "specification": {
      "task_slots": 14
    },
    "metrics": [
      {
        "type": "proc_physical_memory",
        "value": 34521088
      },
      {
        "type": "proc_virtual_memory",
        "value": 454171721728
      }
    ]
  }
]
```

```
➜  ~ curl -s http://localhost:50050/api/executor/aa68ce95-4175-404e-a036-db12d7399e79 | jq
{
  "executor_info": {
    "id": "aa68ce95-4175-404e-a036-db12d7399e79",
    "host": "127.0.0.1",
    "port": 50051,
    "last_seen": 1777154562000,
    "specification": {
      "task_slots": 14
    },
    "metrics": [
      {
        "type": "proc_physical_memory",
        "value": 36339712
      },
      {
        "type": "proc_virtual_memory",
        "value": 454207553536
      }
    ]
  },
  "os_info": {
    "system_name": "Darwin",
    "kernel_ver": "Darwin 25.4.0",
    "os_ver": "26.4.1",
    "os_ver_long": "macOS 26.4.1",
    "physical_cores": 14,
    "num_disks": 2,
    "total_disk_space": 1989325168640,
    "total_available_disk_space": 1342051923950,
    "open_files_limit": 10240
  }
}
```

one note, @sandugood peak memory disappeared from this version, not sure if i'm doing something wrong ?

we had a

```
    "peaks": {
      "physical_memory": 0,
      "virtual_memory": 0
    }
```

previously and we said we're going to merge it with metrics adding `peak_physical_memory` and `peak_virtual_memory` ?

_Originally posted by @milenkovicm in https://github.com/apache/datafusion-ballista/issues/1547#issuecomment-4320686016_

martin-g:

Closes #1642.

# Rationale for this change

https://github.com/apache/datafusion-ballista/pull/1547 exposed some extra details about the executors' system and process metrics.
This PR visualize them in the TUI app.

Pressing `Enter` on a selected executor opens a popup with the executor's extra info and some info of its operating system.

The Executors page now shows extra info per executor.


ballista-cli/src/tui/ui/main/executors/executor_details_popup.rs
```diff
@@ -0,0 +1,189 @@
...
+        Span::raw(os.open_files_limit.to_string()),
+    ]));
+
+    lines
+}
+
+fn metric_name(typ: &str) -> &str {
+    match typ {
+        "proc_physical_memory" => "Physical Memory",
+        "proc_virtual_memory" => "Virtual Memory",
```
martin-g: I am not sure whether the `Virtual Memory` is useful to the app user.
Here on MacOS its value is 400+ Gb (available RAM - 36Gb, i.e. more than 10x bigger).
And 2.3Gb on Linux with 64Gb RAM.
I am not sure what exactly it measures.

milenkovicm: maybe we could ignore it until somebody asks for it

milenkovicm:
would suggest moving `Physical Memory` and `Peak Physical Memory`  to main executor screen, maybe after executor id, wdyt @martin-g ?

edit: `Tasks Slots` & `CPU` cores as well

milenkovicm:
three unrelated requests with this PR 

image

can we change format `10.00K/10.00K/ ...` instead of `10.00 K/ 10.00 K/ ...` as we have time units stick with value, it would be great to be consistent 

same in 

image

can we stick K with the number (there is same pattern in task view)?

also, could we get job duration or end time ?

image

if possible, thanks @martin-g, great tool 

one additional note, can we add screenshot of `tui` in main `README.md` (@andygrove  proposed) ?

martin-g:
> can we change format `10.00K/10.00K/ ...` instead of `10.00 K/ 10.00 K/ ...` as we have time units stick with value,

> can we stick K with the number (there is same pattern in task view)?

Done!


> also, could we get job duration or end time ?

Added `Duration` column!

> one additional note, can we add screenshot of `tui` in main `README.md` (@andygrove proposed) ?

I will do it in a separate PR

milenkovicm:
> would suggest moving `Physical Memory` and `Peak Physical Memory` to main executor screen, maybe after executor id, wdyt @martin-g ?
> 
> edit: `Tasks Slots` & `CPU` cores as well

any thought on this one @martin-g ?

martin-g:
> `CPU` cores as well

This is a bit more involved.
At the moment the Executors view uses the `/api/executors` REST API to load its data.
To be able to show the CPU count I will need to make an extra HTTP request to `/api/executor/<id>` for every executor to load its extra details (the OS info).

I see these options:
1) make the extra HTTP requests in the TUI code
2) update the REST API to return all the information with `/api/executors`
3) do nothing, and show the CPU count only in the details popup

milenkovicm:
> > `CPU` cores as well
> 
> This is a bit more involved. At the moment the Executors view uses the `/api/executors` REST API to load its data. To be able to show the CPU count I will need to make an extra HTTP request to `/api/executor/<id>` for every executor to load its extra details (the OS info).
> 
> I see these options:
> 
> 1. make the extra HTTP requests in the TUI code
> 2. update the REST API to return all the information with `/api/executors`
> 3. do nothing, and show the CPU count only in the details popup

we can do it as a follow up

milenkovicm:
image

there is new error

martin-g:
We have discussed this one some time ago. Before it was panicking to install tracing, now it just prints to stderr.

The problem is that the CLI app uses `env_logger` to log to stderr/stdout but since the TUI app paints on the terminal I use the `tracing` crate to log to a file instead.
Neither `log` nor `tracing` have APIs to uninstall the global logger they install :-/

So, we need to start using just one logger for both CLI and TUI.
I will see whether the `tracing` crate provides a way to use a subscriber dynamically - if TUI then log to a file else log to stdout.

milenkovicm (approved):
Thanks @martin-g 

I'd have one suggestion as a follow up, instead of using `centered` for cells using numerical values, use `right` align. I find it easier to to compare neighbouring rows.

Merge when ready

PR #1672, opened 2026-05-09
minor: add pending stage indicator for `AdaptiveDatafusionExec`

milenkovicm:

Closes #.

 # Rationale for this change

Looking at `AdaptiveDatafusionExec` it was hard to understand if stage has been resolved or not.
This change prints pending stage indicator, making behaviour similar to `ExchangeExec`

### DIFF
--- a/ballista/scheduler/src/state/aqe/execution_plan.rs
+++ b/ballista/scheduler/src/state/aqe/execution_plan.rs
@@ -411,12 +411,13 @@ impl DisplayAs for AdaptiveDatafusionExec {
             DisplayFormatType::Default | DisplayFormatType::Verbose => {
                 write!(
                     f,
-                    "AdaptiveDatafusionExec: is_final={:?}, plan_id={}, stage_id={}",
+                    "AdaptiveDatafusionExec: is_final={:?}, plan_id={}, stage_id={}, stage_resolved={}",
                     self.is_final,
                     self.plan_id,
                     self.stage_id()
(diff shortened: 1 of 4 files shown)


ballista/scheduler/src/state/aqe/execution_plan.rs
```diff
@@ -411,12 +411,13 @@ impl DisplayAs for AdaptiveDatafusionExec {
...
                 write!(
                     f,
-                    "AdaptiveDatafusionExec: is_final={:?}, plan_id={}, stage_id={}",
+                    "AdaptiveDatafusionExec: is_final={:?}, plan_id={}, stage_id={}, stage_resolved={}",
                     self.is_final,
                     self.plan_id,
                     self.stage_id()
                         .map(|stage_id| format!("{}", stage_id))
                         .unwrap_or_else(|| "pending".to_string()),
+                    self.shuffle_partitions.lock().is_some()
```
andygrove: nit: looks like there is an existing `self.shuffle_created()` helper that could be used?

andygrove (approved):
LGTM. Thanks @milenkovicm

ballista/scheduler/src/state/aqe/execution_plan.rs
milenkovicm: good catch, 
`self.shuffle_created()` should probably be `self.stage_resolved()`

metegenez (approved):
LGTM as well.

PR #1673, opened 2026-05-09
fix: remove double-counting of write_time in shuffle writer

### LINKED ISSUE #631: `write_time` of ShuffleWriterExec is incorrect
**Describe the bug**
It shows the complete time of the stream, rather than the time the writing takes.

**To Reproduce**
Observe `write_time` metric, it takes always around the time for the tasks to finish.

**Expected behavior**
`write_time` should only record the time it takes to write

**Additional context**

kimjune01:

Closes #631

# Rationale for this change

`write_time` in `ShuffleWriterExec` reports the total stream duration instead of actual write time. The `None` partitioning branch wraps the entire `write_stream_to_disk` call in a timer, but `write_stream_to_disk` already records `write_time` internally per batch. The outer timer double-counts every batch write and also includes time spent waiting for upstream data.

Removed the outer `write_time` timer in the `None` partitioning branch of `ShuffleWriterExec::execute_shuffle_write`. `write_stream_to_disk` handles its own `write_time` measurement per batch, so the outer timer was redundant.

`write_time` now accurately reflects disk write duration rather than total stream processing time.

### DIFF
--- a/ballista/core/src/execution_plans/shuffle_writer.rs
+++ b/ballista/core/src/execution_plans/shuffle_writer.rs
@@ -211,8 +211,6 @@ impl ShuffleWriterExec {
 
             match output_partitioning {
                 None => {
-                    let timer = write_metrics.write_time.timer();
-
                     let path = create_shuffle_path(
                         &self.work_dir,
                         &self.job_id,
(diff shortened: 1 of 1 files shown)


andygrove:
Thanks @kimjune01. I think these changes may have already been made in main branch. Could you rebase this PR?

kimjune01:
You're right — these changes are already in main. Closing. Thanks for the review!

PR #1675, opened 2026-05-09
fix: rest api calculates stage running time correctly

### LINKED ISSUE #1671: How to check the execution time of stages?
How should I check the execution time of each stage in Ballista? I used the REST API method, such as curl http://xxx.xx.xx.xx:50050/api/job/LWlhbrV/stages, and found that the results seem to be inaccurate.
Like 

``
        {
            "stage_id":"5",
            "stage_status":"Successful",
            "input_rows":27787616,
            "output_rows":55575232,
            "elapsed_compute":"16ns"
        }
``

but actually stage 5 took 65 seconds,I calculated it by checking the scheduler logs for the scheduling time of task 0 in stage 5 and the completion time of stage 5.

milenkovicm:

Closes #1671

 # Rationale for this change

### DIFF
--- a/ballista/scheduler/src/api/handlers.rs
+++ b/ballista/scheduler/src/api/handlers.rs
@@ -725,34 +725,31 @@ fn format_job_status(status: &Option<Status>, elapsed_ms: u64) -> (String, Strin
 fn get_running_stage_time(task_infos: &[Option<TaskInfo>]) -> String {
     let min_start = task_infos
         .iter()
-        .map(|t| {
-            t.as_ref()
-                .map(|t| t.start_exec_time)
-                .unwrap_or_else(|| get_current_time())
-        })
+        .flat_map(|t| t.as_ref().map(|t| t.start_exec_time))
         .min();
 
-    let max_end = task_infos
-        .iter()
-        .map(|t| {
-            t.as_ref()
-                .map(|t| t.end_exec_time)
-                .unwrap_or_else(|| get_current_time())
-        })
-        .max();
-    let elapsed_ms = match (min_start, max_end) {
-        (Some(start), Some(end)) if end > start => end - start,
+    let elapsed_ms = match (min_start, get_current_time()) {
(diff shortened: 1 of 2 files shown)


ballista/scheduler/src/api/handlers.rs
```diff
@@ -539,11 +541,9 @@ pub async fn get_query_stages<
                             .as_ref()
                             .map(|m| get_combined_count(m.as_slice(), "output_rows"))
                             .unwrap_or(0);
-                        summary.elapsed_compute = running_stage
-                            .stage_metrics
-                            .as_ref()
-                            .map(|m| get_elapsed_compute_nanos(m.as_slice()))
-                            .unwrap_or_default();
+                        summary.elapsed_compute = get_running_stage_time(&running_stage
```
metegenez, line 544: maybe making this elapsed_wall_time would make more sense?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -722,19 +722,43 @@ fn format_job_status(status: &Option<Status>, elapsed_ms: u64) -> (String, Strin
     }
 }
 
-fn get_elapsed_compute_nanos(metrics: &[MetricsSet]) -> String {
-    let nanos: usize = metrics
+fn get_running_stage_time(task_infos: &[Option<TaskInfo>], current_time: u128) -> String {
```
metegenez, line 725: Should this also filter out `.filter(|t| *t > 0)` zeros?

metegenez (approved):
LGTM.

ballista/scheduler/src/api/handlers.rs
milenkovicm: i agree, will do it as a follow up as `tui` will be impacted

ballista/scheduler/src/api/handlers.rs
milenkovicm: I agree

PR #1647, opened 2026-05-02
feat(scheduler): broadcast-style hash join for small-side joins

andygrove:
# Which issue does this PR close?

Closes #.

# Rationale for this change

Today, when DataFusion plans a hash join with a small side, Ballista's distributed planner relies on `CoalescePartitionsExec` to satisfy `CollectLeft`'s single-partition requirement. That collapses build-side parallelism onto a single executor and creates a hot spot for both producing and serving the build shuffle file. To work around this, Ballista's default session sets DataFusion's `hash_join_single_partition_threshold` to 0, which avoids the hot spot but also gives up small-side optimization entirely.

This PR adds a true broadcast-style hash join. For small-side joins, the build subtree keeps its natural M-partition layout and writes M shuffle files; each downstream join task fetches all M files in parallel and builds its hash table locally. Build data flows executor-to-executor; the scheduler is not on the data path.

# What changes are included in this PR?

- New session config `ballista.optimizer.broadcast_join_threshold_bytes` (default 10 MB, `0` disables).
- `UnresolvedShuffleExec` and `ShuffleReaderExec` get a `broadcast` flag and an explicit `upstream_partition_count`. Proto/codec round-trip the new fields.
- `ShuffleReaderExec::execute` and `partition_statistics` aggregate across all upstream locations when broadcast.
- `DefaultDistributedPlanner` adds:
  - `maybe_promote_to_broadcast` that promotes eligible `HashJoinExec(Partitioned)` to `CollectLeft` when one side is under the threshold.
  - A `HashJoinExec(CollectLeft)` lowering branch that strips `CoalescePartitionsExec` over the build, writes M-partition build shuffle, and inserts a broadcast `UnresolvedShuffleExec`.
  - A broadcast branch in `remove_unresolved_shuffles` that flattens upstream locations.
  - A broadcast branch in `rollback_resolved_shuffles` that preserves `broadcast` and `upstream_partition_count` on retry.
- Unit tests in the scheduler planner (broadcast on / off, M>1 build).
- Codec round-trip tests for the broadcast variants.
- End-to-end correctness check in `examples/`.

AQE is deferred: the `AdaptivePlanner` does not yet apply this lowering. The `hash_join_single_partition_threshold` zero workaround is left in place.

# Are there any user-facing changes?

[...]

Issue #1055, opened 2024-09-15
Left/full outer join incorrect 


Dandandan:
**Describe the bug**
See discussion here https://github.com/apache/datafusion/issues/12454

The "broadcast join" (CollectLeft) is wrong for certain join types which produce results on unmatched left rows.

**To Reproduce**
Run a broadcast join with left / full outer on more than one node

**Expected behavior**

**Additional context**

milenkovicm:
should we take this once it gets merged in DF ?

milenkovicm:
#1647 fixed this issue

Dandandan:
I don't fully understand how https://github.com/apache/datafusion-ballista/pull/1647 would fix the issue? 🤔

andygrove:
It looks like this issue was closed prematurely. The PR that was merged does state that "AQE is deferred". Let's reopen this issue.

milenkovicm:
I apologise, I mistakenly thought this has been fixated

PR #1676, opened 2026-05-11
feat: Pinot-style colocated-join optimizer for hash-bucketed tables

### LINKED ISSUE #1677: Pinot-style colocated-join optimizer for hash-bucketed tables
## Problem

Ballista shuffles data on every join whose keys don't already match the input partitioning, because the planner has no way to learn that a table is hash-bucketed by some column. For workloads on pre-bucketed data (e.g. Iceberg / Parquet datasets where producers already hash-partition by the join key), the network shuffle is pure overhead: the data is *already* co-located, the planner just doesn't know it.

Concretely:
- `BallistaQueryPlanner` ships the logical plan to the scheduler.
- `AdaptivePlanner` runs DataFusion's physical planner + a custom rule chain ending in `DistributedExchangeRule`, which inserts an `ExchangeExec` above every `RepartitionExec(Hash)`.
- `DefaultDistributedPlanner` cuts a stage at every shuffle.
- The only join-aware rule today is `JoinSelection`, which only swaps build sides based on row/byte stats — it has no notion of partition co-location.
- The catalog has no concept of hash bucketing; only `ListingTable` directory partitions exist.

## Proposed solution

Bring three Pinot V2 Physical Optimizer ideas to Ballista (Pinot's `RelToPRelConverter` → `WorkerExchangeAssignmentRule`):

1. **Colocated joins** — when both join inputs declare matching hash bucketing on the join keys, skip the shuffle entirely and let the join's required hash distribution be satisfied directly by the inputs.
2. **Sub-partitioning** for divisor bucket counts (e.g. 16 vs 8) — locally coalesce the larger side instead of shuffling, since `(hash(k) % 16) % 8 == hash(k) % 8`.
3. **Small-side broadcast** in the AQE path — when one side fits under a configured byte threshold, promote a `Partitioned` `HashJoinExec` to `CollectLeft`. (This complements PR #1647, which added the same lowering to the non-AQE `DefaultDistributedPlanner`; the AQE planner has no equivalent today, per the TODO at `ballista/scheduler/src/state/aqe/mod.rs`.)

## Architecture

A small metadata trait in `ballista-core` lets any `TableProvider` declare on-disk hash bucketing without a DataFusion fork. Two new `PhysicalOptimizerRule`s slot into the existing `AdaptivePlanner` rule chain just before `DistributedExchangeRule`:

```
DataFusion physical planner
  → DataFusion optimizer rules (incl. EnforceDistribution, JoinSelection)
  → ColocatedJoinRule           ← strip redundant RepartitionExec
                                   (also handles divisor sub-partitioning)
  → BroadcastSmallSideRule      ← promote Partitioned join to CollectLeft
  → DistributedExchangeRule     ← existing: maps remaining repartitions to ExchangeExec
  → DefaultDistributedPlanner   ← existing: cuts stages at ExchangeExec
```

## Goals

- Let users declare per-table hash partitioning so the optimizer can reason about it.
- Skip the shuffle when both join inputs are already hash-partitioned on the join keys with matching bucket counts.
- When bucket counts differ but have a divisor relationship, use a sub-partitioning exchange that keeps all data local.
- When one side is below a configurable size threshold, broadcast it instead of repartitioning the other side — for the AQE code path that PR #1647 explicitly left as a TODO.

## Non-goals

- A full PRel-style multi-pass optimizer rewrite (reuse DataFusion's `Partitioning` + `EquivalenceProperties` instead of inventing a `BallistaDataDistribution`).
- Custom DDL grammar — piggyback on DataFusion's existing `OPTIONS (...)` clause and a `PartitionedTableProvider` wrapper.
- Iceberg / Delta integration (design the trait so they can plug in later).
- Rewriting `JoinSelection`'s cost model.

## Implementation

PR #1676.

## Follow-up

#1679 — extend `ColocatedJoinRule` and `BroadcastSmallSideRule` to `SortMergeJoinExec`, since PR #1651 made sort-merge the default join.

wirybeaver:
Closes #1677. Follow-up tracked in #1679.

## Summary

Implements the colocated-join optimizer described in #1677: a small metadata trait in `ballista-core` lets any `TableProvider` declare on-disk hash bucketing, and two new `PhysicalOptimizerRule`s slot into the existing `AdaptivePlanner` rule chain just before `DistributedExchangeRule`. Together they cover three cases — colocated joins (skip the shuffle), divisor sub-partitioning (16 vs 8 buckets, locally coalesce instead of shuffle), and small-side broadcast in the AQE code path.

```
DataFusion physical planner
  → DataFusion optimizer rules (incl. EnforceDistribution, JoinSelection)
  → ColocatedJoinRule           ← new: strip redundant RepartitionExec
                                       (also handles divisor sub-partitioning)
  → BroadcastSmallSideRule      ← new: replace partitioned join with CollectLeft
  → DistributedExchangeRule     ← existing: maps remaining repartitions to ExchangeExec
  → DefaultDistributedPlanner   ← existing: cuts stages at ExchangeExec
```

**Metadata layer (`ballista-core`)**: `BallistaPartitionMetadata` is an optional contract any `TableProvider` can implement to declare on-disk hash bucketing (keys + hash function + bucket count). `PartitionedTableProvider` is a wrapper that attaches a `HashDistribution` to any existing provider without modifying it, and `HashDistributedScanExec` re-advertises the wrapped scan's `output_partitioning()` as `Partitioning::Hash` so optimizer rules can read it. `BucketSubPartitionExec` chains input partitions per output partition via `stream::iter` + `flatten` — pure local concat, no network — for the divisor case (relies on the identity `(hash(k) % 16) % 8 == hash(k) % 8`).

**Optimizer rules (`ballista-scheduler`)**: `ColocatedJoinRule` walks each `HashJoinExec` and, when both inputs declare matching keys + hash function and either equal or divisor-related bucket counts, strips the `RepartitionExec` above each input or wraps the larger side in `BucketSubPartitionExec`. `BroadcastSmallSideRule` promotes a `Partitioned` `HashJoinExec` to `CollectLeft` when one side's `total_byte_size` is below the configured threshold, using `HashJoinExec::swap_inputs` when the small side is on the right. It is restricted to `JoinType::Inner` pending #1055. `default_optimizers` now takes the `SessionConfig` so the broadcast threshold flows through, and `ColocatedJoinRule` runs before `BroadcastSmallSideRule` so colocation wins when both could apply.

**Relationship to PR #1647**: that PR added small-side broadcast lowering to the non-AQE `DefaultDistributedPlanner::plan_query_stages_internal`, and the implementation explicitly noted (TODO at `state/aqe/mod.rs`) that the same lowering does not fire in the AQE path. `BroadcastSmallSideRule` is the AQE-side counterpart that resolves that TODO; this PR removes the resolved comment. The two rules cover disjoint code paths (AQE is opt-in via `ballista.planner.adaptive.enabled`, default false), share the same `BALLISTA_BROADCAST_JOIN_THRESHOLD_BYTES` config key, and behave identically.

**Both features are opt-in.** Tables only get colocation behavior when the user wraps them with `PartitionedTableProvider`. `BroadcastSmallSideRule::from_session_config` returns a no-op rule (threshold = 0) when the `BallistaConfig` extension is not registered on the `SessionConfig`, so existing tests using plain `SessionConfig::new()` are unaffected.

## Test plan

- [x] `cargo test -p ballista-core` — 92 passed (trait, wrapper, scan adapter, `BucketSubPartitionExec` correctness + rejection of non-divisor inputs)
- [x] `cargo test -p ballista-scheduler` — 108 passed (5 `ColocatedJoinRule` cases, 4 divisor cases, 5 `BroadcastSmallSideRule` cases, 3 end-to-end plan-snapshot tests, plus all pre-existing)
- [x] `cargo check -p ballista-scheduler --tests` — clean
- [x] End-to-end plan snapshots cover three scenarios: matching bucket counts → no `ExchangeExec`; 8/4 divisor → `BucketSubPartitionExec(out=4, factor=2)`, no shuffle; plain `MemTable` → `ExchangeExec` retained, optimizer is silent
- [ ] TPC-H Q5 against bucketed customer/orders/lineitem — environment-dependent, planned as a follow-up benchmark validation

## Caveats / risks

- **Sort-merge is now the default join** (#1648 / PR #1651). Both rules currently match `HashJoinExec` only, so under default settings the rules don't fire — users opt back into hash join via `SET datafusion.optimizer.prefer_hash_join = true`. Extending the rules to `SortMergeJoinExec` is tracked in #1679.
- **Bucket-to-file alignment** is by filename convention in `PartitionedTableProvider`; documented but no manifest format yet — a follow-up could add one.
- **Wrong declaration**: trust + document. If a user declares `BUCKETS=N` but data isn't actually bucketed that way, results may be wrong. We chose trust over verify to keep the metadata layer lightweight.
- **AQE replan loop** re-runs the optimizer after each stage. All three new rules are stateless tree rewrites, so they're idempotent.

### DIFF
--- a/ballista/core/src/lib.rs
+++ b/ballista/core/src/lib.rs
@@ -50,6 +50,8 @@ pub mod extension;
 #[cfg(feature = "build-binary")]
 /// Object store configuration and utilities for distributed file access.
 pub mod object_store;
+/// Hash-partition (bucketing) metadata for distributed query optimization.
+pub mod partitioning;
 /// Query planning utilities for distributed execution.
 pub mod planner;
 /// Runtime registry for codec and function registration.
--- a/ballista/core/src/partitioning.rs
+++ b/ballista/core/src/partitioning.rs
@@ -1,0 +1,644 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Hash-partition (bucketing) metadata for distributed query optimization.
+//!
+//! Pinot-style colocated joins, sub-partitioning, and small-side broadcast
+//! all need to know whether a table's data is already hash-bucketed by some
+//! column. This module defines the contract by which a [`TableProvider`] can
+//! declare that bucketing, plus a thin wrapper that lets users attach the
+//! metadata to any existing provider without writing a custom one.
+
+use std::any::Any;
+use std::fmt;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::catalog::{Session, TableProvider};
+use datafusion::common::{Constraints, Statistics};
+use datafusion::error::{DataFusionError, Result};
+use datafusion::execution::context::TaskContext;
+use datafusion::logical_expr::{Expr, TableProviderFilterPushDown, TableType};
+use datafusion::physical_expr::EquivalenceProperties;
+use datafusion::physical_expr::expressions::Column;
+use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion::physical_plan::{
+    DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, PhysicalExpr,
+    PlanProperties, SendableRecordBatchStream,
+};
+use futures::StreamExt;
+
+/// Hash function used to bucket a table's rows on disk.
+///
+/// The optimizer uses this to verify that two co-located inputs were bucketed
+/// by the same function before eliding a shuffle. Crucially, it does **not**
+/// have to match DataFusion's internal `RepartitionExec` hasher — the
+/// declaration is a promise about the on-disk layout, and the colocated-join
+/// rule trusts it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum HashFn {
+    /// Murmur3 (32-bit), as used by Spark/Hive bucketing.
+    Murmur3,
+    /// xxHash64.
+    XxHash64,
+}
+
+impl fmt::Display for HashFn {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            HashFn::Murmur3 => write!(f, "murmur3"),
+            HashFn::XxHash64 => write!(f, "xxhash64"),
+        }
+    }
+}
+
+/// Declares that a table is hash-bucketed across `num_buckets` partitions on
+/// the columns named in `keys`, using `hash_fn`.
+///
+/// Two tables are *co-located* for a join when they share the same `keys`
+/// (matching join keys positionally), the same `hash_fn`, and the same
+/// `num_buckets` — bucket *k* of one matches bucket *k* of the other.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct HashDistribution {
+    /// Column names used as partition keys.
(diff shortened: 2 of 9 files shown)


metegenez:
hi @wirybeaver , I am working on coalesce partition into target binary sizes, your work looks good, just can you verify against sf10 or sf100, let's check if they compile well and there is no big regression? It d help reviewers to assess.

wirybeaver:
> hi @wirybeaver , I am working on coalesce partition into target binary sizes, your work looks good, just can you verify against sf10 or sf100, let's check if they compile well and there is no big regression? It d help reviewers to assess.

Hi @metegenez, thanks for the contribution to coalesce partition. Yeah, I will verify with sf10 and sf100 and paste the result. Recently I was occupied by other tasks. I recall there're a few AQE rules were added recently around join and coalesce. I will merge those changes.

milenkovicm:
closing as open for too long without progress, please reopen if still needed and updated

PR #1649, opened 2026-05-03
feat(aqe): Lazy stage evaluation in AQE

milenkovicm:
# Which issue does this PR close?

Mentioned in in https://github.com/apache/datafusion-ballista/pull/1647#issuecomment-4365479608 

 # Rationale for this change

AQE planner eagerly splits plan into stages:

```
AdaptiveDatafusionExec: is_final=false, plan_id=4, stage_id=pending
      ProjectionExec: expr=[sum(t0.c0)@1 as sum(t0.c0)]
        AggregateExec: mode=FinalPartitioned, gby=[c0@0 as c0], aggr=[sum(t0.c0)]
          ExchangeExec: partitioning=Hash([c0@0], 2), plan_id=3, stage_id=pending, stage_resolved=false
            AggregateExec: mode=Partial, gby=[c0@0 as c0], aggr=[sum(t0.c0)]
              HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(p2@0, c2@1)], projection=[c0@1]
                CoalescePartitionsExec
                  ExchangeExec: partitioning=None, plan_id=1, stage_id=pending, stage_resolved=false
                    ProjectionExec: expr=[c@0 as p2]
                      AggregateExec: mode=FinalPartitioned, gby=[c@0 as c], aggr=[]
                        ExchangeExec: partitioning=Hash([c@0], 2), plan_id=0, stage_id=pending, stage_resolved=false
                          AggregateExec: mode=Partial, gby=[c@0 as c], aggr=[]
                            DataSourceExec: partitions=1, partition_sizes=[1]
                ProjectionExec: expr=[min(t.a)@1 as c0, c@0 as c2]
                  AggregateExec: mode=FinalPartitioned, gby=[c@0 as c], aggr=[min(t.a)]
                    ExchangeExec: partitioning=Hash([c@0], 2), plan_id=2, stage_id=pending, stage_resolved=false
                      AggregateExec: mode=Partial, gby=[c@1 as c], aggr=[min(t.a)]
                        DataSourceExec: partitions=1, partition_sizes=[1]
```

plan exchange/execution is predefined based on little to no information, which makes re-planing a bit harder (if not impossible). This might have implications on broadcasting, partition coalescing .... 

This PR changes AQE to plan one step at the time:

```
AdaptiveDatafusionExec: is_final=false, plan_id=2, stage_id=pending
      ProjectionExec: expr=[sum(t0.c0)@1 as sum(t0.c0)]
[...]

PR #1684, opened 2026-05-11
feat(aqe): CoalescePartitionsRule — shuffle-partition coalescing on resolved stats

metegenez:
Adds a new AQE physical-optimizer rule, **`CoalescePartitionsRule`**, that runs once per stage as the stage becomes runnable. It reads the per-partition byte stats published by upstream stages, runs Spark's `ShufflePartitionsUtil` bin-pack against an advisory target, and attaches a shared `CoalescePlan` to every leaf `ExchangeExec` feeding the stage. The adapter consumes that decision and builds the stage's `ShuffleReaderExec` with `K < M` partitions instead of `M`.

Built alongside @milenkovicm's lazy AQE (#1649, merged 2026-05-09) and orthogonal to @andygrove's broadcast hash join (#1647, merged). All three use DataFusion's standard `PhysicalOptimizerRule` trait; no new traits, no upstream DataFusion changes, no executor data-path changes.

## Surface

Opt-in via a single boolean. **`ballista.coalesce.enabled=false`** is the default — the rule short-circuits, the plan flows through untouched, parallelism is preserved. Users who want fewer/larger downstream tasks turn the rule on; this packs toward `ballista.coalesce.target_partition_bytes` (64 MB by default — Spark's `advisoryPartitionSizeInBytes` analog).

| `coalesce.enabled` | behavior | Spark analog |
|---|---|---|
| `false` (default) | no-op, parallelism preserved | `parallelismFirst=true` outcome |
| `true` (opt-in) | pack toward advisory target | `parallelismFirst=false` |

## Where the rule fires — per stage, not globally

The rule is invoked inside `actionable_stages()` right before each newly-runnable stage is handed to `BallistaAdapter`. Each call sees **one stage's plan as its root** (an `ExchangeExec` or `AdaptiveDatafusionExec`), descends, and treats every leaf `ExchangeExec` it finds as the stage's alignment group.

It is **not** registered in `default_optimizers()`. That chain runs on the entire residual plan tree on every `replan_stages()` pass, which would cause two problems we hit during validation:

1. **Cross-stage alignment groups.** The walker would collect "topmost ExchangeExecs", which spans multiple future stages' inputs. A K decision driven by stage N's byte distribution gets stamped onto stage N+1's inputs too — they're not in the same join, so they shouldn't share K.
2. **Stale state across fires.** `set_coalesce` uses interior mutability on a shared `Arc<ExchangeExec>`. A later fire that bails on K=1 degenerate doesn't invalidate the K=5 a previous fire wrote on the same exchange. The next stage then sees mismatched leg Ks (`SortMergeJoinExec: left:16 right:5 → assertion fails`).

Per-stage invocation fixes both: each call's leaves are scoped to one join's inputs, and the K decision is local to that stage. The ''bail on unresolved leaf'' path becomes unreachable by construction — a stage becomes runnable precisely when all its inputs are resolved.

## Algorithm

For each per-stage call:

1. Collect every leaf `ExchangeExec` in this stage's subtree — the alignment group. All members share the upstream partition count `M` and were hash-partitioned on the same join key upstream, so partition `i` means the same logical hash bucket on every leaf.
2. Sum per-partition byte stats element-wise across the group.
3. Bin-pack the summed sizes via Spark's `ShufflePartitionsUtil` algorithm (`split_size_list_by_target_size`, with small/merged factors).
4. If `1 < K < M`, attach the *same* `CoalescePlan` (with K partition groups) to every leaf via `set_coalesce(..)`.

Sharing one `CoalescePlan` across the whole alignment group is what makes the rule safe for hash-Partitioned joins and SortMergeJoin: hash buckets that were aligned at M stay aligned at K, so DataFusion's `EnforceDistribution` doesn't insert remediation repartitions.

## Plan-tree examples (from real TPC-H Q8 at SF=100)

**Stage 11 — coalesce attaches, both join legs aligned at K=5:**

```
AdaptiveDatafusionExec
  ProjectionExec
    SortMergeJoinExec: on=[(o_custkey@3, c_custkey@0)]
      SortExec: [o_custkey@3 ASC]
        ExchangeExec: Hash([o_custkey@3], 16) plan_id=11 resolved=true coalesce=5 of 16
          ...stage 10 writer output (~19 MB × 16 partitions)...
      SortExec: [c_custkey@0 ASC]
        ExchangeExec: Hash([c_custkey@0], 16) plan_id=4 resolved=true coalesce=5 of 16
          ...customer table scan...
```

Rule's debug trace for this fire:
```
[coalesce-rule] collected 2 leaf ExchangeExec(s)
  leaf: plan_id=11 M=16 resolved=true existing_coalesce=None
  leaf: plan_id=4  M=16 resolved=true existing_coalesce=None
[coalesce-rule] bin-pack result: K=5 M=16
[coalesce-rule] set_coalesce(K=5) on plan_id=11
[coalesce-rule] set_coalesce(K=5) on plan_id=4
```

Both legs get the SAME `CoalescePlan` → SMJ runs with 5 partitions per side, hash buckets stay aligned.

**Stage 12 — K=1 degenerate, rule no-ops, no spurious coalesce:**

```
SortMergeJoinExec: on=[(c_nationkey@4, n_nationkey@0)]
  ExchangeExec: Hash([c_nationkey@4], 16) plan_id=12 resolved=true        <- no coalesce
  ExchangeExec: Hash([n_nationkey@0], 16) plan_id=5  resolved=true        <- no coalesce
```

Rule's debug trace:
```
[coalesce-rule] collected 4 leaf ExchangeExec(s)
  leaf: plan_id=12 M=16 resolved=true existing_coalesce=None
  leaf: plan_id=5  M=16 resolved=true existing_coalesce=None
  leaf: plan_id=6  M=16 resolved=true existing_coalesce=None
  leaf: plan_id=7  M=16 resolved=true existing_coalesce=None
[coalesce-rule] summed bytes per upstream partition: total ~76 MB
[coalesce-rule] bin-pack result: K=1 M=16
[coalesce-rule] K degenerate; bail without setting coalesce
```

All four legs retain native M=16. The SMJ runs at full parallelism. No partition-count mismatch.

## Components

| Area | Change |
|---|---|
| **Rule** | `CoalescePartitionsRule` in `state/aqe/optimizer_rule/coalesce_partitions.rs`. Unit struct. Invoked per-stage in `AdaptivePlanner::actionable_stages()`. |
| **Carrier** | `ExchangeExec` gains an `Arc<Mutex<Option<Arc<CoalescePlan>>>>` slot with `set_coalesce` / `coalesce` accessors. `coalesce=K of M` annotation appears in `DisplayAs` output only when attached. |
| **Adapter** | `BallistaAdapter::adapt_to_ballista` checks `exchange.coalesce()`; on `Some(cp)` it builds a K-partition reader via `ShuffleReaderExec::try_new_coalesced` instead of the default `try_new`. |
| **Reader** | `ShuffleReaderExec::try_new_coalesced` constructs the K-shape pre-concatenated reader; legacy `try_new` path is byte-for-byte unchanged when no `CoalescePlan` is attached. Coexists with `try_new_broadcast` from PR #1647 (field numbers 7/8 on the proto, not clashing with 5/6). |
| **Algorithm** | Pure-CPU bin-pack helpers in `state/aqe/coalesce/algorithm.rs` — `split_size_list_by_target_size` (verbatim Rust port of Spark's `splitSizeListByTargetSize`, including merged-factor early flush and small-tail post-loop merge) + `start_indices_to_partition_groups`. |
| **Config** | New keys: `ballista.coalesce.enabled` (default `false`), `target_partition_bytes` (default 64 MB), `small_partition_factor` (default 0.2), `merged_partition_factor` (default 1.2). |
| **Proto** | `CoalescePlan` + `PartitionGroup` round-trip via `BallistaPhysicalExtensionCodec`. |

## Test plan

- [x] `cargo test --workspace --no-fail-fast` — ~600 tests pass, 0 failures.
- [x] `cargo clippy --workspace --all-targets --tests` — 0 warnings.
- [x] 7 functional tests in `state::aqe::test::coalesce_rule` snapshot displayed plan trees and assert on the `coalesce=K of M` field for: happy path (`K=2`), disabled rule (`coalesce=none`), K=M degenerate (`coalesce=none`), partitioned hash join with both legs aligned, two stacked hash joins with three legs aligned, sort-merge join, and post-finalize `ShuffleReaderExec` showing the coalesced K baked into the runnable plan.
- [x] Algorithm-level `#[test]` cases for bin-pack edge cases (small-tail folding, merged-factor early flush, strict-overshoot boundary).
- [x] **TPC-H SF=100 sanity, 22 queries × 2 join variants** (`prefer_hash_join=true/false`), `coalesce.enabled=true`: **44/44 rc=0**, row counts identical across variants, 0 panics, 0 ERROR-level lines in scheduler or executor logs. 9 of 22 queries get measurable K reductions: Q2, Q5, Q7, Q8, Q9, Q10, Q11, Q15, Q16, Q21 — observed Ks of 2, 3, 5, 8, 12, 13 across various join shapes.


ballista/core/proto/ballista.proto
```diff
@@ -88,13 +90,30 @@ message ShuffleReaderExecNode {
   datafusion.Partitioning partitioning = 4;
   bool broadcast = 5;
   uint32 upstream_partition_count = 6;
+  // Optional coalesce metadata. Absent means "no coalesce" (legacy one-to-one read behavior).
+  optional CoalescePlan coalesce = 7;
```
metegenez, line 94: I am planning to make this field cover broadcast as well, but i didnt want to make the scope bigger than it should be.

milenkovicm:
thanks @metegenez will review this PR asap

ballista/core/src/config.rs
```diff
@@ -86,6 +86,21 @@ pub const BALLISTA_SHUFFLE_SORT_BASED_MEMORY_LIMIT_PER_TASK_BYTES: &str =
 pub const BALLISTA_BROADCAST_JOIN_THRESHOLD_BYTES: &str =
     "ballista.optimizer.broadcast_join_threshold_bytes";
 
+/// Configuration key to enable AQE coalesce-shuffle-partitions rule.
+/// Disabled by default — opt in when the workload benefits from larger
+/// downstream tasks more than from preserved parallelism.
+pub const BALLISTA_COALESCE_ENABLED: &str = "ballista.coalesce.enabled";
```
milenkovicm: can we prefix configuration values with `ballista.planner.` (`ballista.planner.coalesce...`)?

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -296,8 +298,13 @@ impl AdaptivePlanner {
...
                     .map(|plan| {
-                        // TODO: we need to find input stages for given stage
-                        //       thus result should change
+                        // Run the coalesce rule per-stage: the root of `plan` is
+                        // the stage's wrapper exchange, so the rule's walker sees
+                        // only this stage's input exchanges as the alignment
+                        // group. This avoids cross-stage gluing and stale state
+                        // that would arise if the rule walked the entire residual
+                        // plan in `default_optimizers()`.
+                        let plan = CoalescePartitionsRule.optimize(plan, config)?;
```
milenkovicm, line 307: Current AQE approach with centralised rules execution has its downsides, it probably bends use of rules in a way which is not really intended and making rules overcomplicated. 

We had an issue before where running a rule detached from other rules creates invalid plan as rules which could "fix" the plan do not run, you probably remember.

A corner case I see is when coalesce partition produce single partition output. In that case rest of the plan could probably be re-optimised due to fact there is single partition now. We would miss that opertunity with this approach, as the fact there is single partition may be catched in next "full" optimizer pass (next stage generatio)

having said that, at the moment i dont have better proposal

ballista/core/src/config.rs
```diff
@@ -383,6 +429,36 @@ impl BallistaConfig {
...
+    pub fn coalesce_target_partition_bytes(&self) -> u64 {
+        self.get_usize_setting(BALLISTA_COALESCE_TARGET_PARTITION_BYTES) as u64
+    }
+
+    /// Returns the small-partition merge factor (Spark legacy).
+    ///
+    /// Stored as Utf8 in CONFIG_ENTRIES because BallistaConfig::parse_value does not
+    /// support DataType::Float64. Falls back to 0.2 if the stored string fails to parse.
+    pub fn coalesce_small_partition_factor(&self) -> f64 {
+        self.get_string_setting(BALLISTA_COALESCE_SMALL_PARTITION_FACTOR)
```
milenkovicm: maybe we should extract get_flot_setting?

ballista/core/src/config.rs
```diff
@@ -178,7 +193,38 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
...
+            DataType::UInt64,
+            Some((64 * 1024 * 1024_usize).to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_COALESCE_SMALL_PARTITION_FACTOR.to_string(),
+            "Small-partition merge factor (Spark legacy). Stored as Utf8 because \
+             BallistaConfig::parse_value does not support Float64; parsed back via \
+             f64::from_str in the accessor."
+                .to_string(),
+            DataType::Utf8,
```
milenkovicm: should it be float?

ballista/core/src/config.rs
```diff
@@ -178,7 +193,38 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
...
+            Some("0.2".to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_COALESCE_MERGED_PARTITION_FACTOR.to_string(),
+            "Merged-partition early-flush factor (Spark legacy). Stored as Utf8 — \
+             see small_partition_factor for rationale."
+                .to_string(),
+            DataType::Utf8,
```
milenkovicm: should it be float

ballista/core/src/config.rs
milenkovicm: probably not

ballista/scheduler/src/state/aqe/optimizer_rule/coalesce_partitions.rs
```diff
@@ -0,0 +1,291 @@
...
+//!      `ShuffleReaderExec`s.
+//!
+//! # Carrier semantics
+//!
+//! The `CoalescePlan` lives on the upstream `ExchangeExec`; the rule does
+//! not rewrite the plan tree. Idempotency is structural — `set_coalesce`
+//! overwrites the slot with an equivalent plan on re-entry, and the
+//! bin-pack is a pure function of the resolved byte sizes, so the second
+//! pass produces the same decision.
+//!
```
milenkovicm, line 92: maybe we should note that it will binpack neighboring partitions only

milenkovicm (commented):
thanks @metegenez 
i think this is good addition to AQE, it will help us moving things, few comments and one note

 running TPCH SF1 Q22 i got 

```

thread 'tokio-rt-worker' (3376074) panicked at ballista/scheduler/src/state/aqe/optimizer_rule/coalesce_partitions.rs:239:23:
index out of bounds: the len is 14 but the index is 14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ballista/scheduler/src/state/aqe/planner.rs
metegenez: Inner stage optimisation will have its limits, for sure. Depending on the follow up broadcast understanding, I can move this into a general rule. I believe we will add more inner stage decisions at some point.

metegenez:
> thanks @metegenez i think this is good addition to AQE, it will help us moving things, few comments and one note
> 
> running TPCH SF1 Q22 i got
> 
> ```
> 
> thread 'tokio-rt-worker' (3376074) panicked at ballista/scheduler/src/state/aqe/optimizer_rule/coalesce_partitions.rs:239:23:
> index out of bounds: the len is 14 but the index is 14
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

I couldnt reproduce it by running 
```
pkill -9 -f "ballista-(scheduler|executor)" 2>/dev/null; sleep 2
rm -rf /tmp/ballista-work && mkdir -p /tmp/ballista-work
rm -f /tmp/sch.log /tmp/exec.log
```
```
RUST_LOG="ballista=info,ballista_scheduler::state::aqe::optimizer_rule=debug" \
    ./target/release/ballista-scheduler --bind-port 50150 --bind-host 0.0.0.0 &>
  /tmp/sch.log &
```
```
RUST_LOG=ballista=info ./target/release/ballista-executor --bind-port 50151
  --bind-host 0.0.0.0 \
  --scheduler-host localhost --scheduler-port 50150 \
  --concurrent-tasks 8 --memory-pool-size 4GB \
  --work-dir /tmp/ballista-work &> /tmp/exec.log &
```
  Trials 
Hash Join preferred,
```
./target/release/tpch benchmark ballista --host localhost --port 50150   --path benchmarks/data --format parquet --partitions 16 --iterations 1 -c   "ballista.planner.adaptive.enabled=true" -c "ballista.coalesce.enabled=true" -c   "datafusion.optimizer.prefer_hash_join=true"
   ```
and not preferred,

  ```
./target/release/tpch benchmark ballista --host localhost --port 50150   --path benchmarks/data --format parquet --partitions 16 --iterations 1 -c   "ballista.planner.adaptive.enabled=true" -c "ballista.coalesce.enabled=true" -c   "datafusion.optimizer.prefer_hash_join=false"
```

What was your config on running this?

milenkovicm:
```
cargo run --bin tpch -- benchmark ballista -p /Users/.../tpch-data-sf1/ -f parquet -i 1 --port 50050 --host 127.0.0.1 -c datafusion.execution.target_partitions=48
```

havent change any of the configuration

milenkovicm:
also:

- `ballista.planner.coalesce.enabled=true`
- `ballista.planner.adaptive.enabled=true`

ballista/core/src/serde/scheduler/mod.rs
```diff
@@ -270,7 +270,10 @@ pub struct ExecutorDataChange {
 pub struct PartitionStats {
     pub(crate) num_rows: Option<u64>,
     pub(crate) num_batches: Option<u64>,
-    pub(crate) num_bytes: Option<u64>,
+    /// Per-partition byte size reported by the shuffle writer. Read by the
+    /// AQE coalesce rule (in `ballista-scheduler`) to bin-pack alignment
+    /// groups, so this field is `pub` rather than `pub(crate)`.
+    pub num_bytes: Option<u64>,
```
milenkovicm: Nit, there is public getter method, do we need this property public?

ballista/scheduler/src/state/aqe/adapter.rs
```diff
@@ -57,10 +58,54 @@ impl BallistaAdapter {
...
+                None => ShuffleReaderExec::try_new(
+                    stage_id,
+                    partitions,
+                    schema,
+                    partitioning,
+                )?,
+            };
+
+            Ok(Transformed::yes(Arc::new(reader)))
+        } else if let Some(reader) = plan.as_any().downcast_ref::<ShuffleReaderExec>() {
```
milenkovicm: I don't see this case in the rule, maybe we should remove this branch

milenkovicm (approved):
Thanks @metegenez few minor comments.

Let's merge this, we can iterate on it.

I have few ideas about broadcast as well, we can discuss the best approach

metegenez:
ready for merge, thanks for reviewing marko.

milenkovicm:
thanks @metegenez epic contribution, 
we can iterate on top of this

PR #1685, opened 2026-05-11
test: cover push scheduling in client integration tests

### LINKED ISSUE #1463: Improve integration tests to test push scheduler mode as well

At the moment, integration tests, setup cluster in `pull mode` 

- https://github.com/apache/datafusion-ballista/blob/0396c58aa58a546395fa59bf6658e17c621c0cc3/ballista/client/tests/common/mod.rs#L104

as we make `push` scheduling default in https://github.com/apache/datafusion-ballista/pull/1461 we need to improve test to cover `push` scheduling.

create new test setup which would start cluster using `push` scheduling and include it in the tests, preferably adding additional case to each test

**Additional context**

abhinavgautam01:

Closes #1463.
# Rationale for this change

Issue #1463 asked for integration tests to cover **push-staged** scheduling as well as **pull-staged**. The in-process test cluster was effectively pull-only: the standalone scheduler used a fixed identity while listening on a random port, and the standalone executor always used the pull `poll_loop`. That gap meant we did not exercise the same scheduler/executor path as the default push policy in real deployments.
In addition, exposing push mode in tests surfaced a real correctness issue: push executors report task status via gRPC clients keyed by the scheduler’s advertised `host:port`; a hard-coded placeholder did not match the actual listener, which could cause hangs or failed status updates.

- **Scheduler (`standalone`)**: bind `127.0.0.1:0` first, then build the in-memory cluster / `SchedulerServer` with `scheduler_endpoint = addr.to_string()` so task metadata and curator IDs match the real gRPC listener.
- **Executor (`standalone`)**: add push-staged support using the existing `executor_server::startup` path; keep pull-staged behavior on `poll_loop` as before.
- **`executor_server::startup`**: wait until the executor gRPC address accepts a TCP connection before `register_executor`, reducing a race with asynchronously started listeners.
- **Client integration tests**: helpers and `rstest` cases for push (`remote_push`, `remote_state_push`); `context_setup` remote tests parameterized over `TaskSchedulingPolicy`; align test URLs with `127.0.0.1` for consistency with the binder.
- **Examples test common**: same `127.0.0.1` host alignment for cluster URLs.
- **Minor**: cast ephemeral `grpc_port` to `u32` for protobuf registration fields.
Tests: `cargo test -p ballista --features standalone` (with `protoc` available).

**Yes, but small and additive.** New public APIs were added on the scheduler and executor `standalone` modules (policy-aware constructors / `_with_scheduling_policy` helpers). Existing entry points retain prior behavior where applicable (e.g. default standalone scheduler helpers still target pull-staged integration defaults as before).

No changes to CLI flags or documented user workflows beyond optional use of these APIs by embedders/tests.


ballista/executor/src/standalone.rs
```diff
@@ -146,8 +259,108 @@ pub async fn new_standalone_executor_from_builder(
...
+        concurrent_tasks,
+        task_scheduling_policy: TaskSchedulingPolicy::PushStaged,
+        grpc_max_decoding_message_size: max_message_sz,
+        grpc_max_encoding_message_size: max_message_sz,
+        ..ExecutorProcessConfig::default()
+    };
+
+    let shutdown_notifier: &'static ShutdownNotifier =
+        Box::leak(Box::new(ShutdownNotifier::new()));
+    let (stop_send, _stop_recv) = mpsc::channel::<bool>(10);
```
martin-g: What is the idea behind the immediate dropping of the Receiver ?
The Sender will fail when one tries to use it

ballista/executor/src/standalone.rs
```diff
@@ -146,8 +259,108 @@ pub async fn new_standalone_executor_from_builder(
...
+        "Ballista v{BALLISTA_VERSION} Rust Executor (push) listening on {flight_addr:?}"
+    );
+
+    let grpc_probe = TcpListener::bind("127.0.0.1:0").await?;
+    let grpc_port = grpc_probe.local_addr()?.port();
+    drop(grpc_probe);
+
+    let executor_meta = ExecutorRegistration {
+        id: Uuid::new_v4().to_string(),
+        host: Some("localhost".to_string()),
```
martin-g: Some times you use `127.0.0.1` and other times `localhost`. Why ?
I guess you replaced `localhost` with `127.0.0.1` to make it explicit that IPv4 should be used. But the new usages of `localhost` confuse me.

ballista/executor/src/standalone.rs
```diff
@@ -146,8 +259,108 @@ pub async fn new_standalone_executor_from_builder(
...
+) -> Result<()> {
+    let flight_listener = TcpListener::bind("localhost:0").await?;
+    let flight_addr = flight_listener.local_addr()?;
+    info!(
+        "Ballista v{BALLISTA_VERSION} Rust Executor (push) listening on {flight_addr:?}"
+    );
+
+    let grpc_probe = TcpListener::bind("127.0.0.1:0").await?;
+    let grpc_port = grpc_probe.local_addr()?.port();
+    drop(grpc_probe);
```
martin-g: Probe and Drop is a subject to TOCTOU race condition.
Some other process may bind to the now available port before you try to do it.

ballista/executor/src/standalone.rs
```diff
@@ -146,8 +259,108 @@ pub async fn new_standalone_executor_from_builder(
...
+        grpc_port,
+        concurrent_tasks,
+        task_scheduling_policy: TaskSchedulingPolicy::PushStaged,
+        grpc_max_decoding_message_size: max_message_sz,
+        grpc_max_encoding_message_size: max_message_sz,
+        ..ExecutorProcessConfig::default()
+    };
+
+    let shutdown_notifier: &'static ShutdownNotifier =
+        Box::leak(Box::new(ShutdownNotifier::new()));
```
martin-g: Could you use `static` + `Lazy` instead of `Box::leak()` ?
This will leak the notifier on every call to this method.

ballista/executor/src/standalone.rs
```diff
@@ -146,8 +259,108 @@ pub async fn new_standalone_executor_from_builder(
...
+        specification: Some(
+            ExecutorSpecification::default()
+                .with_task_slots(concurrent_tasks as u32)
+                .into(),
+        ),
+        os_info: Some(ExecutorOperatingSystemSpecification::default().into()),
+    };
+
+    let config_snap = config_producer();
+    let max_message_sz = config_snap.ballista_grpc_client_max_message_size() as u32;
```
martin-g:
```suggestion
    let max_message_size = config_snap.ballista_grpc_client_max_message_size() as u32;
```
`sz` is not very clear. Also it is not consistent with [pull](https://github.com/apache/datafusion-ballista/pull/1685/changes#diff-2bdd8710486575e8337fd9ad57940750b92aa9566a88bd34d811e119f7386571R230)

ballista/executor/src/standalone.rs
```diff
@@ -93,16 +115,72 @@ pub async fn new_standalone_executor_from_builder(
...
     let listener = TcpListener::bind("localhost:0").await?;
     let address = listener.local_addr()?;
     info!("Ballista v{BALLISTA_VERSION} Rust Executor listening on {address:?}");
 
     let executor_meta = ExecutorRegistration {
-        id: Uuid::new_v4().to_string(), // assign this executor a unique ID
+        id: Uuid::new_v4().to_string(),
         host: Some("localhost".to_string()),
         port: address.port() as u32,
-        // TODO Make it configurable
```
martin-g, line 105: Why this comment is removed ?

ballista/executor/src/executor_server.rs
```diff
@@ -148,6 +169,8 @@ pub async fn startup<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>(
         })
     };
 
+    wait_executor_grpc_listen(&config.bind_host, config.grpc_port).await?;
```
martin-g: Is `config.grpc_port` the same as `executor_meta.grpc_port` ?
https://github.com/apache/datafusion-ballista/pull/1685/changes#diff-219ce667b5c79a620f1b44a8ba862134f4e1f7cfee78c737deaeeb50be71ac14R149

ballista/executor/src/standalone.rs
martin-g: Why the cast to `u32` ?

ballista/executor/src/executor_server.rs
martin-g: You have `addr` here - a `std::net::SocketAddr`. No need to pass `host`+`port` to parse them again in the body of `wait_executor_grpc_listen()`

ballista/executor/src/standalone.rs
abhinavgautam01: agreed, that would close the channel and later sends on stop_send could fail. There is now a lightweight task that recv()s in a loop so the Sender stays usable...

ballista/executor/src/standalone.rs
abhinavgautam01: Standalone paths now use 127.0.0.1 for binds and the advertised host so behavior matches explicit IPv4 loopback everywhere (no resolver ambiguity)...

ballista/executor/src/standalone.rs
abhinavgautam01: push standalone no longer reserves a port by probing and dropping the listener; it passes a TcpListener already bound on 127.0.0.1:0 into ExecutorGrpcListen::Bound so nothing else can take that port between probe and bind...

ballista/executor/src/standalone.rs
abhinavgautam01: Removed Box::leak. Standalone uses Arc::new(ShutdownNotifier::new()) and keeps one Arc clone alive inside the task that awaits the server handle so the notifier isn’t freed while gRPC is still running—no leak and no process-global static state...

ballista/executor/src/standalone.rs
abhinavgautam01: Renamed to max_message_size (session config / Flight, same idea as pull) and grpc_max_message_size for ExecutorProcessConfig, with a one-line comment tying them together....

ballista/executor/src/standalone.rs
abhinavgautam01: restored an explicit TODO next to the hard-coded grpc_port on standalone pull registration...

ballista/executor/src/executor_server.rs
abhinavgautam01: For the real executor process they stay aligned because structure_executor_metadata copies options.grpc_port into protobuf grpc_port; startup still has a debug_assert_eq! linking executor.metadata.grpc_port and config.grpc_port. Standalone push builds ExecutorRegistration and ExecutorProcessConfig from the same bound TcpListener port.

ballista/executor/src/standalone.rs
abhinavgautam01: ExecutorProcessConfig’s gRPC encode/decode limits are u32 (matching tonic’s sizing APIs), while ballista_grpc_client_max_message_size() is usize on SessionConfig, we clamp with min(u32::MAX as usize) then cast so oversized session values don’t truncate silently...

ballista/executor/src/executor_server.rs
abhinavgautam01: wait_executor_grpc_listen takes a single SocketAddr; start_executor_process parses bind_host + executor.metadata.grpc_port once into grpc_listen_addr for the dynamic listen path. For standalone push, ExecutorGrpcListen::Bound skips the TCP probe entirely because the listener is already bound. startup now passes &ShutdownNotifier and clones only the broadcast::Sender into the gRPC task instead of Arc-wrapping the whole notifier just to satisfy 'static...

ballista/client/tests/context_checks.rs
```diff
@@ -162,6 +165,7 @@ mod supported {
     #[rstest]
     #[case::standalone(standalone_context())]
     #[case::remote(remote_context())]
+    #[case::remote_push(remote_context_push_scheduling())]
```
killzoner: i think you should use a dedicated fixture with values ` #[values(remote_context(), remote_context_push_scheduling())]` (something like this, don't know the exact syntax with `rtest`) for this to avoid repeating everywhere

killzoner:
You added standalone push scheduler, but this is not covered by tests ATM ?

ballista/client/tests/context_checks.rs
abhinavgautam01: done. i replaced the repeated remote/push `#[case]` entries with rstest `#[values(...)]` on the async context parameter. `rstest` 0.26 does not accept `#[values]` on fixture parameters, so this uses the supported syntax while keeping the context variants grouped at the parameter...

abhinavgautam01:
> You added standalone push scheduler, but this is not covered by tests ATM ?

added coverage in `context_setup.rs`: `standalone_push_scheduling_executes_sql` starts the standalone scheduler/executor with push scheduling and verifies a real SQL query against it...

killzoner:
> > You added standalone push scheduler, but this is not covered by tests ATM ?
> 
> added coverage in `context_setup.rs`: `standalone_push_scheduling_executes_sql` starts the standalone scheduler/executor with push scheduling and verifies a real SQL query against it...

Sorry, not quite what I meant. I meant having 0 change on the test cases themselves, and just splitting what you did to 
```rust
#[fixture]
   async fn remote(
       #[values(TaskSchedulingPolicy::PullStaged, TaskSchedulingPolicy::PushStaged)]
       policy: TaskSchedulingPolicy,
   ) -> SessionContext {
       common::remote_context_with_scheduling(policy).await
   }
```

 and the same for `standalone`.
This way you keep the 2 remote vs standalone, do 0 change on the existing code, and cover pull / push for both paths.

Also now you have diverging code style with `ballista/client/tests/context_unsupported.rs`

abhinavgautam01:
updated... i removed the standalone one-off test and changed the context tests to keep the two context cases (`standalone` and `remote`) while expanding `TaskSchedulingPolicy::{PullStaged, PushStaged}` through `#[values(...)]`...

i also aligned `context_unsupported.rs` to use the same pattern... the exact `#[fixture] async fn remote(#[values(...)] ...)` syntax does not compile with the repo’s `rstest` 0.26, so this uses the supported equivalent with context builders plus policy values...

abhinavgautam01:
ping @martin-g

andygrove:
Thanks @abhinavgautam01. I plan on reviewing this PR this week.

PR #1686, opened 2026-05-11
Minor improvements for #1675

martin-g:

Follow-up of #1675.

# Rationale for this change

Return `0ns`, i.e. with a time unit, as a fallback

* Time::to_string() returns the unit
* Use .expect() with an error message if the system time is before Unix epoch (highly unlikely)


ballista/scheduler/src/api/handlers.rs
```diff
@@ -729,14 +728,14 @@ fn get_running_stage_time(task_infos: &[Option<TaskInfo>], current_time: u128) -
...
 
+    let time = Time::new();
     match (min_start, current_time) {
         (Some(start), end) if end > start => {
-            let t = Time::new();
-            t.add_duration(Duration::from_millis((end - start) as u64));
-            t.to_string()
+            time.add_duration(Duration::from_millis((end - start) as u64));
         }
-        _ => "0".to_string(),
```
martin-g, line 750: Here `"0"` was not matching the formatting of `Time::to_string()`

milenkovicm: I did that on purpose to identify that its has not been run. `0ns` does not make a lot of sense when we use millisecond granularity.not sure if empty string would be better fit. 

Do you prefer `0ns` instead of ` ` ?

martin-g: I see!
How about "N/A" ?

I had the same issue in the TUI app. When `end_exec_time` is `0` (i.e. unfinished job) I was wondering between "N/A" and "a running duration", i.e. "now() - start_exec_time" on every tick/refresh. But the TUI's system time may be different than the Scheduler's system time and I decided to show "N/A".

martin-g: Or maybe the field type should be `Option<u64>` and None telling the client that it is unavailable yet.

milenkovicm: `Option<u64>` makes sense

milenkovicm:
@martin-g is our final answer `Option<u64>` ? 😀

martin-g:
Yes!
I will wait for https://github.com/apache/datafusion-ballista/pull/1693 to be merged first and then I will re-work this PR!

milenkovicm (commented):
LGTM thanks @martin-g

PR #1691, opened 2026-05-12
fix(join-selection): guard CollectLeft swap when right has multiple partitions

### LINKED ISSUE #1681: TPC-H Q2 fails on latest `main` when hash-based shuffle is enabled
**Describe the bug**

TPC-H Q2 seems to have a problem on latest `main` (May 11 2026).
It was working fine last week.

**To Reproduce**

```
datafusion-ballista/benchmarks on  1642-executor-metrics [$!?⇡] is 📦 v53.0.0 via 🐍 v3.12.13 (ballista) via 🦀 v1.95.0
❯ python3 tpch.py --query q2 --path tpch-data
Registering table part at path tpch-data/part.parquet
Registering table supplier at path tpch-data/supplier.parquet
Registering table partsupp at path tpch-data/partsupp.parquet
Registering table customer at path tpch-data/customer.parquet
Registering table orders at path tpch-data/orders.parquet
Registering table lineitem at path tpch-data/lineitem.parquet
Registering table nation at path tpch-data/nation.parquet
Registering table region at path tpch-data/region.parquet
Job k2Jf6V6 failed: Job failed due to stage 11 failed: Task failed due to runtime execution error: DataFusionError(Internal("Assertion failed: self.mode != PartitionMode::CollectLeft || left_partitions == 1: Invalid HashJoinExec, the output partition count of the left child must be 1 in CollectLeft mode,consider using CoalescePartitionsExec or the EnforceDistribution rule"))

Traceback (most recent call last):
  File "/Users/mgrigorov/git/apache/datafusion/datafusion-ballista/benchmarks/tpch.py", line 55, in <module>
    df.show()
  File "/Users/mgrigorov/git/apache/datafusion/datafusion-ballista/python/python/ballista/extension.py", line 64, in method_wrapper
    return getattr(df, func)(*argz, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Exception: DataFusion error: Arrow error: External error: Execution error: Job k2Jf6V6 failed: Job failed due to stage 11 failed: Task failed due to runtime execution error: DataFusionError(Internal("Assertion failed: self.mode != PartitionMode::CollectLeft || left_partitions == 1: Invalid HashJoinExec, the output partition count of the left child must be 1 in CollectLeft mode,consider using CoalescePartitionsExec or the EnforceDistribution rule"))
```

**Expected behavior**

**Additional context**

andygrove:

Closes #1681

# Rationale for this change

TPC-H Q2 with `prefer_hash_join=true` fails on latest `main` with:

```
Invalid HashJoinExec, the output partition count of the left child must be 1
in CollectLeft mode, consider using CoalescePartitionsExec or the
EnforceDistribution rule
```

The Ballista-customized `JoinSelection` rule in
`ballista/scheduler/src/physical_optimizer/join_selection.rs` calls
`try_collect_left(..., ignore_threshold=true, 0, 0)` on already-resolved
`HashJoinExec(CollectLeft)` joins. When `should_swap_join_order(left, right)`
returns `true` (left has bigger byte-size than right), the `(true, true)` arm
calls `hash_join.swap_inputs(PartitionMode::CollectLeft)`. `swap_inputs`
keeps `mode = CollectLeft` and swaps the children, so if the old right was a
multi-partition shuffle reader (e.g. `Hash([…], 14)`), the new left ends up
with 14 partitions while still in `CollectLeft` mode. Ballista has no
post-`JoinSelection` `EnforceDistribution` pass to reshape this, so the
invalid plan is serialized to the executor and the assertion fires at
`HashJoinExec::execute`.

The bug is reachable today because PR #1647's broadcast-style hash join
lowering produces `HashJoinExec(CollectLeft)` joins where the build side
is a broadcast `ShuffleReaderExec` (1 logical partition). On TPC-H Q2 SF1,
stage 11's broadcast reader (full supplier row, 7 columns × 10K rows ≈
1 MB) happens to be heavier in bytes than its hash probe reader
(post-join 3K rows × 4 columns ≈ 230 KB), so the swap triggers and the
plan becomes invalid. Stage 22 has the same join shape but a smaller
broadcast side, so it doesn't swap and runs fine. The existing comment in
`execution_stage.rs:386–392` already flags this category of risk:

```rust
// ballista specific JoinSelection, as datafusion rule can't be used here.
// Datafusion JoinSelection may produce plans which need change of partitions
// in order to be valid.
```

- `try_collect_left`: guard the `(true, true)` and `(false, true)` swap
  arms with `right.output_partitioning().partition_count() == 1`. A swap
  into `CollectLeft` is only safe when the would-be new left will have a
  single partition. The non-swap rebuild paths are untouched.
- Regression test
  `collect_left_swap_preserves_one_partition_build`: builds a
  `HashJoinExec(CollectLeft)` with a 1-partition, heavier left and a
  2-partition, lighter right, runs `JoinSelection::optimize`, and asserts
  the resulting join still has a 1-partition build side. The test fails
  on `main` and passes with this change.

Verified end-to-end: TPC-H Q2 SF1 with `prefer_hash_join=true` and
`target_partitions=14` now returns 100 rows instead of failing at stage 11.
Other queries in a sanity sweep (Q1, Q3, Q5, Q7, Q9, Q18, Q21 under
sort-merge default) continue to pass. There is a separate, pre-existing
correctness issue where Q21 with `prefer_hash_join=true` returns 0 rows; it
is independent of this assertion bug and out of scope here.

No public API changes. Plans that previously crashed at the executor with
the `CollectLeft` assertion will now keep the original (1-partition build)
shape instead of being swapped into an invalid form.

### DIFF
--- a/ballista/scheduler/src/physical_optimizer/join_selection.rs
+++ b/ballista/scheduler/src/physical_optimizer/join_selection.rs
@@ -237,11 +237,17 @@ pub(crate) fn try_collect_left(
             threshold_num_rows,
         );
 
+    // Swapping into `CollectLeft` puts the current right onto the build side,
+    // and `HashJoinExec` requires the build side to have exactly one partition
+    // in that mode. Ballista has no post-`JoinSelection` distribution pass to
+    // fix this, so any swap that would break the invariant must be skipped.
+    let right_one_partition = right.output_partitioning().partition_count() == 1;
     match (left_can_collect, right_can_collect) {
         (true, true) => {
             // Don't swap null-aware anti joins as they have specific side requirements
             if hash_join.join_type().supports_swap()
                 && !hash_join.null_aware
+                && right_one_partition
                 && should_swap_join_order(&**left, &**right)?
             {
                 Ok(Some(hash_join.swap_inputs(PartitionMode::CollectLeft)?))
@@ -262,7 +268,10 @@ pub(crate) fn try_collect_left(
         ))),
         (false, true) => {
             // Don't swap null-aware anti joins as they have specific side requirements
-            if hash_join.join_type().supports_swap() && !hash_join.null_aware {
+            if hash_join.join_type().supports_swap()
+                && !hash_join.null_aware
+                && right_one_partition
+            {
                 hash_join.swap_inputs(PartitionMode::CollectLeft).map(Some)
             } else {
                 Ok(None)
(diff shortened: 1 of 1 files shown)


martin-g:
Issue #1681 is fixed in this branch!

ballista/scheduler/src/physical_optimizer/join_selection.rs
```diff
@@ -746,6 +755,108 @@ mod test {
 
     /// Create join filter for NLJoinExec with expression `big_col > small_col`
     /// where both columns are 0-indexed and come from left and right inputs respectively
```
milenkovicm: has this comment been hijacked from next method down `nl_join_filter` ?

ballista/scheduler/src/physical_optimizer/join_selection.rs
```diff
@@ -746,6 +755,108 @@ mod test {
...
+            find_hash_join(&optimized).expect("HashJoinExec missing from optimized plan");
+
+        assert_eq!(*hj.partition_mode(), PartitionMode::CollectLeft);
+        assert_eq!(
+            hj.left().output_partitioning().partition_count(),
+            1,
+            "JoinSelection swapped a CollectLeft join's inputs and ended up \
+             with a multi-partition reader on the build side, which violates \
+             CollectLeft's invariant",
+        );
```
milenkovicm, line 854: minor, would it make sense to check if `hj.right().output_partitioning().partition_count() == 2` just in case  ?

milenkovicm (approved):
thanks @andygrove 
two minor comments

milenkovicm:
hey @andygrove i just addressed those two minor comments to merge this, i hope you dont mind. will merge when CI green

PR #1693, opened 2026-05-12
Fix REST API panic on job list/detail when end_time < start_time

### LINKED ISSUE #1689: if the job fail rest api will panic with: `attempt to subtract with overflow`
**Describe the bug**

```
thread 'tokio-rt-worker' (9926839) panicked at ballista/scheduler/src/api/handlers.rs:334:55:
attempt to subtract with overflow
```

https://github.com/apache/datafusion-ballista/blob/8b4efb6d4191b40d9f05e9bcc065d7464ec9716d/ballista/scheduler/src/api/handlers.rs#L334

```rust
let jobs: Vec<JobResponse> = jobs
        .iter()
        .map(|job| {
            let (plain_status, job_status) =
                format_job_status(&job.status.status, job.end_time - job.start_time);
```

**To Reproduce**

**Expected behavior**

**Additional context**

abhinavgautam01:

Closes #1689
 # Rationale for this change

The REST job list and job-detail handlers computed elapsed time as `end_time - start_time` with `u64`. If `end_time` is before `start_time` (for example `end_time` is still `0` while `start_time` is set, or other inconsistent timestamps on failed jobs), the subtraction underflows and the scheduler API can panic with `attempt to subtract with overflow`. Saturating subtraction matches how we already treat inconsistent timestamps elsewhere in the same module and avoids the panic.

- Add `job_elapsed_ms(start_time, end_time)` using `end_time.saturating_sub(start_time)`.
- Use it in `get_jobs` and `get_job` when calling `format_job_status`.
- Add unit tests for normal duration and `end_time < start_time`.

Yes, but only in the sense of **bug fix behavior**: the jobs REST API should no longer panic when listing or fetching a job whose stored timestamps are inconsistent; elapsed duration used in status text is clamped to **0 ms** when `end_time < start_time` instead of crashing.

### DIFF
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -39,7 +39,6 @@ use datafusion::prelude::SessionConfig;
 use log::{debug, error, info, warn};
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
 use std::vec;
 
 // TODO: the AQE planner runs DataFusion's DefaultPhysicalPlanner with a
@@ -1196,10 +1195,7 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
             .map(|l| l.try_into())
             .collect::<ballista_core::error::Result<Vec<_>>>()?;
 
-        self.end_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_millis() as u64;
+        self.end_time = timestamp_millis();
 
         self.status = JobStatus {
             job_id: self.job_id.clone(),
@@ -1292,10 +1288,7 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
                 let task_attempt = stage.task_failure_numbers[partition_id];
                 let task_info = crate::state::execution_graph::TaskInfo {
                     task_id,
-                    scheduled_time: SystemTime::now()
-                        .duration_since(UNIX_EPOCH)
-                        .unwrap()
-                        .as_millis(),
+                    scheduled_time: timestamp_millis() as u128,
                     // Those times will be updated when the task finish
                     launch_time: 0,
                     start_exec_time: 0,
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -1167,6 +1167,9 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
 
     /// fail job with error message
     fn fail_job(&mut self, error: String) {
+        // Same as non-AQE graph: persist end time when the job fails.
+        self.end_time = timestamp_millis();
+
         self.status = JobStatus {
             job_id: self.job_id.clone(),
             job_name: self.job_name.clone(),
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -1167,7 +1167,6 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
 
     /// fail job with error message
     fn fail_job(&mut self, error: String) {
-        // Same as non-AQE graph: persist end time when the job fails.
         self.end_time = timestamp_millis();
 
         self.status = JobStatus {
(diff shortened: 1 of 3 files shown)


milenkovicm (commented):
maybe we should also set the `end_time` in case of job failure ?

abhinavgautam01:
> maybe we should also set the `end_time` in case of job failure ?

Good point — succeed_job already sets end_time, but fail_job never did, so FailedJob.ended_at could stay 0 while started_at was set. I’ve updated both StaticExecutionGraph and the AQE graph’s fail_job to set self.end_time to timestamp_millis() before building the failed status, mirroring success handling. The REST handlers still use saturating elapsed ms so inconsistent timestamps can’t panic if they ever appear again.

ballista/scheduler/src/state/aqe/mod.rs
```diff
@@ -1167,6 +1167,9 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
 
     /// fail job with error message
     fn fail_job(&mut self, error: String) {
+        // Same as non-AQE graph: persist end time when the job fails.
```
milenkovicm: do we need this comment ?

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -1348,6 +1348,10 @@ impl ExecutionGraph for StaticExecutionGraph {
 
     /// fail job with error message
     fn fail_job(&mut self, error: String) {
+        // Match `succeed_job`: record wall-clock end time so job metadata and REST consumers
```
milenkovicm: do we need this comment ?

milenkovicm (commented):
thanks @abhinavgautam01 
it would be good to add tests for `fail_job` if possible

ballista/scheduler/src/state/aqe/mod.rs
abhinavgautam01: no, removed it. Calling timestamp_millis() before building FailedJob is clear from the code and parity with succeed_job is obvious....

ballista/scheduler/src/state/execution_graph.rs
abhinavgautam01: no,  removed it for the same reason; the behaviour is self-explanatory and the new test documents the invariant....

abhinavgautam01:
> maybe we should also set the `end_time` in case of job failure ?

added test_fail_job_sets_end_time_and_failed_metadata in ballista/scheduler/src/state/execution_graph.rs: it calls ExecutionGraph::fail_job on a StaticExecutionGraph, checks the job is Failed with the expected message, that end_time is set and >= start_time, and that FailedJob.started_at / FailedJob.ended_at match the graph’s times....
AdaptiveExecutionGraph::fail_job still mirrors the same two lines (end_time = timestamp_millis() + status); this test locks in the behaviour for the trait contract used by both paths....

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -1772,6 +1774,38 @@ mod test {
...
+        assert!(
+            matches!(
+                graph.status().status.as_ref(),
+                Some(job_status::Status::Failed(f)) if f.error == "test failure"
+            ),
+            "expected FailedJob status after fail_job"
+        );
+        assert!(
+            graph.end_time() >= start,
+            "end_time {} should be set and >= start_time {}",
```
martin-g:
```suggestion
            "end_time ({}) should be set and >= start_time ({})",
```

ballista/scheduler/src/state/aqe/mod.rs
martin-g:
```suggestion
self.end_time = timestamp_millis();
```

ballista/scheduler/src/state/execution_graph.rs
martin-g:
```suggestion
self.end_time = timestamp_millis();
```

martin-g (approved):

ballista/scheduler/src/state/execution_graph.rs
abhinavgautam01: done, the assertion panic message now uses parentheses: end_time ({}) ... start_time ({})...

ballista/scheduler/src/state/aqe/mod.rs
abhinavgautam01: applied: self.end_time = timestamp_millis();. Also switched the test-only TaskInfo.scheduled_time path to timestamp_millis() as u128 and dropped the SystemTime / UNIX_EPOCH import so we don’t get unused-import warnings when pop_next_task is not compiled....

ballista/scheduler/src/state/execution_graph.rs
abhinavgautam01: applied the same timestamp_millis() assignment for consistency with fail_job and the shared helper...!

martin-g (approved):

milenkovicm (approved):
thanks @abhinavgautam01 and @martin-g

PR #1697, opened 2026-05-13
[CLI/TUI] Use only tracing crate for logging in CLI and TUI

### LINKED ISSUE #1683: CLI/TUI: Use a single logger library
We have discussed this one some time ago. Before it was panicking to install tracing, now it just prints to stderr.

The problem is that the CLI app uses `env_logger` to log to stderr/stdout but since the TUI app paints on the terminal I use the `tracing` crate to log to a file instead.
Neither `log` nor `tracing` have APIs to uninstall the global logger they install :-/

So, we need to start using just one logger for both CLI and TUI.
I will see whether the `tracing` crate provides a way to use a subscriber dynamically - if TUI then log to a file else log to stdout.

_Originally posted by @martin-g in https://github.com/apache/datafusion-ballista/issues/1670#issuecomment-4419580315_

martin-g:

Closes #1683.

# Rationale for this change

When CLI - log to stdout
When TUI - log to a file

The TUI app needs the whole screen. Logging to stdout breaks its UI.

Drop the usage of `env_logger` by the CLI app for logging to stdout.
Use a dynamic logger (custom impl of MakeWriter trait) that depending on a flag logs either to stdout or a file

No.
In the CLI app now the user can start the TUI app without an error that the tracing logger cannot be installed because the one from env_logger is already there.


milenkovicm (approved):
thanks @martin-g

PR #1698, opened 2026-05-13
fix: no executor warning, correct prometheus feature name in TUI

killzoner:

I was looking at testing some stuff locally in distributed mode.
I made : 

```bash
cargo run --bin ballista-scheduler --features prometheus-metrics # first shell
cargo run --bin ballista-cli -- --host localhost --port 50050 # second shell
```

then submit `SELECT 1;` in the cli.

This hangs forever (cannot even close the cli) and nothing appears.
This is because there is no executor, but it's not obvious looking at metrics or terminal output.

 # Rationale for this change

This avoids silent job submission without user feedback. I can also imagine a cluster with all executor not reachable, the task would be submitted but zero warning, and 0 visibility in metrics.

- promote `debug` message to `warn` when no executor (I guess this should never happen and makes the issue visible even running with no extra config)
- fix a typo in the (great!) TUI on prometheus metrics feature name to enable
- ~expose pending task in metrics (metric existed but what never filled)~

- did *not* address the cli hang while waiting for scheduler response (might be a follow up if deemed useful)

~Existing `pending_tasks_queue_size` is now filled~

### DIFF
--- a/ballista/scheduler/src/state/executor_manager.rs
+++ b/ballista/scheduler/src/state/executor_manager.rs
@@ -114,7 +114,7 @@ impl ExecutorManager {
         }
         let alive_executors = self.get_alive_executors();
         if alive_executors.is_empty() {
-            debug!("There's no alive executors for binding tasks");
+            warn!("There's no alive executors for binding tasks");
             return Ok(vec![]);
         }
         self.cluster_state
--- a/ballista/scheduler/src/state/executor_manager.rs
+++ b/ballista/scheduler/src/state/executor_manager.rs
@@ -114,7 +114,7 @@ impl ExecutorManager {
         }
         let alive_executors = self.get_alive_executors();
         if alive_executors.is_empty() {
-            warn!("There's no alive executors for binding tasks");
+            warn!("There are no alive executors to bind tasks");
             return Ok(vec![]);
         }
         self.cluster_state
--- a/ballista-cli/src/tui/ui/main/metrics/mod.rs
+++ b/ballista-cli/src/tui/ui/main/metrics/mod.rs
@@ -88,7 +88,7 @@ pub fn render_metrics(f: &mut Frame, area: Rect, app: &App) {
         render_no_metrics(
             f,
             rects[1],
-            "The scheduler is built with 'prometheus_metric' feature disabled.",
+            "The scheduler is built with 'prometheus-metrics' feature disabled.",
         );
     }
 }


ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -114,7 +114,7 @@ impl ExecutorManager {
         }
         let alive_executors = self.get_alive_executors();
         if alive_executors.is_empty() {
-            debug!("There's no alive executors for binding tasks");
+            warn!("There's no alive executors for binding tasks");
```
martin-g:
```suggestion
            warn!("There are no alive executors to bind tasks");
```

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -88,6 +88,10 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
             time_recorder = Some((Instant::now(), event.clone()));
         };
         let event_sender = EventSender::new(tx_event.clone());
+        self.metrics_collector.set_pending_tasks_queue_size(
+            self.state.task_manager.pending_task_number().await as u64,
```
martin-g: Should this be updated here ?
It looks expensive to iterate the jobs on every event.

Also, the metrics should be updated **after** processing the event.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -270,6 +270,16 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
         self.active_job_cache.len()
     }
 
+    /// Get the total number of pending tasks across all active jobs.
+    pub async fn pending_task_number(&self) -> usize {
+        let mut count = 0;
+        for pair in self.active_job_cache.iter() {
+            let graph = pair.value().execution_graph.read().await;
```
martin-g: DashMap::iter() acquires a Read lock on the map.
The lock is kept here while `.await`-ing and this prevents any writers to proceed.

```suggestion
            let execution_graph = pair.value().execution_graph.clone();
            let graph = execution_graph.read().await;
```

killzoner: Thanks ! I'm not very used to `dashmap`, this makes sense

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
killzoner: At first I tried to put it in specific stages, but I couldn't figure exactly what stages would make it reliable and not potentially stale (i think even executor lost can yield back tasks in pending?)

Ended with this out of simplicity, will have a second look to be more selective

killzoner: I ended up with https://github.com/apache/datafusion-ballista/pull/1698/changes/d46c134215b75c224624bc8eb2ef528423636b7f

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -270,6 +270,17 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
         self.active_job_cache.len()
     }
 
+    /// Get the total number of pending tasks across all active jobs.
+    pub async fn pending_task_number(&self) -> usize {
```
milenkovicm: I don't like that we iterate through all jobs all the time to calculate pending task numbers

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -358,6 +371,13 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                 );
             }
         }
+
+        if update_pending_tasks_metric {
```
milenkovicm: In which case would we use this metrics? If this number is 200 or 2000 current what does it say to observer? How can I interpret it?

Could we use scheduling task latency metric instead?   Like a 1min 5min 10min moving average of it?

I'm not sure but task update even may have it (to be confirmed) so we could build that metric without iterating through all the active jobs 

This is just an idea for discussion

milenkovicm (commented):
Thanks @killzoner i have a comment for discussion

Maybe we could split this pr in typo and warning part and take metric as a separate pr where we need to think about the metric itself and way it's collected

killzoner:
> Thanks @killzoner i have a comment for discussion
> 
> Maybe we could split this pr in typo and warning part and take metric as a separate pr where we need to think about the metric itself and way it's collected

Sure I will split, I'm not completely satisfied either with the current solution for the metrics

EDIT: done, i will open later another PR and link the existing pending comments

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
killzoner: Honestly i'm not sure the metric is *that* useful. This accumulates only in case of no executor being able to launch tasks.

We would probably need to rework the way `bind_schedulable_tasks` is being done to have a valid and not perf impacting metric.

My take is that it's probably more useful to have a metric saying "no executor available" than having the number of pending tasks, which is more focus and also easier to implement (you can probably set this metric only in `revive_offers`) then *remove* the existing unfilled `pending_tasks_queue_size`

ballista/scheduler/src/state/task_manager.rs
killzoner: Everyone seems to agree on that along with @martin-g , and the added value for this metric is not high

martin-g (approved):

PR #1704, opened 2026-05-14
[TUI] Add a config setting for rendering job stage's plan as a tree

martin-g:

Related-to: #1650

# Rationale for this change

Make it possible to render the stage plans as a tree too.

* Add a config setting to render job stage's plans as a tree.
* Add scrolling support to the Stage Plan popup

Yaml: `job.stage.plan.tree: true`
Env var: `BALLISTA_JOB_STAGE_PLAN_TREE=true`

No

### DIFF
--- a/ballista-cli/src/tui/domain/jobs/stages.rs
+++ b/ballista-cli/src/tui/domain/jobs/stages.rs
@@ -77,6 +77,7 @@ pub struct JobStagesPopup {
     pub stages: JobStagesResponse,
     pub table_state: TableState,
     details_view: StageDetailsView,
+    pub scroll_position: u16,
 }
 
 impl JobStagesPopup {
@@ -86,6 +87,7 @@ impl JobStagesPopup {
             stages,
             table_state: TableState::default(),
             details_view: StageDetailsView::None,
+            scroll_position: 0,
         }
     }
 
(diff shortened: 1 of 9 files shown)


ballista-cli/src/tui/domain/jobs/stages.rs
```diff
@@ -114,36 +121,44 @@ impl JobStagesPopup {
...
+                }
+            } else {
+                self.table_state.select(Some(0));
+            }
+        } else if self.is_plan_view() {
+            self.scroll_position = self.scroll_position.saturating_add(1);
         }
     }
 
     pub fn scroll_up(&mut self) {
```
milenkovicm, line 144: i believe we should be able to scroll to right / left as well, as some plans can't really fit in allocated space, and get trimmed. I've noticed that with text based plans? 

 would that be possible ?

milenkovicm (approved):
thanks @martin-g 
one request for left/right scroll in plan mode, we can take it as a follow up if you prefer

ballista-cli/src/tui/domain/jobs/stages.rs
martin-g: > would that be possible ?

Yes! It is possible!
I will do it in a follow-up!

PR #1711, opened 2026-05-15
[TUI] Add support for horizontal scrolling to the job/stage plan popups

### LINKED ISSUE #1710: [TUI] Support horizontal scrolling in job/stage plan popups
i believe we should be able to scroll to right / left as well, as some plans can't really fit in allocated space, and get trimmed. I've noticed that with text based plans? 

 would that be possible ?

_Originally posted by @milenkovicm in https://github.com/apache/datafusion-ballista/pull/1704#discussion_r3247279919_

martin-g:

Closes #1710 

# Rationale for this change

Some job/stage plans are wider than the popup area and their right part is not visible. 

Add support for scrolling horizontally with Left/Right arrows

The user can scroll left/right in plan popups in TUI

### DIFF
--- a/ballista-cli/src/tui/app.rs
+++ b/ballista-cli/src/tui/app.rs
@@ -206,6 +206,8 @@ impl App {
                     KeyCode::Esc => popup.set_no_details_view(),
                     KeyCode::Up => popup.scroll_up(),
                     KeyCode::Down => popup.scroll_down(),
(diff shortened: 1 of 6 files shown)


milenkovicm (approved):
thanks @martin-g

milenkovicm:
one note, it would be great if you add a TUI screenshot to README before we release 53.

PR #1695, opened 2026-05-12
Right align all numeric columns in the TUI tables

martin-g:
# Which issue does this PR close?

Closes #1692.

# Rationale for this change

It is easier to compare cells in a column by their value.

# What changes are included in this PR?

Use right alignment for all numeric header and data cells.

# Are there any user-facing changes?

Visual representation of the data in the table columns

PR #1713, opened 2026-05-15
[TUI] Executor's id is not a numeric column. It should be center aligned

martin-g:

Follow-up of #1695 .

# Rationale for this change

The Executor's ID column is a UUID, not a number.
It should be centered, not right aligned.

Center the executor's ID column and give it more space.

Only visual changes in the TUI


milenkovicm (approved):
thanks @martin-g

PR #1716, opened 2026-05-16
Merge Executor's brief and extended properties

### LINKED ISSUE #1682: make `/api/executor/<id>` return number of cpus & update tui to show it
> This is a bit more involved. At the moment the Executors view uses the `/api/executors` REST API to load its data. To be able to show the CPU count I will need to make an extra HTTP request to `/api/executor/<id>` for every executor to load its extra details (the OS info).

1. update the REST API to return all the information with `/api/executors`
2. update ballista tui to display cpus in executor view 

_Originally posted by @milenkovicm in https://github.com/apache/datafusion-ballista/issues/1670#issuecomment-4419175918_

this is follow up on #1670

martin-g:

Closes #1682

# Rationale for this change

Provide all the information about an executor in the `/api/executors` REST endpoint.

* /api/executors now returns all the information per executor
* visualize the `CPU cores` property in the Executors table

The REST API is changed but it is has been introduced in an earlier PR that is not yet released.

### DIFF
--- a/ballista-cli/src/tui/app.rs
+++ b/ballista-cli/src/tui/app.rs
@@ -343,32 +343,36 @@ impl App {
                 if self.is_jobs_view() {
                     self.sort_jobs_by(JobsSortColumn::Status);
                 } else if self.is_executors_view() {
-                    self.sort_executors_by(ExecutorsSortColumn::TaskSlots);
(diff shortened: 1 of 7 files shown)


milenkovicm (approved):
thanks @martin-g

PR #1718, opened 2026-05-18
feat(aqe): SplitPartitionsRule — file-list sharding for skewed shuffle partitions (v1)

### LINKED ISSUE #1643: TPC-H Q2 parallelism bottleneck: hash partitioning on low-cardinality keys causes severe data skew
## Summary

**note that this is AI generated and there are probably mistakes, but I think the overall points made are valid**

When running TPC-H Q2 at SF1000 with 32 executors (8 concurrent tasks each, 64 partitions), the query starts with all executors busy but quickly drops to a single busy executor for a significant portion of the query. Root cause is hash-partitioning intermediate shuffle data on low-cardinality columns (`n_regionkey` with 5 distinct values, `n_nationkey` with 25 distinct values) into 64 buckets.

## Environment

- Ballista with 32 executors, 8 concurrent tasks each
- 64 hash partitions
- TPC-H SF1000, Parquet/zstd format
- PushStaged scheduling policy

### Stage execution flow

The query is broken into stages 3–17. The critical path is:

| Stage | Operation | Shuffle Key | Duration |
|-------|-----------|-------------|----------|
| 13 | Scan `partsupp` (800M rows) | `Hash(ps_suppkey, 64)` | ~4s |
| 3 | Scan `supplier` (10M rows) | `Hash(s_suppkey, 64)` | <1s |
| 6 | HashJoin `part ⋈ partsupp` | `Hash(ps_suppkey, 64)` | ~2s |
| 14 | Join supplier+partsupp | `Hash(s_nationkey, 64)` | ~2s |
| **15** | **HashJoin `nation ⋈ (supplier+partsupp)`** | **`Hash(n_regionkey, 64)`** | **~9s** |
| **16** | **HashJoin `region` + partial agg** | `Hash(ps_partkey, 64)` | ~3s |
| 17 | Final agg `min(ps_supplycost)` | `Hash(ps_partkey, 64)` | ~1s |

### The bottleneck: Stage 15 data skew

Stage 15 joins `nation` (25 rows) with the supplier+partsupp result (800M rows) on `n_nationkey`, then repartitions the output by `Hash(n_regionkey, 64)`.

Since `n_regionkey` has only **5 distinct values** mapped into 64 hash buckets, only 5 of 64 partitions receive any data. The rest are empty. Observed row distribution on one executor:

| Partition | Input Rows | Write Time |
|-----------|-----------|------------|
| 56, 57, 59, 60, 62 | 0 | ~100µs each |
| 58 | 31.96M | 2.10s |
| 61 | 63.93M | 5.53s |

The wall-clock skew ratio is **8670x** (8.7s max vs ~0s min).

### Stage 16 compounds the problem

Stage 16 joins `region` (1 row after filtering, e.g., `r_name = 'EUROPE'`) with stage 15 output on `r_regionkey`. Only 1 of the 5 non-empty regionkey partitions produces output, but all partitions still read their full shuffle data to perform the hash join — spending seconds reading millions of rows only to discard them.

### 27-second idle gap

After this executor finishes its stage 15/16 partitions at `15:48:25`, it sits completely idle until `15:48:52` (27 seconds) waiting for the straggler executor(s) that received the heaviest `n_regionkey` partitions. This alone accounts for **51% of the total executor wall time**.

```
Total wall time:   51.8s
Actual work:       25.2s
Idle (blocked):    26.5s
Utilization:       49%
```

### Shuffle write throughput is fine

The shuffle write mechanism itself performs at a consistent 10–16 Mrows/s across the heavy stages. The problem is not shuffle I/O — it's the skewed partition assignment.

### Four 23.87 GB shuffles of the same 800M-row dataset

The `ShuffleReaderExec` stats show four distinct 800M-row / 23.87 GB shuffle reads flowing through the query plan. The `partsupp` and `supplier+partsupp` intermediate data is reshuffled multiple times through the join tree.

## Suggested improvements

1. **Broadcast join for small dimension tables**: `nation` (25 rows) and `region` (5 rows) should be broadcast rather than hash-partitioned. This eliminates the skewed `Hash(n_regionkey, 64)` shuffle entirely.

2. **Push region filter earlier**: If the region filter (`r_name = 'EUROPE'`) is pushed down before the nation join, only ~5 nations participate, reducing intermediate data by ~4/5.

3. **Detect low-cardinality partition keys**: The planner could check column statistics (NDV) before choosing a hash partition key. When NDV << partition count, it should prefer an alternative key or switch to broadcast.

4. **Reduce redundant reshuffles**: The 800M-row supplier+partsupp result is shuffled 4 times through the join tree. Reordering joins or combining stages could reduce this.

wirybeaver:
Adds **`SplitPartitionsRule`** — the inverse of #1684's `CoalescePartitionsRule`. When upstream stats show one shuffle partition is far larger than the median, the rule fans that partition out across multiple reader tasks via round-robin assignment over its file list, instead of folding small partitions together. Same per-stage invocation, same alignment-group leaf walk, same carrier-slot-on-`ExchangeExec` pattern as #1684 — strict architectural mirror.

> **Stacks on #1684.** Until that lands, this PR's diff shows the 5 coalesce commits + 1 split commit. Once #1684 merges I'll rebase and the diff will reduce to the single split commit.

Part of the AQE epic #1359. Motivating bug: #1643 (TPC-H Q2 SF1000, one partition 8670× larger than the median). **v1 does NOT close #1643** — see "v1 scope" below.

## Mechanism — file-list sharding (v1)

The shuffle reader side already lists multiple `PartitionLocation`s per output partition (one per upstream map task). Splitting just means handing those locations to several reader tasks via round-robin assignment by file index. No row-range reads, no protobuf changes, no executor data-path changes — pure scheduler/adapter work.

Tradeoff: a partition backed by only one file can't be split this way. The rule bails on that idx (factor stays at 1). Row-range reads to lift that restriction live in a v2 task doc (linked at the bottom).

## v1 scope — and what it does NOT cover

File-list sharding produces `UnknownPartitioning(K')` output. Rows that used to land in the same hash bucket on the M-side end up scattered across multiple K'-side partitions (the round-robin assignment is file-keyed, not row-keyed). That breaks any downstream operator with a hash or single-partition input requirement:

- `HashJoinExec(Partitioned)` and `SortMergeJoinExec` — both legs must agree on hash buckets.
- `AggregateExec(FinalPartitioned)` — assumes each downstream partition holds a closed set of group keys.
- Any future operator whose `required_input_distribution()` returns `HashPartitioned(_)` or `SinglePartition`.

Rather than enumerate operator types, the rule walks the stage subtree and inspects `required_input_distribution()` directly. If any node above the leaves demands hash or single-partition input, the rule bails the whole stage. Strictly correct, future-proof against new DataFusion operators.

**TPC-H Q2's skew sits behind a `FinalPartitioned` aggregate**, so v1 cannot help it. v1 helps the narrower set of stages where the consumer is distribution-agnostic (`FilterExec`, `ProjectionExec`, `LocalLimitExec`, single-input scans into `UnknownPartitioning` sinks). The infrastructure is the win; v2 (row-range reads + aggregate-aware plan rewriting) is where Q2 lands.

## Surface

Opt-in via a single boolean. **`ballista.planner.split.enabled=false`** is the default — the rule short-circuits, the plan flows through untouched. Users who want the skew-handling turn it on.

| key | default | meaning |
|---|---|---|
| `ballista.planner.split.enabled` | `false` | master gate |
| `ballista.planner.split.skew_factor` | `5.0` | Spark `OptimizeSkewedJoin` default — partition is "skewed" if `bytes > skew_factor × median` |
| `ballista.planner.split.min_split_bytes` | 64 MiB | absolute floor — don't fan-out trivially small partitions |
| `ballista.planner.split.max_split_factor` | `8` | per-partition fan-out cap to limit executor pressure |

## Algorithm

For each per-stage call (mirrors the coalesce rule line-for-line):

1. Bail if disabled.
2. Single subtree walk that does two things: bail if any operator demands `HashPartitioned`/`SinglePartition` input, AND collect every leaf `ExchangeExec` (`Jump` after each hit, same convention as #1684).
3. Conflict guard: bail if any leaf already has `coalesce()` or `split()` set.
4. Alignment-group invariant: every leaf shares M (Q22 guard — same one #1684 added in `fix(coalesce): bail on heterogeneous M`).
5. Sum byte sizes element-wise across the alignment group; capture leaf-0's file-count vector.
6. `decide_split_factors(summed_bytes, file_counts, skew, min_bytes, max_factor)` — three guards (single-file → factor 1; below min-bytes → factor 1; below skew ratio → factor 1) then `ceil(bytes / median)` capped at `max_factor`, floored at 2.
7. If every factor is 1, bail.
8. Build one `SplitPlan` with `shards = factors_to_shards(&factors)` and attach uniformly to every leaf via `set_split` — sharing the plan (not just K') keeps per-idx fan-out identical across leaves, which matters for non-join multi-leaf shapes (UNION).

Sharing the plan across the alignment group is the same invariant #1684 enforces for coalesce — preserves uniform K' downstream.

## Components

| Area | Change |
|---|---|
| **Rule** | `SplitPartitionsRule` in `state/aqe/optimizer_rule/split_partitions.rs`. Unit struct. Invoked per-stage in `AdaptivePlanner::actionable_stages()` right after `CoalescePartitionsRule`. |
| **Carrier** | `ExchangeExec` gains an `Arc<Mutex<Option<Arc<SplitPlan>>>>` slot next to the existing `coalesce` slot, with `set_split` / `split` accessors. `split=K' of M` annotation appears in `DisplayAs` output only when attached. Mutually exclusive with `coalesce` — rule's conflict guard enforces it. |
| **Adapter** | `BallistaAdapter::transform_children` checks `exchange.split()`; on `Some(sp)` it pre-shards the M-shape upstream `Vec<Vec<PartitionLocation>>` into K'-shape via `SplitShard::owns_file(file_idx)` and builds the reader via `ShuffleReaderExec::try_new_split` with `UnknownPartitioning(K')`. |
| **Reader** | `ShuffleReaderExec::try_new_split` constructor + `split: Option<SplitPlan>` field. Threaded through `with_work_dir`, `with_client_pool`, `with_new_children`, `partition_statistics`, `DisplayAs`. `execute()` needs no changes — `self.partition[idx]` already returns the per-output-partition `Vec<PartitionLocation>`, which the adapter has pre-sharded. The reader is oblivious to whether sharding happened. |
| **Algorithm** | Pure-CPU helpers in `state/aqe/split/algorithm.rs` — `decide_split_factors` (median-based skew detection mirroring Spark's `OptimizeSkewedJoin`) + `factors_to_shards` (per-idx factor → flat shard list). |
| **Config** | Four new keys (table above). |

`SplitPlan` is NOT round-tripped through proto in v1 — the rule attaches it on the scheduler side, the adapter consumes it inline, and the resulting `ShuffleReaderExec` ships already-sharded `Vec<Vec<PartitionLocation>>`. A protobuf round-trip would only be needed if the rule outcome had to survive serialization to the executor, which it doesn't.

## Test plan

- [x] `cargo test --workspace --no-fail-fast` — workspace tests pass, 0 failures (includes the 7 coalesce regressions from #1684 plus the 8 new split tests).
- [x] `cargo clippy --workspace --all-targets --tests` — 0 warnings.
- [x] `cargo fmt --all` — clean.
- [x] 9 unit tests in `state/aqe/split/algorithm.rs` covering the decision function
- [x] 8 functional tests in `state::aqe::test::split_rule`:
- [ ] **TPC-H SF=100 sanity sweep, 22 queries × 2 join variants, `split.enabled=true`**: deferred to PR validation env. Expected outcome is mostly non-regression — most queries hit a hash or `FinalPartitioned` consumer and the rule bails. If any query *does* fire the rule (a stage ending in `Filter` / `Projection` / `LocalLimit` over a hash exchange), I'll record the K → K' increase in a comment.

## v2 follow-up

The honest scope limitation in v1 (file-list sharding requires ≥2 files per partition; bails on hash/single consumers) is lifted by v2 — row-range reads (`PartitionLocation::row_range: Option<(u64, u64)>`, batch-count IPC reader on the executor) AND aggregate-aware splitting (rewrite `AggregateExec(FinalPartitioned)` into per-shard partial + reshuffle + final). That's the path #1643 will actually land on. Task doc written and cross-linked.

### DIFF
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -101,6 +101,23 @@ pub const BALLISTA_COALESCE_SMALL_PARTITION_FACTOR: &str =
 pub const BALLISTA_COALESCE_MERGED_PARTITION_FACTOR: &str =
     "ballista.planner.coalesce.merged_partition_factor";
 
+/// Configuration key to enable AQE split-skewed-partitions rule.
+/// Disabled by default — opt in for workloads with severe partition skew
+/// downstream of distribution-agnostic operators (Filter/Projection/LocalLimit).
+/// v1 bails on joins and FinalPartitioned aggregates because file-list sharding
+/// breaks hash colocation.
+pub const BALLISTA_SPLIT_ENABLED: &str = "ballista.planner.split.enabled";
+/// Multiplier over the median per-partition byte size at which a partition is
+/// considered skewed enough to split. Mirrors Spark's `skewedPartitionFactor`.
+pub const BALLISTA_SPLIT_SKEW_FACTOR: &str = "ballista.planner.split.skew_factor";
+/// Minimum partition byte size below which the split rule never fires, even
+/// when the skew ratio would otherwise qualify. Prevents pointless splitting
+/// of tiny outliers.
+pub const BALLISTA_SPLIT_MIN_BYTES: &str = "ballista.planner.split.min_split_bytes";
+/// Upper bound on the per-partition split factor. Caps task fan-out to avoid
+/// overwhelming the executor when a single partition is many medians larger.
+pub const BALLISTA_SPLIT_MAX_FACTOR: &str = "ballista.planner.split.max_split_factor";
+
 /// Result type for configuration parsing operations.
 pub type ParseResult<T> = result::Result<T, String>;
 use std::sync::LazyLock;
@@ -219,6 +236,40 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
             "Merged-partition early-flush factor (Spark legacy).".to_string(),
             DataType::Float64,
             Some("1.2".to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_SPLIT_ENABLED.to_string(),
+            "Enables the AQE split-skewed-partitions rule. Disabled by default — \
+             opt in for severe per-partition skew downstream of distribution-agnostic \
+             operators (Filter/Projection/LocalLimit). v1 bails on joins and \
+             FinalPartitioned aggregates."
+                .to_string(),
+            DataType::Boolean,
+            Some(false.to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_SPLIT_SKEW_FACTOR.to_string(),
+            "Multiplier over the median per-partition byte size at which a partition \
+             is considered skewed enough to split. Mirrors Spark's skewedPartitionFactor."
+                .to_string(),
+            DataType::Float64,
+            Some("5.0".to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_SPLIT_MIN_BYTES.to_string(),
+            "Minimum partition byte size below which the split rule never fires, \
+             even when the skew ratio would otherwise qualify."
+                .to_string(),
+            DataType::UInt64,
+            Some((64 * 1024 * 1024_usize).to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_SPLIT_MAX_FACTOR.to_string(),
+            "Upper bound on the per-partition split factor. Caps task fan-out \
+             for very-large outliers."
+                .to_string(),
+            DataType::UInt32,
+            Some(8.to_string()),
         ),
     ];
     entries
@@ -448,6 +499,28 @@ impl BallistaConfig {
     /// Returns the merged-partition early-flush factor (Spark legacy).
     pub fn coalesce_merged_partition_factor(&self) -> f64 {
         self.get_float_setting(BALLISTA_COALESCE_MERGED_PARTITION_FACTOR)
+    }
+
+    /// Returns whether the AQE split-skewed-partitions rule is enabled.
+    pub fn split_enabled(&self) -> bool {
+        self.get_bool_setting(BALLISTA_SPLIT_ENABLED)
+    }
+
+    /// Returns the multiplier over the per-partition byte median at which a
+    /// partition qualifies as skewed (Spark's `skewedPartitionFactor`).
+    pub fn split_skew_factor(&self) -> f64 {
+        self.get_float_setting(BALLISTA_SPLIT_SKEW_FACTOR)
+    }
+
+    /// Returns the minimum partition byte size below which the split rule
+    /// never fires.
+    pub fn split_min_split_bytes(&self) -> u64 {
+        self.get_usize_setting(BALLISTA_SPLIT_MIN_BYTES) as u64
+    }
+
+    /// Returns the upper bound on per-partition split factor.
+    pub fn split_max_split_factor(&self) -> u32 {
+        self.get_usize_setting(BALLISTA_SPLIT_MAX_FACTOR) as u32
     }
 
     /// Should client employ pull or push job tracking strategy
--- a/ballista/core/src/execution_plans/mod.rs
+++ b/ballista/core/src/execution_plans/mod.rs
@@ -31,7 +31,9 @@ use std::path::{Path, PathBuf};
 use datafusion::common::exec_err;
 pub use distributed_explain_analyze::DistributedExplainAnalyzeExec;
 pub use distributed_query::DistributedQueryExec;
-pub use shuffle_reader::{CoalescePlan, PartitionGroup, ShuffleReaderExec};
+pub use shuffle_reader::{
+    CoalescePlan, PartitionGroup, ShuffleReaderExec, SplitPlan, SplitShard,
+};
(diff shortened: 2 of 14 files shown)


milenkovicm:
Please rebase as downstream PR has been merged

wirybeaver:
Closing in favor of porting Spark's `OptimizeSkewedJoin` AQE rule. File-list sharding bails on any hash/single-partition consumer (joins, FinalPartitioned aggregates, global limits), which covers most realistic workloads — including the TPC-H Q2 case this PR aimed to address. Spark's split+replicate pattern handles the highest-value skew shape (skewed joins) directly and is correct by construction. New work will land on branch `optimize-skewed-join`.

PR #1722, opened 2026-05-18
feat(aqe): dynamic switching from streaming agg to hash aggregation

wirybeaver:

Part of #1359 (AQE epic) — roadmap line: *"switch from streaming aggregation to hash aggregation (extended rules)"*

## What does this PR do?

Adds `DynamicAggregateAlgorithmRule`, an AQE physical optimizer rule that re-derives `InputOrderMode` for each `AggregateExec` after a shuffle stage resolves and rewrites the operator when the derived mode differs from the cached one.

### Problem

DataFusion freezes `InputOrderMode` (`Linear` / `Sorted` / `PartiallySorted`) at plan time. In Ballista's AQE this creates two issues:

1. **Wasted memory**: a downstream aggregate stays `Linear` (hash table, O(distinct-groups) memory) even when an upstream stage completes and grants ordering on the group-by columns — where `Sorted` (streaming, O(1) memory) would suffice.
2. **Stale correctness assumption**: after a subtree rewrite, an aggregate may hold a `Sorted` claim when its input is no longer ordered.

### Approach

`AggregateExec::with_new_children` already calls `try_new_with_schema` which re-derives `input_order_mode` from the current input `EquivalenceProperties`. The rule:

1. Walks the plan with `transform_up`.
2. At each `AggregateExec`, skips if no resolved `ExchangeExec` exists in the subtree (idempotence guard).
3. Calls `with_new_children([same_input])` to force re-derivation.
4. Returns `Transformed::yes(rebuilt)` only when the derived mode differs from the cached one.

No upstream DataFusion changes are required.

### Configuration

```sql
SET ballista.aqe.dynamic_aggregate.enabled = true;
```

Disabled by default (`false`) pending benchmarking.

## Checklist

- [x] New rule registered in `default_optimizers()` before `DistributedExchangeRule`
- [x] Config key `ballista.aqe.dynamic_aggregate.enabled` added to `BallistaConfig`
- [x] 9 unit tests: gate off, no-exchange skip, unresolved-exchange skip, sorted/unsorted input, schema preservation, idempotence
- [x] All 42 AQE optimizer rule tests pass

### DIFF
--- a/ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
+++ b/ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
@@ -1,0 +1,338 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::state::aqe::execution_plan::ExchangeExec;
+use ballista_core::config::BallistaConfig;
+use datafusion::common::tree_node::{Transformed, TreeNode};
+use datafusion::config::ConfigOptions;
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::aggregates::AggregateExec;
+use log::info;
+use std::sync::Arc;
+
+/// AQE rule that re-derives `InputOrderMode` for each `AggregateExec` after a
+/// stage completes and rewrites the operator when the derived mode differs from
+/// the cached one.
+///
+/// DataFusion freezes `InputOrderMode` (Linear = hash table / Sorted = streaming /
+/// PartiallySorted) at plan time. After a shuffle stage resolves, the input's
+/// `EquivalenceProperties` may expose ordering that wasn't visible earlier, making
+/// streaming aggregation possible (or, after a subtree rewrite, making the previously
+/// assumed ordering stale). This rule repairs both cases on each AQE replan.
+///
+/// Relies on `AggregateExec::with_new_children` calling `try_new_with_schema`,
+/// which re-derives `input_order_mode` from the current input equivalence
+/// properties — no upstream DataFusion changes required.
+///
+/// Enabled via `ballista.aqe.dynamic_aggregate.enabled` (default: false).
+#[derive(Debug, Clone, Default)]
+pub struct DynamicAggregateAlgorithmRule {}
+
+impl DynamicAggregateAlgorithmRule {
+    fn transform(
+        plan: Arc<dyn ExecutionPlan>,
+        enabled: bool,
+    ) -> datafusion::error::Result<Transformed<Arc<dyn ExecutionPlan>>> {
+        if !enabled {
+            return Ok(Transformed::no(plan));
+        }
+        let Some(agg) = plan.as_any().downcast_ref::<AggregateExec>() else {
+            return Ok(Transformed::no(plan));
+        };
+        // Only act when a resolved exchange exists below — if nothing resolved,
+        // the ordering is the same as at initial plan time and no switch is needed.
+        if !subtree_has_resolved_exchange(&plan) {
+            return Ok(Transformed::no(plan));
+        }
+
+        // Re-derive InputOrderMode by reconstructing the aggregate with the same
+        // input. with_new_children → try_new_with_schema re-runs the derivation
+        // against the current input EquivalenceProperties.
+        let rebuilt = plan
+            .clone()
+            .with_new_children(vec![agg.children()[0].clone()])?;
+        let rebuilt_agg = rebuilt
+            .as_any()
+            .downcast_ref::<AggregateExec>()
+            .expect("with_new_children on AggregateExec must return AggregateExec");
+
+        if rebuilt_agg.input_order_mode() == agg.input_order_mode() {
+            return Ok(Transformed::no(plan));
+        }
+
+        info!(
+            "DynamicAggregateAlgorithmRule: switching InputOrderMode {:?} → {:?}",
+            agg.input_order_mode(),
+            rebuilt_agg.input_order_mode(),
+        );
+        Ok(Transformed::yes(rebuilt))
+    }
+}
+
+impl PhysicalOptimizerRule for DynamicAggregateAlgorithmRule {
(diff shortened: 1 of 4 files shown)


ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+/// Returns true if any `ExchangeExec` in `plan`'s subtree is resolved
+/// (i.e. its shuffle has completed). Recurses through all nodes, including
+/// through exchange boundaries — unlike `nearest_exchange_status` in
+/// `distributed_exchange.rs` which stops at the first exchange, here we
+/// want to know if *any* upstream stage has finished.
+fn subtree_has_resolved_exchange(plan: &Arc<dyn ExecutionPlan>) -> bool {
+    if let Some(exchange) = plan.as_any().downcast_ref::<ExchangeExec>() {
+        exchange.shuffle_created() && !exchange.inactive_stage
```
martin-g, line 117: The doc above says `Recurses through all nodes, including through exchange boundaries` but the implementation stops recursion on the first ExchangeExec

ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        // the SortExec — so it may already be Sorted. What matters is the rule
+        // doesn't break it (idempotence when already correct).
+        let original_mode = agg_exec.input_order_mode().clone();
+
+        let rule = DynamicAggregateAlgorithmRule::default();
+        let result = rule.optimize(agg.clone(), &config_with_flag(true)).unwrap();
+        let result_agg = result.as_any().downcast_ref::<AggregateExec>().unwrap();
+
+        // After the rule the mode must match what re-derivation produces.
+        assert_eq!(result_agg.input_order_mode(), &original_mode);
```
martin-g, line 262: assert that the order mode is Sorted

ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        assert_eq!(
+            DynamicAggregateAlgorithmRule::default().name(),
+            "DynamicAggregateAlgorithmRule"
+        );
+    }
+
+    #[test]
+    fn schema_check_is_false() {
+        assert!(!DynamicAggregateAlgorithmRule::default().schema_check());
+    }
```
martin-g, line 337: None of the tests validates order mode change from Linear to Sorted.

milenkovicm:
Sorry but I do not understand this PR, can you please give a bit more explanation what are you trying to do in this PR? 

In which case can we assume sorting is going to be preserved after an exchange ?

milenkovicm:
> Wasted memory: a downstream aggregate stays Linear (hash table, O(distinct-groups) memory) even when an upstream stage completes and grants ordering on the group-by columns — where Sorted (streaming, O(1) memory) would suffice.

In which cases will this happen? When can w guarantee preserving sorting across shuffles ?


> Stale correctness assumption: after a subtree rewrite, an aggregate may hold a Sorted claim when its input is no longer ordered.

Can you please give an example when would this claim  be true?

ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
wirybeaver: You're right — the doc contradicts the implementation. The code stops at the first `ExchangeExec` and returns based on its resolved state, which is what we actually want (the nearest exchange gates the agg's input; a resolved exchange further down behind an unresolved one doesn't help). The doc should describe the actual behavior.

That said, please see my reply to @milenkovicm on the main thread — the broader design has problems that may make this PR moot.

ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
wirybeaver: Right — the assertion is missing, but more importantly the test as written can't exercise a Linear→Sorted transition. `aggregate_on_k(exchange)` calls `AggregateExec::try_new`, which derives `input_order_mode` from the exchange's eq_properties at construction time. Since the exchange wraps a `SortExec(k)` and propagates its sort ordering, the agg is derived as `Sorted` immediately. The rule then re-derives the same mode, producing no transition — so the test passes trivially and doesn't validate what its name claims. Full root-cause analysis in my reply to @milenkovicm.

ballista/scheduler/src/state/aqe/optimizer_rule/dynamic_aggregate_algorithm.rs
wirybeaver: Confirmed — and this gap isn't just test coverage. `with_new_children` re-derivation produces the same mode the constructor would, because `ExchangeExec.equivalence_properties()` doesn't change between construction and resolution. So the rule can't, by design, produce a Linear↔Sorted transition. Details in my reply to @milenkovicm.

wirybeaver:
@milenkovicm Thanks for pushing back — your questions exposed that my reasoning was wrong. Let me walk through what I now see.

**The flawed premise.** I assumed `ExchangeExec::equivalence_properties()` would reflect post-shuffle ordering once the exchange resolves. It doesn't. From `ballista/scheduler/src/state/aqe/execution_plan.rs:118`:

```rust
let eq_properties = input.properties().eq_properties.clone();
```

The eq_properties are captured at construction time from the **pre-shuffle input** and never updated when `resolve_shuffle_partitions()` is called.

**Why the rule is a no-op (or worse).**

1. **Linear → Sorted can't usefully fire.** If the agg's input advertises sort properties, those were visible at the agg's construction time too — `try_new_with_schema` derived the right mode then. The rule's `with_new_children` re-derivation produces the same mode.

2. **When it would fire, it would be incorrect.** A hash repartition wrapping a `SortExec` carries the sort eq_properties on the exchange, even though the actual post-shuffle data is no longer globally sorted (hash shuffle destroys global ordering). Switching the agg to `Sorted` based on those properties would cause streaming aggregation to produce wrong results.

3. **In valid replan scenarios it's redundant.** If a sibling rule changes the agg's subtree during `transform_up`, the framework already calls `with_new_children` on the parent, which re-derives the mode — no separate rule needed.

**What the roadmap line actually requires.**

For AQE to re-derive useful information after a stage resolves, the resolved exchange (or whatever node represents the post-shuffle reader) needs to advertise *post-shuffle* properties — driven by shuffle mode:

- Hash shuffle: destroys global ordering, only preserves partitioning on the hash key
- Sort-based shuffle: preserves per-partition ordering on the sort key
- Single-partition coalesce: preserves input ordering

Once those properties reflect reality, no dedicated re-derivation rule is needed — `with_new_children` during normal optimizer passes naturally picks up the change.

**Proposal.** Close this PR. I'd like to open a separate one (or RFC discussion) to update `ExchangeExec.equivalence_properties()` to translate properties on resolve based on shuffle mode. Happy to take direction from you on whether that's the right next step or if I'm still misreading the AQE design.

Apologies for the misfire — should have validated the premise with a runnable example before opening.

wirybeaver:
Closing per my analysis above — the design is wrong at the foundation. Will reopen with a different approach (likely updating `ExchangeExec.equivalence_properties()` to reflect post-shuffle reality on resolve) once direction is settled.

PR #1723, opened 2026-05-18
feat(aqe): empty join handling

sandugood:

Handles #1359, particularly this part:

> verify [if we can reuse instead](https://github.com/spiceai/spiceai/blob/348b7cc7eaac7eaebbce277e8569d8119989a921/crates/datafusion-optimizer-rules/src/physical_plan/hash_join_optimization.rs#L39) of creating our own rule for eliminating joins

 # Rationale for this change
Handling cases of different physical types of joins and adding branching for different logical parts (i.e Left or LeftAnti) without adding separate AQE rule - adding logic straight into the `PropagateEmptyExecRule`

1. Added `JoinInfo` struct to contain information about the join (both left and right part)
2. Added `as_join` function, which can be extended in the future to add support for other physical joins. Right now it handles both SortMerge and HashJoin
3. Reused `is_guaranteed_empty` from the Spice-AI implementation of eliminating empty joins.

No

# Were this changes tested?
Yes, tests were included in both `propagate_empty.rs` and `join_info.rs`


milenkovicm:
thanks @sandugood will have a look soon

ballista/scheduler/src/state/aqe/optimizer_rule/join_info.rs
```diff
@@ -0,0 +1,226 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
milenkovicm: can we move this helpers to `propagate_empty.rs` ?

milenkovicm (commented):
sorry for late review @sandugood 
few recommendations 

this is extra on top of your current contribution that would make sense to be added, please add it.


## `empty_exec!` should produce correct partition num

```
macro_rules! empty_exec {
    ($e:expr) => {
        Ok(Transformed::yes(Arc::new(
            EmptyExec::new($e.schema())
                .with_partitions($e.properties().output_partitioning().partition_count()),
        )))
    };
}
```

## add configuration value 

`ballista.planner.propagate_empty.enabled=true` as we may need to disable this rule in some cases

ballista/scheduler/src/state/aqe/optimizer_rule/join_info.rs
sandugood: Yeah, I think it makes sense, since it is basically related to the propagate_empty rule

milenkovicm:
#1752 has few other rules added, could you please copy paste them here. Will merge this pr before that one and makes sense to include them

sandugood:
Seems like there is an error when trying to get packages from crates.io
Could you, please, rerun the CI in a while?

milenkovicm:
@sandugood my apologies I got stuck with #1752 and needed a bit of a break. I hope to get back and review this tomorrow. Also please check that pr as there are some comments regarding empty exec and I'm not sure if this pr covers them

sandugood:
No need to hurry, take your time and thanks for the review
Going to check those comments on #1752 regarding the emptyexec and see what can be (or need to be) done

ballista/scheduler/src/state/aqe/optimizer_rule/propagate_empty.rs
```diff
@@ -85,12 +98,47 @@ impl PropagateEmptyExecRule {
...
+            let right_empty = is_guaranteed_empty(join.right);
+
+            // Checking whether join would produce an empty result
+            let should_eliminate = match join.join_type {
+                JoinType::Inner => left_empty || right_empty,
+                JoinType::Full => left_empty && right_empty,
+                JoinType::Left
+                | JoinType::RightSemi
+                | JoinType::LeftAnti
+                | JoinType::LeftMark => left_empty,
```
milenkovicm: should we add left semi? 
https://github.com/apache/datafusion/blob/7843ab385bdc90db02a80ccfd410eb74c1aeb0fb/datafusion/optimizer/src/propagate_empty_relation.rs#L143-L153

ballista/scheduler/src/state/aqe/optimizer_rule/propagate_empty.rs
```diff
@@ -85,12 +98,47 @@ impl PropagateEmptyExecRule {
...
+                JoinType::Right
+                | JoinType::LeftSemi
+                | JoinType::RightAnti
+                | JoinType::RightMark => right_empty,
```
milenkovicm: should we add right semi? 
https://github.com/apache/datafusion/blob/7843ab385bdc90db02a80ccfd410eb74c1aeb0fb/datafusion/optimizer/src/propagate_empty_relation.rs#L143-L153

milenkovicm (commented):
thanks @sandugood 
few comments and a bit more recommendations.


| Join type    | left_empty                       | right_empty                       |
|--------------|----------------------------------|-----------------------------------|
| Inner        | → empty                          | → empty                           |
| Full         | → null-pad (keep right, null left) | → null-pad (keep left, null right) |
| Full (both)  | → empty                          |                                   |
| Left         | → empty                          | → null-pad (keep left, null right) |
| Right        | → null-pad (keep right, null left) | → empty                          |
| LeftSemi     | → empty                          | → empty                           |
| RightSemi    | → empty                          | → empty                           |
| LeftAnti     | → empty                          | → return left child as-is         |
| RightAnti    | → return right child as-is       | → empty                           |


reading https://github.com/apache/datafusion/blob/main/datafusion/optimizer/src/propagate_empty_relation.rs and comparing  claude had few more recommendations,

### Missed optimisation 1 — `Left join + right_empty`

- **Upstream**: builds a null-padded `Projection` (all left rows survive with NULLs for right columns)
- **Ballista**: falls through to `Transformed::no` — the join is left as-is
- This is not a correctness bug (the join will still produce the right answer) but it is a missed simplification that avoids executing the join entirely

### Missed optimisation 2 — `Right join + left_empty`

- Same pattern as #3 in the mirror direction.
- **Upstream**: builds a null-padded `Projection` (all right rows survive with NULLs for left columns)
- **Ballista**: falls through

### Missed optimisation 3 — `Full join + one side empty`

- **Upstream**: if only one side is empty, converts to a null-padded projection of the surviving side
- **Ballista**: only eliminates when BOTH sides are empty; single-side-empty cases fall through
- Lines 121–122 implement `left_empty && right_empty` but the `left_empty XOR right_empty` cases are not handled

### Missed optimisation 4 — `LeftAnti + right_empty`

- **Upstream**: returns the left child plan directly (empty right means no rows are excluded → all left rows pass)
- **Ballista**: falls through (join is not simplified)

### Missed optimisation 5 — `RightAnti + left_empty`

- **Upstream**: returns the right child plan directly (empty left means no rows are excluded → all right rows pass)
- **Ballista**: falls through

ballista/scheduler/src/state/aqe/optimizer_rule/propagate_empty.rs
sandugood: Added, as well as left-semi and full

sandugood:
Addressed the comments, added test-cases as a sanity-check
The last test case simulates the mentioned flaky test fail in #1752, but it would be nice if you could rerun the same logic against this PR @metegenez 
Thank you in advance and thanks @milenkovicm for the review

milenkovicm (approved):
thanks @sandugood

metegenez:
> Addressed the comments, added test-cases as a sanity-check The last test case simulates the mentioned flaky test fail in #1752, but it would be nice if you could rerun the same logic against this PR @metegenez Thank you in advance and thanks @milenkovicm for the review

I will check it soon, a bit busy these days. Thanks for the contribution.

PR #1726, opened 2026-05-18
feat: task_manager delegats physical plan creation to execution graph

milenkovicm:

Closes #.

# Rationale for this change

Currently, ballista schedule creates and optimizes physical plan before it is delegated to execution graph for execution. this makes plan change a bit more complicated than it needs to be.

Propagating logical plan to execution graph will give ability to it to transform logical plan (or non optimized physical plan) in any way it needs, simplifying planning rules.

There is a bit of refactoring of existing code, moving some of the code (`EXPLAIN` handling) to its own method.

- change method parameters across few methods around task manager and execution graph creation
- move `EXPLAIN` related code to a function

yes if users are implementing `execution graph` interface

### DIFF
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -139,7 +139,6 @@ impl AdaptiveExecutionGraph {
         ctx: &SessionContext,
         logical_plan: &LogicalPlan,
         queued_at: u64,
-        session_config: Arc<SessionConfig>,
     ) -> ballista_core::error::Result<Self> {
         let session_id = ctx.session_id();
 
@@ -151,7 +150,7 @@ impl AdaptiveExecutionGraph {
         //       do we handle it now or just ignore it?
         let physical_plan = handle_explain_plan(job_id, ctx, logical_plan, plan).await?;
         let logical_plan = Some(logical_plan.display_indent().to_string());
-
+        let session_config = Arc::new(ctx.copied_config());
         // TODO: run planner set of optimizers in try_new
         let mut planner = AdaptivePlanner::try_new(
             &session_config,
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -19,6 +19,7 @@ use crate::display::print_stage_metrics;
 use crate::scheduler_server::event::QueryStageSchedulerEvent;
 use crate::scheduler_server::timestamp_millis;
 use crate::state::aqe::planner::AdaptivePlanner;
+use crate::state::distributed_explain::handle_explain_plan;
 use crate::state::execution_graph::{
     ExecutionGraph, ExecutionGraphBox, ExecutionStage, ResolvedStage, RunningTaskInfo,
     StageOutput,
@@ -34,6 +35,8 @@ use ballista_core::serde::protobuf::{
     job_status, task_status,
 };
 use ballista_core::serde::scheduler::{ExecutorMetadata, PartitionLocation};
+use datafusion::execution::context::SessionContext;
+use datafusion::logical_expr::LogicalPlan;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::prelude::SessionConfig;
 use log::{debug, error, info, warn};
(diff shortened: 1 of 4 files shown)


ballista/scheduler/src/state/mod.rs
```diff
@@ -380,120 +366,20 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
-            if let LogicalPlan::TableScan(scan) = plan {
-                let provider = source_as_provider(&scan.source)?;
-                if let Some(table) = provider.as_any().downcast_ref::<ListingTable>() {
-                    let local_paths: Vec<&ListingTableUrl> = table
-                        .table_paths()
-                        .iter()
-                        .filter(|url| url.as_str().starts_with("file:///"))
-                        .collect();
-                    if !local_paths.is_empty() {
-                        // These are local files rather than remote object stores, so we
```
martin-g, line 407: Is it intentional that this check for local files is no more made ?

ballista/scheduler/src/state/aqe/mod.rs
```diff
@@ -129,20 +132,33 @@ impl AdaptiveExecutionGraph {
...
+    pub async fn try_new(
         scheduler_id: &str,
         job_id: &str,
         job_name: &str,
-        session_id: &str,
-        plan: Arc<dyn ExecutionPlan>,
+        ctx: &SessionContext,
+        logical_plan: &LogicalPlan,
         queued_at: u64,
         session_config: Arc<SessionConfig>,
```
martin-g: The `session_config` parameter could be removed. It is available via `ctx.copied_config()`
https://github.com/apache/datafusion-ballista/pull/1726/changes#diff-b96609436c205ce10f64d745f73eb8043c695665b84679d2d606432603f4884cR285

ballista/scheduler/src/state/mod.rs
```diff
@@ -380,120 +366,20 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
         queued_at: u64,
         subscriber: Option<JobStatusSubscriber>,
     ) -> Result<()> {
         let start = Instant::now();
-        let session_config = Arc::new(session_ctx.copied_config());
-        if log::max_level() >= log::Level::Debug {
-            // optimizing the plan here is redundant because the physical planner will do this again
-            // but it is helpful to see what the optimized plan will be
-            let optimized_plan = session_ctx.state().optimize(plan)?;
-            debug!("Optimized plan: {}", optimized_plan.display_indent());
```
martin-g, line 393: It would be good to still log the optimized plan for better diagnostics.

ballista/scheduler/src/state/mod.rs
```diff
@@ -380,120 +366,20 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
-        let plan = session_ctx.state().create_physical_plan(plan).await?;
-        debug!(
-            "Physical plan: {}",
-            DisplayableExecutionPlan::new(plan.as_ref()).indent(false)
-        );
-
-        let plan = plan.transform_down(&|node: Arc<dyn ExecutionPlan>| {
-            if node.output_partitioning().partition_count() == 0 {
-                let empty: Arc<dyn ExecutionPlan> =
-                    Arc::new(EmptyExec::new(node.schema()));
```
martin-g, line 458: Should this logic be preserved somewhere in the new implementation ?

ballista/scheduler/src/state/mod.rs
milenkovicm: plan has been optimized but not used later, so log does not have a great value

ballista/scheduler/src/state/mod.rs
milenkovicm: yes, i don't think this rule get triggered at all

ballista/scheduler/src/state/mod.rs
milenkovicm: there should be no nodes with output partitioning 0,

metegenez (commented):
Should this change also affect StaticExecutionGraph? @milenkovicm

milenkovicm:
> Should this change also affect StaticExecutionGraph? @milenkovicm

it could but i dont want to change it in this PR

metegenez (approved):
That make sense, so overall this make ExecutionGraph trait more responsible for getting optimal physical plans. I couldnt find anything that doesnt make sense here. LGTM.

ballista/scheduler/src/state/mod.rs
metegenez: Try to reproduce a case for this, couldnt make it. I also think this is safe.

milenkovicm:
thanks @metegenez & @martin-g 
i plan to keep this PR open until after ballista 53 release which i plan for later

milenkovicm:
thanks @metegenez and @martin-g  for review

PR #1725, opened 2026-05-18
feat: TUI make task popup scrollable

milenkovicm:
# Which issue does this PR close?

Closes #1724.

 # Rationale for this change

# What changes are included in this PR?

# Are there any user-facing changes?

PR #1729, opened 2026-05-19
feat(tui): Use separate areas for the table and its associated scrollbar

martin-g:

Addresses an issue explained at https://github.com/apache/datafusion-ballista/pull/1725#issuecomment-4485893557

# Rationale for this change

Use a separate area for the table and its associated scrollbar.
This way the scrollbar is not rendered over the last column.

Add horizontal area for the table's scrollbar.

Only UI.


milenkovicm (approved):
thanks @martin-g

PR #1730, opened 2026-05-19
minor: [TUI] Extract a helper method for the table/scrollbar area splitter

martin-g:

Follow-up of #1729 

# Rationale for this change

Cleanup and code reuse.

1. Extract a duplicated code for splitting an Area into Table and Scrollbar to a helper method
2. Remove unused Scrollbar area in Jobs and Executors that was not really used.

Only UI

### DIFF
--- a/ballista-cli/src/tui/ui/main/executors/executors_table.rs
+++ b/ballista-cli/src/tui/ui/main/executors/executors_table.rs
@@ -17,8 +17,9 @@
 
 use crate::tui::app::App;
 use crate::tui::domain::executors::{Executor, SortColumn};
+use crate::tui::ui::vertical_scrollbar;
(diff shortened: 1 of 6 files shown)


martin-g:
Converted to Draft because I want to move the Scrollbar inside the Border widget.

martin-g:
https://github.com/ratatui/ratatui/blob/44fbfb4634b3df82326ca72424b50e544c917f30/ratatui-widgets/src/scrollbar.rs#L74 - to put the Scrollbar inside the Table's Block/Border we need to use `Margin` but this leads to the [overlaps](https://github.com/apache/datafusion-ballista/pull/1725#issuecomment-4485893557) in the Table's last column

milenkovicm (approved):
@martin-g thanks a lot

PR #1731, opened 2026-05-19
minor: [TUI] Sort the metrics before rendering them

martin-g:

N/A

# Rationale for this change

Each render of the `Metrics` table may "blink" due to the different order of the metrics/rows.

Sort the metrics in the Prometheus response before visualizing them in the Metrics table.

Only UI


milenkovicm (approved):
LGTM, thanks @martin-g

PR #1749, opened 2026-05-23
fix(python): clear latent lint debt and gate wheel jobs on lint/tests

andygrove:

Closes #.

# Rationale for this change

The Python Release Build workflow had \`startup_failure\` on every trigger from 2026-03-21 (when ASF Infra began enforcing the third-party-actions allowlist) until #1748 landed. During that 2-month CI blackout:

- The lint gate was effectively disabled. PR #1625 merged two unused imports in \`python/python/tests/test_context.py\` that \`ruff check\` would have caught.
- \`ruff format --check\` debt accumulated in \`python/python/ballista/extension.py\` and \`python/python/ballista/jupyter.py\`.
- More importantly, the workflow's wheel-building jobs only depend on \`generate-license\` — they would have built and published artifacts even on a red lint/test run, which is what triggered this investigation (a real lint failure surfaced after #1748, but the wheels jobs kept going).

This PR clears the lint debt and tightens the job graph so the lint/test gate is real.

\`python/python/tests/test_context.py\`
- Remove unused \`InsertOp, DataFrameWriteOptions\` imports added by #1625 (the \`test_write_*\` tests there don't pass them; #1625's description even notes \`write_options\` isn't respected yet).
- Apply \`ruff format\` (one missing blank line between function definitions).

\`python/python/ballista/extension.py\` and \`python/python/ballista/jupyter.py\`
- Apply \`ruff format\`. Trailing commas in imports, collapsed-or-wrapped function signatures, blank-line separators between methods. No behavior change.

\`python/pyproject.toml\`
- Add \`[tool.ruff.format] exclude = [\"*.ipynb\"]\`. Jupyter saves notebooks with one JSON indentation style and \`ruff format\` wants another, so without this every notebook commit would either churn ~800 lines of whitespace through ruff or red the CI. \`ruff check\` still applies to notebook cells.

\`.github/workflows/build.yml\`
- Add \`build\` (ruff) and \`test-release\` (pytest) to the \`needs:\` list of all four wheel-building jobs (\`build-python-mac-win\`, \`build-manylinux-x86_64\`, \`build-manylinux-aarch64\`, \`build-sdist\`). Wheels can no longer be produced when lint or pytest fails.

No code changes for ballista users. CI behavior change: the wheel-publish pipeline now correctly gates on lint and tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,6 +99,7 @@ jobs:
           uv run --no-project ruff format --check .
 
   generate-license:
+    if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6.0.2
@@ -113,6 +114,7 @@ jobs:
           path: LICENSE.txt
 
   build-python-mac-win:
+    if: github.event_name != 'pull_request'
     needs: [generate-license, build, test-release]
     name: Mac/Win
     runs-on: ${{ matrix.os }}
@@ -177,6 +179,7 @@ jobs:
           path: python/target/wheels/*
 
   build-manylinux-x86_64:
+    if: github.event_name != 'pull_request'
     needs: [generate-license, build, test-release]
     name: Manylinux x86_64
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


andygrove:
merged to unblock other PRs

PR #1760, opened 2026-05-24
docs(release): cover TestPyPI verification and fix stale wheel-download notes

andygrove:

Closes #.

# Rationale for this change

Worked through the Python wheel release docs end-to-end on the `53.0.0-rc1-pypitest-3` tag and hit a few rough edges and stale pieces:

- The verification docs (\"Verifying Release Candidates\") only cover the Rust tarball script. They do not mention that RC verifiers can also install the wheels from TestPyPI to sanity-check the Python artifacts before voting.
- `dev/release/download-python-wheels.py` aborts in non-interactive shells because the gpg signing step needs a TTY for pinentry. There is no warning in the docs and the failure mode is confusing.
- The expected-files list pins the `manylinux_2_17` glibc tag, but the runner image now produces `manylinux_2_39` wheels (and that number will keep moving as the GitHub-hosted runner advances).
- The fallback advice says \"the `build-sdist` job currently does not upload its output as a workflow artifact.\" That is no longer true after #1753, which added a `dist-sdist` artifact. The local-rebuild command in that same block uses `maturin build --release --sdist --out dist`, which was the broken command (it produces both a wheel and an sdist) fixed in #1759.
- The per-platform fallback download still references `dist-windows-latest`, but #1751 renamed it to `dist-windows-2022`. It also omits the new `dist-sdist`.
- The TestPyPI smoke test runs `python -m venv` against whatever `python` is on PATH. On macOS that is 3.9, and pip then reports a misleading \"No matching distribution found\" because the wheels are `cp310-abi3` (Python ≥ 3.10).

`dev/release/README.md` only:

- New \"(Optional) Verify the Python wheels from TestPyPI\" subsection under \"Verifying Release Candidates\".
- Note next to `download-python-wheels.py` explaining the gpg/non-TTY failure mode and that the traceback is harmless for a TestPyPI-only dry-run.
- Wheel-filename list uses `manylinux_*` globs instead of the hard-coded `manylinux_2_17` tag, with a one-line explanation of why.
- Fallback per-platform `gh run download` adds `dist-sdist` and renames `dist-windows-latest` → `dist-windows-2022`.
- Drop the stale \"build-sdist does not upload\" paragraph; replace with a shorter note that local rebuild is rarely needed, using the correct `maturin sdist --out dist` command.
- Smoke-test venv uses `python3.10` and a comment explains why `python` alone is not enough.

No code, script, or workflow changes.

Documentation only. No runtime or API impact.

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -203,11 +203,41 @@ For the release to become "official" it needs at least three PMC members to vote
 
 ### Verifying Release Candidates
 
-The `dev/release/verify-release-candidate.sh` is a script in this repository that can assist in the verification process. Run it like:
+The `dev/release/verify-release-candidate.sh` is a script in this repository that can assist in the verification process. It downloads the source tarball from the ASF dev SVN, verifies the GPG signature and checksums, and builds the Rust workspace. Run it like:
 
 ```
 ./dev/release/verify-release-candidate.sh 0.11.0 0
 ```
+
+#### (Optional) Verify the Python wheels from TestPyPI
+
+If the release manager has uploaded the RC's Python wheels to
+[test.pypi.org](https://test.pypi.org/project/ballista/) as part of their
+pre-vote dry-run (see [Publish Python Wheels to PyPI](#publish-python-wheels-to-pypi)
+below), verifiers can install them in a throwaway virtualenv to sanity-check
+the artifacts that will ship to real PyPI. The wheels there are byte-identical
+to what would be uploaded to pypi.org if the vote passes.
+
+The wheels are built as `cp310-abi3`, so the venv needs Python ≥ 3.10:
+
+```bash
+export BALLISTA_VERSION=0.11.0    # version under vote
+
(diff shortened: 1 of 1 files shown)


dev/release/README.md
```diff
@@ -203,12 +203,42 @@ For the release to become "official" it needs at least three PMC members to vote
...
+```bash
+export BALLISTA_VERSION=53.0.0    # version under vote
+
+python3.10 -m venv /tmp/ballista-rc-verify
+source /tmp/ballista-rc-verify/bin/activate
+pip install -i https://test.pypi.org/simple/ \
+    --extra-index-url https://pypi.org/simple/ \
+    ballista==${BALLISTA_VERSION}
+python -c "from ballista import BallistaSessionContext; print('ok')"
+deactivate
```
martin-g, line 232: Is it expected/guaranteed that the new version will be at test.pypi.org ?
If the release manager always uploads the wheel to test.pypi.org then this snippet could be added to `dev/release/verify-release-candidate.sh` and done automatically.

PR #1768, opened 2026-05-25
Improve changelog generator and split CHANGELOG.md into per-version files

andygrove:

Closes #.

# Rationale for this change

- Simplify changelog generation so it no longer required manual effort
- Use the change log generator script from Comet that includes credits at the end showing commits per contributor
- Regenerated the change log for 53.0.0 using the new script - [here](https://github.com/andygrove/datafusion-ballista/blob/improve-changelog-generator/docs/source/changelog/53.0.0.md)
- Split the single every-growing change log page into one file per release and move to the documentation site so that it is easer for people to find

- Port `dev/release/generate-changelog.py` to the Comet version, adapted to Ballista's project name and CLI shape (`<tag1> <tag2> <version>`). The new script also resolves refs to SHAs before handing them to the GitHub API, which fixes a latent bug where `HEAD` was resolved by the API to the default branch instead of the caller's local HEAD.
- Add 19 per-version files under `docs/source/changelog/`, generated from the existing root `CHANGELOG.md`. Drop the noisy pre-split `apache/arrow-datafusion` umbrella-repo entries (`6.0.0`, `6.0.0-rc0`, `7.0.0`, `7.0.0-rc2`, `7.1.0-rc1`); the pre-migration `CHANGELOG.md` remains in git history if anyone needs to re-run the split.
- Add `docs/source/changelog/index.md` with a Sphinx toctree listing every kept release newest-first.
- Add a `Changelog` toctree caption to `docs/source/index.rst` so the per-version pages appear in the rendered site.
- Replace the root `CHANGELOG.md` with a short stub pointing at the new location.
- Delete the unused `.github_changelog_generator` config (Ruby tool, no longer authoritative).
- Update the `### Change Log` section of `dev/release/README.md` to document the single-command flow.

The published docs now include a "Changelog" caption in the top-level TOC linking to a separate page per release. The root `CHANGELOG.md` becomes a one-screen stub pointing readers at `docs/source/changelog/index.md`. No code or runtime behavior changes.


dev/release/generate-changelog.py
```diff
@@ -75,26 +75,83 @@ def generate_changelog(repo, repo_name, tag1, tag2):
...
+under the License.
+-->\n""")
+
+    print(f"# Apache DataFusion Ballista {version} Changelog\n")
+
+    # get the number of commits
+    commit_count = subprocess.check_output(f"git log --pretty=oneline {tag1}..{tag2} | wc -l", shell=True, text=True).strip()
+
+    # get number of contributors
+    contributor_count = subprocess.check_output(f"git shortlog -sn {tag1}..{tag2} | wc -l", shell=True, text=True).strip()
```
martin-g:
```suggestion
    shortlog_output = subprocess.check_output(["git", "shortlog", "-sn", f"{tag1}..{tag2}"], text=True)
    contributor_count = len(shortlog_output.strip().splitlines())
```

dev/release/generate-changelog.py
```diff
@@ -75,26 +75,83 @@ def generate_changelog(repo, repo_name, tag1, tag2):
...
+    print_pulls(repo_name, "Fixed bugs", bugs)
     print_pulls(repo_name, "Performance related", performance)
     print_pulls(repo_name, "Implemented enhancements", enhancements)
-    print_pulls(repo_name, "Fixed bugs", bugs)
     print_pulls(repo_name, "Documentation updates", docs)
-    print_pulls(repo_name, "Merged pull requests", all_pulls)
+    print_pulls(repo_name, "Other", other)
+
+    # show code contributions
+    credits = subprocess.check_output(f"git shortlog -sn {tag1}..{tag2}", shell=True, text=True).rstrip()
```
martin-g:
```suggestion
    credits = shortlog_output.rstrip()
```

dev/release/generate-changelog.py
martin-g:
```suggestion
        parts = re.findall(r'^([a-zA-Z]+)(\([a-zA-Z0-9_-]+\))?(!)?:', pull.title)
```
Conventional commits allow upper-case too

dev/release/generate-changelog.py
```diff
@@ -75,26 +75,83 @@ def generate_changelog(repo, repo_name, tag1, tag2):
...
+def resolve_ref(ref):
+    """Resolve a git ref (e.g. HEAD, branch name) to a full commit SHA."""
+    try:
+        return subprocess.check_output(
+            ["git", "rev-parse", ref], text=True
+        ).strip()
+    except subprocess.CalledProcessError:
+        # If it can't be resolved locally, return as-is (e.g. a tag name
+        # that the GitHub API can resolve)
+        return ref
```
martin-g:
```suggestion
        # If it can't be resolved locally, return as-is (e.g. a remote tag).
        # The GitHub API will attempt to resolve it.
        print(f"Note: Could not resolve '{ref}' locally; passing to GitHub API as-is", file=sys.stderr)
        return ref
```

dev/release/generate-changelog.py
```diff
@@ -75,26 +75,83 @@ def generate_changelog(repo, repo_name, tag1, tag2):
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->\n""")
+
+    print(f"# Apache DataFusion Ballista {version} Changelog\n")
+
+    # get the number of commits
+    commit_count = subprocess.check_output(f"git log --pretty=oneline {tag1}..{tag2} | wc -l", shell=True, text=True).strip()
```
martin-g:
```suggestion
    commit_count = subprocess.check_output(["git", "rev-list", "--count", f"{tag1}..{tag2}"], text=True).strip()
```

`git rev-list --count` is more efficient since there is no need to pipe to wc -l (also more Windows-friendly).
Also, using an array + string format only for the tags + shell=False prevents command injection.

dev/release/generate-changelog.py
```diff
@@ -75,26 +75,83 @@ def generate_changelog(repo, repo_name, tag1, tag2):
             cc_breaking = parts_tuple[2] == '!'
 
         labels = [label.name for label in pull.labels]
```
milenkovicm: Can we ignore `auto-dependencies` PRs? Currently i'm filtering them out manually

andygrove: Done in 3f327863 — the categorization loop now skips any PR labeled `auto-dependencies`, so Dependabot PRs are dropped automatically and no longer need manual filtering.

andygrove:
Thanks for the reviews @martin-g @milenkovicm. I addressed feedback now.

milenkovicm (approved):
Thanks @andygrove

martin-g (approved):

PR #1778, opened 2026-05-26
feat: runtime stage's metrics rendering

### LINKED ISSUE #1743: REST API to expose job runtime metrics

At the moment job info returns stage information but without runtime metrics which is collected after stage has run. Adding runtime metrics would help understanding job execution better 

Expose runtime metrics as part of job info

**Additional context**

open to proposal how this runtime metrics will be delivered

sandugood:

Closes #1743 .

 # Rationale for this change
Right now we don't expose stage-level metrics via Scheduler's REST API interface. We only have task level metrics (which are essential and explanatory by themselves, but not full). 

Introduced a new query parameter (`PlanFormat`) enum, which can be `Default` (no metrics and tree-style rendering; for backward compatibility), `Tree` (tree-style rendering) and `Metrics` (metrics of each stage of the execution)

`format_node` is responsible for rendering the metrics in right order

Results for a `tpchgen-cli -s 1 --format parquet --output-dir testdata` generated dataset with metrics rendering.

```
curl -s localhost:50050/api/job/cjaZtGS/stages?plan_format=metrics | jq -r ".stages.[1].stage_plan"
SortShuffleWriterExec: partitioning=Hash([c_custkey@0], 4), metrics=[write_time=158.15ms, output_rows=150.0 K, spill_count=0, repart_time=11.19ms, spill_time=4ns, input_rows=150.0 K, spill_bytes=0]
  AggregateExec: mode=Partial, gby=[c_custkey@0 as c_custkey], aggr=[sum(test_parquet.c_acctbal)], metrics=[elapsed_compute=94.16ms, aggregation_time=5.38ms, output_rows=150.0 K, peak_mem_used=10.16 M, output_bytes=45.0 MB, start_timestamp=2026-05-26 20:59:43.356634094 UTC, skipped_aggregation_rows=0, emitting_time=62.03µs, output_batches=20, aggregate_arguments_time=147.62µs, end_timestamp=2026-05-26 20:59:43.546108325 UTC, spilled_bytes=0.0 B, spill_count=0, reduction_factor=100% (150.0 K/150.0 K), spilled_rows=0, time_calculating_group_ids=87.23ms]
    DataSourceExec: ...
```

No, because users can continue to use `/api/job/{job_id}/stages` without a query parameter. Using `plan_format=tree` or `plan_format=metrics` is optional


milenkovicm (commented):
thanks @sandugood 

suggestion, can we have three options for this 

- no metrics, (default) same behaviour like we had 
- tree render
- with metrics, user requested

ballista/scheduler/src/api/handlers.rs
```diff
@@ -523,11 +523,16 @@ pub async fn get_query_stages<
...
+                        summary.stage_plan = running_stage.stage_metrics.as_deref().map(|m| {
+                            format_stage_plan_with_metrics(running_stage.plan.as_ref(), m, render_tree)
+                        }).or_else(|| {
+                            // no metrics yet
+                            Some(if render_tree {
+                                displayable(running_stage.plan.as_ref()).tree_render().to_string()
+                            } else {
+                                displayable(running_stage.plan.as_ref()).indent(false).to_string()
+                            })
+                        });
```
martin-g: I think this could be simplified to:
```rust
                        summary.stage_plan = Some(format_stage_plan_with_metrics(
                            running_stage.plan.as_ref(),
                            running_stage.stage_metrics.as_deref().unwrap_or(&[]),
                            render_tree,
                        ));
```

ballista/scheduler/src/api/handlers.rs
```diff
@@ -682,6 +687,85 @@ fn task_duration_percentiles(tasks: &[Option<TaskSummary>]) -> Option<Percentile
...
+        "metric count mismatch: consumed {} but stage_metrics has {}",
+        metric_idx,
+        stage_metrics.len()
+    );
+
+    result
+}
+
+/// Formatting the node in DFS fashion
+/// It is constructed in the same way it is collected (using in-order traversal)
```
martin-g:
```suggestion
/// It is constructed in the same way it is collected (using pre-order traversal)
```
`root then children` is pre-order, no ?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -682,6 +687,85 @@ fn task_duration_percentiles(tasks: &[Option<TaskSummary>]) -> Option<Percentile
...
+        impl std::fmt::Display for SingleNode<'_> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                self.0.fmt_as(DisplayFormatType::Default, f)
+            }
+        }
+        SingleNode(plan).to_string()
+    };
+
+    let prefix = "  ".repeat(indent);
+    let mut result = format!("{}{}{}\n", prefix, node_line, metric_str);
```
martin-g: This could be optimized by passing a mutable String at the render start call and using `write!()` instead of `format!()`. This will avoid many String allocations.

ballista/scheduler/src/api/handlers.rs
```diff
@@ -682,6 +687,85 @@ fn task_duration_percentiles(tasks: &[Option<TaskSummary>]) -> Option<Percentile
...
+/// For reference how the metrics are collected on the executor's side - see ballista-core/utils.rs
+fn format_node(
+    plan: &dyn ExecutionPlan,
+    stage_metrics: &[MetricsSet],
+    metric_idx: &mut usize,
+    indent: usize,
+) -> String {
+    let metrics_set = if plan.metrics().is_some() {
+        let m = stage_metrics.get(*metric_idx);
+        *metric_idx += 1;
```
martin-g: Shouldn't this be incremented even when `plan.metrics().is_none()` ?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -682,6 +687,85 @@ fn task_duration_percentiles(tasks: &[Option<TaskSummary>]) -> Option<Percentile
...
+        m
+    } else {
+        None
+    };
+
+    let metric_str = metrics_set
+        .map(|m| {
+            let aggregated = m.aggregate_by_name();
```
martin-g:
```suggestion
            let aggregated = m.aggregate_by_name().sorted_for_display().timestamps_removed();
```
as at https://github.com/apache/datafusion-ballista/blob/7a96c9480632c865c6e387a93dcba8457dca3bc2/ballista/scheduler/src/display.rs#L136-L139

ballista/scheduler/src/api/handlers.rs
sandugood: Addressed this with a comment above the code section

`collect_plan_metrics()`, which behavior I am copying,works only with `Some(metrics)`

ballista/scheduler/src/api/handlers.rs
sandugood: Yes, fixed. Thanks

sandugood:
Introduced a new enum (which has a default value). Going to add info about the API change to PR's overview

ballista/scheduler/src/api/handlers.rs
```diff
@@ -204,8 +205,20 @@ pub struct QueryStageSummary {
...
+    pub plan_format: Option<PlanFormat>,
+}
+
+#[derive(Debug, serde::Deserialize, Default, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum PlanFormat {
+    /// ?plan_format=default => plain indent, no metrics
+    #[default]
+    Default,
+    /// ?plan_format=tree => tree render, no metrics
```
martin-g, line 218: Please also update the parameter name at https://github.com/apache/datafusion-ballista/blob/main/ballista-cli/src/tui/http_client.rs#L132
We can add support for `metrics` in a follow-up.

martin-g (approved):

ballista/scheduler/src/api/handlers.rs
sandugood: Fixed. After this one gets merged I can take care of the tui part, if you dont mind

milenkovicm: https://github.com/apache/datafusion-ballista/blob/main/dev/bin/showplan.sh

Has param as well

ballista/scheduler/src/api/handlers.rs
```diff
@@ -382,16 +395,17 @@ pub async fn get_job<
     let percent_complete =
         ((completed_stages as f32 / num_stages as f32) * 100_f32) as u8;
 
-    let render_tree = query.render_tree.unwrap_or(false);
+    let plan_format = query.plan_format.clone().unwrap_or(PlanFormat::Default);
```
milenkovicm: Is there unwrap or default method ?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -490,7 +504,7 @@ pub async fn get_query_stages<
     Path(job_id): Path<String>,
     query: Query<JobQueryParams>,
 ) -> Result<impl IntoResponse, SchedulerErrorResponse> {
-    let render_tree = query.render_tree.unwrap_or(false);
+    let plan_format = query.plan_format.clone().unwrap_or(PlanFormat::Default);
```
milenkovicm: Unwrap or default?

milenkovicm (commented):
Few minor comments, will try it tomorrow, may have a bit more

ballista/scheduler/src/api/handlers.rs
sandugood: Yeah, unwrap_or_default() seems like an appropriate thing here. Thanks

ballista/scheduler/src/api/handlers.rs
sandugood: Changed that script with a new `-m` param

milenkovicm (approved):
I think this makes sense, thanks @sandugood 

One comment for investigation before we merge this

there is similar code at : 

- https://github.com/apache/datafusion-ballista/blob/fe2e37a0915167671caf8a9d9a8c24545659ba2b/ballista/core/src/execution_plans/distributed_explain_analyze.rs#L265
- https://github.com/apache/datafusion-ballista/blob/fe2e37a0915167671caf8a9d9a8c24545659ba2b/ballista/scheduler/src/display.rs#L133

would it be possible to create a single function and used across three places?

sandugood:
> I think this makes sense, thanks @sandugood
> 
> One comment for investigation before we merge this
> 
> there is similar code at :
> 
> * https://github.com/apache/datafusion-ballista/blob/fe2e37a0915167671caf8a9d9a8c24545659ba2b/ballista/core/src/execution_plans/distributed_explain_analyze.rs#L265
> * https://github.com/apache/datafusion-ballista/blob/fe2e37a0915167671caf8a9d9a8c24545659ba2b/ballista/scheduler/src/display.rs#L133
> 
> would it be possible to create a single function and used across three places?

Would dig into it today

sandugood:
Thank you for pointing out the display implementation. It turns out that we already had a battle-tested implementation of node visiting with metrics🙂. 
Reused that, which improved the code structure. This is about addressing the second point in your comment

About the first comment, tackling the explain analyze part might be suitable for another PR? wdyt?

milenkovicm:
> About the first comment, tackling the explain analyze part might be suitable for another PR? wdyt?

fine with me, thanks @sandugood

milenkovicm:
thanks @sandugood and @martin-g

PR #1779, opened 2026-05-26
[WIP]feat(aqe): port Spark's OptimizeSkewedJoin

wirybeaver:
## Summary

Ports Spark's [`OptimizeSkewedJoin`](https://github.com/apache/spark/blob/v3.5.1/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/OptimizeSkewedJoin.scala) AQE rule into Ballista. When a shuffle partition is disproportionately large (skewed), the rule splits it into smaller sub-ranges and replicates the matching partition on the other side of the join, so the work that was a single straggler task becomes N parallel tasks.

Replaces the file-list sharding approach (#1718), which had to bail on any `HashPartitioned` / `SinglePartition` consumer — effectively all joins and `FinalPartitioned` aggregates.

### Design

| Spark | Ballista |
|---|---|
| `OptimizeSkewedJoin` | `OptimizeSkewedJoinRule` in `state/aqe/optimizer_rule/` |
| `MapOutputStatistics` | `ExchangeExec::shuffle_partitions()` + `partition_stats.num_bytes()` |
| `PartialReducerPartitionSpec(reducerIdx, [startMap, endMap))` | `SkewJoinShard { upstream_idx, start_map_idx, end_map_idx }` |
| Cartesian pair-up across legs | Pre-computed `Vec<SkewJoinShard>` of length N on both legs |

Rule is **opt-in**: `SET ballista.planner.skew_join.enabled = true`. Defaults match Spark (factor=5.0, threshold=256 MiB, advisory=64 MiB).

### Current limitations

- **v1 only handles single-join stages.** Bails when a stage subtree has more than one binary join. Per-join iteration is future work.
- **Mutually exclusive with DataFusion's dynamic-filter pushdown.** The skew rewrite breaks the hash-colocation invariant that dynamic filters assume. A guard enforces `SET datafusion.optimizer.enable_dynamic_filter_pushdown = false`.
- **Split granularity is per upstream task, not per row group.** Single-file tables produce 1 upstream task per partition — nothing to subdivide. Tables must span multiple files. (Spark doesn't have this limitation because it splits large files into multiple scan tasks automatically.)

---

### Setup

```bash
cd /home/user/optimize-skewed-join
cargo build --release -p ballista-scheduler -p ballista-executor -p ballista-cli

# Scheduler with AQE debug logging
RUST_LOG=info,ballista_scheduler::state::aqe=debug \
  ./target/release/ballista-scheduler --bind-port 50050 > /tmp/scheduler.log 2>&1 &
sleep 3

# Executor
RUST_LOG=info \
  ./target/release/ballista-executor \
    --scheduler-host localhost --scheduler-port 50050 \
    --bind-port 50051 --concurrent-tasks 4 > /tmp/executor.log 2>&1 &
```

Generate synthetic skewed data — 20 files, ~90% of rows have key=0:
```bash
mkdir -p /tmp/skewed_dir
for i in $(seq 0 19); do
  duckdb -c "
    COPY (
      SELECT (random() < 0.9)::int AS key, range AS val
      FROM range(50000)
    ) TO '/tmp/skewed_dir/part_${i}.parquet' (FORMAT parquet);
  "
done

duckdb -c "
  COPY (SELECT range AS key, range AS val FROM range(100))
  TO '/tmp/dim.parquet' (FORMAT parquet);
"
```

> **Note:** `ballista-cli` must be invoked with `--host localhost --port 50050`. Without these flags it runs queries in-process and the scheduler is never contacted.

### Baseline (rule OFF)

```bash
cat > /tmp/run-baseline.sql << 'EOSQL'
SET ballista.planner.adaptive.enabled = true;
SET ballista.planner.skew_join.enabled = false;
SET datafusion.optimizer.enable_dynamic_filter_pushdown = false;

CREATE EXTERNAL TABLE fact STORED AS PARQUET LOCATION '/tmp/skewed_dir/';
CREATE EXTERNAL TABLE dim  STORED AS PARQUET LOCATION '/tmp/dim.parquet';

SELECT dim.val, count(*) FROM fact JOIN dim ON fact.key = dim.key GROUP BY dim.val;
EOSQL

./target/release/ballista-cli --host localhost --port 50050 \
  -f /tmp/run-baseline.sql > /tmp/baseline.out 2>&1
```

### Rewritten (rule ON)

```bash
cat > /tmp/run-rewritten.sql << 'EOSQL'
SET ballista.planner.adaptive.enabled = true;
SET ballista.planner.skew_join.enabled = true;
SET ballista.planner.skew_join.skewed_partition_factor = 1.5;
SET ballista.planner.skew_join.skewed_partition_threshold_bytes = 1024;
SET ballista.planner.skew_join.advisory_partition_bytes = 1024;
SET datafusion.optimizer.enable_dynamic_filter_pushdown = false;

CREATE EXTERNAL TABLE fact STORED AS PARQUET LOCATION '/tmp/skewed_dir/';
CREATE EXTERNAL TABLE dim  STORED AS PARQUET LOCATION '/tmp/dim.parquet';

SELECT dim.val, count(*) FROM fact JOIN dim ON fact.key = dim.key GROUP BY dim.val;
EOSQL

./target/release/ballista-cli --host localhost --port 50050 \
  -f /tmp/run-rewritten.sql > /tmp/rewritten.out 2>&1
```

### Verify

```bash
# Results match
diff <(grep -E '^\|' /tmp/baseline.out | sort) \
     <(grep -E '^\|' /tmp/rewritten.out | sort)

# Rule attached
grep -a 'attaching SkewJoinPlan' /tmp/scheduler.log
```

### Results

```
attaching SkewJoinPlan: M=16 K'=34
  left_skewed=[…, true, …, true, …]   (partitions 3 and 5)
  left_sizes=[…, 10817136, …, 1207824, …]
  right_skewed=all false               (dim table — 100 rows)
```

- **Correctness:** rows identical between baseline and rewritten.
- **Skew detection:** 2 of 16 partitions flagged (10.8 MB and 1.2 MB vs near-zero for the rest).
- **Split:** 34 downstream join tasks instead of 16. Non-skewed side replicated for each sub-range.
- **Mutual-exclusion guard:** verified separately — `enable_dynamic_filter_pushdown=true` triggers bail as expected.

Supersedes #1718.

### DIFF
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -101,6 +101,31 @@ pub const BALLISTA_COALESCE_SMALL_PARTITION_FACTOR: &str =
 pub const BALLISTA_COALESCE_MERGED_PARTITION_FACTOR: &str =
     "ballista.planner.coalesce.merged_partition_factor";
 
+/// Configuration key to enable AQE optimize-skewed-join rule (Spark's
+/// `OptimizeSkewedJoin`). Disabled by default — opt in for workloads where
+/// per-key skew creates straggler tasks on join inputs.
+pub const BALLISTA_SKEW_JOIN_ENABLED: &str = "ballista.planner.skew_join.enabled";
+/// Multiplier over the per-side median below which a partition is not
+/// considered skewed. Mirrors Spark's
+/// `spark.sql.adaptive.skewJoin.skewedPartitionFactor`.
+pub const BALLISTA_SKEW_JOIN_SKEWED_PARTITION_FACTOR: &str =
+    "ballista.planner.skew_join.skewed_partition_factor";
+/// Absolute byte threshold below which a partition is never split, even when
+/// the ratio-to-median qualifies. Mirrors Spark's
+/// `spark.sql.adaptive.skewJoin.skewedPartitionThresholdInBytes`.
+pub const BALLISTA_SKEW_JOIN_SKEWED_PARTITION_THRESHOLD_BYTES: &str =
+    "ballista.planner.skew_join.skewed_partition_threshold_bytes";
+/// Advisory target sub-shard byte size used by the bin-packer when splitting
+/// a skewed partition's per-mapper outputs. Decoupled from the coalesce
+/// target so the two rules can be tuned independently.
+pub const BALLISTA_SKEW_JOIN_ADVISORY_PARTITION_BYTES: &str =
+    "ballista.planner.skew_join.advisory_partition_bytes";
+/// Tail-merge factor: a trailing sub-shard whose size is below
+/// `small_partition_factor * advisory` folds back into its predecessor
+/// (Spark's `splitSizeListByTargetSize` legacy).
+pub const BALLISTA_SKEW_JOIN_SMALL_PARTITION_FACTOR: &str =
+    "ballista.planner.skew_join.small_partition_factor";
+
 /// Result type for configuration parsing operations.
 pub type ParseResult<T> = result::Result<T, String>;
 use std::sync::LazyLock;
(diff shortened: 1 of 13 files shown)


martin-g:
@wirybeaver Please resolve the conflicts!
Update: Sorry, I didn't notice that it is still a draft!

milenkovicm:
closing as open for too long without progress, please reopen if still needed and updated

PR #1792, opened 2026-05-29
feat: add `/api/job/{job_id}/config` endpoint to retrieve session configuration

### LINKED ISSUE #1783: REST API to provide job configuration information

When debugging applications it is hard to understand which `SessionConfiguration` values are used with the job, making job a bit harder

Add additional endpoint `/api/job/{job_id}/config` which should print map if job `SessionConfig` values. 
To be investigated to we keep configuration with the job info, and if we don't to be added.

**Additional context**

killzoner:

Closes https://github.com/apache/datafusion-ballista/issues/1783.

 # Rationale for this change

Being able to debug configuration for any already executed job.

Add `/api/job/{job_id}/config` endpoint in REST API

Documentation is updated

Possible follow-up : add this to TUI to allow inspecting the config

### DIFF
--- a/ballista/scheduler/src/state/task_manager.rs
+++ b/ballista/scheduler/src/state/task_manager.rs
@@ -35,6 +35,7 @@ use ballista_core::serde::protobuf::{
 };
 use ballista_core::serde::scheduler::ExecutorMetadata;
 use dashmap::DashMap;
+use datafusion::execution::config::SessionConfig;
 use datafusion::execution::context::SessionContext;
 use datafusion::logical_expr::LogicalPlan;
 use datafusion::physical_plan::ExecutionPlan;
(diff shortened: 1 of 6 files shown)


ballista/scheduler/src/state/task_manager.rs
```diff
@@ -440,6 +440,16 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
         }
     }
 
+    /// Get the session configuration for a job.
+    pub async fn get_job_config(&self, job_id: &str) -> Result<HashMap<String, String>> {
```
milenkovicm: can we make this method return `SessionConfig` and we we map to hash map in the handler code ?

milenkovicm (approved):
thanks @killzoner 

one comment to address if you agree

ballista/scheduler/src/state/task_manager.rs
killzoner: will do!

killzoner: done

milenkovicm:
thanks @killzoner 
we can leave TUI integration until @martin-g finishes #1742

killzoner:
Thank you for the review!

PR #1805, opened 2026-06-02
minor: code cleanup and log level changes 

milenkovicm:

Closes #.

 # Rationale for this change

- log statements are too noisy especially if there are many tasks running (high partitioning)
-  scheduler `substrait` feature is affecting compilation time without actually being used 

-  scheduler `substrait` feature is disabled by default, as its not commonly used but affects code compilation time 
- removal of `WarnOnDuplicateExecRule` as not very useful anymore
- as we have quite a lot of job information exposed we do not need such noisy loggers anymore.

yes, if `substrait` scheduler support is needed it should be enabled


milenkovicm:
merging this as i want to continue with dynamic join improvements

PR #1806, opened 2026-06-02
minor(aqe): backport few datafusion physical rules identified as not idempotent 

milenkovicm:

Closes #.

 # Rationale for this change

- https://github.com/apache/datafusion/pull/22523
- ~https://github.com/apache/datafusion/pull/22522~ (removed as it produces wrong result for some reason)

rules intensified by @wirybeaver as non idempotent and fixed in datafusion. as it will not get into datafusion 54 i've backported them to ballista codebase for now.

- ~two rules copied over~
- a rule copied over
- creation of default datafusion optimizer set is done locally in ballista at the moment 

no, only affecting AQE which is disabled by default


milenkovicm:
fyi  @wirybeaver 
will merged change on green ci

milenkovicm:
for some reason it produces wrong result, keeping it on hold for now

PR #1808, opened 2026-06-03
feat(tui): Show shimmer animation for "Queued" and "Running" jobs

### LINKED ISSUE #1787: TUI: Show a spinner for the Running jobs

It would be nice to add a "loading indicator" for the jobs with "Running" status. To show that something is happening on the background.

We may use one of the small spinners at https://github.com/sorinirimies/tui-spinner#fluxspinner--all-frame-presets next to the `Running` label in the `Status` column

**Additional context**

We have to see how the re-rendering of the spinner works. If it needs the application tick event then we have to decide whether to decrease the tick time from 2000ms to a lower value because this will increase the frequency of the REST calls to the Scheduler.

martin-g:

Closes #1787 

# Rationale for this change

Use an animation to tell the user that the job is not finished yet.

Use https://github.com/vinhnx/tui-shimmer as an indicator that the job is not finished yet

https://github.com/user-attachments/assets/f2343c7e-14c0-4b05-bd79-5032834c2bb3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1061,6 +1061,7 @@ dependencies = [
  "tracing-subscriber",
  "tracing-web",
  "tui-big-text",
+ "tui-shimmer",
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-sys",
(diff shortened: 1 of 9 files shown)


martin-g:
Set as a Draft because the web tui breaks for some reason.

martin-g:
https://github.com/user-attachments/assets/9fe43ab2-5bda-451f-b16c-fe50e64f198e

The tui-shimmer widget does not work so well in the web version :-/

martin-g:
https://github.com/user-attachments/assets/57cc1f64-506f-417e-8621-6186a16b8ebb

fixed it!

milenkovicm (approved):
cool, thanks @martin-g

PR #1818, opened 2026-06-04
feat(scheduler): Use command line options for CORS origins/methods instead of env vars

### LINKED ISSUE #1804: Replace CORS env vars with config properties
The Scheduler's Axum setup uses environment variables to configure custom CORS allowed origins and/or methods in #1742.
The usage of env vars is not consistent with the rest of the application. The Scheduler config should be used instead!

_Originally posted by @milenkovicm in https://github.com/apache/datafusion-ballista/pull/1742#discussion_r3339974868_

martin-g:

Closes #1804

# Rationale for this change

Use command line options instead of environment variables for consistency.

Add http://localhost:8080 (local dev) and
https://nightlies.apache.org (convenience deployment) to the allowed origins

The scheduler binary now has two more options:
* `--cors-allowed-origins "..."` (`*` allows any, `""` uses the defaults, anything else sets custom ones)
* `--cors-allowed-methods "..."` (`*` allows all methods, `""` uses the defaults - GET, OPTIONS, PATCH, anything else sets custom ones)

### DIFF
--- a/ballista/scheduler/src/api/routes.rs
+++ b/ballista/scheduler/src/api/routes.rs
@@ -91,8 +91,8 @@ fn allowed_origins(config: Arc<SchedulerConfig>) -> AllowOrigin {
     };
 
     match config.cors_allowed_origins.trim() {
-        origins if origins.is_empty() => AllowOrigin::list(default_origins()),
-        origins if origins == "*" => AllowOrigin::any(),
+        "" => AllowOrigin::list(default_origins()),
+        "*" => AllowOrigin::any(),
         origins => {
             let mut header_values = Vec::new();
             for origin in origins.split(',') {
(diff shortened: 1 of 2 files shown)


milenkovicm:
> Add http://localhost:8080/ (local dev) and
> https://nightlies.apache.org/ (officially supported on-premise deployment) to the allowed origins

i don't think we should say "officially supported", its just a convenience method used for development purposes, which users should be able to override. 

I'd suggest to expose / with redirect, which can be overriden so by default it sends users to nightlies.apache.org with some kind of versioning, or users can point redirect to wherever they have started their tui service, 

wdyt @martin-g ?

martin-g:
I don't quite understand the suggestion. Please explain further!

milenkovicm:
sorry for confusion, 

make a scheduler `/` redirect by default to `https://nightlies.apache.org/ballsita/<ballista version>` but we can expose option for users to change that redirect to, lets say `/tui` if they want to ship local version. 

this way users will have web tui available without need for them to compile it and deploy it

Issue #1807, opened 2026-06-02
[EPIC] TUI Improvements


milenkovicm:
As we have so many great improvements and ideas on TUI it would be great to add an epic to track them all

- [x] #1797 
- [x] #1787 (PR: https://github.com/apache/datafusion-ballista/pull/1808)
- [x] #1785 
- [x] #1804 
- [x] #1849
- [x] #1997

Ideas to be discussed

- [x] `t` binding to switch plan to tree
- [x] `m` binding to add plan metrics 
- [x] update readme with `tui web` part 
- [ ] create a blog post
- [ ] SSE job modifications (i.e. update the tui state with Server Side Events (push) instead of HTTP polls).
- [ ] deploy tui to a known url that we can trust, forward to and be publicly available

milenkovicm:
i hope you approve adding epic for TUI @martin-g

martin-g:
I added the TUI label few days ago 🙂

jiengup:
Maybe I can help with this.

martin-g:
> Maybe I can help with this.

Thanks for offering your help!
Just let us know when you start working on something, so we don't make double efforts!

jiengup:
> > Maybe I can help with this.
> 
> Thanks for offering your help! Just let us know when you start working on something, so we don't make double efforts!

I want to try to close https://github.com/apache/datafusion-ballista/issues/1797 if you haven't started yet.

martin-g:
Sure! Go ahead! 
It should be similar to showing the jobs' stages popup.
Don't hesitate to propose improvements in the old/current code if you see something!

sandugood:
Would like to work on the `plan_format` (of the REST API) integration into TUI, if that is not currently taken
So we could switch between different rendering modes there

milenkovicm:
is there a way to make web tui not so dark?

martin-g:
> is there a way to make web tui not so dark?

<img width="1559" height="842" alt="Image" src="https://github.com/user-attachments/assets/368e305e-e3e5-4c4b-a1fd-1287e181da9f" />

With the new [theming](https://github.com/apache/datafusion-ballista/issues/1849) support you could set a custom background color with:
```json
"theme": {
  "name": "dark",
  "overrides": {
    "app_background": {"bg": "#rrggbb"}
  }
}
```

milenkovicm:
we have redundant information on job page

[...]

PR #1828, opened 2026-06-07
feat(TUI): enable various plan rendering formats

sandugood:

Adresses #1807

 # Rationale for this change
`Scheduler`'s REST API interface now supports different styles of plan's rendering for stages. Beforehand, in the Ballista's TUI we could only render the "default" looking plan.

Switching between different formatting styles for stage's plan.

Yes. 
Now we have three possible variants of stage's plan formatting:
1. Default (which is enabled by default when pressing `p` in the job's Stages section. Also we can switch back to it by pressing `d`
2. Tree-format - by pressing `t`
3. Metrics - by pressing `m`

## Here is the demonstration:
https://github.com/user-attachments/assets/50b9844a-65c9-4a9b-b6c9-ba1c7cfea81a

### DIFF
--- a/ballista-cli/src/tui/app.rs
+++ b/ballista-cli/src/tui/app.rs
@@ -249,8 +249,6 @@ impl App {
                     _ => {}
                 }
             } else if popup.is_plan_view() {
-                use crate::tui::domain::jobs::stages::StagePlanTab;
-
                 // Collect job_id + tab to fetch while popup is still borrowed,
                 // then drop the borrow before the async call.
                 let fetch = match key.code {
@@ -287,7 +285,7 @@ impl App {
                 };
 
                 if let Some((job_id, tab)) = fetch
-                    && let Err(e) = load_stage_plan(self, &job_id, tab).await
+                    && let Err(e) = load_stage_plan(self, &job_id, &tab).await
                 {
                     tracing::error!("Failed to load stage plan: {e:?}");
                 }
@@ -988,8 +986,6 @@ impl App {
                     _ => {}
                 }
             } else if popup.is_plan_view() {
-                use crate::tui::domain::jobs::stages::StagePlanTab;
-
                 let fetch = match key.code {
                     KeyCode::Char('d') => popup
                         .set_tab(StagePlanTab::Default)
--- a/ballista-cli/src/tui/app.rs
+++ b/ballista-cli/src/tui/app.rs
@@ -53,6 +53,7 @@ use crate::tui::http_client::HttpClient;
 use crate::tui::ui::{
     load_executor_details_popup, load_executors_data, load_job_details, load_job_dot,
     load_job_stages_popup, load_jobs_data, load_metrics_data,
+    load_stage_plan,
 };
 
 const INVALID_DATE: &str = "Invalid date";
@@ -221,13 +222,32 @@ impl App {
                     _ => {}
                 }
             } else if popup.is_plan_view() {
-                match key.code {
-                    KeyCode::Esc => popup.set_no_details_view(),
-                    KeyCode::Up => popup.scroll_up(),
-                    KeyCode::Down => popup.scroll_down(),
-                    KeyCode::Left => popup.scroll_left(),
-                    KeyCode::Right => popup.scroll_right(),
-                    _ => {}
+                use crate::tui::domain::jobs::stages::StagePlanTab;
+
+                // Collect job_id + tab to fetch while popup is still borrowed,
+                // then drop the borrow before the async call.
+                let fetch = match key.code {
+                    KeyCode::Char('d') => popup
+                        .set_tab(StagePlanTab::Default)
+                        .map(|tab| (popup.job_id.clone(), tab)),
+                    KeyCode::Char('t') => popup
+                        .set_tab(StagePlanTab::Tree)
+                        .map(|tab| (popup.job_id.clone(), tab)),
+                    KeyCode::Char('m') => popup
+                        .set_tab(StagePlanTab::Metrics)
+                        .map(|tab| (popup.job_id.clone(), tab)),
+                    KeyCode::Esc => { popup.set_no_details_view(); None }
+                    KeyCode::Up => { popup.scroll_up();    None }
+                    KeyCode::Down => { popup.scroll_down();  None }
+                    KeyCode::Left => { popup.scroll_left();  None }
+                    KeyCode::Right => { popup.scroll_right(); None }
+                    _ => None,
+                };
+
+                if let Some((job_id, tab)) = fetch {
+                    if let Err(e) = load_stage_plan(self, &job_id, tab).await {
+                        tracing::error!("Failed to load stage plan: {e:?}");
+                    }
                 }
             } else if popup.is_no_details_view() {
                 match key.code {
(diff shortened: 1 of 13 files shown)


milenkovicm:
Can we change to tree on job plan as well?

sandugood:
Right now this is set via config. 
Also, when pressing `p` in the main `Jobs` menu (with all of the jobs listed), the popup window already shows:
- stage plan
- physical plan
- logical plan

Not sure if there is really room for another window (that would be switching between default stage plan and tree-style stage plan), but I'm opened to suggestions.

milenkovicm:
When physica plan shown for a specific job can we press t it similar to show it as a tree ?

sandugood:
> When physica plan shown for a specific job can we press t it similar to show it as a tree ?

Yeah, going to add it to the job list plan as well. So users might switch physical plan rendering between "default" and "tree"
After committing going to show a video snip as well

martin-g:
The banner text does not render well for you for some reason.

ballista-cli/src/tui/app.rs
```diff
@@ -797,6 +831,11 @@ impl App {
             UiData::JobStagesData(job_id, stages) => {
                 self.job_stages_popup = Some(JobStagesPopup::new(job_id, stages));
             }
+            UiData::JobStagesPlanData(_job_id, tab, stages) => {
```
martin-g: I was going to suggest to not pass `job_id` if it won't be used but I think it might be better to check that it is the id of the currently edited job. To prevent a late Ajax response modifying something unrelated.

ballista-cli/src/tui/domain/jobs/stages.rs
```diff
@@ -92,6 +108,10 @@ impl JobStagesPopup {
         Self {
             job_id,
             scrollbar_state: ScrollbarState::new(stages.stages.len()),
+            plan_cache: PlanCache {
+                default: Some(stages.clone()),
```
martin-g, line 133: This is not always true.
There is an TUI config `job.stage.plan.tree` that may say that the `tree` rendering is the default.
We can drop the config now since it will be very easy to change the plan rendering type.

ballista-cli/src/tui/app.rs
```diff
@@ -221,13 +221,47 @@ impl App {
...
-                    KeyCode::Down => popup.scroll_down(),
-                    KeyCode::Left => popup.scroll_left(),
-                    KeyCode::Right => popup.scroll_right(),
-                    _ => {}
+                use crate::tui::domain::jobs::stages::StagePlanTab;
+
+                // Collect job_id + tab to fetch while popup is still borrowed,
+                // then drop the borrow before the async call.
+                let fetch = match key.code {
+                    KeyCode::Char('d') => popup
```
martin-g, line 255: You will have to do the same for the Web TUI.
At the moment the key event handing is duplicated due to differences between Crossterm/Ratzilla event handling (async/sync) :-/

ballista-cli/src/tui/ui/main/jobs/mod.rs
```diff
@@ -102,6 +103,40 @@ pub async fn load_job_stages_popup(app: &App, job_id: &str) -> TuiResult<()> {
...
+#[cfg(not(feature = "web"))]
+pub async fn load_stage_plan(
+    app: &App,
+    job_id: &str,
+    tab: crate::tui::domain::jobs::stages::StagePlanTab,
+) -> TuiResult<()> {
+    use crate::tui::domain::jobs::stages::StagePlanTab;
+
+    let fmt = match tab {
+        StagePlanTab::Default => None,
```
martin-g:
```suggestion
        StagePlanTab::Default => Some(""),
```
or `Some("default")`

to bypass the TUI config setting `job.stage.plan.tree`

ballista-cli/src/tui/domain/jobs/stages.rs
```diff
@@ -102,6 +122,36 @@ impl JobStagesPopup {
...
+            StagePlanTab::Tree => self.plan_cache.tree = Some(resp.clone()),
+            StagePlanTab::Metrics => self.plan_cache.metrics = Some(resp.clone()),
+        }
+        // If we're currently on that tab, update the live stages too so the
+        // selection / scroll state is preserved.
+        let active_fmt = self.active_plan_format();
+        if active_fmt == Some(fmt) {
+            self.stages = resp;
+        }
+    }
```
martin-g, line 156: The cloning can be avoided if the tab is not the current active one:

```suggestion
    pub fn cache_plan_response(&mut self, fmt: StagePlanTab, resp: JobStagesResponse) {
        let active_fmt = self.active_plan_format();
        if active_fmt.as_ref() == Some(&fmt) {
            self.stages = resp.clone();
        }
        match fmt {
            StagePlanTab::Default => self.plan_cache.default = Some(resp),
            StagePlanTab::Tree => self.plan_cache.tree = Some(resp),
            StagePlanTab::Metrics => self.plan_cache.metrics = Some(resp),
        }
    }
```

ballista-cli/src/tui/domain/jobs/stages.rs
```diff
@@ -102,6 +122,36 @@ impl JobStagesPopup {
...
+
+    #[allow(dead_code)]
+    pub fn cached_response(&self, tab: &StagePlanTab) -> Option<JobStagesResponse> {
+        match tab {
+            StagePlanTab::Default => self.plan_cache.default.clone(),
+            StagePlanTab::Tree => self.plan_cache.tree.clone(),
+            StagePlanTab::Metrics => self.plan_cache.metrics.clone(),
+        }
+    }
```
martin-g: Better return a reference
```suggestion
    pub fn cached_response(&self, tab: &StagePlanTab) -> Option<&JobStagesResponse> {
        match tab {
            StagePlanTab::Default => self.plan_cache.default.as_ref(),
            StagePlanTab::Tree => self.plan_cache.tree.as_ref(),
            StagePlanTab::Metrics => self.plan_cache.metrics.as_ref(),
        }
    }
```

ballista-cli/src/tui/ui/footer.rs
```diff
@@ -80,6 +80,12 @@ pub(super) fn render_footer(f: &mut Frame, area: Rect, app: &App) {
                         } else if app.is_job_stage_plan_popup_open() {
                             current_view_key_bindings
                                 .push(Span::from("[↑↓] Scroll up/down, "));
```
martin-g: I see we missed to add the horizontal scrolling binding here.
Please add it too!

```suggestion
                                .push(Span::from("[↑↓←→] Scroll, "));
```

sandugood:
Thanks for review @martin-g 
Converting to draft and going to tackle it in subsequent days

ballista-cli/src/tui/app.rs
sandugood: You are right, it won't be used
Removed for now

ballista-cli/src/tui/domain/jobs/stages.rs
sandugood: Removed the config fully, because now user has the possibility to switch between different styles of physical plan formatting. 

Note: I am using the `default` as a naming convention (like default-style rendering of physical plan) and because pressing `d` on the keyboard seemed like a good choice

sandugood:
Here is the demonstration of both:
1. TUI

https://github.com/user-attachments/assets/62befecd-335c-4f76-ada8-f531907f4a54

2. Web TUI


https://github.com/user-attachments/assets/b585ce2c-5554-454d-981e-482057b8e10c

cc @martin-g

sandugood:
> The banner text does not render well for you for some reason.

By the way, have no idea why it renders gibberish on my Ubuntu :disappointed:

martin-g (approved):

martin-g:
Thank you, @sandugood !

PR #1844, opened 2026-06-09
chore: re enable approx distinct test

coderfender:

Closes #.

 # Rationale for this change


milenkovicm:
`id` sorted is 

```
+----+
| id |
+----+
| 0  |
| 1  |
| 2  |
| 3  |
| 4  |
| 5  |
| 6  |
| 7  |
+----+
```

i'm not sure what the correct value should be, but it complains about the expected one

coderfender:
Yeah . I ran the checks locally and the expected answer is 3 and not 1 . Pushing a commit to rectify that soon

coderfender:
Bumped expected value and verified on local that DF produces the new value `3` instead of `1`

milenkovicm (approved):
Thanks @coderfender good catch

PR #1848, opened 2026-06-09
refactor: make use of `async_trait::async_trait` consistent across project

### LINKED ISSUE #1847: Inconsistent usage of `#[tonic::async_trait]`,  `#[async_trait]` and `#[async_trait::async_trait]`.
This is obviously just a nit, and I am only opening this to reference in a PR if that is desired. Going through the codebase (working on #1829 ), I have noticed that there is an inconsistent usage of the `async_trait` macro, where at some places it is using the re-export of tonic (e.g. `#[tonic::async_trait]`) and in other places it uses the macro from async_trait directly (e.g. `#[async_trait::async_trait]` or a `use` + `#[async_trait]`). 

## Using tonic's re-export
I guess it makes sense to use tonic's re-export in case of gRPC-related service impls, to signal the relation with gRPC. But I found myself confused by it in two specific places:

- `trait ClusterState`:  [mod.rs](https://github.com/apache/datafusion-ballista/blob/7be5a3f7668cce29c6a53afb3be155003a3a2273/ballista/scheduler/src/cluster/mod.rs#L154-L158)
- `trait JobState`: [mod.rs](https://github.com/apache/datafusion-ballista/blob/7be5a3f7668cce29c6a53afb3be155003a3a2273/ballista/scheduler/src/cluster/mod.rs#L277-L281)

The impl's of both traits are **not** using tonic's re-export, so I propose to change those two instances to use the `async_trait` directly. 

## Using `#[async_trait::async_trait]` or a `use` + `#[async_trait]`)
The project uses the fully qualified version 7 times and the `use` + unqualified (do I say that correctly?) version 14 times. Not really a preference but using the fully qualified version will shorten the `use` blocks and makes it visually consistent. 

Is it okay to put a PR in for this? Or rather not?

JarroVGIT:

Closes #1847 

 # Rationale for this change
Inconsistent usage of `#[tonic::async_trait]`, `#[async_trait]` and `#[async_trait::async_trait]` leads to unnecessary cognitive load on developer.

- Changed all `#[async_trait]` into `#[async_trait::async_trait]` and removed the `use async_trait::async_trait;` statements.
- Changed two occurences of `#[tonic::async_trait]` to `#[async_trait::async_trait]` on trait definitions that are not gRPC related.

No.

### DIFF
--- a/ballista/scheduler/src/test_utils.rs
+++ b/ballista/scheduler/src/test_utils.rs
@@ -25,7 +25,6 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
-use async_trait::async_trait;
 
 use crate::config::SchedulerConfig;
 use crate::metrics::SchedulerMetricsCollector;
--- a/ballista/scheduler/src/test_utils.rs
+++ b/ballista/scheduler/src/test_utils.rs
@@ -25,7 +25,6 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
-
 use crate::config::SchedulerConfig;
 use crate::metrics::SchedulerMetricsCollector;
 use crate::planner::DefaultDistributedPlanner;
--- a/ballista/scheduler/src/test_utils.rs
+++ b/ballista/scheduler/src/test_utils.rs
@@ -80,7 +80,7 @@ const TEST_SCHEDULER_NAME: &str = "localhost:50050";
 #[derive(Debug)]
 pub struct ExplodingTableProvider;
 
-#[async_trait]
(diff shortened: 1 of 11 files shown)


ballista/scheduler/src/test_utils.rs
```diff
@@ -25,7 +25,6 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
-use async_trait::async_trait;
```
martin-g:
```suggestion
```

martin-g (approved):

martin-g:
Thank you, @JarroVGIT !

PR #1850, opened 2026-06-09
feat(tui): Add theming support

### LINKED ISSUE #1849: [TUI]: Add support for UI themes

The TUI/WebTUI use should be able to specify custom colors for all UI elements in the application, like text labels, table headers, table's selected row, etc.

Provide default dark and light themes.
Provide ways to inherit themes and customize just parts of them.

Introduce a Theme with fields for all styled UI elements in the application. 

**Additional context**

martin-g:

Closes #1849 

# Rationale for this change

Some users prefer light terminal background. Others - dark,
It would be good to support UI themes, so the users could change the colors of the application

Introduced Theme and settings.
A user could add something like:
```
  theme:
    name: "dark"
    overrides:
      table_header:
        fg: "DarkBlue"
        bg: "White"
        add_modifier: "UNDERLINED|ITALIC"
      status_running:
        fg: "#rrggbb"
        bg: "108"
``` 
to their config and overwrite the defaults.
Two pre-defined base themes are available: `dark` and `light`. The user could use any of them and override any/all of their properties.

New settings are available in the application config.


milenkovicm:
image

i got that log, not sure if its related

milenkovicm (approved):
thanks @martin-g 

not sure if we can pre-pack `light` and `dark` theme ?

milenkovicm:
ah sorry, did not see its still draft

martin-g:
> i got that log, not sure if its related

It is not related to the theming. It is introduced with https://github.com/apache/datafusion-ballista/pull/1808. I will fix !

martin-g:
Here is the default light theme:

image

martin-g:
This is ready for review!

ballista-cli/src/tui/infrastructure/config.rs
```diff
@@ -81,10 +91,10 @@ impl Settings {
                         .format(FileFormat::Yaml)
                         .required(false),
                 )
-                // Add in settings from the environment (with a prefix of BALLISTA_)
-                // E.g. `BALLISTA_SCHEDULER_URL=http://localhost:50051 ballista_cli`
+                // Add in settings from the environment (with a prefix of `BALLISTA`.)
+                // E.g. `BALLISTA__SCHEDULER__URL=http://localhost:50051 ballista_cli`
```
milenkovicm, line 95: Why double underscores ?

milenkovicm (approved):
Thanks @martin-g 
TUI is becoming datafusion sub project on its own 😀

One question, how do I change theme of web tui?

ballista-cli/src/tui/infrastructure/config.rs
martin-g: Because some of the properties use underscore, e.g. `table_header` and `app_background` and the `config` crate cannot resolve `BALLISTA_THEME_OVERRIDES_TABLE_HEADER_BG=color`
There was no such issue with the top level settings. BALLISTA_DATA_RELOAD_INTERVAL_MS=123 properly sets the `data_reload_interval_ms` setting.

martin-g:
> One question, how do I change theme of web tui?

At the moment the only way is via query string: 
`?ballista_theme_name=dark&ballista_theme_overrides_app_background_bg=darkgrey`
The theme name is optional, since `dark` is the default anyway.

Only these two theme related settings are supported in the WebTUI. It is bit painful to parse those from the query string.

I am considering adding a new page/popup to the application for editing the config. In the first version it would be a simple TextArea that dumps the Settings struct as Yaml. On edit the Settings is updated and also stored in ~/.config/... (for TUI) and in localStorage for WebTUI. Asking the users to manipulate the localStorage via the browser console would be bad.
But as you said - the TUI is becoming more and more complex :-/

milenkovicm:
i think this is good enough, having a way to hack the settings is good for now

milenkovicm:
let others contribute as well 😀

Issue #1851, opened 2026-06-09
`ExecutorMetatada` too verbose


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`ExecutorMetadata` contains `ExecutorSpecification` & `ExecutorOperatingSystemSpecification `

```rust
pub struct ExecutorMetadata {
    /// Unique executor identifier.
    pub id: String,
    /// Hostname or IP address of the executor.
    pub host: String,
    /// Port number for data transfer.
    pub port: u16,
    /// Port number for gRPC communication.
    pub grpc_port: u16,
    /// Resource specification for this executor.
    pub specification: ExecutorSpecification,
    /// OS and hardware info for this executor
    pub os_info: ExecutorOperatingSystemSpecification,
}
```

where `ExecutorOperatingSystemSpecification ` has more data (quite few of them strings

```rust
pub struct ExecutorOperatingSystemSpecification {
    /// System name
    pub system_name: String,
    /// Kernel version
    pub kernel_ver: String,
    /// OS version
    pub os_ver: String,
    /// OS version (long)
    pub os_ver_long: String,
    /// Number of physical cores available on this executor
    pub physical_cores: u32,
    /// Number of physical disks available on this executor
    pub num_disks: u32,
    /// Total disk space on this executor, in bytes
    pub total_disk_space: u64,
    /// Total available disk space on this executor, in bytes
    pub total_available_disk_space: u64,
    /// Open files limit on this executor
    pub open_files_limit: u64,
}
```

problem is that `ExecutorMetadata` is part of `PartitionLocation`

```rust
#[derive(Debug, Clone)]
pub struct PartitionLocation {
    /// The source partition ID from the map stage.
    pub map_partition_id: usize,
    /// The partition identifier.
    pub partition_id: PartitionId,
    /// Metadata about the executor hosting this partition.
    pub executor_meta: ExecutorMetadata,
    /// Statistics about the partition data.
    pub partition_stats: PartitionStats,
    /// shuffle file id
    pub file_id: Option<u64>,
    /// whether this partition uses sort shuffle
    pub is_sort_shuffle: bool,
}
```

which represent partition locations and which is going around from/to scheduler and executor, also it might be cached as part of stage informations. which would take a bit of space and might attribute #1836 

**Describe the solution you'd like**

Trim down or change information/data structure representing executor metadata, 

**Describe alternatives you've considered**

[...]

PR #1852, opened 2026-06-09
feat: NewType pattern for `job_id`

JarroVGIT:

Draft PR.

Closes #.

 # Rationale for this change
`job_id` and `job_name` (and also, `session_id` and `executor_id`) are all `String` and `&str` throughout the code base. It can make it hard to reason about HashMaps that are keyed by such identifier (`HashMap<String, SomeType>` is not obvious in what it stores). It also makes it easy to make a mistake when passing such items to functions that expect a few of them. For example `fn function(job_name: &str, job_id: &str) -> ()` is valid if called with reversed arguments.

NewType also brings new code and overhead. This PR is created as a draft to see how it would look like if implemented for `job_id` and `job_name`. 

- Notably: the wire format remains unchanged (e.g. the generated code still expects `String`s). 
- Created the newtypes (through a macro that also implements all the nice traits, such as `Display`, `From<String>`, `From<&str>`, `From<JobId> for String` (for unwrapping back into proto structs), `AsRef<str>` and `Borrow<str>` (which allows for `map.get("job-123")` without allocating a JobId per lookup.
- Replaced all `job_name: String` and `job_id: String` in internal structs (non-proto)
- Followed the compiler and fixed the errors. 

I am going to assume there will be, yes, but I haven't paid any attention to it yet.

### DIFF
--- a/ballista/core/src/execution_plans/distributed_query.rs
+++ b/ballista/core/src/execution_plans/distributed_query.rs
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::JobId;
 use crate::client::BallistaClient;
 use crate::config::BallistaConfig;
 use crate::extension::{BallistaConfigGrpcEndpoint, SessionConfigExt};
@@ -84,7 +85,7 @@ pub struct DistributedQueryExec<T: 'static + AsLogicalPlan> {
     /// - job_scheduling_in_ms: Time job waited in scheduler queue (started_at - queued_at)
     metrics: ExecutionPlanMetricsSet,
     /// The scheduler job id after the query has been accepted.
-    job_id: Arc<Mutex<Option<String>>>,
+    job_id: Arc<Mutex<Option<JobId>>>,
 }
 
 impl<T: 'static + AsLogicalPlan> DistributedQueryExec<T> {
@@ -136,7 +137,7 @@ impl<T: 'static + AsLogicalPlan> DistributedQueryExec<T> {
     }
 
     /// Returns the scheduler job id after the query has been accepted.
-    pub fn job_id(&self) -> Option<String> {
+    pub fn job_id(&self) -> Option<JobId> {
         self.job_id.lock().clone()
     }
 
@@ -336,7 +337,7 @@ async fn execute_query_pull(
     max_message_size: usize,
     grpc_config: GrpcClientConfig,
     metrics: Arc<ExecutionPlanMetricsSet>,
-    job_id_handle: Arc<Mutex<Option<String>>>,
+    job_id_handle: Arc<Mutex<Option<JobId>>>,
     partition: usize,
     session_config: SessionConfig,
 ) -> Result<impl Stream<Item = Result<RecordBatch>> + Send> {
@@ -394,7 +395,7 @@ async fn execute_query_pull(
         "Session id inconsistent between Client and Server side in DistributedQueryExec."
     );
 
-    let job_id = query_result.job_id;
+    let job_id: JobId = query_result.job_id.into();
     *job_id_handle.lock() = Some(job_id.clone());
     let mut prev_status: Option<job_status::Status> = None;
 
@@ -404,7 +405,7 @@ async fn execute_query_pull(
             flight_proxy,
         } = scheduler
             .get_job_status(GetJobStatusParams {
-                job_id: job_id.clone(),
+                job_id: job_id.clone().into(),
             })
             .await
             .map_err(|e| DataFusionError::Execution(format!("{e:?}")))?
@@ -509,7 +510,7 @@ async fn execute_query_push(
     max_message_size: usize,
     grpc_config: GrpcClientConfig,
     metrics: Arc<ExecutionPlanMetricsSet>,
-    job_id_handle: Arc<Mutex<Option<String>>>,
+    job_id_handle: Arc<Mutex<Option<JobId>>>,
     partition: usize,
     session_config: SessionConfig,
 ) -> Result<impl Stream<Item = Result<RecordBatch>> + Send> {
@@ -568,11 +569,12 @@ async fn execute_query_push(
             status,
             flight_proxy,
         } = item;
-        let job_id = status
+        let job_id: JobId = status
             .as_ref()
             .map(|s| s.job_id.to_owned())
-            .unwrap_or("unknown_job_id".to_string()); // should not happen
-        if !job_id.starts_with("unknown_") {
+            .unwrap_or("unknown_job_id".to_string()) // should not happen
+            .into();
+        if !job_id.as_str().starts_with("unknown_") {
             let mut shared_job_id = job_id_handle.lock();
             if shared_job_id.is_none() {
                 *shared_job_id = Some(job_id.clone());
@@ -761,6 +763,7 @@ async fn fetch_partition(
 
 #[cfg(test)]
 mod test {
+    use crate::JobId;
     use crate::config::BallistaConfig;
     use crate::execution_plans::distributed_query::{
         DistributedQueryExec, get_client_host_port,
@@ -835,7 +838,7 @@ mod test {
             LogicalPlan::default(),
             "session".to_string(),
         ));
-        *exec.job_id.lock() = Some("job-123".to_string());
+        *exec.job_id.lock() = Some("job-123".to_owned().into());
 
         let new_exec = exec.clone().with_new_children(vec![]).unwrap();
         let new_exec = new_exec
@@ -843,6 +846,6 @@ mod test {
             .downcast_ref::<DistributedQueryExec<LogicalPlanNode>>()
             .unwrap();
 
-        assert_eq!(new_exec.job_id().as_deref(), Some("job-123"));
+        assert_eq!(new_exec.job_id(), Some(JobId::new("job-123")));
     }
 }
--- a/ballista/core/src/execution_plans/distributed_query.rs
+++ b/ballista/core/src/execution_plans/distributed_query.rs
@@ -838,7 +838,7 @@ mod test {
             LogicalPlan::default(),
             "session".to_string(),
         ));
-        *exec.job_id.lock() = Some("job-123".to_owned().into());
+        *exec.job_id.lock() = Some("job-123".into());
 
         let new_exec = exec.clone().with_new_children(vec![]).unwrap();
         let new_exec = new_exec
--- a/ballista/core/src/execution_plans/sort_shuffle/spill.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/spill.rs
@@ -73,7 +73,7 @@ impl SpillManager {
     /// * `compression` - Compression codec for spill files
     pub fn new(
         work_dir: &str,
-        job_id: &str,
+        job_id: &JobId,
         stage_id: usize,
         input_partition: usize,
         schema: SchemaRef,
--- a/ballista/core/src/execution_plans/sort_shuffle/spill.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/spill.rs
@@ -21,6 +21,7 @@
 //! At finalization, the spill bytes are concatenated verbatim into the
 //! consolidated output file alongside the in-memory remainder.
 
+use crate::JobId;
 use crate::error::{BallistaError, Result};
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::arrow::ipc::reader::StreamReader;
@@ -80,7 +81,7 @@ impl SpillManager {
         compression: CompressionType,
     ) -> Result<Self> {
         let mut spill_dir = PathBuf::from(work_dir);
-        spill_dir.push(job_id);
+        spill_dir.push(job_id.as_str());
         spill_dir.push(format!("{stage_id}"));
         spill_dir.push(format!("{input_partition}"));
         spill_dir.push("spill");
@@ -273,7 +274,7 @@ mod tests {
 
         let mut manager = SpillManager::new(
             temp_dir.path().to_str().unwrap(),
-            "job1",
+            &JobId::new("job1"),
             1,
             0,
             schema.clone(),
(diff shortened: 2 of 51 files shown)


milenkovicm:
Does it make sense to have a type for job name? It does not have anu special meaning

JarroVGIT:
Alright it compiles again. It was quite interesting to walk through the entire codebase and I ran into several places where I had to dig around to find out if `HashMap<String, HashMap<String, SomeType>>` contained a job id. The following was the most interesting (to me):

- [scheduler - state/aqe/planner.rs](https://github.com/apache/datafusion-ballista/blob/fe36e9301e68c22e5b63c01eeb59f4a9698deb67/ballista/scheduler/src/state/aqe/planner.rs#L167-L169): This calls a function that takes a `job_id` parameter, but a `job_name` is passed.
- [scheduler - scheduler_server/event.rs](https://github.com/apache/datafusion-ballista/blob/0e897e86a877b71adf578a21e1d8536283d551af/ballista/scheduler/src/scheduler_server/event.rs#L153-L155): The composed string claims this variant contains a job id, while the call sites only populate it with an executer id.
- Some signatures take values by owned value, others take it by reference. In some cases, I'd argue that we could do a bit of a clean up. For example, any `Struct::new()` should take parameters that are going to be owned values in the struct as owned values, not as references. (See also [here](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control) in the Rust API guidelines).
- Although it was a few hours work (of which I am sure most of it could've been automated or LLM-ed), it was relatively easy to just (1) change struct fields with find+replace, (2) change signatures with find+replace and (3) fix all compiler errors. In the last step you'll eventually run into the places where it actually began for me; the `HashMap<String, _>` and such. It was quite fun, actually.

@milenkovicm :
> Does it make sense to have a type for job name? It does not have anu special meaning

I thought I pick JobId and JobName because they are often closely used (at call sites of functions). JobId was definitely the most important one due to its usage in several container types though. I do think that ExecutorId would also benefit from this. 

The downside of this is the massive PR that it requires (you kinda have to change it all, or none of it). I had to touch 50 files apparently to make it whole again. I guess it can be split in the different Id's but then still, I think for `JobId` alone it was already 46 files. So, not sure if that is worth the risk/effort to be honest.

milenkovicm (commented):
thanks @JarroVGIT I will have a look in details, if you agree, i would suggest not to introduce type for `JobName` as it has no specific meaning nor limitations (perhaps we should remove it from public interfaces where appropriate). 

regarding `ExecutorId` should we consider using uuid which that id is (and avoid use of strings for uuid)

milenkovicm:
changing `executor_id` to uuid can be done as separate PR, once we wrap up this one and #1851

JarroVGIT:
> thanks @JarroVGIT I will have a look in details, if you agree, i would suggest not to introduce type for `JobName` as it has no specific meaning nor limitations (perhaps we should remove it from public interfaces where appropriate).

Agree, I will remove `JobName` from this draft. Will be later today I think, or tomorrow morning.

> regarding `ExecutorId` should we consider using uuid which that id is (and avoid use of strings for uuid)

I will open an issue to discuss ExecutorId separately, if that's okay with you

ballista/core/src/execution_plans/sort_shuffle/spill.rs
```diff
@@ -273,7 +274,7 @@ mod tests {
 
         let mut manager = SpillManager::new(
             temp_dir.path().to_str().unwrap(),
-            "job1",
+            &JobId::new("job1"),
```
martin-g: IMO this would be friendlier with `"job1".into()`

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -151,7 +152,7 @@ impl AdaptivePlanner {
     pub async fn try_new(
         ctx: &SessionContext,
         logical_plan: &LogicalPlan,
-        job_name: String,
+        job_name: JobName,
```
martin-g: Maybe this should be changed to `JobId` and update the callers ?!
Or pass both if needed.

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -359,12 +364,10 @@ impl AdaptivePlanner {
...
                         // plan in `default_optimizers()`.
                         let plan = CoalescePartitionsRule.optimize(plan, config)?;
-                        BallistaAdapter::adapt_to_ballista(
-                            plan,
-                            self.job_name.as_str(),
-                            config,
-                        )
-                        .map(|w| (w.plan.stage_id(), w))
+                        // adapt_to_ballista takes an job_id, we are passing a job_name. Need to transform to fix compiler.
+                        let job_id = self.job_name.clone().into_inner().into();
```
martin-g: I have the feeling `self.job_name` really should have been named `self.job_id`

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -568,11 +569,12 @@ async fn execute_query_push(
             status,
             flight_proxy,
         } = item;
-        let job_id = status
+        let job_id: JobId = status
             .as_ref()
             .map(|s| s.job_id.to_owned())
-            .unwrap_or("unknown_job_id".to_string()); // should not happen
-        if !job_id.starts_with("unknown_") {
+            .unwrap_or("unknown_job_id".to_string()) // should not happen
```
martin-g, line 575:
```suggestion
            .unwrap_or_else(|| "unknown_job_id".to_string()) // should not happen
```

ballista/executor/src/execution_loop.rs
martin-g:
```suggestion
    let job_id = task.job_id.into();
```

ballista/executor/src/execution_loop.rs
```diff
@@ -275,7 +275,7 @@ async fn run_received_task<T: 'static + AsLogicalPlan, U: 'static + AsExecutionP
         })?;
 
     let query_stage_exec = executor.execution_engine.create_query_stage_exec(
-        job_id.clone(),
+        job_id.clone().into(),
```
martin-g:
```suggestion
        job_id.clone(),
```

ballista/executor/src/execution_loop.rs
```diff
@@ -285,7 +285,7 @@ async fn run_received_task<T: 'static + AsLogicalPlan, U: 'static + AsExecutionP
     dedicated_executor.spawn(async move {
         use std::panic::AssertUnwindSafe;
         let part = PartitionId {
-            job_id: job_id.clone(),
+            job_id: job_id.clone().into(),
```
martin-g:
```suggestion
            job_id: job_id.clone(),
```

ballista/core/src/execution_plans/distributed_query.rs
```diff
@@ -835,14 +838,14 @@ mod test {
             LogicalPlan::default(),
             "session".to_string(),
         ));
-        *exec.job_id.lock() = Some("job-123".to_string());
+        *exec.job_id.lock() = Some("job-123".to_owned().into());
```
martin-g: All `"job_id".to_owned().into()` could be shortened to `"job_id".into()`

ballista/scheduler/src/scheduler_server/event.rs
```diff
@@ -150,6 +150,7 @@ impl Debug for QueryStageSchedulerEvent {
             QueryStageSchedulerEvent::JobDataClean(job_id) => {
                 write!(f, "JobDataClean : job_id={job_id}.")
             }
+            // TODO: This is not job_id but Executor ID (based on usage).
```
martin-g, line 153: Shall we introduce ExecutorId too ? I think it would be good!

ballista/core/src/execution_plans/sort_shuffle/spill.rs
```diff
@@ -73,14 +74,14 @@ impl SpillManager {
     /// * `compression` - Compression codec for spill files
     pub fn new(
         work_dir: &str,
-        job_id: &str,
+        job_id: &JobId,
```
JarroVGIT, line 77: Here is an example where references are passed, only for the associated function to make it owned values. If there is interest in changing this into all `::new()` and alike functions, I can do a pass in another PR. Low value stuff I reckon, but I find it's the kind of stuff that also helps me get a feel of navigating the codebase so I don't mind doing it.

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -164,7 +165,11 @@ impl AdaptivePlanner {
         );
 
         let plan = state.create_physical_plan(logical_plan).await?;
-        let plan = handle_explain_plan(&job_name, ctx, logical_plan, plan)
+
+        // Note: the signature requires a JobId, but we are passing a JobName. The below is a
```
JarroVGIT, line 169: Should I make the `handle_explain_plan` a bit more generic (e.g. name the parameter `identifier: &str`)?

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -359,12 +364,10 @@ impl AdaptivePlanner {
                         // that would arise if the rule walked the entire residual
                         // plan in `default_optimizers()`.
                         let plan = CoalescePartitionsRule.optimize(plan, config)?;
-                        BallistaAdapter::adapt_to_ballista(
-                            plan,
-                            self.job_name.as_str(),
-                            config,
-                        )
-                        .map(|w| (w.plan.stage_id(), w))
+                        // adapt_to_ballista takes an job_id, we are passing a job_name. Need to transform to fix compiler.
```
JarroVGIT, line 367: Not sure how to deal with this, is there a preference or is a comment and the current solution sufficient?

JarroVGIT:
@milenkovicm I have removed `JobName` again.

@martin-g I saw your comments and started to fix a few, but also saw you have resolved them all afterwards. I have made the following changes:
- Replace `&JobId::new("job-1")` with `&"job-1".into()`.  It actually saves another import in some files, which was a nice benefit.
- Replaced several `to_owned().into()` with `into()` where appropriate. I remember contemplating if I should or should not implement `From<&str>`, decided to implement it and then apparently did not use it (whoops). 

There are a few comments in there with questions, where I am not sure how to approach it best.

martin-g:
I played with some new Github UI button that resolved in one click all my comments. There was no way to undo/unresolve :-)

I agree with @milenkovicm that only ids should have their own types. I think even the executor_id should be ExecutorId(Uuid). Uuid is an implementation detail. It might change to Ulid later if needed and the places using ExecutorId won't notice. 
Or if another Id type is introduced that is also based on Uuid. It would be good to use different (new)types for them, to not mix them.

ballista/core/src/execution_plans/sort_shuffle/spill.rs
JarroVGIT: Fixed, though because we have to pass a reference, it is `&"job1".into()`.

ballista/scheduler/src/state/aqe/planner.rs
JarroVGIT: Yeah I agree, that makes more sense. I will create a new issue for it an draft an PR.

ballista/scheduler/src/state/aqe/planner.rs
JarroVGIT: Correct, I will take this in another PR to fix the `AdaptivePlanner` and the call sites.

ballista/core/src/execution_plans/distributed_query.rs
JarroVGIT: I am totally okay with applying that change but I didn't change that particular code. Out of curiosity, is `unwrap_or_else` over `unwrap_or`  convention for Ballista?

ballista/scheduler/src/state/aqe/planner.rs
JarroVGIT: See #1859

ballista/scheduler/src/state/aqe/planner.rs
JarroVGIT: See #1859

ballista/executor/src/execution_loop.rs
JarroVGIT: Applied

ballista/executor/src/execution_loop.rs
JarroVGIT: Applied

ballista/core/src/execution_plans/distributed_query.rs
JarroVGIT: Yes, that was a good catch! I have applied it everywhere.

ballista/scheduler/src/scheduler_server/event.rs
JarroVGIT: Agreed, I will make another issue/PR

JarroVGIT:
@martin-g Ha yes that happens sometimes, I was able to see your comments nevertheless and I have applied them.

ballista/core/src/execution_plans/distributed_query.rs
martin-g: The idea is to postpone the String allocation until really needed.
It is not Ballista specific.

ballista/core/src/lib.rs
```diff
@@ -47,6 +50,8 @@ pub mod event_loop;
 pub mod execution_plans;
 /// Extension traits and utilities for DataFusion integration.
 pub mod extension;
+/// Strongly-typed string identifiers (job ids, job names, ...).
```
martin-g, line 53:
```suggestion
/// Strongly-typed string identifiers (job ids, ...).
```

martin-g (approved):

milenkovicm (approved):
LGTM thanks @JarroVGIT

milenkovicm:
thanks @JarroVGIT and @martin-g

milenkovicm:
oops, ive merged PR with few suggestions, I apologise, should not started reviewing without finishing my first coffee :) please open follow up PR if needed

PR #1866, opened 2026-06-15
fix: use DataFrame.__name__ instead of hardcoded string in metaclass method wrapping

djanand:

Fixes an existing TODO, which required removing the string comparison 

Closes #.

 # Rationale for this change
The metaclass in extension.py identifies DataFrame-returning methods by comparing` __annotations__["return"]` against the hardcoded string `"DataFrame"`. This works only by coincidence. It happens to match because the class name is DataFrame. If the class were renamed or imported under an alias, the comparison would silently stop matching and methods would no longer be wrapped, breaking distributed execution without any error.

  Replacing `"DataFrame"` with `DataFrame.__name__` ties the comparison to the actual imported class, so it stays correct automatically if the name ever changes. Two unit tests are added to guard this behavior. They verify at class-creation time that all DataFrame-returning methods on DataFrame and SessionContext are correctly wrapped in DistributedDataFrame and BallistaSessionContext respectively, without requiring a live cluster.

Touched upon in the rationale section above. 

N/A

### DIFF
--- a/python/Cargo.lock
+++ b/python/Cargo.lock
@@ -642,6 +642,7 @@ dependencies = [
  "futures",
  "http",
  "insta",
+ "itertools",
  "log",
  "object_store",
  "parking_lot",
@@ -652,6 +653,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tonic",
+ "tower-http",
  "uuid",
 ]
 
--- a/python/python/ballista/extension.py
+++ b/python/python/ballista/extension.py
@@ -72,12 +72,11 @@ class RedefiningDataFrameMeta(type):
 
         for base_name, base_value in bases[0].__dict__.items():
             #
-            # TODO: could we not use 'DataFrame' as a string here?
             #
             if (
                 callable(base_value)
                 and not base_name.startswith("__")
-                and base_value.__annotations__.get("return") == "DataFrame"
+                and base_value.__annotations__.get("return") == DataFrame.__name__
             ):
                 #
                 # functions which return DataFrame are redefined
@@ -112,12 +111,11 @@ class RedefiningSessionContextMeta(type):
 
         for base_name, base_value in bases[0].__dict__.items():
             #
-            # could we not use 'DataFrame' as a string here?
             #
             if (
                 callable(base_value)
                 and not base_name.startswith("__")
-                and base_value.__annotations__.get("return") == "DataFrame"
+                and base_value.__annotations__.get("return") == DataFrame.__name__
             ):
                 #
                 # functions which return DataFrame are redefined
--- a/python/python/ballista/extension.py
+++ b/python/python/ballista/extension.py
@@ -71,8 +71,6 @@ class RedefiningDataFrameMeta(type):
             return method_wrapper
 
         for base_name, base_value in bases[0].__dict__.items():
-            #
-            #
             if (
                 callable(base_value)
                 and not base_name.startswith("__")
@@ -110,8 +108,6 @@ class RedefiningSessionContextMeta(type):
             return method_wrapper
 
         for base_name, base_value in bases[0].__dict__.items():
-            #
-            #
             if (
                 callable(base_value)
                 and not base_name.startswith("__")
--- a/python/python/tests/test_context.py
+++ b/python/python/tests/test_context.py
@@ -16,6 +16,7 @@
 # under the License.
 
 from ballista import BallistaSessionContext, setup_test_cluster
+from ballista.extension import DataFrame, DistributedDataFrame, SessionContext
 from datafusion import col, lit
 import pytest
 import pyarrow as pa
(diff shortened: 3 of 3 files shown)


coderfender:
LGTM pending CI . Thank you @djanand

coderfender (approved):
Does this mean that if someone changes `__name__` attribute, this would still break ?

python/python/tests/test_context.py
```diff
@@ -138,3 +138,33 @@ def test_write_json(ctx, tmp_path):
     df.write_json(out_dir)
     json_files = list((tmp_path / "out").glob("*.json"))
     assert len(json_files) > 0
+
+
+def test_distributed_dataframe_wraps_dataframe_returning_methods():
+    from ballista.extension import DistributedDataFrame, DataFrame
```
coderfender: We could generalize imports in both tests to top level and DRY

python/python/tests/test_context.py
```diff
@@ -138,3 +138,33 @@ def test_write_json(ctx, tmp_path):
...
+    should_be_wrapped = {
+        name
+        for name, val in SessionContext.__dict__.items()
+        if callable(val)
+        and not name.startswith("__")
+        and val.__annotations__.get("return") == DataFrame.__name__
+    }
+
+    assert should_be_wrapped
+    assert should_be_wrapped.issubset(BallistaSessionContext.__dict__)
```
coderfender: Can we add more exhaustive checks in the class assertion here ?

python/python/tests/test_context.py
```diff
@@ -138,3 +138,33 @@ def test_write_json(ctx, tmp_path):
...
+        name
+        for name, val in DataFrame.__dict__.items()
+        if callable(val)
+        and not name.startswith("__")
+        and val.__annotations__.get("return") == DataFrame.__name__
+    }
+
+    assert should_be_wrapped
+    assert should_be_wrapped.issubset(DistributedDataFrame.__dict__)
+
```
coderfender, line 162: nit : a lot of code here seems duped . Could be extracted to a test method @djanand

djanand:
> oes this mean that if someone changes `__name__` attribute, this would still break ?

 Both DataFrame.__name__ and the annotation string "DataFrame" would change together if DataFusion renames the class, so they stay in sync. But if
  they hardcoded "DataFrame" and the class was renamed to e.g. `DataFrameV2`, the annotation would update but the hardcoded string wouldn't, causing silent breakage.

python/python/tests/test_context.py
djanand: agree. moved the imports to the top

python/python/tests/test_context.py
djanand: added a bit more exhaustive checks. ptal

python/python/tests/test_context.py
djanand: @coderfender makes sense. Moved it to a common fn.

python/python/ballista/extension.py
```diff
@@ -112,12 +111,11 @@ def method_wrapper(*args, **kwargs):
 
         for base_name, base_value in bases[0].__dict__.items():
             #
```
martin-g:
```suggestion
```
Let's remove the whole comment. Not just the middle of it

python/python/ballista/extension.py
```diff
@@ -112,12 +111,11 @@ def method_wrapper(*args, **kwargs):
...
-            # could we not use 'DataFrame' as a string here?
             #
```
martin-g:
```suggestion

```

python/Cargo.lock
```diff
@@ -642,6 +642,7 @@ dependencies = [
  "futures",
  "http",
  "insta",
+ "itertools",
```
martin-g, line 645: Why are those dependencies added ?
They do not look related to the changes above

python/python/ballista/extension.py
```diff
@@ -72,12 +72,11 @@ def method_wrapper(*args, **kwargs):
 
         for base_name, base_value in bases[0].__dict__.items():
             #
```
martin-g:
```suggestion
```

python/python/ballista/extension.py
```diff
@@ -72,12 +72,11 @@ def method_wrapper(*args, **kwargs):
...
-            # TODO: could we not use 'DataFrame' as a string here?
             #
```
martin-g:
```suggestion

```

python/Cargo.lock
djanand: @martin-g - The `python/` workspace depends on `ballista-core` and `ballista-scheduler` via path deps, and those crates recently gained `itertools` and `tower-http` as direct dependencies. The lock file just reflects that. i.e **no new** deps were introduced by this PR.
The failed CI run without lock change is [here](https://github.com/apache/datafusion-ballista/actions/runs/27530389236/job/81366868681)

Did a bit of digging and `pyproject.toml` has `locked = true` under `[tool.maturin]`, which passes --locked to cargo, requiring python/Cargo.lock to be fully up to date and committed. 
Let me know if I'm missing anything

python/python/ballista/extension.py
djanand: got it. updated - ptal

python/python/ballista/extension.py
djanand: Sure. addressed - ptal

python/python/ballista/extension.py
djanand: updated. - ptal

python/python/ballista/extension.py
djanand: updated. - ptal

martin-g (approved):

martin-g:
Thank you, @djanand & @coderfender !

PR #1870, opened 2026-06-16
chore(ci): Use Ubuntu ARM64 for some tests on Linux

martin-g:

N/A

# Rationale for this change

Linux AMD64 runners are in shortage.
Experiment with ARM64

Change 'ubuntu-latest` with `ubuntu-24.04-arm` for few Linux CI jobs

No

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -138,7 +138,9 @@ jobs:
           submodules: true
           fetch-depth: 1
       - name: Setup Rust toolchain
-        uses: ./.github/actions/setup-builder
+        run: |
+          rustup toolchain install stable
+          rustup default stable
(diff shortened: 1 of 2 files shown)


martin-g:
Close + reopen to test the build caches

martin-g:
@milenkovicm WDYT about this ? The Linux ARM64 runners seems to start as soon as the macOs ones.
We can migrate some more Linux AMD64 to ARM64 if you see no issues with this.

milenkovicm:
i have no issues, but it would be good to keep some on x64

milenkovicm (approved):
lets try this @martin-g if you still want to do it, sorry i totally missed to approve it

martin-g:
I will close it for now.

The CI check will be as fast as the slowest CI job. 
The Linux x86_64 runners at Github have problems lately but this CPU arch is the most used for production deployments, so we must to test on it.

PR #1875, opened 2026-06-17
perf: address scheduler crash on high partition SF data

### LINKED ISSUE #1851: `ExecutorMetatada` too verbose

coderfender:

Closes https://github.com/apache/datafusion-ballista/issues/1851

 # Rationale for this change

We currently have more than what is needed in PartitionMetadata causing scheduler OOM on high cardinality data

Implement PartitionLocation a lean, address-only struct instead of the full metadata:

```
  #[derive(Debug, Clone, PartialEq, Eq)]
  pub struct PartitionLocationMetadata {
      pub id: String,
      pub host: String,
      pub port: u16,
      pub grpc_port: u16,
  }

  pub struct PartitionLocation {
      ...
      pub executor_meta: Arc<PartitionLocationMetadata>,
  }

```
  - New From<&ExecutorMetadata> for PartitionLocationMetadata is the single place
  that maps an executor to its location address.
  - Arc-wrapped so all locations for the same executor (built in one
  partition_to_location call) share one allocation rather than N copies.
  - Proto PartitionLocation now carries only the four address fields, shrinking
  the serialized payload sent to executors/clients. (no contract changes per say)

No

No

### DIFF
--- a/ballista/core/src/serde/scheduler/from_proto.rs
+++ b/ballista/core/src/serde/scheduler/from_proto.rs
@@ -38,7 +38,7 @@ use crate::serde::protobuf::{NamedPruningMetrics, NamedRatio};
 use crate::serde::scheduler::{
     Action, BallistaFunctionRegistry, ExecutorData, ExecutorMetadata,
     ExecutorOperatingSystemSpecification, ExecutorSpecification, PartitionId,
-    PartitionLocation, PartitionStats, TaskDefinition,
+    PartitionLocation, PartitionLocationMetadata, PartitionStats, TaskDefinition,
 };
 
 use crate::RuntimeProducer;
@@ -108,14 +108,19 @@ impl TryInto<PartitionLocation> for protobuf::PartitionLocation {
                     )
                 })?
                 .into(),
-            executor_meta: self
-                .executor_meta
-                .ok_or_else(|| {
+            partition_location_metadata: {
+                let m = self.executor_meta.ok_or_else(|| {
                     BallistaError::General(
                         "executor_meta in PartitionLocation is missing".to_owned(),
                     )
-                })?
-                .into(),
+                })?;
+                Arc::new(PartitionLocationMetadata {
+                    id: m.id,
+                    host: m.host,
+                    port: m.port as u16,
+                    grpc_port: m.grpc_port as u16,
+                })
+            },
             partition_stats: self
                 .partition_stats
                 .ok_or_else(|| {
--- a/ballista/core/src/serde/scheduler/from_proto.rs
+++ b/ballista/core/src/serde/scheduler/from_proto.rs
@@ -41,8 +41,8 @@ use crate::serde::scheduler::{
     PartitionLocation, PartitionLocationMetadata, PartitionStats, TaskDefinition,
 };
 
-use crate::RuntimeProducer;
 use crate::serde::{BallistaCodec, protobuf};
+use crate::{JobId, RuntimeProducer};
 use protobuf::{NamedCount, NamedGauge, NamedTime, operator_metric};
 
 impl TryInto<Action> for protobuf::Action {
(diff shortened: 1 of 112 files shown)


coderfender:
@milenkovicm please take a look whenever you get a chance

milenkovicm:
thanks @coderfender will have a look

ballista/core/src/serde/scheduler/from_proto.rs
```diff
@@ -108,14 +108,19 @@ impl TryInto<PartitionLocation> for protobuf::PartitionLocation {
...
-                .ok_or_else(|| {
+            partition_location_metadata: {
+                let m = self.executor_meta.ok_or_else(|| {
                     BallistaError::General(
                         "executor_meta in PartitionLocation is missing".to_owned(),
                     )
-                })?
-                .into(),
+                })?;
+                Arc::new(PartitionLocationMetadata {
```
milenkovicm, line 117: my idea was to have single instance of partition location and then Arc-ed behind shared pointers.  At the moment, we have instance of executor metadata per partition location which for scenarios where we have just a few executor and many partitions may produce a lot of data (id, host ...). also when serialised we would serialize executor metadata for each of partition, which is hitting gprc issues. not sure if we can somehow reduce executor metadata instance count and its serialization

milenkovicm (commented):
thanks @coderfender, i guess we have two goals with this effort 

- reduce number of executor metadata instances keept in memory (we do not need to have exact one instance, but it would ne good not to have instance per partition location)
- reduce serialised payload as we're hitting gprc threshoulds with many partitions

ballista/core/src/serde/scheduler/from_proto.rs
coderfender: Interesting

coderfender: Would it be okay to create a followup PR to handle the second part  and port all common info in `ShuffleReaderExec` proto level and not `PartitionLocation` ?

milenkovicm:
i think #1853 does that already if im not mistaken

coderfender:
Yeah we had a chat in the original PR . I am okay to merge that PR ahead if that makes more sense this one. Although I think mine is a little bit different in terms of code paths changes

coderfender:
Closing this PR in favor of #1853

PR #1891, opened 2026-06-23
feat(scheduler): expose existing job state event stream

Jeadie:

 # Rationale for this change

## Summary 
Adds a `tokio::sync::broadcast` sender to `SchedulerServer`, subscribable via `subscribe_job_updates()`. On job state changes (queued / running / completed / failed / cancelled) the scheduler broadcasts a `JobStateEvent` carrying the job id and new state. Lets components holding a handle to the scheduler observe job lifecycle transitions without polling.

### Relationship to the existing JobStatusSubscriber
Similar to the existing `JobStatusSubscriber` (mpsc) mechanism wired through `submit_job(..., subscriber)`. This broadcast channel is a separate, fan-out style notification path; worth deciding the best DX between these.

| | `JobStatusSubscriber` (existing) | `JobStateEvent` broadcast (this PR) |
|---|---|---|
| Channel | `mpsc` - single consumer | `broadcast` - fan-out |
| Scope | 1 subscriber per submitted job, request-scoped | Global; fires for all jobs |
| Consumer | submitting gRPC client, via `ExecuteQuery` / `GetJobStatusResult`. | in-process component(s), via `subscribe_job_updates()` |
| Payload | Full protobuf `JobStatus`  | Lightweight `JobStateEvent { job_id: String, state }` |

Both fire from the same lifecycle points in `query_stage_scheduler.rs`.

Ported from spiceai/datafusion-ballista#15.


ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -85,9 +88,20 @@ pub struct SchedulerServer<T: 'static + AsLogicalPlan, U: 'static + AsExecutionP
...
+    /// `subscribe_job_updates()`.
+    job_state_sender: broadcast::Sender<job_state_event::JobStateEvent>,
 }
 
 impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T, U> {
+    /// Default capacity for the job state broadcast channel.
+    ///
+    /// This determines how many job state events can be buffered before
+    /// slow receivers start lagging behind.
+    const JOB_STATE_CHANNEL_CAPACITY: usize = 256;
```
andygrove: The 256 capacity is a sensible default. For a scheduler with high job throughput and a slow or briefly stalled subscriber, the buffer could fill and subscribers would start seeing `Lagged`. Would it be worth pulling this into `SchedulerConfig` so operators with heavier job rates can size it, keeping 256 as the default? Different deployments run very different workloads, so a fixed buffer may not fit everyone.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -211,6 +236,10 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                     .record_completed(&job_id, queued_at, completed_at);
 
                 info!("Job finished successfully: [{job_id}]");
+
+                // Broadcast job completed state
+                self.broadcast_job_state(JobStateEvent::completed(&job_id));
```
andygrove: One thing I noticed about ordering. The `Completed` event is broadcast here just before `succeed_job` runs, and the same pattern holds for `queued` (before `queue_job`), `failed`, and `cancelled`. The `Running` event, by contrast, fires after `submit_job` succeeds. So most events are emitted before the corresponding state transition is committed, which means a subscriber that reacts by querying the scheduler could briefly see stale state. There's also the case where `succeed_job` or `cancel_job` returns an error after the event already went out, so a subscriber would have been told the job completed when the commit actually failed. Would it make sense to broadcast after the state transition succeeds, so events line up with committed state and with how `Running` already behaves?

andygrove (commented):
Thanks for upstreaming this. It's cleanly structured, the doc comments are thorough, and I like that it's purely additive with no change to existing behavior. Ignoring send errors when there are no receivers is the right call for a broadcast channel.

On the design question you raised yourself about this versus the existing `JobStatusSubscriber`: agreed it's worth settling. It might help to document when a consumer should reach for each one, since they overlap in purpose but differ in scope and payload. If the broadcast path can eventually cover the in-process cases the mpsc path serves, consolidating later would be nice, but that doesn't need to block this.

Would it also be worth a test that subscribes via `subscribe_job_updates()` and drives a job through its lifecycle, asserting the expected events arrive? The unit tests cover the event types well, and a test on the actual broadcast wiring would guard the lifecycle hookups in `query_stage_scheduler.rs` against future refactors.

One minor note: `JobState` mirrors the protobuf `job_status::Status`. A `From` conversion between them, or a comment tying them together, could help keep the two from drifting over time.

phillipleblanc:
@andygrove this is ready for another review, thanks!

ballista/scheduler/src/scheduler_server/mod.rs
phillipleblanc: This is now configurable.

milenkovicm:
apologies @phillipleblanc for slow review, i'll try to catch up 
one question, current ballista implementation does have 

```rust
JobState::job_state_events(&self) -> Result<JobStateEventStream>;
```

not its a bit of ceremony to subscribe to it. 

would it make sense for `TaskManager` to expose 

```rust
    pub async fn job_state_events(&self) -> Result<JobStateEventStream> {
        self.state.job_state_events().await
    }
```

and then SchedulerServer to expose 

```rust
pub async fn job_state_events(&self) -> Result<JobStateEventStream> {
        self.state.task_manager.job_state_events().await
    }
```

`JobStateEventStream` is defined as:

```rust
pub type JobStateEventStream = Pin<Box<dyn Stream<Item = JobStateEvent> + Send>>; 
```

milenkovicm:
can we align on one, please. i believe @ akshaychitneni  is using existing one

phillipleblanc:
@milenkovicm Agreed, I've pushed a change that exposes the existing stream rather than adding a new one.

milenkovicm (approved):
thanks @phillipleblanc 

can i ask if you could expose another method, `ClusterState` has `cluster_state_events` 

```rust
    async fn cluster_state_events(&self) -> Result<ClusterStateEventStream> {
        Ok(Box::pin(self.cluster_event_sender.subscribe()))
    }
```

which emits cluster related notifications, could you please expose that one to Scheduler as well

milenkovicm:
or if you prefer we can do it in another PR, whatever suits @phillipleblanc

phillipleblanc:
I'll do it in this PR, I need to merge main anyway

milenkovicm:
thanks @phillipleblanc

PR #1892, opened 2026-06-23
feat: allow an external semaphore for executor task concurrency

Jeadie:

 # Rationale for this change

Adds an optional `available_task_slots: Option<Arc<Semaphore>>` parameter to the executor `poll_loop`. When provided, the caller supplies the semaphore that bounds concurrent task execution; when `None`, the loop creates one internally sized to the executor's task slots (unchanged behavior).

Enables:
- sharing the concurrency semaphore across poll loops connected to different schedulers,
- observing executor busy state externally via `available_permits()`.

Ported from spiceai/datafusion-ballista#14.


andygrove (commented):
Thanks for upstreaming this. The opt-in shape is clean and the default behavior is preserved exactly, which makes it easy to reason about. My only notes are a couple of doc-comment clarifications for the shared-semaphore path. Separately, would it be worth a small test that passes in a semaphore and asserts the loop uses it rather than creating its own? I realize that's awkward to set up against `poll_loop`, so feel free to wave it off if it's more trouble than it's worth.

ballista/executor/src/execution_loop.rs
```diff
@@ -75,8 +81,9 @@ where
         .unwrap()
         .clone()
         .into();
-    let available_task_slots =
-        Arc::new(Semaphore::new(executor_specification.task_slots as usize));
+    let available_task_slots = available_task_slots.unwrap_or_else(|| {
```
andygrove: Nice that this stays fully opt-in and the default path is untouched. I had a couple of questions about the shared-semaphore case from the description.

First, on reporting: when two poll loops share one semaphore but poll different schedulers, each loop reports `available_task_slots.available_permits()` as its own `num_free_slots`. So both schedulers can see the same free slots and each dispatch up to that many tasks. The semaphore still bounds actual concurrent execution through `acquire_owned`, so nothing over-runs. The second scheduler's tasks would just sit blocked waiting for a permit. Is that the intended design, with the semaphore acting purely as an execution throttle and the over-reporting being acceptable?

Second, on sizing: `DedicatedExecutor` further down is still sized to this executor's `task_slots`. If a caller passes a semaphore with more permits than `task_slots`, it would admit more concurrent tasks than the task runner has threads for. That's probably a fine tradeoff for the caller to own.

Both might be worth a line in the doc comment, noting that a shared semaphore is counted by each loop and that it controls admission only, not the task runner's thread count.

andygrove (approved):
Thanks @Jeadie

ballista/executor/src/execution_loop.rs
martin-g: Since the `Semaphore` could be provided by a third-party now we should handle `AcquireError` instead of panicking when the Semaphore is closed.
At the least map it to BallistaError.

martin-g: same here - handle AcquireError

martin-g: If the externally provided Semaphore has more permits (`N`) than the executor's task slots (`M`) then here it will request N tasks from the Scheduler and later acquire their permits (line 144) but when trying to execute them it will be able to actually run just `M` of them and the rest (`N-M`) will be queued with their permits acquired.

ballista/executor/src/execution_loop.rs
```diff
@@ -75,8 +81,9 @@ where
...
+        Arc::new(Semaphore::new(executor_specification.task_slots as usize))
+    });
```
martin-g, line 103:
```suggestion
    });
    assert!(available_task_slots.available_permits() > 0 || ..., "semaphore must have > 0 permits");
```
Passing a Semaphore with 0 permits will lead to a deadlock.

ballista/executor/src/execution_loop.rs
Jeadie: Ideally we could do something like
```rust
min(semaphore.available_permits(), dedicated_executor.available_threads())
```
but `DedicatedExecutor` doesn't expose an equivalent interface.

Currently, we mention this tension in docs.

ballista/executor/src/execution_loop.rs
Jeadie: On Reporting: Yes, intended design as described. Designed for simplicity and HA without coordination. Optimisations/improvements needed to reduce occurence of this (we expect # executors >> # schedulers, and possibly for most requests to be handled by a single scheduler). 

On Sizing: Yes, added documentation for users to avoid this footgun

milenkovicm:
thanks @Jeadie, looks like one clippy issue left

phillipleblanc:
@milenkovicm this should be good now

milenkovicm:
thanks @Jeadie @phillipleblanc & @andygrove

PR #1893, opened 2026-06-23
feat: add `poll_now_notify` to `poll_loop` and `on_work_available` callback

Jeadie:

 # Rationale for this change

Adds two related hooks for push-style executor scheduling:
- An optional `poll_now_notify: Option<Arc<Notify>>` parameter on the executor `poll_loop`. When provided, the idle wait becomes a `tokio::select!` over the poll interval and the notify, so the scheduler can wake an idle executor immediately instead of waiting for the next poll tick. `None` preserves the existing timer-only behavior.
- An `OnWorkAvailableFn` callback on `SchedulerConfig` (`on_work_available`), invoked by the query stage scheduler when new work becomes available (after `JobSubmitted` and when new stages become runnable).

Ported from spiceai/datafusion-ballista#12


ballista/scheduler/src/config.rs
```diff
@@ -34,6 +34,16 @@ use datafusion_proto::physical_plan::PhysicalExtensionCodec;
...
 use std::sync::Arc;
 
+/// Callback invoked when new work becomes available for executors.
+///
+/// This is called after:
+/// - A job is submitted and tasks are ready to be scheduled
+/// - Tasks complete and new stages become runnable
+///
+/// This allows external systems to notify executors to poll immediately
+/// rather than waiting for their next poll interval.
```
martin-g:
```suggestion
/// rather than waiting for their next poll interval.
///
/// # Warning
///
/// This callback is executed synchronously within the scheduler's main event loop.
/// Implementations **must be non-blocking** and should offload any blocking or
/// long-running operations (such as network I/O) to a separate task or thread.
```

The callback is synchronous and a blocking implementation may affect badly the Scheduler;w work.

ballista/executor/src/execution_loop.rs
```diff
@@ -204,7 +208,20 @@ where
         }
 
         if !active_job {
-            tokio::time::sleep(Duration::from_millis(50)).await;
+            // Wait for either the poll interval or a poll_now notification
+            match &poll_now_notify {
+                Some(notify) => {
+                    tokio::select! {
+                        () = tokio::time::sleep(Duration::from_millis(50)) => {}
```
martin-g: The duration here could be longer.
The idea is that `notify` will notify us as soon as there is some work, right ?
So, there is no need to poll so frequently if an external mechanism will notify us when to do it.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -295,6 +300,13 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                                 .await?;
                         }
 
+                        // Notify external systems when new stages become runnable
+                        if !stage_events.is_empty()
+                            && let Some(ref callback) = self.config.on_work_available
+                        {
+                            callback("tasks_completed:new_stages_runnable");
+                        }
+
```
martin-g: Shouldn't this be done after line 312 ? I.e. after posting the events.
Currently this will tell executors to poll but it seems to early.

ballista/scheduler/src/config.rs
```diff
@@ -546,6 +560,7 @@ impl TryFrom<Config> for SchedulerConfig {
             cors_allowed_origins: opt.cors_allowed_origins,
             #[cfg(feature = "rest-api")]
             cors_allowed_methods: opt.cors_allowed_methods,
+            on_work_available: None,
```
martin-g, line 648: How is this new setting supposed to be set ?
I'd expect at least a new `with_on_work_available` setter to be added if there is no new command line option.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -180,6 +180,11 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
                         .post_event(QueryStageSchedulerEvent::ReviveOffers)
                         .await?;
                 }
+
+                // Notify external systems that new work is available
+                if let Some(ref callback) = self.config.on_work_available {
+                    callback(&format!("job_submitted:{job_id}"));
```
martin-g: The `event_sender.post_event(QueryStageSchedulerEvent::ReviveOffers)` is non-blocking and the events might be still in the channel when this callback is executed. I.e. the first poll may see no events.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
martin-g: The `stage_events` are not necessarily "new stages are runnable". They might be JobRunningFailed or JobFinished, neither of which represents new runnable work.

ballista/scheduler/src/config.rs
```diff
@@ -34,6 +34,16 @@ use datafusion_proto::physical_plan::PhysicalExtensionCodec;
...
 
+/// Callback invoked when new work becomes available for executors.
+///
+/// This is called after:
+/// - A job is submitted and tasks are ready to be scheduled
+/// - Tasks complete and new stages become runnable
+///
+/// This allows external systems to notify executors to poll immediately
+/// rather than waiting for their next poll interval.
+pub type OnWorkAvailableFn = Arc<dyn Fn(&str) + Send + Sync>;
```
martin-g: The function receives a plain `&str` as a command/message. IMO it would be better to use an Enum is all possible variants of commands.

martin-g: Is there a need to use an Arc ?
It could be a Box.

ballista/scheduler/src/config.rs
phillipleblanc: Comment added.

ballista/executor/src/execution_loop.rs
phillipleblanc: Agreed, implemented to use a 1s fallback if a `poll_now_notify` is provided, otherwise unchanged at 50ms to preserve existing behavior.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
phillipleblanc: This is correct, changed to remove the callback from `TaskUpdating` and moved into the `JobUpdated` arm, assuming `update_job` returns `new_tasks > 0`. So this only notifies when stages genuinely had new scheduleable tasks, and only after they are updated.

ballista/scheduler/src/config.rs
phillipleblanc: Added `SchedulerConfig::with_on_work_available`

ballista/scheduler/src/config.rs
phillipleblanc: Replaced the `&str` with a `WorkAvailableReason` enum (`JobSubmitted { job_id }` / `NewStagesRunnable { job_id }`), so consumers get structured data instead of parsing strings.

It also needs to stay `Arc` because `SchedulerConfig` derives `Clone` and `Box<dyn Fn>` isn't cloneable. We could drop the Clone requirement if we want to keep it as Box.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
phillipleblanc: Fixed. The callback now fires only after `graph.revive()` has run (in `update_job`, and `submit_plan` revives before caching), which is the exact point tasks become visible to poll_work. Now a woken executor's first poll always finds the work.

phillipleblanc:
@martin-g Thanks for the thorough review, this should be ready now for another look.

milenkovicm (approved):
thanks @phillipleblanc, @Jeadie and @martin-g looks like all comments has been resolved, and i have nothing more to add.  Please rebase so we can merge it

phillipleblanc:
> thanks @phillipleblanc, @Jeadie and @martin-g looks like all comments has been resolved, and i have nothing more to add. Please rebase so we can merge it

I've merged the latest changes and pushed. Thanks!

PR #1899, opened 2026-06-26
Fix/python cargo lock drift

### LINKED ISSUE #1887: Improve python/Cargo.lock drifts - CI deps
**python/** is a separate Cargo workspace (excluded from the main one) with its own Cargo.lock, and it path-depends on the main ballista crates. But:

- **Dependabot.yml** only manages directory: "/" (the main workspace) there's no /python entry. So every dep bump updates /Cargo.lock and never python/Cargo.lock. Acts like a safety net. It catches any drift regardless of source and stops it landing on main.
- CI does not catches the drift early, it only surfaces later as the cryptic --locked maturin error.
So the lock goes stale every time deps move, and the next contributor inherits a red build. 

1)  Add a fast CI guard in dependencies.yml - when python Cargo.toml/Cargo.lock change (push + PR), catch a stale python lock on the PR that introduces it, with actionable message. 
2) Add /python to `dependabot.yml` so its direct deps stay fresh and dep-bump PRs touch both locks:

* Extend the existing `dev/update_ballista_versions.py.` The repo already has a script that bumps python's version in lockstep with the workspace. Add a cd python && cargo generate-lockfile step to it (or a sibling dev/ script), so regenerating both locks is one documented command. Pair with _fast CI guard_  to enforce it. This fits the repo's existing conventions perhaps.

**Additional context**
Build Failure ex: [link](https://github.com/apache/datafusion-ballista/actions/runs/27927974489/job/82634133304?pr=1883)

djanand:

Closes #1887.

 # Rationale for this change

`python/` is a **separate Cargo workspace** (excluded from the root workspace) with its own `python/Cargo.lock`, and it path-depends on the main ballista crates. Two gaps make that lock drift out of sync:

1. **Dependabot only covers the root workspace** (`directory: "/"`)- there is no `/python` entry, so dep bumps update `/Cargo.lock` but never `python/Cargo.lock`.
2. **No CI catches the drift early.** `rust.yml` runs `--locked` only against the main workspace (python is excluded). A stale python lock only surfaces *later* and cryptically, in `build.yml`'s maturin steps (which run with `locked = true` from `python/pyproject.toml`): error: cannot update the lock file python/Cargo.lock because --locked was passed
3. Added an explicit least-privilege permissions: contents: read block to dependencies.yml (resolves a CodeQL finding; matches the other workflows).

Net effect: when a dep moves in a ballista crate manifest (e.g. `itertools`, `tower-http` were bumped to `0.15`/`0.7`), `python/Cargo.lock` is left behind on `main`, and the next contributor inherits a red maturin build with an unhelpful message. This PR fixes the current drift and adds a guard so future drift is caught on the PR that introduces it.

- **`python/Cargo.lock`** - minimally reconciled with the manifests (picks up the already-published `itertools 0.15.0` / `tower-http 0.7.0` that flow in via the ballista path-deps; lockfile format v4 preserved). This fixes the current drift so the new guard passes from day one.
- **`.github/workflows/dependencies.yml`** - new `python-lock` job that runs `cargo metadata --locked` in `python/` and fails with an actionable message (`run 'cd python && cargo update' and commit`) if the lock is stale. The workflow already triggers on `**/Cargo.toml` / `**/Cargo.lock`, so both main-workspace and python changes exercise it.
- **`.github/dependabot.yml`** - new `cargo` entry for `directory: "/python"` (mirroring the root block's manual-bump ignore list for arrow/datafusion/sqlparser), so python's **direct** deps stay fresh.

No runtime/API changes - this is CI/tooling only.

One contributor-facing note: dependabot bumps a **transitive** dep into python via the ballista path-deps (which python doesn't declare directly, so dependabot can't sync python's lock for it). Such a PR will now trip the `python-lock` guard and needs a one-line `cd python && cargo update` + commit on that branch. This isn't new breakage - that drift already broke the next `python/**` PR cryptically; the guard just relocates it to the PR that caused it, with a clear fix message.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,6 +32,21 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: "sqlparser"
         update-types: ["version-update:semver-major"]
+  - package-ecosystem: cargo
+    directory: "/python"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+    target-branch: main
+    labels: [auto-dependencies]
+    ignore:
+      # arrow and datafusion are bumped manually
+      - dependency-name: "arrow*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "datafusion*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "sqlparser"
+        update-types: ["version-update:semver-major"]
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -53,3 +53,23 @@ jobs:
         run: |
           cd dev/msrvcheck
           cargo run
+
+  python-lock:
+    name: python Cargo.lock up to date
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v7.0.0
+        with:
+          fetch-depth: 1
+      - name: Setup Rust toolchain
+        uses: ./.github/actions/setup-macos-builder
+        with:
+          rust-version: stable
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
+      - name: Verify python/Cargo.lock is in sync with manifests
+        run: |
+          cd python
+          if ! cargo metadata --locked --format-version 1 >/dev/null; then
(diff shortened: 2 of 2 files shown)


martin-g (approved):

djanand:
This requires https://github.com/apache/datafusion-ballista/pull/1919 to be reviewed & merged since`main` has diverged

coderfender:
@djanand  seems like #1919  is merged now  ?

djanand:
> @djanand seems like #1919 is merged now ?

thanks @coderfender. Can you please review/approve/merge this PR ?

Issue #1770, opened 2026-05-25
Static job planner to respect `datafusion....hash_join_single_partition_threshold` for broadcast decision


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

static job planner uses `ballista.optimizer.broadcast_join_threshold_bytes` configuration value to decide if shuffle should be broadcast (`CollectLeft`), this configuration value diverges from datafusion configurations:

- `datafusion.optimizer.hash_join_single_partition_threshold_rows`
- `datafusion.optimizer.hash_join_single_partition_threshold`

which serve same purpose.

**Describe the solution you'd like**

I would propose to change static planner to to respect same configuration values like datafusion `datafusion.optimizer.hash_join_single_partition_threshold_rows` & `datafusion.optimizer.hash_join_single_partition_threshold`

AQE implementation (#1752) uses the same configuration values as datafusion hence it would be good if we can get both planners (and datafusion) aligned 

suggestion is to change 

`datafusion.optimizer.hash_join_single_partition_threshold_rows` https://github.com/apache/datafusion-ballista/blob/65ec1c479acf9fe7acb19a0780bc0a6100739b9c/ballista/core/src/extension.rs#L773 from 0 to 1M (datafusion `131072`)

and `datafusion.optimizer.hash_join_single_partition_threshold` https://github.com/apache/datafusion-ballista/blob/65ec1c479acf9fe7acb19a0780bc0a6100739b9c/ballista/core/src/extension.rs#L769 from 0 to 10MB (datafusion 1048576)

**Describe alternatives you've considered**

align both planners to  `ballista.optimizer.broadcast_join_threshold_bytes` but that would diverge from datafusion configuration, which is not ideal as ballista tends to align with datafusion as much as possible 

**Additional context**

milenkovicm:
cc @andygrove i hope you approve this configuration change for broadcast planning

ramieali:
Hi @milenkovicm , i would like to work on this issue if it's still up for grabs

milenkovicm:
Hi @ramieali after a bit of consideration I'm not sure if this would be good first time issue

ramieali:
thanks for the heads up! if i am not underscoping it then i think a simple refactor would be changing planner config
https://github.com/apache/datafusion-ballista/blob/65ec1c479acf9fe7acb19a0780bc0a6100739b9c/ballista/scheduler/src/planner.rs#L280-L286
 

[...]

Issue #1375, opened 2026-01-09
Investigate use of `DynamicFilters` in ballista


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

There is great push in DataFusion community to support `DynamicFilters`, looks like there are [some good query improvements](https://datafusion.apache.org/blog/2025/09/10/dynamic-filters/) when dynamic filters are used. 

I'm not sure if they would improve ballista execution due to stage execution but at least we could investigate if they will yield any benefits for ballista

**Describe the solution you'd like**

It would be great if we could some effort to investigate if shuffle read would have any benefits of it.

- try to integrate dynamic filters with shuffle read to start with 
- benchmark benefits (if any) 
- provide road forward 

**Describe alternatives you've considered**

**Additional context**

Huy1Ng:
might need this feature: https://github.com/apache/datafusion/issues/18296

Huy1Ng:
At the current stage, dynamic filter only work in **a single task** . Let' say you have cluster with 1 scheduler and 1 executor. 
You make this query "select * from line_items sort by line_id asc". The schedule plans 2 task, and send to the sole executor. Each task will be able to leverage the new threshold it found in TopK to reduce number of rows scan upstream, *but these tasks cannot share this information to each other*. 

My approach is to solve this local problem with by having tasks on a single executor share this dynamic filter with each others. Then we can proceed to look for ways to sync this information across the network boundary.

Dandandan:
One case where it works well is in broadcasted hash join (collect left mode), as it collects the build side to each executor, it can also push down the filter / hash table to the scan.

andygrove:
A couple of concrete data points from SF100 TPC-H benchmarking that may help scope this:

**Dynamic filter pushdown is currently explicitly disabled in Ballista.** `ballista/core/src/extension.rs:817` sets:

```rust
.set_bool("datafusion.optimizer.enable_dynamic_filter_pushdown", false)
```

So the first step here is likely understanding *why* it's off (presumably because a dynamic filter is computed at runtime on one node and can't cross a serialized stage boundary to reach the probe-side scan on another node) and what it would take to make it survive distributed planning/serialization — which lines up with the "integrate with shuffle read" goal above.

[...]

PR #1900, opened 2026-06-26
feat: broadcast small build sides via non-zero CollectLeft thresholds

andygrove:

Part of #1770.

This PR makes the default-value change suggested in #1770 (restore non-zero
thresholds), and adds the join-type safety guard that broadcasting those build
sides turns out to require. The other part of #1770 — having the *static* job
planner respect `datafusion.optimizer.hash_join_single_partition_threshold(_rows)`
instead of `ballista.optimizer.broadcast_join_threshold_bytes` — is not included
here, so #1770 is left open.

# Rationale for this change

Ballista pins `datafusion.optimizer.hash_join_single_partition_threshold` and
`..._threshold_rows` to `0`, which disables `CollectLeft` (broadcast) hash joins.
While the threshold is `0`, the AQE join resolver (`DynamicJoinSelectionExec` /
`to_actual_join`) never promotes a small build side to a broadcast `CollectLeft`
join — every join is repartitioned through a shuffle, even single-row dimension
tables. Raising the thresholds lets the resolver broadcast small build sides,
which is a large TPC-H win (see numbers below).

The `0` pin was originally a workaround for #1055 ("Left/full outer join
incorrect for CollectLeft / broadcast"). Review on this PR established that #1647
never actually fixed the underlying problem for distributed execution: a
`CollectLeft` join replicates the build (left) side to *every* probe task, and
each task processes only its slice of the probe (right) side. That is correct
only for join types whose output is driven entirely by the probe side. Any join
type that emits rows on behalf of the build side — unmatched outer rows, or
semi/anti/mark rows that depend on a build row's global match status — would have
those rows emitted independently by every probe task, producing duplicate or
spurious results.

So we cannot simply un-zero the thresholds. We un-zero them *and* restrict
`CollectLeft` promotion to the join types that are safe to broadcast. This guard
is needed on both planning paths, including the static planner, where the same
unsafe promotion was a latent bug (the #1647 path) that could fire whenever
`prefer_hash_join=true`.

Measured on TPC-H SF10 (1 scheduler + 1 executor, 8 task slots, 8 partitions,
3 iterations, AQE enabled), comparing the resolver with the threshold at `0` vs.
the values in this PR:

| | threshold = 0 | threshold = 10 MB / 1M rows |
|---|---:|---:|
| full 22-query total | 32.2 s | 16.2 s |

Per-query, broadcast fires for small dimension joins and is left off for
big-vs-big joins. Examples: Q2 0.82 s → 0.12 s, Q8 3.13 s → 0.46 s,
Q9 4.61 s → 1.09 s, Q11 0.59 s → 0.09 s, Q21 5.17 s → 1.49 s. Row counts are
unchanged for all 22 queries.

- Set `datafusion.optimizer.hash_join_single_partition_threshold` to 10 MB and
  `datafusion.optimizer.hash_join_single_partition_threshold_rows` to 1,000,000
  in the Ballista session defaults (previously both `0`).
- Add a shared `collect_left_broadcast_safe(JoinType)` predicate that allows
  `CollectLeft` broadcast only for probe-driven join types: `Inner`, `Right`,
  `RightSemi`, `RightAnti`, `RightMark`. Safety is evaluated on the join type
  *after* any build-side swap, since the build side is always the left input of
  the resulting `HashJoinExec`.
- Apply the guard in the **AQE resolver** (`to_actual_join` in `dynamic_join.rs`):
  a small build side is collected into a `CollectLeft` join only when the
  post-swap join type is broadcast-safe, otherwise it is repartitioned. The swap
  decision reuses `SelectJoinRule::supports_swap_join_order`, the same call the
  resolver later uses to perform the swap, so the predicted and actual join types
  stay in sync.
- Apply the guard in the **static planner** (`maybe_promote_to_broadcast` in
  `planner.rs`), closing the latent #1647 path so it can no longer promote an
  unsafe join type to `CollectLeft`.
- Add tests: an exhaustive safe/unsafe `JoinType` matrix for the predicate,
  resolver tests covering both swap orientations and `prefer_hash_join` on/off,
  static-planner tests across all join types and build-side orientations, a
  `LEFT join` small-build "must not broadcast" case, and a sort-merge-join
  "no broadcast" case.
- Update the client settings test to assert the new defaults (and rename it to
  reflect that it now checks the thresholds are set).

Yes. With AQE enabled, a hash join whose build side fits under these thresholds
and whose join type is broadcast-safe is executed as a broadcast (`CollectLeft`)
join instead of being repartitioned. No SQL semantics change, and results
(including row counts) are unchanged. Join types that emit rows on behalf of the
build side continue to be repartitioned.

The static (non-AQE) planner still uses a separate config
(`ballista.optimizer.broadcast_join_threshold_bytes`) to decide *whether* to
broadcast, so its broadcast frequency is unchanged. Its promotion path now also
applies the broadcast-safety guard, which only removes previously-unsafe
promotions (reachable with `prefer_hash_join=true`).


ballista/core/src/extension.rs
```diff
@@ -760,18 +760,15 @@ impl SessionConfigHelperExt for SessionConfig {
...
-            // "Left/full outer join incorrect for CollectLeft / broadcast"
-            //
-            // In order to make correct results (decreasing performance) CollectLeft
-            // has been disabled until fixed
+            // A build side smaller than these thresholds is collected into a
+            // CollectLeft (broadcast) hash join rather than being repartitioned.
             .set_u64(
                 "datafusion.optimizer.hash_join_single_partition_threshold",
-                0,
+                10 * 1024 * 1024,
```
andygrove, line 774: https://github.com/apache/datafusion-ballista/issues/1055 is fixed now, so we can revert the workaround

Dandandan: I don't really understand yet why https://github.com/apache/datafusion-ballista/pull/1647 fixed this for left/outer joins?

E.g. see 
https://github.com/apache/datafusion/issues/12454 for discussion ot it.

It doesn't really change reporting on unmatched rows (for which it needs to produce null rows)

Dandandan: It probably needs some test cases for those ("duplicated unmatched keys across executors") to make sure they don't regress and then see if we finally solve it / bring it back? :)

andygrove: I hit a regression when running benchmarks locally with AQE enabled, so moving this to draft. I also created https://github.com/apache/datafusion-ballista/pull/1913 so that we have CI tests for AQE to avoid regressions. Looking into feedback now.

andygrove:
@Dandandan This PR has expanded a lot since the last review. Could you take another look? Also, fyi @avantgardnerio

Dandandan (approved):

avantgardnerio (changes requested):
From claude:

```
The static path's safety guard is bypassed when DataFusion's own JoinSelection promotes the join before Ballista sees it. With the PR's session default of
  hash_join_single_partition_threshold = 10 MB, DF's try_collect_left (datafusion/physical-optimizer/src/join_selection.rs:200) will set PartitionMode::CollectLeft on any
  small-build-side join — it only swaps when both sides are collectable; for small LEFT JOIN big with the small side already on the left, it falls into the (true, false) arm and just
  stamps CollectLeft on the LEFT join. No swap, no join-type check.
```

Test in a stacked PR incoming

avantgardnerio:
I'm not super familiar with joins yet, so this could be complete AI slop, but here's a failing test: https://github.com/andygrove/datafusion-ballista/pull/89

Sorry if it's a false alarm, I figured better to share than let a bug hit main. I'll flip my review to approve now, merge if I'm wrong!

avantgardnerio (approved):
Approved pending review of stacked PR

andygrove:
@avantgardnerio found a real gap in the broadcast-safety guard via a stacked test PR (andygrove/datafusion-ballista#89), now fixed here in b40a6701.

DataFusion's own `JoinSelection` runs during `create_physical_plan` and, with the non-zero `hash_join_single_partition_threshold` this PR sets, stamps `CollectLeft` without restricting by join type. The guard only covered the `Partitioned -> CollectLeft` promotion path, so a pre-promoted unsafe join (e.g. `small LEFT JOIN big` with the small side already on the left) bypassed it and reached the broadcast-lowering branch, replicating the outer side across every probe task.

The fix moves the broadcast-safety check to the front of `maybe_promote_to_broadcast`: any `CollectLeft` hash join whose join type isn't `collect_left_broadcast_safe` is demoted back to a `Partitioned` (shuffle) join, re-hash-partitioning both inputs on the join keys. It runs ahead of the threshold check so the correctness guard still fires even if Ballista's own broadcast promotion is disabled while DF's threshold is non-zero. avantgardnerio's reproducer test is pulled in verbatim (with the explanatory comment swapped for a current-state one to match the repo convention) and now passes.

andygrove:
The TPC-H CI job appears to be hanging. This is likely catching a real issue. I will investigate.

andygrove:
> The TPC-H CI job appears to be hanging. This is likely catching a real issue. I will investigate.

Benchmark hangs on some queries when AQE is enabled due to changes in this PR

andygrove:
The new AQE-on TPC-H CI job was hanging (not the AQE-off job). Root-caused and fixed in 382b8e1e.

## Symptom

Under the adaptive (AQE) planner, the job completed the AQE-off suite, then hung on the first multi-join query (Q2) and never recovered. It reproduces locally on a 1-executor / 4-slot / 16-partition cluster and is independent of slot/partition counts (hangs at `-c 4`/4, `-c 8`/8, `-c 16`/16).

## Root cause: DataFusion 54 dynamic filter pushdown

DataFusion 54 populates dynamic filters at runtime from an upstream operator (a hash-join build side, a TopK heap, a partial aggregate) and reads them in a downstream scan **within the same plan**. Ballista splits a plan into stages at shuffle/broadcast boundaries that run as **independent tasks**, so when the producing operator and the consuming scan land in different stages, the filter is never populated across the boundary and the consuming scan blocks forever.

This deadlocks every multi-join query under AQE, because broadcast (`CollectLeft`) hash joins create the cross-stage producer/consumer split (the probe-side scan carries `predicate=DynamicFilter [ empty ]`). The static planner escapes it only because it plans sort-merge joins, which carry no join dynamic filter — which is why the AQE-off suite passed and AQE-on hung.

The regression entered when this branch merged `main`: the DataFusion 54 upgrade (#1906) and the AQE-on CI (#1913) landed in the same merge, and AQE-on had never been exercised in CI before.

## How it was confirmed

- `main` (same DF54 + same AQE-on CI) passes the TPC-H job in ~24 min; only this branch hung → the un-zeroed thresholds (broadcast under AQE) are what trigger it.
- DF53-vs-DF54 bisect: on the pre-merge DF53 commit, Q2/Q5/Q8/Q9 all run fine (Q2 0.16 s, 100 rows). On DF54 they all hang.
- `enable_dynamic_filter_pushdown=false` makes Q2 complete and removes the `DynamicFilter` from the plan; the per-type sub-flags (join/topk/aggregate) each individually leave the hang.

## Fix

Pin `datafusion.optimizer.enable_dynamic_filter_pushdown=false` in the Ballista session defaults, alongside the other DataFusion options Ballista already disables because they rely on in-process state that cannot cross stage boundaries (e.g. `enable_physical_uncorrelated_scalar_subquery`). Long-term, carrying dynamic filters across stage boundaries is tracked by #1375.

Verified locally: all 21 AQE-on queries (Q16 omitted, as in CI) complete with the default config at the CI topology.

Orthogonal note: Q11 and Q15 return 0 rows under **both** AQE on and off at SF10 — a pre-existing correctness issue, not introduced here and not caught by CI (which doesn't validate row counts). Worth a separate issue.

andygrove:
Moving to draft until https://github.com/apache/datafusion-ballista/pull/1917 is merged. I want to make sure we are not introducing any regressions

PR #1902, opened 2026-06-26
fix: honor client datafusion.* session config overrides in scheduler planning

### LINKED ISSUE #1901: datafusion.* session config overrides from the client are ignored by scheduler-side (AQE) planning
## Describe the bug

Client-set DataFusion session config (the `datafusion.*` namespace) is ignored by scheduler-side query planning, while `ballista.*` settings in the same session do take effect.

## To Reproduce

Run any TPC-H query with AQE enabled and a `datafusion.optimizer` override, e.g. via the tpch benchmark:

```
cargo run --release --bin tpch -- benchmark ballista \
  --host localhost --port 50050 \
  --query 5 --path <data> --format parquet --partitions 8 \
  -c datafusion.optimizer.hash_join_single_partition_threshold=10485760 \
  -c datafusion.optimizer.hash_join_single_partition_threshold_rows=1000000
```

## Expected behavior

The AQE join resolver (`DynamicJoinSelectionExec::to_actual_join`) should see the overridden threshold and promote small build sides to `CollectLeft`.

## Actual behavior

The resolver still sees the threshold as `0` — every join resolves to `Partitioned`, no `CollectLeft`, even for single-row build sides. The `datafusion.*` override has no effect. In the **same command**, the `ballista.planner.adaptive.enabled=true` override **does** take effect (AQE engages), so the session settings are reaching the scheduler — only the `datafusion.*` ones are being dropped.

## Additional context

The obvious path looks correct end-to-end:

- `SessionConfigHelperExt::to_key_value_pairs` (ballista/core/src/extension.rs) serializes **all** config entries, including `datafusion.*`.
- `create_update_session` (ballista/scheduler/src/scheduler_server/grpc.rs) calls `produce_config()` then `update_from_key_value_pair(&settings)`, applying the client overrides on top of the scheduler base config.

So the override reaches the stored session config, but the value appears to be lost when the per-job planning `SessionState` is built — the suspicion is that Ballista's session defaults (`upgrade_for_ballista`, which sets `hash_join_single_partition_threshold = 0`) are re-applied over the client config. `ballista.*` extension config survives because it lives in the `BallistaConfig` extension rather than in `ConfigOptions.optimizer`. (Suspected cause — not yet confirmed to a specific line.)

**Impact:** users cannot tune DataFusion optimizer behavior per session; changing these thresholds currently requires recompiling the defaults (see #1770 / PR #1900).

andygrove:

Closes #1901.

# Rationale for this change

Client-set `datafusion.*` session config (e.g. `SET
datafusion.optimizer.prefer_hash_join = true`, or `-c` overrides in the tpch
benchmark) had no effect on scheduler-side planning, while `ballista.*` settings
in the same session worked.

The cause is `SessionConfigExt::upgrade_for_ballista`, which calls
`ballista_restricted_configuration()` to apply Ballista's opinionated DataFusion
defaults (`prefer_hash_join = false`, `hash_join_single_partition_threshold = 0`,
the Utf8View flags). `SessionConfig::new_with_ballista()` already applies these
once at construction. When `remote_with_state` later calls
`upgrade_for_ballista` again, the defaults are re-applied *after* the user has
set their own values, silently reverting them. `ballista.*` settings survive only
because the restricted config does not touch them.

Notably, the restricted-config comments state that users can opt back into
`prefer_hash_join` and override the view-type flags via `SET` — but the
re-application defeated exactly that.

- `upgrade_for_ballista` now applies `ballista_restricted_configuration()` only
  when the config has not already been through Ballista setup (detected by the
  absence of the `BallistaConfig` extension). A config that already carries the
  extension keeps the user's values.
- Tests: a user override of `prefer_hash_join` / `hash_join_single_partition_threshold`
  survives `upgrade_for_ballista`; a plain config still receives Ballista's
  defaults on upgrade.

Yes. `datafusion.*` session settings set on the client (via `SET` or config
overrides) are now honored by the scheduler. `round_robin_repartition` remains
effectively enforced because the scheduler forces it off when building the
execution context. No SQL semantics change.

### DIFF
--- a/ballista/core/src/extension.rs
+++ b/ballista/core/src/extension.rs
@@ -355,15 +355,22 @@ impl SessionConfigExt for SessionConfig {
     }
 
     fn upgrade_for_ballista(self) -> SessionConfig {
-        // if ballista config is not provided
-        // one is created and session state is updated
-        let ballista_config = self.ballista_config();
+        // Ballista's opinionated DataFusion defaults are applied once, when a
+        // plain config is first upgraded. The `BallistaConfig` extension marks a
+        // config that has already been through this path; re-applying the
+        // defaults on such a config would overwrite any values the user set
+        // afterwards (e.g. via `-c`/`SET`), so they are left untouched.
+        let already_upgraded =
+            self.options().extensions.get::<BallistaConfig>().is_some();
 
-        // session config has ballista config extension and
-        // default datafusion configuration is altered
-        // to fit ballista execution
-        self.with_option_extension(ballista_config)
-            .ballista_restricted_configuration()
+        let ballista_config = self.ballista_config();
+        let config = self.with_option_extension(ballista_config);
+
+        if already_upgraded {
+            config
+        } else {
+            config.ballista_restricted_configuration()
+        }
     }
 
     fn ballista_config(&self) -> BallistaConfig {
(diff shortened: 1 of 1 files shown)


andygrove:
I went ahead and merged this to try and help diagnose the issues in the DF54 upgrade PR

Issue #1795, opened 2026-05-30
Improve job failure handling


milenkovicm:
**Describe the bug**
In case of job failure job gets proper status "Failed"

<img width="1146" height="105" alt="Image" src="https://github.com/user-attachments/assets/7ab6db4e-9211-481b-8dbf-a97b7077e12d" />

but underlying stages and task get stuck with status running:

<img width="1042" height="101" alt="Image" src="https://github.com/user-attachments/assets/c85b70c8-2235-4127-81cc-0cebdb84b296" />

<img width="1167" height="134" alt="Image" src="https://github.com/user-attachments/assets/6c0b4312-d2ed-4d44-a8e3-ba1f8ea97eb4" />

**To Reproduce**

running tpch benchmark with failure generator enabled 

```
cargo run --bin tpch -- benchmark ballista -p /Users/marko/TMP/tpch_data/tpch-data-sf10/ -f parquet -i 1 --port 50050 --host 127.0.0.1 -c datafusion.execution.target_partitions=24 -c ballista.planner.adaptive.enabled=true -c ballista.testing.chaos_execution.enabled=true  -c ballista.testing.chaos_execution.fault_type=fatal -c ballista.testing.chaos_execution.probability=0.35  -q 1
```

should trigger job failure 

**Expected behavior**

When job is cancelled or failed, running stages and tasks should be canceled and marked as cancelled. 

It has to be confirmed if job failure actually issues "cancel" stages 

**Additional context**

#1793 & #1789  to be merged beforehand as they would simplify testing

killzoner:
Hello @milenkovicm I would be interested by working on this one, possibly after the 2 linked PRs are merged (i see one of them is closed so maybe there is still pending work?)

milenkovicm:
this has been addressed this morning with #1871

milenkovicm:
ah sorry @killzoner ... this is different, i think there is no other pr pending, you could take it, all required pr has been merged

killzoner:
Thank you for the confirmation, i will take this one then!

milenkovicm:
when you're into this area @killzoner please have a look how can we track number of failed tasks, as at the moment we do not see that information on UI (this could be a separate pr)

killzoner:
Thanks for the additionnal hints, will do

killzoner:
~~One question I have on the design: currently a task has `Running`, `Successful`, `Failed`, no `Cancelled` state.~~

[...]

PR #1903, opened 2026-06-26
feat: cancel running stages and tasks on job failure or cancellation

killzoner:

Closes partially https://github.com/apache/datafusion-ballista/issues/1795 (needs a follow up REST/TUI fix)

# Rationale for this change

When a job fails or is cancelled, only the job status was set to `Failed`. Its running stages and tasks stayed stuck in `Running` (`fail_stage` was never called).

- Add `ExecutionGraph::abort_running`, called from `task_manager::abort_job` (used by both the failure and cancel paths): it fails the job and runs `fail_stage` over the running stages.
- `RunningStage::to_failed` marks still-running tasks as `Failed(TaskKilled)`.
- Tests for both static and adaptive graphs.

1626af4872587b400d1dc6e588f383160f3c4918 can be dropped if you prefer a default trait method over two per-impl copies (i figured there was no default trait method so far in existing code)

Tested locally with TPC-H under the chaos fault generator:

```bash
cargo run --bin ballista-scheduler --features prometheus-metrics
cargo run --bin ballista-executor
cargo run --bin tpch -- benchmark ballista -p /path/to/tpch/sf10 -f parquet -i 1 \
  --port 50050 --host 127.0.0.1 \
  -c datafusion.execution.target_partitions=24 \
  -c ballista.testing.chaos_execution.enabled=true \
  -c ballista.testing.chaos_execution.fault_type=fatal \
  -c ballista.testing.chaos_execution.probability=0.6 \
  -q 1
```

Failed job in the TUI:

**Global view** (job shows `Failed`):
Screenshot from 2026-06-26 17-03-39

**Stage detail** (stage now shows `Failed`, was stuck `Running`):
Screenshot from 2026-06-26 17-04-07

**Stage tasks** (empty for a failed stage, surfaced in the follow-up):
Screenshot from 2026-06-26 17-04-19

I will do a follow-up PR to update the REST API and expose the failed /cancelled task over a failed stage, with TUI handling (I have a draft already)

Note: this PR made use of LLM tooling

Failed/cancelled jobs now show their stages as `Failed` instead of stuck `Running`.

### DIFF
--- a/ballista/scheduler/src/state/execution_graph.rs
+++ b/ballista/scheduler/src/state/execution_graph.rs
@@ -2161,6 +2161,52 @@ mod test {
         Ok(())
     }
 
+    // Aborting a running job (failure or cancellation) must transition every
+    // running stage to Failed and return its in-flight tasks for cancellation.
+    // `abort_running` is the shared teardown invoked by `abort_job`.
+    #[tokio::test]
+    async fn test_abort_running_cancels_stages_and_returns_inflight_tasks() -> Result<()>
+    {
+        let executor = mock_executor("executor-id1".to_string());
+        let mut graph = test_join_plan(2).await;
+
+        // Call revive to move the two leaf Resolved stages to Running
+        graph.revive();
+        assert!(
+            graph.running_stages().len() >= 2,
+            "expected two concurrently running leaf stages, found {:?}",
+            graph.running_stages()
+        );
+
+        // Dispatch a task so there is an in-flight task to cancel
+        let _task = graph.pop_next_task(&executor.id)?.unwrap();
+
+        // Aborting cancels every running stage and returns its in-flight tasks
+        let cancelled = graph.abort_running("job aborted".to_string());
+
+        assert!(
+            !cancelled.is_empty(),
+            "abort_running must return the in-flight tasks to cancel"
+        );
+        assert!(
+            graph.running_stages().is_empty(),
+            "every running stage must be cancelled, found {:?}",
+            graph.running_stages()
+        );
+        assert!(
+            matches!(
+                graph.status(),
+                JobStatus {
+                    status: Some(job_status::Status::Failed(_)),
+                    ..
+                }
+            ),
+            "the job must be Failed after abort"
+        );
+
+        Ok(())
+    }
+
     #[tokio::test]
     async fn test_long_delayed_failed_task_after_executor_lost() -> Result<()> {
         let executor1 = mock_executor("executor-id1".to_string());
--- a/ballista/scheduler/src/state/execution_graph.rs
+++ b/ballista/scheduler/src/state/execution_graph.rs
@@ -222,7 +222,14 @@ pub trait ExecutionGraph: Debug {
     /// Abort a running job: fail it, transition every running stage to Failed,
     /// and return the in-flight tasks that should be cancelled. Used for both the
     /// failure and cancellation teardown paths.
-    fn abort_running(&mut self, error: String) -> Vec<RunningTaskInfo>;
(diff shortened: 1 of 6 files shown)


ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -219,6 +219,11 @@ pub trait ExecutionGraph: Debug {
     /// fail job with error message
     fn fail_job(&mut self, error: String);
 
+    /// Abort a running job: fail it, transition every running stage to Failed,
+    /// and return the in-flight tasks that should be cancelled. Used for both the
+    /// failure and cancellation teardown paths.
+    fn abort_running(&mut self, error: String) -> Vec<RunningTaskInfo>;
```
andygrove: On your open question about the default trait method versus two per-impl copies: I'd lean toward the default method. The body only calls `running_tasks`, `fail_job`, `running_stages`, and `fail_stage`, which are all methods on this trait, so a default implementation here should compile and let both `StaticExecutionGraph` and `AdaptiveExecutionGraph` drop their identical copies. That keeps the teardown logic in one place so the two graph types can't drift apart later. The lack of an existing default method in the trait is fine, this seems like a reasonable first one. Happy either way though if you'd rather keep them explicit.

ballista/scheduler/src/state/execution_stage.rs
```diff
@@ -554,15 +554,40 @@ impl RunningStage {
...
     pub fn to_failed(&self, error_message: String) -> FailedStage {
+        let task_infos = self
+            .task_infos
+            .iter()
+            .map(|task_info| {
+                task_info.as_ref().map(|info| {
+                    if matches!(info.task_status, task_status::Status::Running(_)) {
+                        TaskInfo {
+                            task_status: task_status::Status::Failed(FailedTask {
+                                error: "cancelled".to_string(),
```
andygrove: Small thing. Since `abort_running` is shared between the failure and cancellation paths, a task killed because the job *failed* (rather than being cancelled by a user) will still get `error: "cancelled"`. The `TaskKilled` reason is accurate either way, so this is minor, but would it be worth threading the actual reason through, or using something more neutral like "killed"? Not a blocker, just a clarity thought for anyone reading task statuses later.

andygrove (commented):
Nice fix. Leaving stages and tasks stuck in `Running` after a job died was a real wart, and routing both the failure and cancellation paths through a single `abort_running` is a clean way to handle it. I also appreciate the tests covering both the static and adaptive graphs, and that you ran it under the chaos generator. CI is green across the board including the TPC-H SF10 job.

My only substantive note is on your own open question about the default trait method, which I think is worth taking. There's also a tiny wording thought on the killed-task error string. Both are inline.

ballista/scheduler/src/state/execution_graph.rs
killzoner: Thanks that was my preference too, but i didn't want to deviate from existing conventions. I will drop/revert the commit then

killzoner:
Thank you for the speedy review @andygrove, I will adjust

ballista/scheduler/src/state/execution_graph.rs
killzoner: Done

ballista/scheduler/src/state/execution_stage.rs
killzoner: I just put `killed` as per your suggestion, this might change depending on the follow up PR from what we want to display in the UI

killzoner:
I think the review items have been addressed

andygrove (approved):
LGTM. Thanks @killzoner

PR #1904, opened 2026-06-26
feat: broadcast small build side of SortMergeJoinExec in the static planner

### LINKED ISSUE #1679: Extend ColocatedJoinRule and BroadcastSmallSideRule to SortMergeJoinExec
## Background

PR #1676 (tracking #1677) introduces `ColocatedJoinRule` (skip the shuffle when both join inputs are co-bucketed) and `BroadcastSmallSideRule` (promote `Partitioned` to `CollectLeft` when one side fits under a threshold). Both rules currently only match `HashJoinExec`.

PR #1651 made `SortMergeJoinExec` the default join in Ballista (issue #1648), because DataFusion's hash join has no spill support. Users now opt back into hash join via `SET datafusion.optimizer.prefer_hash_join = true`.

The combined effect: under default settings, neither of the rules from #1676 fire for the planned `SortMergeJoinExec`, so co-bucketed tables still get a network shuffle and small-side joins still get repartitioned.

## Proposed work

1. **`ColocatedJoinRule`** — extend the matcher in `try_elide_join_repartitions` to handle `SortMergeJoinExec` in addition to `HashJoinExec`. The hash-distribution-satisfaction logic (matching keys, equal or divisor-related bucket counts) is the same; only the rewrite step (stripping the `RepartitionExec` and/or wrapping a side in `BucketSubPartitionExec`) needs the new variant. Note: `SortMergeJoinExec` requires sorted inputs, so the `SortExec` above each input needs to be preserved or re-derived after the rewrite — verify it doesn't get lost when the underlying source already advertises a compatible sort order.
2. **`BroadcastSmallSideRule`** — broadcast doesn't apply to sort-merge in the same way (no `CollectLeft` mode for `SortMergeJoinExec`). Instead, when one side is small enough to fit, consider lowering to `HashJoinExec(CollectLeft)` directly — which means a join-mode swap, not just a partition-mode swap. Needs a small spike to confirm this is sound for sort-merge inputs.
3. **Tests** — add e2e snapshot cases mirroring the three in `colocated_join_e2e.rs` (matching → no exchange; divisor → `BucketSubPartitionExec`; unbucketed control), but with the default `prefer_hash_join = false` so the planned join is `SortMergeJoinExec`.

## Why this matters

Without this extension, the optimizations from #1676 are effectively gated behind a non-default config (`prefer_hash_join = true`). Users running Ballista with default settings against pre-bucketed tables won't benefit from colocation or small-side broadcast.

## References

- #1677 — feature request for the colocated-join optimizer
- PR #1676 — implementation
- PR #1651 / #1648 — sort-merge as default join
- PR #1647 — broadcast threshold config (non-AQE path)

andygrove:

Closes #1679.

# Rationale for this change

The static `DefaultDistributedPlanner` can broadcast a small build side
(`HashJoinExec(Partitioned)` → `CollectLeft`, lowered as a broadcast shuffle),
but only for `HashJoinExec`. Ballista defaults `prefer_hash_join = false`, so
TPC-H joins are planned as `SortMergeJoinExec` and broadcast never fires on the
default (non-AQE) path: small dimension tables get a full hash shuffle and force
the large side to reshuffle on the join key.

A `SortMergeJoinExec` has no `CollectLeft` mode, so when a side is small enough
this PR converts it to a `HashJoinExec(CollectLeft)` (the build side fits in
memory by definition, so the no-spill reason Ballista prefers sort-merge does not
apply) and reuses the existing broadcast lowering. This mirrors what the AQE
resolver already does for sort-merge joins.

- New config `ballista.optimizer.broadcast_sort_merge_join_enabled`
  (default `false`).
- `maybe_promote_to_broadcast` converts a small-side `SortMergeJoinExec` to a
  `HashJoinExec(Partitioned)` (dropping the now-redundant input `SortExec`) and
  runs it through the existing threshold/swap/promotion path. Handles the
  `ProjectionExec` that `swap_inputs` inserts to preserve column order.
- Tests: config default; SMJ promoted to a broadcast `CollectLeft` with sorts
  stripped; SMJ left unchanged when the flag is off or no side is under the
  threshold.

Yes — a new opt-in config. With it enabled, a `SortMergeJoinExec` whose build
side fits under `broadcast_join_threshold_bytes` is executed as a broadcast
(`CollectLeft`) hash join instead of two hash-partitioned shuffles. No SQL
semantics change.

TPC-H SF10 (AQE off, 8 partitions): ~16% faster across the 22-query suite
(q8 1.6×, q11 1.5×, q2 1.4×, q17 1.4×, q9 1.3×), with identical row counts.

A follow-up can also strip the join-key `RepartitionExec` from the converted
inputs (build broadcast from natural partitions, probe kept in place) for a
further gain; the existing hash-join broadcast path shares that limitation.


milenkovicm:
~~For the record I don't remember implementing this for AQE planner, I was thinking to implement it but don't think it has been implemented ( broadcasting probe side for sort merge )~~

AQE does support this, AQE does not support broadcast of build side in SMJ

ballista/scheduler/src/planner.rs
```diff
@@ -999,12 +1075,150 @@ order by
...
+        let plan = df.into_optimized_plan()?;
+        let plan = ctx.state().create_physical_plan(&plan).await?;
+
+        let mut planner = DefaultDistributedPlanner::new();
+        let job_uuid = Uuid::new_v4();
+        let stages =
+            planner.plan_query_stages(&job_uuid.to_string().into(), plan, &options)?;
+
+        let mut found_smj = false;
+        for stage in &stages {
```
milenkovicm: I find it easier to understand tests if they use plan represented as a string, not sure if it could be done here

milenkovicm (approved):
Thanks @andygrove this makes sense.

Just slight correction on my side, I have mistakenly assumed build side of sort merge join has been broadcasted. 

I'm not 100% sure but I believe sparks supports broadcasting build side of SMJ, perhaps that could be a good follow up (I think I saw it in TPCDS Q72 SF10)

ballista/scheduler/src/planner.rs
andygrove: Good call — I've rewritten all three SortMergeJoinExec broadcast tests to assert the stage plan as a string via the `assert_plan!` macro instead of walking the plan tree by hand. The expected plan is now spelled out inline: a broadcast `CollectLeft` hash join with the sorts stripped when the conversion fires, vs. the unchanged `SortMergeJoinExec` over sorted inputs when the flag is off or no side is under the threshold. Pushed in 7f868fca.

andygrove:
> I'm not 100% sure but I believe sparks supports broadcasting build side of SMJ, perhaps that could be a good follow up (I think I saw it in TPCDS Q72 SF10)

I filed https://github.com/apache/datafusion-ballista/issues/1922 to explore this

PR #1863, opened 2026-06-15
bug: fix empty projection serde ambiguity bug

coderfender:
# Which issue does this PR close?

Closes https://github.com/apache/datafusion-ballista/issues/1838

 # Rationale for this change

# What changes are included in this PR?

Wraps empty filterexec inside empty projection to avoid serde ambiguity
Added unit tests to verify behavior 

# Are there any user-facing changes?

PR #1906, opened 2026-06-26
chore(deps): upgrade to DataFusion 54

### LINKED ISSUE #1776: Update to datafusion 54

### LINKED ISSUE #1907: DataFusion 54: DataSourceExec shared work-queue makes each Ballista scan task read the whole table (wrong results + ~Nx slowdown)
## Describe the bug

After upgrading to DataFusion 54 (PR #1906), distributed TPC-H execution is dramatically slower. CI's TPC-H SF10 job went from ~12 min on `main` to ~52 min and then failed with `No space left on device`. Example per-query (CI, SF10, `prefer_hash_join=false`):

| query | main (DF53) | DF54 branch |
|---|---:|---:|
| q1 | 2.95 s | 40 s |
| q2 | 3.38 s | 600 s |
| q3 | 5.48 s | 143 s |
| q5 | 15.8 s | died (disk full) |

This reproduces locally and is well-characterized below. Single-node DataFusion 54 is fast, so the regression is in Ballista's distributed execution path.

## To Reproduce

SF10 parquet, 1 scheduler + 1 executor (`-c 8`), release build:

```
tpch benchmark ballista --query 1 --partitions 16 --iterations 1 -c datafusion.optimizer.prefer_hash_join=false
```

## The core finding: tasks queued behind the first wave run ~8× slower

With an executor of N task slots, the **first N tasks (wave 1) run fast; every subsequent wave runs ~8× slower for identical work.**

| executor slots | partitions | waves | q1 time |
|---:|---:|---:|---:|
| 8 | 8 | 1 | 0.50 s |
| 8 | 16 | 2 | 4.2 s |
| 8 | 24 | 3 | 8.0 s |
| 16 | 16 | 1 | 0.30 s |
| 16 | 32 | 2 | 4.0 s |

- Single-node DataFusion 54 (same data, 16 target partitions): **0.21 s**.
- DataFusion 53 (same Ballista commit), distributed, q1 @16part/8slots: **0.56 s** (no wave penalty).
- It resets per query and is purely timing/wave-dependent, not partition-index-dependent (with 16 slots, partitions 8–15 run in wave 1 and are fast).

Per-task timing (q1 @16part/8slots): wave-1 tasks ~0.45 s each; wave-2 tasks start immediately after wave 1 (no dispatch gap) but take ~3.6 s each of **compute** time.

## What the slow tasks are doing

A CPU sample during the slow wave shows all 8 worker threads **busy** (not parked) in:
`execute_query_stage → GroupedHashAggregateStream → ProjectionStream → FilterExecStream → CooperativeStream → FileStream → ParquetRecordBatchReader → snappy/RLE decode`.

So later-wave tasks are fully parallel and CPU-bound, but the **same parquet decode is ~8× slower** than in wave 1 for the same rows/task. The merged scan stage metrics also report very low `scan_efficiency_ratio` (~6%) and inflated `output_rows`/`files_opened` that scale with partition count (may be partly a metric-merge artifact, but the wall-clock slowdown is real).

## Ruled out

All tested with config propagation working (after merging the fix for client `datafusion.*` overrides):

- Stripping `CooperativeExec` from the executed plan — no effect.
- `tokio` version — unchanged (1.52.3) between DF53 and DF54 branches.
- Executor worker-thread priority (the `DedicatedExecutor` low priority) — no effect.
- Worker thread count (×8) — no effect.
- Sort-based vs hash-based shuffle — identical.
- `datafusion.optimizer.repartition_file_scans=false` — no effect (didn't change the plan).
- `datafusion.optimizer.repartition_file_min_size` huge (whole-file groups, no byte-range splitting) — made it **much worse** (51 s), so byte-range splitting is helping, not hurting.
- `datafusion.execution.parquet.pushdown_filters=true` — no effect.
- Sharing one `RuntimeEnv` across tasks (so the DF54 `FileMetadataCache` persists instead of being fresh per task) — no effect.

Caveat: a test bypassing the `DedicatedExecutor` was inadvertently applied to the pull/poll path while the cluster ran PushStaged, so "runtime bypass" is still inconclusive.

## Question / context

Has anyone seen DataFusion 54 parquet/scan behavior where tasks run after an initial batch decode the same data significantly slower on a shared process? (We hit DF54 parquet footer/metadata-caching changes when upgrading Comet.) Any pointers on what changed in 54's parquet scan or runtime that could cause later, fully-parallel, CPU-bound scan tasks to be ~8× slower would help. Pinning this likely needs a DF53↔54 scan-level comparison or CPU-counter profiling.

Blocks #1906.

### LINKED ISSUE #1909: DataFusion 54: uncorrelated scalar subqueries (q11/q15/q22) fail — ScalarSubqueryExpr cannot be deserialized in a split stage plan
## Describe the bug

On DataFusion 54, TPC-H queries with an **uncorrelated scalar subquery** fail in distributed execution: **q11, q15, q22**. The executor cannot deserialize the per-stage physical plan it receives:

```
DataFusion error: Internal error: ScalarSubqueryExpr can only be deserialized as part
of a surrounding ScalarSubqueryExec.
```

(Correlated / `IN` subqueries that decorrelate to joins — e.g. q2, q17, q20 — are unaffected.) These queries worked on DataFusion 53, so this is a DF54 regression for Ballista.

## To Reproduce

SF10 parquet, 1 scheduler + 1 executor (`-c 8`), release build:

```
tpch benchmark ballista --query 15 --partitions 16 -c datafusion.optimizer.prefer_hash_join=false
```

The query never completes (see #1908 — the scheduler marks the executor dead on the launch failure).

## Root cause

DataFusion 54 plans an uncorrelated scalar subquery as a physical `ScalarSubqueryExec` that wraps a `ScalarSubqueryExpr`. In `datafusion-proto`, a `ScalarSubqueryExpr` can only be **deserialized inside its surrounding `ScalarSubqueryExec`**, which seeds the per-scope results handle in the decode context:

- `datafusion-proto` `physical_plan/from_proto.rs` (`ExprType::ScalarSubquery`) requires `ctx.scalar_subquery_results()` to be present, else it returns the error above.
- That handle is only set while decoding the `ScalarSubqueryExec` node (`PhysicalPlanDecodeContext::with_scalar_subquery_results`).

A **whole** plan round-trips fine. But Ballista splits the physical plan into stages at shuffle boundaries, and the scalar subquery's input contains a shuffle (e.g. an aggregate), so the `ScalarSubqueryExec` and the operator that consumes its value land in **different stages**. The consuming stage is then serialized with a *bare* `ScalarSubqueryExpr`, and the executor decodes it without the surrounding `ScalarSubqueryExec` → failure.

## Possible directions

- Evaluate the uncorrelated scalar subquery as its own job/stage and inline its single value as a literal into the main plan before stage splitting (closest to how a distributed engine usually handles a broadcast scalar).
- Or keep the `ScalarSubqueryExec` and all of its consumers within a single stage so the proto round-trip stays whole.
- Or thread the scalar-subquery results handle through Ballista's per-stage decode context so a split stage can reconstruct the expr.

Input from maintainers on the preferred approach would be welcome; this likely needs design discussion (and possibly an upstream `datafusion-proto` change).

## Impact

q11, q15, q22 cannot run in distributed mode on DataFusion 54. Blocks #1906.

andygrove:

Closes #1776. Supersedes the earlier draft #1777 (which pinned a `branch-54` git rev); this is a fresh upgrade off the latest `main` against the published `datafusion = "54"`.

Also fixes two DataFusion 54 distributed-execution regressions surfaced by this upgrade: Closes #1907 and Closes #1909.

# Rationale for this change

DataFusion 54.0.0 is now released on crates.io, so Ballista can track it with a normal version dependency instead of a git pin. Re-doing the upgrade fresh on current `main` (rather than rebasing the stale draft) keeps the API migration consistent with all the code that has landed since, and validates against the released crate (whose API differs from the pre-release `branch-54` commit the draft targeted — e.g. `parse_protobuf_partitioning` now takes a `PhysicalPlanDecodeContext` rather than the extra argument the draft removed).

The upgrade also surfaced two distributed-execution regressions (caught by the `TPC-H SF10` CI job) that are fixed here so the upgrade is correct and performant end to end — see "Distributed-execution fixes" below.

- Workspace deps: `datafusion`/`datafusion-*` `54`, `arrow`/`arrow-flight` `58.3`, `object_store` `0.13.2`, and `rustyline` `18` in `ballista-cli`.
- API migration to DataFusion 54:
  - `as_any` was removed from `ExecutionPlan`/`DataSource`/`PhysicalExpr`; downcast directly on the trait object, and upcast `dyn ShuffleWriter` to `dyn ExecutionPlan` before downcasting. `as_any` is retained where it still exists (Arrow arrays, `UserDefinedLogicalNode`, `ExtensionOptions`).
  - `ExecutionPlan::partition_statistics` now returns `Result<Arc<Statistics>>`.
  - `parse_protobuf_partitioning` / `parse_protobuf_hash_partitioning` take a `PhysicalPlanDecodeContext`; `serialize_partitioning` takes the codec plus the proto converter.
  - `TaskContext::new` and `FunctionRegistry` gained higher-order-function parameters/methods, wired with empty defaults in Ballista.
  - `BatchPartitioner::new_hash_partitioner` is now fallible.
- Test updates for DataFusion 54 behavior changes:
  - `approx_percentile_cont_with_weight` result value.
  - Shuffle-writer tests made robust to hash distribution (54 changed the hash split) — assert row conservation and valid file paths rather than an exact per-partition split.
  - Execution-graph / dot tests build their physical plan with the Ballista session config (or `hash_join_single_partition_threshold = 0`) so the small test join keeps its `Partitioned` plan, matching production.
  - Regenerated AQE plan snapshots (54 folds a `ProjectionExec` into `HashJoinExec` as a `projection=` field).

## Distributed-execution fixes

- **Scan over-read / wrong results (#1907).** DataFusion 54's `DataSourceExec` hands file groups to partition streams from a shared work-queue that is only divided across partitions when all partitions of one plan instance are polled concurrently. Ballista executes one partition per task on its own decoded plan instance, so a task polling a single partition in isolation drained the whole queue and scanned the entire table — producing N-times-inflated aggregates and N-times the work for every task after the first wave. The executor now restricts each task's `DataSourceExec` to the file group for its `partition_id` before execution.
- **Uncorrelated scalar subqueries (#1909).** DataFusion 54 plans uncorrelated scalar subqueries as a physical `ScalarSubqueryExec` whose `ScalarSubqueryExpr` can only be deserialized inside that exec; Ballista's stage splitting separated them, so the executor could not decode the stage plan (TPC-H q11/q15/q22). The Ballista restricted configuration now disables `datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery`, so the optimizer rewrites these to joins, which Ballista distributes correctly.

Ballista now runs on DataFusion 54. No Ballista API or SQL semantics change.

Verified: `cargo build --workspace --all-targets`, `cargo clippy --all-targets --workspace -- -D warnings`, and `cargo test --workspace` all clean (including new unit tests for both distributed-execution fixes). The full 22-query TPC-H SF10 suite runs end to end on a distributed cluster with correct results and no regression.


andygrove:
There seems to be a major performance regression. This was caught by the `TPC-H SF10` job in CI. I am investigating.

andygrove:
The Parquet scans are reading too many rows. Still trying to find the root cause.

andygrove:
Following up on my notes above (the perf regression / "parquet scans reading too many rows") — root-caused and fixed, and it turned out to be a correctness bug, not just performance.

**What was happening:** DataFusion 54's `DataSourceExec` distributes file groups across partitions via a shared work-queue that is only divided when *all* partitions of one plan instance are polled concurrently. Ballista runs one partition per task on its own decoded plan instance, so a task polling a single partition in isolation drained the whole queue and **scanned the entire table**. That is the "too many rows" I was seeing — every task dispatched after the first wave re-read the whole table, which inflated aggregates by N× (e.g. q1 `count_order` was exactly 8× too high on a 2-wave run) and burned N× the CPU. The benchmark harness only validates results at SF1, so it slipped through as a "perf" symptom. Details in #1907.

**Fix:** restrict each task's `DataSourceExec` to the file group for its `partition_id` before execution. q1 on `-c8` is correct again (A/F `count_order` 14,804,077) and back to ~0.5 s.

While verifying the full suite I hit a second, independent DF54 regression: uncorrelated scalar subqueries (q11/q15/q22) failed to deserialize once the plan is split into stages, because DF54 plans them as a `ScalarSubqueryExec` whose inner `ScalarSubqueryExpr` can't be decoded on its own. Filed as #1909 and fixed here by decorrelating them to joins (`enable_physical_uncorrelated_scalar_subquery = false` in Ballista's restricted config).

Both fixes are in this PR with unit tests, and the **full 22-query TPC-H SF10 suite now runs end to end with correct results and no regression**. I've updated the PR description accordingly.

One related robustness issue I noticed but left out of this PR (it's pre-existing and orthogonal): a deterministic task *decode* failure makes the scheduler mark the executor dead, which with a single executor hangs the job instead of failing the query. Filed separately as #1908.

ballista/core/src/extension.rs
```diff
@@ -789,6 +789,22 @@ impl SessionConfigHelperExt for SessionConfig {
...
+            // splits a plan into stages the expr is serialized without its exec
+            // and the executor fails to decode it. Disabling this option makes
+            // the optimizer rewrite uncorrelated scalar subqueries to joins,
+            // which Ballista distributes correctly.
+            //
+            // See https://github.com/apache/datafusion-ballista/issues/1909
+            .set_bool(
+                "datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery",
+                false,
+            )
```
andygrove, line 807: I filed follow on issue https://github.com/apache/datafusion-ballista/issues/1910 to optimize this

ballista/core/src/registry.rs
```diff
@@ -106,6 +107,19 @@ impl FunctionRegistry for BallistaFunctionRegistry {
         self.window_functions.keys().cloned().collect()
     }
 
+    fn higher_order_function_names(&self) -> HashSet<String> {
+        HashSet::new()
```
martin-g: This will break the built-in HOFs from DataFusion.
IMO it would be better to copy them from the state at https://github.com/apache/datafusion-ballista/blob/85805db74ffaa7aca25009099ecd06819163dc22/ballista/core/src/registry.rs#L154-L166

ballista/executor/src/execution_engine.rs
```diff
@@ -101,19 +106,61 @@ impl DefaultExecutionEngine {
...
+/// Returns `None` for any node that is not a file-backed `DataSourceExec`, and
+/// for a `partition_id` outside the source's file groups (e.g. when an operator
+/// between the scan and the stage output changed the partition count).
+fn restrict_scan_to_partition(
+    plan: &Arc<dyn ExecutionPlan>,
+    partition_id: usize,
+) -> Option<Arc<dyn ExecutionPlan>> {
+    let exec = plan.downcast_ref::<DataSourceExec>()?;
+    let source: &dyn Any = exec.data_source().as_ref();
+    let config = source.downcast_ref::<FileScanConfig>()?;
```
martin-g: If the config is different type, e.g. `MemorySourceConfig` then [Ok(Transformed::no(p))](https://github.com/apache/datafusion-ballista/pull/1906/changes#diff-708c49b754b125819c73edc3dd46dee0f92f2031bd09426a11990ecd05251601R179) will be executed and it may return duplicate results, no ?
Maybe log a WARN here before returning `None` ?!

ballista/core/src/serde/mod.rs
```diff
@@ -381,15 +382,15 @@ impl PhysicalExtensionCodec for BallistaPhysicalExtensionCodec {
                 )
             })?;
         let converter = DefaultPhysicalProtoConverter {};
+        let decode_ctx = PhysicalPlanDecodeContext::new(ctx, self.default_codec.as_ref());
```
martin-g:
```suggestion
        let decode_ctx = PhysicalPlanDecodeContext::new(ctx, self);
```

martin-g (commented):
Should https://github.com/apache/datafusion-ballista/pull/1863/changes#diff-a52fcf388ba654263ab5f17dbf84452ffb97bda2c0a082fc347bb5ffe6c55591R422 be reverted with this PR too ? Or in a follow-up ?

andygrove:
Thanks for the review @martin-g. I have addressed feedback so far.

martin-g (approved):

PR #1911, opened 2026-06-28
perf: prune irrelevant partitions task

### LINKED ISSUE #1880: Improve stage encoding size by removing unrelated partition location

`ShuffleReaderExec` holds information about each partition (`PartitionLocation`) even this information is not required for a specific task (partition). This will make plan very big when partition number is big, making serialistaion hitting grpc message size (https://github.com/apache/datafusion-ballista/pull/1853#issuecomment-4741128632)

Currently, once serialised stage plan is cached (and reused for each task), this functionallity could be disabled using `disable-stage-plan-cache` scheduler feature, but even disabled it will still encode information about all partitions 

Implement stage encoding logic which would take into account actual partition and remove unrelated data 

**Additional context**

This work is orthogonal to #1853

Scheduler feature `disable-stage-plan-cache` might disable plan caching but current  implementation will not take into account actual partition information,

I had a quick experiment with 

```diff
diff --git a/ballista/scheduler/Cargo.toml b/ballista/scheduler/Cargo.toml
index f882d26f..3a60d897 100644
--- a/ballista/scheduler/Cargo.toml
+++ b/ballista/scheduler/Cargo.toml
@@ -34,7 +34,7 @@ required-features = ["build-binary"]
 
 [features]
 build-binary = ["clap", "tracing-subscriber", "tracing-appender", "tracing", "ballista-core/build-binary"]
-default = ["build-binary", "rest-api"]
+default = ["build-binary", "rest-api", "disable-stage-plan-cache"]
 # job info can cache stage plans, in some cases where
 # task plans can be re-computed, cache behavior may need to be disabled.
 disable-stage-plan-cache = []
diff --git a/ballista/scheduler/src/state/task_manager.rs b/ballista/scheduler/src/state/task_manager.rs
index f04c10e6..255aa27e 100644
--- a/ballista/scheduler/src/state/task_manager.rs
+++ b/ballista/scheduler/src/state/task_manager.rs
@@ -167,6 +167,7 @@ impl JobInfoCache {
     fn encode_stage_plan<U: AsExecutionPlan>(
         &mut self,
         stage_id: usize,
+        partition_id: usize,
         plan: &Arc<dyn ExecutionPlan>,
         codec: &dyn PhysicalExtensionCodec,
     ) -> Result<Vec<u8>> {
@@ -186,13 +187,57 @@ impl JobInfoCache {
     fn encode_stage_plan<U: AsExecutionPlan>(
         &mut self,
         _stage_id: usize,
+        partition_id: usize,
         plan: &Arc<dyn ExecutionPlan>,
         codec: &dyn PhysicalExtensionCodec,
     ) -> Result<Vec<u8>> {
+        use ballista_core::execution_plans::ShuffleReaderExec;
+        use datafusion::common::tree_node::{Transformed, TreeNode};
+
+        let transformed = plan.clone().transform_up(|node| {
+            if let Some(reader) = node.as_any().downcast_ref::<ShuffleReaderExec>() {
+                if reader.broadcast {
+                    return Ok(Transformed::no(node));
+                }
+                let new_partition: Vec<Vec<_>> = reader
+                    .partition
+                    .iter()
+                    .enumerate()
+                    .map(|(i, locs)| {
+                        if i == partition_id || locs.len() <= 1 {
+                            locs.clone()
+                        } else {
+                            vec![]
+                        }
+                    })
+                    .collect();
+                let new_reader = if let Some(coalesce) = reader.coalesce.clone() {
+                    ShuffleReaderExec::try_new_coalesced(
+                        reader.stage_id,
+                        new_partition,
+                        coalesce,
+                        reader.schema(),
+                        reader.properties().output_partitioning().clone(),
+                    )?
+                } else {
+                    ShuffleReaderExec::try_new(
+                        reader.stage_id,
+                        new_partition,
+                        reader.schema(),
+                        reader.properties().output_partitioning().clone(),
+                    )?
+                };
+                Ok(Transformed::yes(
+                    Arc::new(new_reader) as Arc<dyn ExecutionPlan>
+                ))
+            } else {
+                Ok(Transformed::no(node))
+            }
+        })?;
+
         let mut plan_buf: Vec<u8> = vec![];
-        let plan_proto = U::try_from_physical_plan(plan.clone(), codec)?;
+        let plan_proto = U::try_from_physical_plan(transformed.data, codec)?;
         plan_proto.try_encode(&mut plan_buf)?;
-
         Ok(plan_buf)
     }
 }
@@ -656,6 +701,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
         if let Some(mut job_info) = self.active_job_cache.get_mut(&job_id) {
             let plan = job_info.encode_stage_plan::<PhysicalPlanNode>(
                 stage_id,
+                task.partition.partition_id,
                 &task.plan,
                 self.codec.physical_extension_codec(),
             )?;
@@ -733,6 +779,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
             if let Some(mut job_info) = self.active_job_cache.get_mut(&job_id) {
                 let plan = job_info.encode_stage_plan::<PhysicalPlanNode>(
                     stage_id,
+                    task.partition.partition_id,
                     &task.plan,
                     self.codec.physical_extension_codec(),
                 )?;
```

and there is no encoding exception as seen in https://github.com/apache/datafusion-ballista/pull/1853#issuecomment-4742357916 but for some reason it returns 0 result, which i believe its due to last stage for which, my quick change, does not work 

Image

coderfender:

Closes #1880

 # Rationale for this change

The current version in ballista sends out all partition information to every task on the executor. This becomes a bottleneck when the total partitions are high essentially reaching gRPC limit causing scale issues

The changes essentially eliminate redundant partition information being sent to the tasks. This is achieved by walking the plan and pruning partition id the task would be working on. There are special edge cases where a task works on more than ine partition (broadcast / coalesce) which are handled in the code

Added some unit tests as well

No

No


ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +165,64 @@ impl JobInfoCache {
             encoded_stage_plans: HashMap::new(),
         }
     }
+
+    #[allow(dead_code)]
```
andygrove:
```suggestion
    #[cfg(feature = "disable-stage-plan-cache")]
```

andygrove:
Thanks @coderfender. Could you fill out the PR description

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +165,64 @@ impl JobInfoCache {
...
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let n = plan.output_partitioning().partition_count();
+        let wanted: HashSet<usize> = partition_ids.iter().copied().collect();
+        Ok(plan
+            .clone()
+            .transform_up(|node| {
+                let Some(r) = node.as_any().downcast_ref::<ShuffleReaderExec>() else {
+                    return Ok(Transformed::no(node));
+                };
+                if r.broadcast || r.partition.len() != n {
```
martin-g, line 189:
```suggestion
                if r.broadcast || r.partition.len() != partition_ids.len() {
```
isn't this more correct ?

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +165,64 @@ impl JobInfoCache {
...
         }
     }
+
+    #[cfg(feature = "disable-stage-plan-cache")]
+    fn partition_prune_helper(
+        partition_ids: &[usize],
+        plan: &Arc<dyn ExecutionPlan>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let n = plan.output_partitioning().partition_count();
+        let wanted: HashSet<usize> = partition_ids.iter().copied().collect();
```
martin-g, line 179: nit: how many partition ids are expected here ?
If it is a small number then `partition_ids.contains(&i)` is OK too and it will avoid the extra allocation.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +165,64 @@ impl JobInfoCache {
...
+                    .enumerate()
+                    .map(|(i, loc)| {
+                        if wanted.contains(&i) {
+                            loc.clone()
+                        } else {
+                            vec![]
+                        }
+                    })
+                    .collect();
+
```
martin-g, line 213: consider returning early if there was no pruning:

```suggestion
             if wanted.len() == r.partition.len() { 
               return Ok(Transformed::no(node));
             }
```

andygrove (commented):
Thanks for tackling this, the gRPC size ceiling at high partition counts is a real scaling limit and pruning per-task locations is a nice way to attack it. I left a few questions, mainly about the partition-alignment invariant and getting the `Clippy` build green. I skipped benchmarking since the change is off by default and the win is serialized-plan size rather than query time, which TPC-H at normal partition counts would not show.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +169,64 @@ impl JobInfoCache {
...
+    fn partition_prune_helper(
+        partition_ids: &[usize],
+        plan: &Arc<dyn ExecutionPlan>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let n = plan.output_partitioning().partition_count();
+        let wanted: HashSet<usize> = partition_ids.iter().copied().collect();
+        Ok(plan
+            .clone()
+            .transform_up(|node| {
+                let Some(r) = node.as_any().downcast_ref::<ShuffleReaderExec>() else {
```
andygrove: This does not compile against current `main`, and it looks like what the `Clippy` job is failing on. DataFusion 54 removed `ExecutionPlan::as_any`, so `node.as_any().downcast_ref::<ShuffleReaderExec>()` no longer resolves. Calling `node.downcast_ref::<ShuffleReaderExec>()` directly should work, since DataFusion 54 added that inherent method on `dyn ExecutionPlan` and it auto-derefs through `Arc`. A rebase on latest `main` will surface this locally.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +169,64 @@ impl JobInfoCache {
...
+                    return Ok(Transformed::no(node));
+                };
+                if r.broadcast || r.partition.len() != n {
```
andygrove, line 189: This is a clean way to scope the pruning to readers that line up with the task, and I want to make sure I follow the safety argument. `n` is the stage plan output partition count and `wanted` is the task output `partition_id`, so keeping only reader partition `i == partition_id` is correct as long as reader partition `i` always feeds output partition `i`. The `r.partition.len() != n` guard handles the coalesce and merge shapes where the counts differ. For the equal-count case, is the reader-partition-`i`-to-output-`i` mapping guaranteed across every stage shape, including multi-reader joins and the AQE planner that also builds `ShuffleReaderExec` in `adapter.rs`? If a stage ever had an `n`-partition reader whose partition `i` did not map to output `i`, this would drop the locations the task needs, and the `vec![]` fallback would make it a silent wrong result rather than an error. A `debug_assert` or a short comment spelling out the invariant would help the next reader trust it.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -163,12 +169,64 @@ impl JobInfoCache {
             encoded_stage_plans: HashMap::new(),
         }
     }
+
+    #[cfg(feature = "disable-stage-plan-cache")]
+    fn partition_prune_helper(
```
andygrove, line 174: Since this can quietly change which partition locations a task receives, it might be worth a unit test or two. For example, build a small stage plan with a `ShuffleReaderExec`, run `partition_prune_helper` for a single partition, and check that the wanted partition keeps its locations while the others are emptied, with the broadcast and coalesce readers left untouched. A serde round-trip on the pruned plan would also lock in that the empty partition vectors encode and decode cleanly.

ballista/scheduler/src/state/task_manager.rs
coderfender: @milenkovicm  's SF1000 test failed with too much metadata gRPC limit per task. So if we have 1000's of partitions processed in this stage ,a hashset's O(1) lookup would be faster and worth the allocation

ballista/scheduler/src/state/task_manager.rs
coderfender: Not really, 
  1.  n = the stage plan's output partition count. It's the correctness guard: it confirms the reader lines up 1:1 with the stage output, so the entries in partition_ids are valid indices into r.partition.
  1. partition_ids.len() = how many partitions this task/batch handles which is a subset. Using it as the guard skips pruning exactly when a task handles fewer than all partitions which is the intent of this PR

andygrove (approved):
Thanks @coderfender. I don't have a suitable benchmark setup yet to really prove any performance benefit, but we know from Comet experience that this is needed at scale.

coderfender:
Thank you. Perhaps @milenkovicm  could run SF100 in turn linked issue ?

andygrove:
> Thank you. Perhaps @milenkovicm  could run SF100 in turn linked issue ?

I ran sf100 but there are only 32 partitions. The issue this PR solves is not a bottleneck at this scale.

milenkovicm:
Just increase number of partitions, and observe grpc message size exception, main will fail after maybe 512ish. Will try later today

coderfender:
Thank you @milenkovicm

milenkovicm:
I've run master with

```
datafusion-ballista git:(master) cargo run --bin tpch -- benchmark ballista -p tpch_data/tpch-data-sf10/ -f parquet -i 1 --port 50050 --host 127.0.0.1 -c datafusion.execution.target_partitions=1024 -q 2
```

1024 partitions, and it works, no GRPC exception, 

thanks @coderfender

milenkovicm:
I spoke too soon, its works much better than it was but still sometimes there is gRPC issue 

```
2026-07-13T21:46:07.497017Z ERROR tokio-rt-worker ThreadId(12) ballista_scheduler::state: Failed to launch new task: Internal Ballista error: Failed to connect to executor 82b9124d-8251-4c03-958b-0e4b061583d1: Status { code: OutOfRange, message: "Error, decoded message length too large: found 120519656 bytes, the limit is: 16777216 bytes", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 13 Jul 2026 21:46:07 GMT"} }, source: None }
2026-07-13T21:46:07.497035Z  INFO tokio-rt-worker ThreadId(12) ballista_scheduler::state::executor_manager: Removing executor 82b9124d-8251-4c03-958b-0e4b061583d1: Some("Failed to launch new task: Internal Ballista error: Failed to connect to executor 82b9124d-8251-4c03-958b-0e4b061583d1: Status { code: OutOfRange, message: \"Error, decoded message length too large: found 120519656 bytes, the limit is: 16777216 bytes\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Mon, 13 Jul 2026 21:46:07 GMT\"} }, source: None }")
```

milenkovicm:
we need to get #1853 in as well

coderfender:
Sure @milenkovicm  , I plan to continue my work on column stats collection

PR #1912, opened 2026-06-28
feat: execute uncorrelated scalar subqueries first instead of rewriting to a join

### LINKED ISSUE #1910: Execute uncorrelated scalar subqueries first and substitute the value, instead of rewriting to a join

#1909 made uncorrelated scalar subqueries work in distributed execution by setting `datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery = false`, which makes the optimizer rewrite them to joins (the previous physical `ScalarSubqueryExec` cannot be serialized across Ballista's stage boundaries).

That fix is **correct but not efficient**. An uncorrelated scalar subquery produces exactly one value, but rewriting it to a join turns it into a (cross / left) join against the subquery's single-row result. Compared to using the value as a constant, the join adds extra shuffle/build/probe work — for example TPC-H q11/q15/q22, where the subquery is a single aggregate that is then compared (`=`, `>`) against the outer rows.

Execute the uncorrelated scalar subquery **first**, then substitute its value into the original plan:

1. Detect uncorrelated scalar subqueries during distributed planning.
2. Run each subquery's plan as its own job/stage(s) and collect its single scalar result (erroring if it returns more than one row, matching scalar-subquery semantics).
3. Replace the `ScalarSubquery` / `ScalarSubqueryExpr` in the main plan with the materialized value as a literal.
4. Plan and execute the (now subquery-free) main plan.

This matches how the value is logically used — a constant — and avoids the join overhead entirely. It also removes the need to disable `enable_physical_uncorrelated_scalar_subquery`, since there would no longer be a `ScalarSubqueryExec` to serialize.

This is essentially sequential/dependent subquery execution in the scheduler: the main query depends on the subquery's result, so the subquery's stages run to completion first and feed a constant into the dependent plan.

- **Decorrelate to a join (current behavior, #1909):** correct but adds join work for what is logically a constant.
- **Keep `ScalarSubqueryExec` and serialize it across stages:** would require teaching `datafusion-proto` / the scheduler to carry the shared in-process results container across processes, which isn't possible — the value has to be materialized and transmitted regardless.

## Additional context

Follow-on to #1909. Once this lands, the `enable_physical_uncorrelated_scalar_subquery = false` default set in #1909 could be revisited.

andygrove:

Closes #1910.

# Rationale for this change

#1909 made uncorrelated scalar subqueries work in distributed execution by
disabling `datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery`,
which makes the optimizer rewrite them to joins (DataFusion 54's physical
`ScalarSubqueryExec` cannot be serialized across Ballista's stage boundaries).

That is correct but not efficient: an uncorrelated scalar subquery produces a
single value, but rewriting it to a join turns it into a (cross / left) join
against the subquery's one-row result, adding shuffle/build/probe work for what
is logically a constant.

This PR runs the subquery first and substitutes its value, which is both
efficient and avoids the `ScalarSubqueryExec` serialization problem entirely.

- In the client query planner (`BallistaQueryPlanner::create_physical_plan`),
  before sending the plan to the scheduler:
  - detect **uncorrelated** `Expr::ScalarSubquery` nodes in the optimized
    logical plan (correlated subqueries are left to DataFusion's existing
    decorrelation rules);
  - run each subquery as its own distributed query and reduce its result to a
    single `ScalarValue` (0 rows → typed null, 1 row → the value, more than one
    row → an error);
  - substitute the value as a literal. Subqueries nested inside a subquery are
    materialized recursively, innermost first.
- Reverts the `enable_physical_uncorrelated_scalar_subquery = false` default
  added in #1909, since the subqueries are now materialized rather than
  decorrelated. This supersedes that workaround.

No API or SQL semantics change. Uncorrelated scalar subqueries (e.g. TPC-H
q11/q15/q22) now execute on a Ballista cluster by computing the subquery first
and inlining its value.

Tested with new unit tests for the detection/substitution helpers (including
that correlated subqueries are skipped) and a standalone + remote integration
test that runs `where id = (select max(id) from test)` end to end. `cargo test`,
`cargo fmt`, and `cargo clippy --all-targets -- -D warnings` are clean.


ballista/client/tests/context_checks.rs
```diff
@@ -77,6 +77,40 @@ mod supported {
...
+        ctx: SessionContext,
+        test_data: String,
+    ) -> datafusion::error::Result<()> {
+        ctx.register_parquet(
+            "test",
+            &format!("{test_data}/alltypes_plain.parquet"),
+            Default::default(),
+        )
+        .await?;
+
```
milenkovicm, line 102: can we explicitly disable `SET datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery = true`, it would give additional context to the test.

ballista/client/tests/context_checks.rs
```diff
@@ -77,6 +77,40 @@ mod supported {
...
+            .await?
+            .collect()
+            .await?;
+        let expected = ["+-----+", "| cnt |", "+-----+", "| 1   |", "+-----+"];
+
+        assert_batches_eq!(expected, &result);
+
+        Ok(())
+    }
+
```
milenkovicm, line 126: can we also add check with AQE enabled, if it works, if not a jira issue or a test in "unsupported" file?

milenkovicm (approved):
thanks @andygrove  just few minor comments

ballista/client/tests/context_checks.rs
andygrove: Done — the test now runs `SET datafusion.optimizer.enable_physical_uncorrelated_scalar_subquery = true` before the query, so it states that it covers the case where the optimizer is free to plan a physical scalar subquery.

ballista/client/tests/context_checks.rs
andygrove: Added `standalone_aqe` / `remote_aqe` cases (new `standalone_context_with_aqe()` / `remote_context_with_aqe()` fixtures set `ballista.planner.adaptive.enabled=true`). All four cases pass — materialization happens in the client query planner before the plan reaches the scheduler, so it is independent of which distributed planner runs. No issue needed.

ballista/core/src/planner.rs
```diff
@@ -151,10 +156,20 @@ impl<T: 'static + AsLogicalPlan> QueryPlanner for BallistaQueryPlanner<T> {
                 _ => {
                     log::debug!("create_physical_plan - handling general statement");
 
+                    // Execute any uncorrelated scalar subqueries first and
+                    // substitute their values, so the plan sent to the scheduler
+                    // is subquery-free. See #1910.
+                    let logical_plan = self
```
phillipleblanc, line 162: I believe this will eagerly execute for `EXPLAIN` plans which shouldn't perform any execution, not a huge deal but something worth calling out.

ballista/core/src/planner.rs
```diff
@@ -164,6 +179,132 @@ impl<T: 'static + AsLogicalPlan> QueryPlanner for BallistaQueryPlanner<T> {
...
+            Arc::new(DistributedQueryExec::<T>::with_extension(
+                self.scheduler_url.clone(),
+                self.config.clone(),
+                plan,
+                self.extension_codec.clone(),
+                session_state.session_id().to_string(),
+            ));
+
+        let batches =
+            datafusion::physical_plan::collect(exec, session_state.task_ctx()).await?;
```
phillipleblanc, line 246: nit: This will collect all of the results into memory before asserting it only has a single row. It should stream the results and stop as soon as it gets more than one row.

phillipleblanc (approved):
Thanks for doing this Andy! Some minor feedback

PR #1913, opened 2026-06-28
ci: run TPC-H SF10 under AQE off + on, and build it faster

andygrove:

N/A — CI coverage + speed improvement.

# Rationale for this change

Two related improvements to the `TPC-H SF10` CI job:

1. **AQE coverage.** The job ran the suite only with AQE off
   (`prefer_hash_join=false`). The adaptive (AQE-on) planner is a different
   planner with different join behavior — it does dynamic broadcast join
   selection — and had **no CI coverage**. A change can pass CI while breaking
   the adaptive path (e.g. a broadcast-under-AQE change failing a query that the
   AQE-off run never exercises).

2. **Faster build.** This job is an integration/correctness check — it asserts
   only that the queries run (no timing gate; real perf numbers come from the
   release profiles and the benchmark-verify job). It was built at opt-level 3
   (`release-nonlto`), and that compile dominates the job's wall time. It does
   not need an opt-level 3 build to check correctness.

- Run the suite a second time with `-c ballista.planner.adaptive.enabled=true`
  (AQE on), in addition to AQE off, using the **same generated data and the same
  cluster** (AQE on/off is a per-query session setting, so the scheduler selects
  the planner per job — no need to regenerate data or restart the cluster). The
  query loop is refactored into a `run_suite` shell function called once per
  mode.
- Add a `tpch-ci` build profile (opt-level 1, codegen-units 256, no LTO) and
  build this job with it. Dependencies compile far faster; SF10 queries stay
  fast enough for a correctness run. Adding the second (AQE-on) pass is cheap
  by comparison, so net job time should drop despite the extra coverage.
- Job renamed to `TPC-H SF10 (all queries, AQE off + on)`.

No change to data generation, query set (q16 still omitted), partitions (16),
iterations, or cluster shape.

No. CI-only change. The new `tpch-ci` profile is for this job only; performance
benchmarking still uses the `release-nonlto` / `release-lto` profiles.

### DIFF
--- a/.github/workflows/tpch.yml
+++ b/.github/workflows/tpch.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Build Ballista binaries
         run: |
-          cargo build --profile release-nonlto --locked \
+          cargo build --profile tpch-ci --locked \
             -p ballista-scheduler \
             -p ballista-executor \
             -p ballista-benchmarks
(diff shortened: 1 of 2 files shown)


milenkovicm (approved):
Thanks @andygrove 
It's good to have AQE enabled, just take a note it may still have a ton of bugs

Dandandan (approved):

martin-g (approved):

PR #1916, opened 2026-06-30
feat: Column statistics support v1

### LINKED ISSUE #1827: Improve shuffle (column) statistics

Currently, shuffle collects basic statistics like the number of rows and bytes. This data is used downstream for some optimisation. However, I understand those statistics aren’t sufficient for join estimations, leading to incorrect decisions in some cases. No column based statistics such as null count or distinct count is collected during exchange.

investigate what would be appropriate approach to collect null and distinct count for columns used in repartition expressions. Collecting null count may not be big problem but distinct count might and approximation may be used for it 

**Additional context**

- collected statistic should be propagated to planners

coderfender:

Closes https://github.com/apache/datafusion-ballista/issues/1827

Introduced new `TaskColumnStats` proto to compute null count , distinct values (TBD) for each column at the partition / stage level. Follow up PRs should leverage this information to make better decisions like broadcast / AQE etc 

 # Rationale for this change


python/Cargo.lock
martin-g, line 1: This removal looks wrong

ballista/scheduler/src/state/execution_stage.rs
```diff
@@ -797,6 +797,7 @@ impl RunningStage {
                         task_status::Status::Successful(SuccessfulTask {
                             executor_id,
                             partitions: _,
+                            task_column_stats: _,
```
andygrove: Following the stats through, they are collected on the executor, put on `SuccessfulTask`, and sent here, but this match drops them with `task_column_stats: _` and nothing else stores them. So today the executor pays the per-batch fold, the extra repeated field on every successful task, and the network transfer, and the scheduler discards the result. Issue #1827 wants these propagated to planners, and right now they do not reach a stored location. Would it be worth landing at least the scheduler-side storage in the same PR, or gating the collection behind a config that defaults off until a consumer exists? That keeps the plumbing you are building without paying for discarded data on every task in the meantime.

ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -368,7 +368,10 @@ impl SortShuffleWriterExec {
                 }
             }
 
-            Ok(results)
+            Ok(ShuffleWriteResult {
+                partitions: results,
+                column_stats: vec![],
```
andygrove: Only the `Hash` repartition path in `ShuffleWriterExec` folds null counts. This sort-based path and the single-partition path both return `column_stats: vec![]`. That is probably fine for v1, but a downstream consumer will get stats for some stages and empty for others. Could this carry a short comment noting that column stats are collected on the hash path only, so whoever consumes them later knows the coverage is partial?

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -126,6 +127,38 @@ struct ShuffleWriteMetrics {
...
+pub struct ShuffleWriteResult {
+    /// partition shuffle write information
+    pub partitions: Vec<ShuffleWritePartition>,
+    /// column stats folded across all partitions for each column
+    pub column_stats: Vec<TaskColumnStats>,
+}
+
+/// Fold the per-column null counts of `batch` into `null_counts` (indexed by column position).
+/// Shared by every shuffle-write path so stats are collected consistently.
+pub(crate) fn accumulate_null_counts(null_counts: &mut [u64], batch: &RecordBatch) {
```
andygrove: Since these two helpers are the heart of the feature, it might be worth a small unit test. Something that folds a couple of `RecordBatch`es with known nulls per column and checks `accumulate_null_counts` sums them per column position, plus a check that `null_counts_to_column_stats` maps positions to columns and leaves `hll_sketch` empty. A `TaskColumnStats` proto encode/decode round-trip would also lock the wire shape in.

andygrove (commented):
Thanks for starting this. Collecting shuffle-time column statistics is a solid foundation for smarter broadcast and AQE decisions, and folding null counts across the input batches is a clean place to gather them. I left a few notes. The main one is that the collected stats are currently dropped at the scheduler, so it might be worth landing a consumer or a storage hook alongside the collection, or gating it behind a config until then. martin-g's note on `python/Cargo.lock` looks right, and it appears to be what is failing the `Test Python Release` jobs, since `pyballista` fails to compile against a mismatched DataFusion version. Reverting that lockfile to `main` and regenerating only if a dependency really changed should get those green.

coderfender:
Thank you . Let me address the PR comments and update here

ballista/scheduler/src/state/execution_stage.rs
coderfender: Thank you for the feedback . I added scheduler side consumption into a DF native `ColumnStats` data structure for future reliance to improve distributed planning. I plan to add other fields `min` `max` etc in a follow up PR (which are left empty for now)

ballista/core/src/execution_plans/sort_shuffle/writer.rs
coderfender: True. I plan to add stat calculation for other repartition types in follow up PR. For now, I've added comments explaining the gap

ballista/core/src/execution_plans/shuffle_writer.rs
coderfender: Agreed. I added couple of unit tests

coderfender:
Created issue for part 2 : https://github.com/apache/datafusion-ballista/issues/1960 ( followup)

coderfender:
Fixing merge conflicts

coderfender:
Rebased with main

coderfender:
Added code to compute and consume col stats (null counts) on scheduler side . The HLL sketch is still not comupted an will be followed up with a PR.

coderfender:
@milenkovicm  , Please take a look whenever you get a chance

ballista/executor/src/lib.rs
```diff
@@ -138,14 +138,16 @@ pub fn as_task_status(
...
-        Ok(partitions) => {
+        Ok(shuffle_write_result) => {
             debug!(
                 "Task {task_id} finished with operator_metrics array size {} \
                  and {} runtime-stats report(s), {} window-state report(s)",
                 metrics.len(),
                 runtime_stats.len(),
                 window_state.len(),
             );
+            let partition = shuffle_write_result.partitions;
```
villebro: nit:
```suggestion
            let partitions = shuffle_write_result.partitions;
```

ballista/executor/src/lib.rs
```diff
@@ -157,8 +159,9 @@ pub fn as_task_status(
                 metrics,
                 status: Some(task_status::Status::Successful(SuccessfulTask {
                     executor_id,
-                    partitions,
+                    partitions: partition,
```
villebro: ..continuation of the previous nit:
```suggestion
                    partitions: partitions,
```

ballista/core/proto/ballista.proto
```diff
@@ -829,6 +829,12 @@ message FailedTask {
   }
 }
 
+message TaskColumnStats {
+  uint32 column      = 1;
+  uint64 null_count  = 2;
+  bytes  hll_sketch  = 3;
```
villebro: At the risk of opening a can of overabstraction, this feels very specific (only single HLL sketch supported). Could we consider generalizing to support arbitrary sketch types and counts?

villebro (commented):
A few comments

ballista/core/proto/ballista.proto
coderfender: Thank you for the review. Totally agree that we could enhance this but this PR was more of the wiring and plan to follow up with further sketch types in follow up PRs

coderfender: Also note that hll_sketch isnt populated and just a place holder :)

PR #1919, opened 2026-07-01
chore(deps): upgrade pyballista to DataFusion 54

djanand:

Closes #.

 # Rationale for this change

#1906 upgraded the main workspace to DataFusion 54, but the `python/` sub-project (a separate Cargo workspace that path-depends on the ballista crates) was left pinned to DataFusion 53. As a result, `python/` now mixes two DataFusion majors: `pyballista` compiles the ballista path-deps against datafusion 54 while its own `datafusion` dependency resolves to 53, which fails to build: 
```
the trait bound datafusion::prelude::SessionConfig: ballista::extension::SessionConfigExt is not satisfied
  = note: there are multiple different versions of crate datafusion_execution in the dependency graph
```
 This PR brings `python/` up to DataFusion 54 so the python build is consistent again.

Bump the `python/` sub-project from DataFusion 53 to 54, keeping both ecosystems in lockstep:

- `python/Cargo.toml` — `datafusion`, `datafusion-proto`, `datafusion-python` → `54`
- `python/pyproject.toml` — `datafusion==54`
- `python/Cargo.lock` — regenerated (now resolves a single datafusion major, `54`)
- `python/uv.lock` — regenerated (`datafusion` python package → `54`)

No source changes were required 
Verified locally: `cargo check`, `maturin develop`, and `pytest` (`73 passed, 1 skipped`) all succeed against DataFusion 54.

The `ballista` Python package now builds against and requires DataFusion 54 (`datafusion==54`). No API changes to the package itself.

No breaking changes to public Rust APIs.

### DIFF
--- a/python/Cargo.lock
+++ b/python/Cargo.lock
@@ -3,54 +3,6 @@
 version = 4
 
 [[package]]
-name = "abi_stable"
-version = "0.11.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69d6512d3eb05ffe5004c59c206de7f99c34951504056ce23fc953842f12c445"
-dependencies = [
- "abi_stable_derive",
- "abi_stable_shared",
- "const_panic",
- "core_extensions",
- "crossbeam-channel",
- "generational-arena",
- "libloading",
- "lock_api",
- "parking_lot",
- "paste",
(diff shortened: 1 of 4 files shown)


coderfender (approved):
LGTM ! Pending CI

PR #1921, opened 2026-07-02
chore(deps): track DataFusion git main and port to its APIs

### LINKED ISSUE #1920: Track DataFusion git main via a pinned rev and bump it periodically

Ballista currently depends on published DataFusion crates.io releases (today, `54`). Because a full release cycle passes between DataFusion versions, breaking changes in DataFusion only surface in Ballista at upgrade time, all at once, making each upgrade a large and risky change (e.g. the DataFusion 54 upgrade in #1906).

Pin the `datafusion*` crates to a git revision of `apache/datafusion` `main` and bump that revision periodically. Tracking unreleased DataFusion continuously means breaking API and behavior changes are caught in small increments, close to when they land upstream, instead of accumulating until the next release.

Each bump would:
- advance the pinned `rev` in the workspace `Cargo.toml`,
- port any DataFusion API changes,
- refresh any snapshots/goldens affected by DataFusion planner/metric changes,
- keep `cargo test` and `cargo clippy --all-targets --workspace -- -D warnings` green.

Continue upgrading only at DataFusion release boundaries. This keeps the dependency on stable releases but concentrates all breakage into infrequent, large upgrades.

### Additional context

Related tracking issue: #1246. First bump implemented in the PR that references this issue.

andygrove:

Closes #1920.

# Rationale for this change

Ballista depends on published DataFusion crates.io releases (today, `54`). Because a full release cycle passes between DataFusion versions, breaking changes only surface in Ballista at upgrade time, all at once — making each upgrade large and risky.

This PR switches the `datafusion*` crates to a pinned git revision of `apache/datafusion` `main`. Tracking unreleased DataFusion continuously lets us catch breaking API and behavior changes in small increments, close to when they land upstream, rather than accumulating them until the next release. The `rev` is bumped periodically (see #1920).

Pin the `datafusion*` crates to a git `rev` of `apache/datafusion` `main` and bump `arrow`/`arrow-flight` to `59` to match.

Port to the DataFusion `main` APIs:

- `TableReference` moved from `datafusion::sql` to `datafusion::common`.
- `PhysicalExtensionCodec::try_decode`/`try_encode` gained a `PhysicalProtoConverterExtension` argument.
- `MetricValue` gained a `PeakMemoryUsage` variant; serialize it as a named gauge.
- `Partitioning` gained a `Range` variant; format it in the DOT graph.
- `EnforceSorting` and `EnforceDistribution` were unified into `EnsureRequirements`; use it for post-stage-split re-optimization.
- `ExecutionPlan::partition_statistics` is deprecated in favour of `statistics_with_args(&StatisticsArgs)`; migrate all call sites and operator overrides. This is a **correctness fix**: DataFusion operators that now override only `statistics_with_args` return unknown statistics through the deprecated method, which would silently disable Ballista's broadcast-threshold checks.
- `ListingOptions` dropped its `target_partitions`/`collect_stat` fields; build it via the builder (`target_partitions` now comes from the session config).
- `Distribution::HashPartitioned` is deprecated; use `KeyPartitioned`.

Disable dynamic filter pushdown in the AQE mock session config to match the Ballista production default, so join-input swaps during re-optimization stay legal. Refresh the EXPLAIN ANALYZE and AQE plan snapshots for DataFusion's dropped `peak_mem_used` metric and its new `RightSemi` join lowering.

Ballista now builds against a git revision of DataFusion `main` rather than a crates.io release. There are no changes to Ballista's own public API.


ballista/scheduler/src/planner.rs
```diff
@@ -1158,12 +1168,13 @@ order by
         assert_plan!(stages[0].as_ref(), @r"
         ShuffleWriterExec: partitioning: None
           AggregateExec: mode=Partial, gby=[], aggr=[count(Int64(1))]
-            ProjectionExec: expr=[]
-              SortMergeJoinExec: join_type=Inner, on=[(k@0, k@0)]
-                SortExec: expr=[k@0 ASC], preserve_partitioning=[false]
-                  DataSourceExec: partitions=1, partition_sizes=[1]
-                SortExec: expr=[k@0 ASC], preserve_partitioning=[false]
-                  DataSourceExec: partitions=1, partition_sizes=[1]
+            RepartitionExec: partitioning=RoundRobinBatch(2), input_partitions=1
```
andygrove, line 1522: Need to check this

ballista/scheduler/tests/tpch_plan_stability/approved/q17.txt
```diff
@@ -16,7 +16,7 @@ SortShuffleWriterExec: partitioning=Hash([l_partkey@0], 16)
 ShuffleWriterExec: partitioning: None
   AggregateExec: mode=Partial, gby=[], aggr=[sum(lineitem.l_extendedprice)]
     ProjectionExec: expr=[l_extendedprice@1 as l_extendedprice]
-      SortMergeJoinExec: join_type=Inner, on=[(p_partkey@2, l_partkey@1)], filter=CAST(l_quantity@0 AS Decimal128(30, 15)) < Float64(0.2) * avg(lineitem.l_quantity)@1
+      SortMergeJoinExec: join_type=LeftSemi, on=[(p_partkey@2, l_partkey@1)], filter=CAST(l_quantity@0 AS Decimal128(30, 15)) < Float64(0.2) * avg(lineitem.l_quantity)@1
```
andygrove, line 19: Need to check if this change in join type is valid. Affects other queries too.

phillipleblanc (approved):
I think the strategy of having the `main` branch track DataFusion's git `main` makes sense to me 👍

PR #1924, opened 2026-07-03
feat: Allow submitting a pre-built physical plan to the scheduler

andygrove:

No dedicated issue. This adds a capability the scheduler is missing: accepting a pre-built physical plan
for distributed execution.

# Rationale for this change

Today the scheduler only accepts a `LogicalPlan` (or a Substrait plan): a submitted plan is turned into a
physical plan on the scheduler via `SessionState::create_physical_plan` before `plan_query_stages` splits
it into stages. That is the right default for SQL/DataFrame clients, but it means a client that has
*already* built a physical `ExecutionPlan` has no way to submit it — most importantly when that plan
contains custom operators that have no logical-plan representation and therefore cannot be expressed as a
`LogicalPlan` or Substrait at all.

Accepting a pre-built physical plan is a generally useful capability: it lets any client that constructs an
`ExecutionPlan` directly — including one with registered custom/extension operators — hand it to Ballista
and get distributed shuffle execution, without a round trip back through logical planning.

One concrete motivating case is native execution of Apache DataFusion Comet plans on Ballista
(apache/datafusion-comet#4796): the driver builds a physical plan whose leaves are foreign (FFI-backed)
Comet fragments and lets Ballista own the shuffle between them. Those fragments have no `LogicalPlan` form,
so the existing submission paths cannot express them. This feature is what makes that possible — but it is
not specific to Comet; it applies to any custom-operator physical plan.

The distribution machinery already supports this: `DefaultDistributedPlanner::plan_query_stages` takes a
physical `Arc<dyn ExecutionPlan>` and splits it at `RepartitionExec(Hash)` into `ShuffleWriterExec` stages,
and custom operators already round-trip through the registered `PhysicalExtensionCodec`. Only the
submission entry point was missing.

- Add a `bytes physical_plan = 7;` variant to the `ExecuteQueryParams.query` oneof.
- Decode it in the scheduler gRPC handler using the session's `PhysicalExtensionCodec`
  (`PhysicalPlanNode::try_into_physical_plan`).
- Thread a small `SubmitPlan { Logical | Physical }` enum through the existing job-submission chain so that,
  for a physical submission, the scheduler **skips** `create_physical_plan` and feeds the decoded
  `Arc<dyn ExecutionPlan>` straight to `plan_query_stages`. The logical and Substrait paths are unchanged.
- Add a client helper (`execute_physical_plan`) to submit a pre-built physical plan and collect results.
- A test submits a `RepartitionExec(Hash)` plan over an in-memory scan through the new path against a
  standalone cluster and asserts the results match local execution (scheduler logs confirm two real shuffle
  stages ran).

Adaptive Query Execution is intentionally **not** supported for physical-plan submissions (returns
`NotImplemented`), because AQE's adaptive re-planning operates on the logical plan. Default AQE is off, so
this only affects clients that opt into both. Extending AQE to physical submissions can be a follow-up if
there is demand.

Yes, additive and opt-in: a new `physical_plan` query variant on `ExecuteQueryParams` and a client helper
to submit a pre-built physical plan. Existing logical-plan and Substrait submission paths are unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,6 +1012,7 @@ dependencies = [
  "tokio",
  "tonic",
  "url",
+ "uuid",
 ]
 
 [[package]]
--- a/ballista/client/Cargo.toml
+++ b/ballista/client/Cargo.toml
@@ -46,6 +46,7 @@ datafusion-proto = { workspace = true }
 env_logger = { workspace = true }
 rstest = { workspace = true }
 tempfile = { workspace = true }
+uuid = { workspace = true }
 tonic = { workspace = true }
 
 [features]
--- a/ballista/client/tests/physical_plan_submission.rs
+++ b/ballista/client/tests/physical_plan_submission.rs
@@ -1,0 +1,132 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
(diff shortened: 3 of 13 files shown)


avantgardnerio (approved):
I support this specifically because:

```
The distribution machinery already supports this
```

I think the XP concept of self-similarity applies here, making this an architecture extension, not violation. 

The logical, physical, and distributed plan layers are self-similar trees. Extending the submission entry point to accept the physical layer is a natural extension of the same abstraction.

andygrove:
Merged. Thanks @avantgardnerio!

PR #1925, opened 2026-07-03
feat: event log and history server (Spark History Server equivalent)

### LINKED ISSUE #1923: Implement equivalent of Spark History Server

**Additional context**

andygrove:

Closes #1923.

# Rationale for this change

Ballista has a live TUI that shows jobs, stages, tasks, executors, and metrics by
reading the scheduler's REST API — but that state is ephemeral. Completed jobs are
cleaned up after `finished_job_state_clean_up_interval_seconds`, and everything is
gone when the scheduler restarts. There is no way to inspect a job after the fact,
the way Spark's History Server lets you.

This PR adds a Spark-History-Server equivalent: the scheduler can record a durable
per-job event log during execution, and a standalone history server replays those
logs and serves the **same** `/api/*` responses the live scheduler does — so the
existing TUI browses completed jobs unchanged, with no scheduler running.

**New `ballista-history` crate** (a leaf crate; depends only on `serde`,
`serde_json`, `tokio`):
- A versioned, JSONL on-disk event schema (`HistoryEvent`): `JobStart`,
  `StageStart`/`StageEnd`, `TaskEnd` (the incremental timeline), and a terminal
  `JobEnd` that embeds the fully-built REST DTOs.
- The REST response DTO types (`JobResponse`, `QueryStagesResponse`, …), shared so
  the live scheduler and the history server serialize byte-identical JSON.
- An async, buffered `EventLogWriter` (appends never block the scheduler hot path)
  and a reader that folds a completed log into the DTO bundle the history server
  serves.

**Scheduler**:
- REST DTOs moved to the shared crate; the graph→DTO construction extracted into
  reusable `dto_build` builders (behavior-preserving — existing API output and
  tests are unchanged).
- A new `event_log_dir` config option (**off by default**). When set, an
  `EventLogWriter` is tee'd off the internal `QueryStageSchedulerEvent` bus in
  `QueryStageScheduler::on_receive`, mapping `JobSubmitted`→`JobStart`,
  `TaskUpdating`→`TaskEnd`, and `JobFinished`/`JobRunningFailed`→`JobEnd`. When
  disabled there is no channel, no task, no file, and no per-event work beyond one
  `Option` check.
- The terminal `JobEnd` is enqueued with a blocking send and flushed before the
  loop proceeds, so a completed job's record cannot be dropped under load; its file
  handle is then closed.

**New `ballista-history-server` binary**:
- `ballista-history-server --event-log-dir <dir> --bind-host <h> --bind-port <p>`
  loads completed logs and serves an axum router over the same `/api/*` paths from
  the stored DTOs. Corrupt/partial logs are skipped rather than failing startup;
  missing jobs return 404 like the live scheduler.

Scope for this first cut: local-filesystem storage and completed jobs only.
Incremental `TaskEnd` timeline events are captured but not yet surfaced in a UI.

Verification: the DTO builders are shared by both the live handler and the writer,
and an end-to-end test asserts the history server serves **byte-identical** JSON to
the live scheduler for the same job.

Yes, all additive and opt-in:
- New scheduler flag `--event-log-dir <dir>` (default: disabled).
- The existing TUI can point at a history server (`--host`/`--port`) to browse
  completed jobs with no live scheduler.

No breaking changes to public APIs; live REST responses are unchanged.

### DIFF
--- a/ballista/history/src/event.rs
+++ b/ballista/history/src/event.rs
@@ -1,0 +1,124 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! The on-disk event-log schema. One `HistoryEvent` is serialized per JSONL line.
+//! This is a frozen public projection of the scheduler's internal events; the
+//! embedded DTOs are the stable contract the history server serves.
+
+use crate::dto::{JobConfig, JobResponse, QueryStagesResponse, TaskStatus};
+use serde::{Deserialize, Serialize};
+
+/// Current on-disk schema version, stamped on `JobStart`/`JobEnd`.
+pub const SCHEMA_VERSION: u32 = 1;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum JobEndStatus {
+    Succeeded,
+    Failed(String),
+}
+
+/// Metrics captured per finished task on the incremental timeline.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TaskEndMetrics {
+    pub input_rows: u64,
+    pub output_rows: u64,
+    pub elapsed_compute_nanos: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "ev")]
+pub enum HistoryEvent {
+    JobStart {
+        version: u32,
+        job_id: String,
+        job_name: String,
+        queued_at: u64,
+        submitted_at: u64,
+        logical_plan: Option<String>,
+        physical_plan: Option<String>,
+    },
+    StageStart {
+        stage_id: usize,
+        partitions: usize,
+    },
+    StageEnd {
+        stage_id: usize,
+        status: String,
+    },
+    TaskEnd {
+        stage_id: u32,
+        partition: u32,
+        executor_id: String,
+        status: TaskStatus,
+        launch_time: u64,
+        start_exec_time: u64,
+        end_exec_time: u64,
+        metrics: TaskEndMetrics,
+    },
+    JobEnd {
+        version: u32,
+        status: JobEndStatus,
+        queued_at: u64,
+        started_at: u64,
+        completed_at: u64,
+        job: Box<JobResponse>,
+        stages: Box<QueryStagesResponse>,
+        config: JobConfig,
+        dot: String,
+    },
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::dto::{JobResponse, QueryStagesResponse};
+    use std::collections::BTreeMap;
+
+    #[test]
+    fn job_end_round_trips_through_jsonl() {
+        let job = JobResponse {
+            job_id: "job-1".into(),
+            job_name: "q1".into(),
+            job_status: "COMPLETED".into(),
+            status: "Successful".into(),
+            num_stages: 2,
+            completed_stages: 2,
+            percent_complete: 100,
+            start_time: 10,
+            end_time: 20,
+            logical_plan: Some("Projection".into()),
+            physical_plan: Some("ProjectionExec".into()),
+            stage_plan: Some("stage plan".into()),
+        };
+        let event = HistoryEvent::JobEnd {
+            version: SCHEMA_VERSION,
+            status: JobEndStatus::Succeeded,
+            queued_at: 5,
+            started_at: 10,
+            completed_at: 20,
+            job: Box::new(job),
+            stages: Box::new(QueryStagesResponse { stages: vec![] }),
+            config: BTreeMap::from([("k".to_string(), "v".to_string())]),
+            dot: "digraph {}".into(),
+        };
+        let line = serde_json::to_string(&event).unwrap();
+        assert!(line.contains("\"ev\":\"JobEnd\""));
+        let back: HistoryEvent = serde_json::from_str(&line).unwrap();
+        // Re-serialize and compare strings (stable, discriminating).
+        assert_eq!(line, serde_json::to_string(&back).unwrap());
+    }
+}
--- a/ballista/history/src/event.rs
+++ b/ballista/history/src/event.rs
@@ -25,10 +25,12 @@ use serde::{Deserialize, Serialize};
 /// Current on-disk schema version, stamped on `JobStart`/`JobEnd`.
 pub const SCHEMA_VERSION: u32 = 1;
 
+/// How a job reached its terminal state.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum JobEndStatus {
     Succeeded,
     Failed(String),
+    Cancelled,
 }
 
 /// Metrics captured per finished task on the incremental timeline.
@@ -39,6 +41,12 @@ pub struct TaskEndMetrics {
     pub elapsed_compute_nanos: u64,
 }
 
+/// A single record in a job's event log.
+///
+/// Stage and partition identifiers are fixed-width (`u32`) throughout rather
+/// than `usize`: this is a durable, cross-machine format, so a log written by a
+/// 64-bit scheduler must mean the same thing to any reader. Callers holding the
+/// scheduler's own `usize` stage ids cast on the way in.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "ev")]
 pub enum HistoryEvent {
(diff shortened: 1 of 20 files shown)


ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -82,6 +110,88 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
+                                ),
+                                *queued_at,
+                                *failed_at,
+                            ),
+                        )
+                        .await;
+                    }
+                    log.finish_job(job_id.as_str()).await;
+                }
+                _ => {}
```
martin-g, line 229: Shouldn't `JobCancel` and `JobPlanningFailed` be handled too ? IMO they should call `log.finish_job(job_id.as_str()).await;` too

ballista/scheduler/src/api/dto_build.rs
```diff
@@ -0,0 +1,654 @@
...
+    let job = graph.as_ref();
+    let (plain_status, job_status) = format_job_status(
+        &job.status().status,
+        job_elapsed_ms(job.start_time(), job.end_time()),
+    );
+
+    let num_stages = job.stage_count();
+    let completed_stages = job.completed_stages();
+    let percent_complete =
+        ((completed_stages as f32 / num_stages as f32) * 100_f32) as u8;
```
martin-g: `num_stages` could be 0

```suggestion
    let percent_complete = if num_stages == 0 {
        0
    } else {
        ((completed_stages as f32 / num_stages as f32) * 100_f32) as u8
    };
```

ballista/history/src/writer.rs
```diff
@@ -0,0 +1,303 @@
...
+            Ok(f) => {
+                handles.insert(job_id.to_string(), f);
+            }
+            Err(e) => {
+                eprintln!("event-log writer: cannot open {}: {e}", path.display());
+                return None;
+            }
+        }
+    }
+    handles.get_mut(job_id)
```
martin-g, line 190: Using https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry will be more idiomatic:

```suggestion
    match handles.entry(job_id.to_string()) {
        std::collections::hash_map::Entry::Occupied(entry) => Some(entry.into_mut()),
        std::collections::hash_map::Entry::Vacant(entry) => {
            let path = log_dir.join(format!("{job_id}.eventlog"));
            match tokio::fs::OpenOptions::new()
                .create(true)
                .append(true)
                .open(&path)
                .await
            {
                Ok(f) => Some(entry.insert(f)),
                Err(e) => {
                    eprintln!("event-log writer: cannot open {}: {e}", path.display());
                    None
                }
            }
        }
    }
```

ballista/scheduler/src/bin/history_server.rs
```diff
@@ -0,0 +1,94 @@
...
+    let log_filter = EnvFilter::new(rust_log.unwrap_or_else(|_| "info".to_string()));
+    tracing_subscriber::fmt()
+        .with_ansi(false)
+        .with_writer(std::io::stdout)
+        .with_env_filter(log_filter)
+        .init();
+
+    let args = Args::parse();
+
+    let store = Arc::new(HistoryStore::load(&args.event_log_dir)?);
```
martin-g, line 63: execute the synchronous `HistoryStore::load()` in `spawn_blocking()`:
```suggestion
    let event_log_dir = args.event_log_dir.clone();
    let store = Arc::new(
        tokio::task::spawn_blocking(move || HistoryStore::load(&event_log_dir))
            .await
            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))??,
    );
```
or make `load()` async and use `tokio::fs::**` APIs inside it.

ballista/history/src/event.rs
```diff
@@ -0,0 +1,124 @@
...
+    StageStart {
+        stage_id: usize,
+        partitions: usize,
+    },
+    StageEnd {
+        stage_id: usize,
+        status: String,
+    },
+    TaskEnd {
+        stage_id: u32,
```
martin-g, line 71: Why `u32` ?
Above it uses `usize` (lines 55 and 59)

ballista/scheduler/src/scheduler_server/event_log.rs
```diff
@@ -0,0 +1,335 @@
...
+) -> Vec<HistoryEvent> {
+    statuses
+        .iter()
+        .filter(|s| !matches!(s.status, Some(task_status::Status::Running(_))))
+        .map(|s| {
+            let status = s
+                .status
+                .as_ref()
+                .map(to_api_task_status)
+                .unwrap_or(ApiTaskStatus::Running);
```
martin-g: At line 72 you filter out the Running tasks but here if the status is None you map it to `Running`. The method doc also says that `Running` should be skipped

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
```diff
@@ -82,6 +110,88 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan>
...
+                }
+                QueryStageSchedulerEvent::JobFinished {
+                    job_id,
+                    queued_at,
+                    completed_at,
+                } => {
+                    if let Ok(Some(graph)) = self
+                        .state
+                        .task_manager
+                        .get_job_execution_graph(job_id)
```
martin-g: What if this returns Err ?
`job_end_event()` won't be called but `log.finish_job(job_id.as_str()).await;` is still executed.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
andygrove: Good catch, and it was worse than just a missing record. Cancellation is terminal for a job (the handler below removes the graph), so a cancelled job never got a `JobEnd`, which meant it was invisible to the history server and its file handle stayed open for the life of the process. Added a `JobCancel` arm that appends a terminal event and calls `finish_job`.

One wrinkle: the tee runs before the cancel is applied to the graph, so the graph still reports the job as running at that point. The new `job_cancel_event` stamps the cancelled status onto the DTO so the history view does not show a cancelled job as perpetually `Running`. Added `JobEndStatus::Cancelled` plus a test for it.

`JobPlanningFailed` I left alone on purpose, and added a comment saying why: it is only posted when `submit_job` fails, so it happens instead of `JobSubmitted`. The job has neither an execution graph nor an open log file at that point, so there is nothing to write or close.

ballista/scheduler/src/api/dto_build.rs
andygrove: Applied. It does not actually panic today since this is float division and `NaN as u8` saturates to 0 in Rust, but relying on that is not great, and `build_job_response_from_overview` right below already guards the same way. Now they match.

ballista/scheduler/src/bin/history_server.rs
andygrove: Agreed on the problem, though I went a slightly different way: the load now happens in `main()` before the tokio runtime is built at all, rather than in `spawn_blocking`. Nothing else is running at startup, so there is no reason for a runtime to exist yet, and it avoids having to map a `JoinError` into `BallistaError`.

ballista/history/src/event.rs
andygrove: The inconsistency was real. `TaskEnd` came straight off the protobuf `TaskStatus` (u32) while the stage events came off the graph (usize), and I never reconciled them.

Resolved it toward fixed width: `StageStart` and `StageEnd` are now `u32` too, with a doc comment on the enum stating the rule. This is a durable, cross-machine format, so a pointer-width type does not really belong in it. Callers holding the scheduler usize stage ids cast on the way in.

ballista/scheduler/src/scheduler_server/event_log.rs
andygrove: Right, that fallback contradicted both the filter above it and the doc comment. Rewrote it as a `filter_map` so a status-less update is skipped outright rather than being recorded as a `TaskEnd` that claims `Running`. Added a test for the `status: None` case.

ballista/scheduler/src/scheduler_server/query_stage_scheduler.rs
andygrove: The `finish_job` call is intentional there, and I have added a comment explaining it. It closes the per-job file handle, so a job whose `JobEnd` could not be built still releases its fd instead of leaking one. The resulting log has no `JobEnd` line, and the reader treats a log without one as incomplete and skips it, so the history server never serves a half-written job.

What was wrong is that the `Err` was swallowed silently. Pulled the lookup into an `event_log_graph` helper that logs both the `Err` case and the `Ok(None)` case at warn level, and reused it in all three arms. Event logging still never fails scheduling, it just costs the job its event.

ballista/history/src/writer.rs
andygrove: I would rather keep this one as is. `entry()` takes an owned key, so it allocates a `String` on every event, including the already-open case which is the overwhelming majority of calls. The `contains_key` plus `get_mut` shape is the usual workaround for the borrow checker rejecting the `get_mut`-then-insert form, and the extra lookup only happens on the first event for a job.

Happy to switch if you feel strongly, the allocation is small, it just seemed like paying it on every event to save one lookup per job.

andygrove:
Thanks for the reviews so far @martin-g. I'm going to break this work down into some smaller PRs to make the review process easier.

andygrove:
First PR split out of this one is https://github.com/apache/datafusion-ballista/pull/2256

andygrove:
More PRs split out from this one:

- #2258 
- #2260

andygrove:
Final PRs, which replace this one

- #2264 
- #2265

PR #1945, opened 2026-07-04
fix: hold pooled shuffle connection for the response stream lifetime

### LINKED ISSUE #1943: Shuffle fetch exhausts ephemeral ports at high target_partitions (client connection caching off by default)
## Describe the bug

```
```

## To Reproduce

  (i.e. `--client-ttl 0`).
  (`target_partitions`):

|---:|---|
| 16 / 32 / 64 | complete |

## Root cause

paths:

  ```rust
  ```

## Expected behavior

## Possible mitigations

  higher partition counts.

## Additional context

andygrove:

Closes #1943.

# Rationale for this change

At high `target_partitions`, distributed TPC-H queries intermittently failed with
`FetchFailed` during shuffle reads. Two symptoms, one root:

- with client connection caching **off** (default, `--client-ttl 0`):
  `EADDRNOTAVAIL` / "Cannot assign requested address" — ephemeral-port exhaustion;
- with caching **on** (`--client-ttl > 0`): an `h2` `BrokenPipe` — "error reading a
  body from connection", i.e. the connection broke **mid-stream while reading data
  batches**.

Instrumented `h2=debug` runs (2 executors × 8 cores, TPC-H SF100 from S3) showed the
shuffle client opening and tearing down **thousands of connections per query**
(client-initiated `GOAWAY`s; zero `RST_STREAM`, so not h2 rapid-reset), and the failure
being an intermittent **race**: the same `target_partitions=128` run failed on some
attempts and passed on others.

Root cause for the broken-pipe case: `fetch_partition_remote` returned the pooled
connection to the pool **as soon as `fetch_partition` returned the response stream**.
The `PooledClient` guard was a local that dropped at that point, returning (a clone of)
the connection to the idle pool while the response **body was still streaming**. That
marks the connection idle/reusable/evictable mid-fetch, racing connection
teardown/reuse against the in-flight body read. And because `execute_do_get`'s retry
only covers the request + first (schema) message — not the body phase — the break
surfaces as a hard failure rather than a retry.

Keep the `PooledClient` checked out for the **lifetime of the response stream**.
`fetch_partition_remote` now wraps the returned stream in a small
`PooledPartitionStream` that owns the `PooledClient`, so the connection is only
returned to the pool — or `discard()`ed, if the stream yields a transport error — once
the stream has been fully consumed. This prevents a connection from being reused,
evicted, or closed while its body is still being read.

Scoped, ~60-line change to `ballista/core/src/execution_plans/shuffle_reader.rs`; no
public API changes.

No API changes. Fixes intermittent shuffle `FetchFailed` at high `target_partitions`.

### DIFF
--- a/ballista/core/src/execution_plans/shuffle_reader.rs
+++ b/ballista/core/src/execution_plans/shuffle_reader.rs
@@ -16,7 +16,7 @@
 // under the License.
 
 use crate::client::BallistaClient;
-use crate::client_pool::BallistaClientPool;
+use crate::client_pool::{BallistaClientPool, PooledClient};
 use crate::error::BallistaError;
 use crate::execution_plans::sort_shuffle::{
     get_index_path, is_sort_shuffle_output, stream_sort_shuffle_partition,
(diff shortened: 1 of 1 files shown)


andygrove:
Superseded by #1946. Instrumenting the failure (h2=debug) showed the root cause is connection *churn* — ~3,500 client-initiated connection closes per query racing in-flight reads — not just a per-connection lifecycle issue. Holding the connection for the stream lifetime (this PR) reduced but didn't eliminate the intermittent broken pipe. #1946 fixes it at the source by reusing a small bounded pool of multiplexed connections per endpoint instead of one-per-fetch. Closing in favor of that.

PR #1946, opened 2026-07-05
fix: reuse a small pool of multiplexed connections per peer for shuffle fetches

### LINKED ISSUE #1943: Shuffle fetch exhausts ephemeral ports at high target_partitions (client connection caching off by default)

andygrove:

Closes #1943.

# Rationale for this change

At high `target_partitions`, distributed queries intermittently failed during shuffle
reads with `FetchFailed` — either `EADDRNOTAVAIL` ("Cannot assign requested address",
with client caching off) or an `h2` `BrokenPipe` ("error reading a body from
connection", with caching on).

Instrumenting the executors with `h2=debug` and reproducing on a 2-executor × 8-core
cluster (TPC-H SF100 from S3) showed the cause: the shuffle client **opened and tore
down thousands of connections per query** — ~3,500 client-initiated `GOAWAY` frames in a
single run, all graceful `NO_ERROR`, with zero `RST_STREAM` (so not h2 rapid-reset). The
failure is an intermittent race: the client's own connection **close racing an in-flight
body read** produces the broken pipe, and the same churn exhausts ephemeral ports when
caching is off.

The churn comes from the pool's model: `acquire` checked out an **exclusive connection
per fetch** and returned it on drop, so a shuffle-heavy stage cycled through thousands of
short-lived connections.

Replace the exclusive-connection-per-fetch pool with a small, bounded set of
**reused, multiplexed connections per endpoint**. `DefaultBallistaClientPool` now keeps up
to `CONNECTIONS_PER_ENDPOINT` (8) shared `BallistaClient`s per `(host, port, config)` and
`acquire` hands out cheap **clones round-robin**. Each underlying `tonic::Channel`
multiplexes concurrent requests over its connection and reconnects transparently, so:

- per-query connections collapse from thousands to a handful, removing the churn-driven
  broken-pipe / port-exhaustion races, and
- load is spread across enough connections that no single one stalls under high
  shuffle-fetch concurrency (a single shared connection deadlocks — hence a small pool
  rather than one).

Dropping a handed-out clone is a no-op (the shared clients stay cached); the background
task still evicts endpoints unused within `idle_timeout`. The change is contained to
`ballista/executor/src/client_pool.rs`; there are no public API changes.

No API changes. Fixes intermittent shuffle `FetchFailed` at high `target_partitions`.

### DIFF
--- a/ballista/executor/src/client_pool.rs
+++ b/ballista/executor/src/client_pool.rs
@@ -17,17 +17,15 @@
 
 //! Connection pool for `BallistaClient` instances.
 //!
-//! `DefaultBallistaClientPool` maintains a `VecDeque` of idle clients per
-//! `(host, port, config)` key backed by a `DashMap`. Callers `BallistaClientPool::acquire`
-//! a `PooledClient` guard; when the guard is dropped the underlying client is
-//! returned to the idle deque automatically.
+//! `DefaultBallistaClientPool` keeps a small set of shared `BallistaClient`s per
+//! `(host, port, config)` endpoint in a `DashMap`. Callers `BallistaClientPool::acquire`
+//! a `PooledClient` that wraps a cheap clone of one of those clients (chosen
+//! round-robin); the underlying `tonic::Channel` multiplexes concurrent requests
+//! and reconnects transparently, so dropping the clone is a no-op and no new
+//! connection is opened per request.
 //!
-//! Connections could be discarded calling `PooledClient::discard` which will result
-//! of dropping connection rather than returning it to the pool. This could be
-//! used for error handling.
-//!
-//! A optional background tokio task evicts idle connections that have not been used
-//! within the configured `idle_timeout`.
+//! A optional background tokio task evicts endpoint connections that have not been
+//! used within the configured `idle_timeout`.
 
 use ballista_core::client::BallistaClient;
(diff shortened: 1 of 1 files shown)


milenkovicm:
#1578 tried using multiplexed connections, the issue was they get stuck if there is a lot of data to be transferred across. Thee were a configuration option which could be tuned, initial message size or something like that and it would help but it was not trivial.

andygrove:
Closing — the small multiplexed pool deadlocks on heavier shuffle stages (h2 connection-level flow-control window starvation). See the full investigation and design tradeoffs in #1943.

andygrove:
> #1578 tried using multiplexed connections, the issue was they get stuck if there is a lot of data to be transferred across. Thee were a configuration option which could be tuned, initial message size or something like that and it would help but it was not trivial.

The next thing I'll try is to mimic Spark's design, which has throttling. https://github.com/apache/datafusion-ballista/issues/1943#issuecomment-4886371492

PR #1947, opened 2026-07-05
fix(cli): sort jobs by stage completion ratio

QuakeWang:

Closes #.

# Rationale for this change

The TUI jobs table sorted the stage completion column with integer division:

`completed_stages / num_stages`

This could panic when `num_stages` was zero, and it also collapsed partial ratios such as `1/4`, `2/4`, and `3/4` to the same value.

This PR updates the jobs table stage-completion sorting to compare the actual completion ratio safely, treating zero-stage jobs as `0%` complete. It also adds regression tests for ratio-based sorting and zero-stage jobs.

Yes. Sorting the TUI jobs table by stage completion now reflects the real completion ratio and no longer panics for jobs with zero stages.

### DIFF
--- a/ballista-cli/src/tui/domain/jobs.rs
+++ b/ballista-cli/src/tui/domain/jobs.rs
@@ -119,14 +119,12 @@ impl JobsData {
             SortColumn::StagesCompleted => jobs.sort_by(|a, b| {
                 let stage_completion = |job: &Job| {
                     if job.num_stages == 0 {
-                        (0_u128, 1_u128)
+                        0.0
                     } else {
-                        (job.completed_stages as u128, job.num_stages as u128)
+                        job.completed_stages as f64 / job.num_stages as f64
                     }
                 };
-                let (a_completed, a_total) = stage_completion(a);
-                let (b_completed, b_total) = stage_completion(b);
-                let cmp = (a_completed * b_total).cmp(&(b_completed * a_total));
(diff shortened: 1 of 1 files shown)


ballista-cli/src/tui/domain/jobs.rs
```diff
@@ -117,9 +117,16 @@ impl JobsData {
...
             }),
             SortColumn::StagesCompleted => jobs.sort_by(|a, b| {
-                let a_stages = a.completed_stages / a.num_stages;
-                let b_stages = b.completed_stages / b.num_stages;
-                let cmp = a_stages.cmp(&b_stages);
+                let stage_completion = |job: &Job| {
+                    if job.num_stages == 0 {
+                        (0_u128, 1_u128)
+                    } else {
+                        (job.completed_stages as u128, job.num_stages as u128)
```
coderfender: Nit : Perhaps 128bit int is an overkill here ?

coderfender (commented):
left comments

ballista-cli/src/tui/domain/jobs.rs
QuakeWang: @coderfender Good point. I switched this to compare f64 completion ratios with total_cmp while keeping the zero-stage guard. The exact u128 fraction comparison was more defensive than this TUI sorting path needs.

martin-g (approved):

martin-g:
Thank you, @QuakeWang !

PR #1948, opened 2026-07-05
feat(scheduler): call try_acquire_job on active cache miss in update_task_statuses

weastel:
No linked issue — this is preparatory work for our distributed scheduler HA. Current Problem is once scheduler is restarted active job cache does not contain previous job state this causes our concern in HA Scheduler. The try_acquire_job hook existed on JobState but was never called from TaskManager.

Closes #

 # Rationale for this change
When an executor sends task status updates, TaskManager::update_task_statuses only processes jobs present in the local active_job_cache. On a cache miss (e.g. after scheduler failover or when a different scheduler receives status for a job it did not originally curate), updates were dropped with an error log and a TODO for “curator changed” handling.

JobState::try_acquire_job is the intended mechanism for a scheduler to take over curation of a still-running job. Calling it on cache miss lets a failover scheduler rehydrate active_job_cache and apply pending task updates, which is the “acquire-on-update” path described in the HA design doc.

Add get_or_acquire_active_execution_graph on TaskManager: returns the cached graph if present; otherwise calls JobState::try_acquire_job and inserts the result into active_job_cache via DashMap::entry().or_insert() (safe under concurrent acquire attempts).
Update update_task_statuses to use this helper instead of failing immediately on cache miss.
Downgrade the cache-miss log from error! to warn! when acquisition is not possible.
No change to the in-memory JobState backend: try_acquire_job still returns None, so single-scheduler behavior is unchanged until a distributed backend implements acquisition.

No user-facing changes. This is internal scheduler behavior only and has no effect until a JobState implementation that supports try_acquire_job is added. No API, config, or documentation updates required for this PR.

### DIFF
--- a/ballista/scheduler/src/state/task_manager.rs
+++ b/ballista/scheduler/src/state/task_manager.rs
@@ -1004,7 +1004,12 @@ mod tests {
 
     #[async_trait::async_trait]
     impl JobState for MockJobState {
-        fn accept_job(&self, _job_id: &JobId, _job_name: &str, _queued_at: u64) -> Result<()> {
+        fn accept_job(
+            &self,
+            _job_id: &JobId,
+            _job_name: &str,
+            _queued_at: u64,
+        ) -> Result<()> {
             Ok(())
         }
 
@@ -1040,7 +1045,11 @@ mod tests {
             Ok(None)
         }
 
-        async fn save_job(&self, _job_id: &JobId, _graph: &ExecutionGraphBox) -> Result<()> {
+        async fn save_job(
+            &self,
+            _job_id: &JobId,
+            _graph: &ExecutionGraphBox,
+        ) -> Result<()> {
             Ok(())
         }
 
@@ -1061,10 +1070,7 @@ mod tests {
             job_id: &JobId,
         ) -> Result<Option<ExecutionGraphBox>> {
             self.try_acquire_calls.fetch_add(1, Ordering::SeqCst);
-            Ok(self
-                .acquirable
-                .remove(job_id)
-                .map(|(_, graph)| graph))
+            Ok(self.acquirable.remove(job_id).map(|(_, graph)| graph))
         }
 
         async fn job_state_events(&self) -> Result<JobStateEventStream> {
@@ -1178,7 +1184,8 @@ mod tests {
         let cached_graph = Box::new(test_aggregation_plan_with_job_id(2, &job_id).await)
             as ExecutionGraphBox;
         let acquirable_graph =
-            Box::new(test_aggregation_plan_with_job_id(4, &job_id).await) as ExecutionGraphBox;
+            Box::new(test_aggregation_plan_with_job_id(4, &job_id).await)
+                as ExecutionGraphBox;
 
         let mock_state = Arc::new(MockJobState::new());
         mock_state.insert_acquirable(acquirable_graph);
--- a/ballista/scheduler/src/state/task_manager.rs
+++ b/ballista/scheduler/src/state/task_manager.rs
@@ -963,3 +963,240 @@ impl From<&ExecutionGraphBox> for JobOverview {
         }
     }
 }
+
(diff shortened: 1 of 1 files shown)


milenkovicm:
thanks for your contribution @weastel will have look ASAP

weastel:
@milenkovicm Any updates on this? Please do let me know if their is any ideological difference

milenkovicm (commented):
sorry for late review @weastel i think this makes sense, would ti be possible to  add a test or two to cover new functionality?

andygrove (commented):
Thanks for picking this up. The direction looks right to me. `get_or_acquire_active_execution_graph` is a clean shape for the helper, the cache-hit path is untouched, and using `entry().or_insert()` is the right instinct for concurrent acquisition. Nice to see the tests added too. I checked them out locally and all four pass.

A few things I noticed.

**A single acquire failure now drops the whole status batch.** This is the one I care most about. `update_task_statuses` groups statuses by job and loops over them. The new `?` on `get_or_acquire_active_execution_graph(...).await?` means that if `try_acquire_job` returns an error, and in a real HA backend that is a remote call that can fail transiently, the whole loop aborts. The caller in `query_stage_scheduler.rs` catches the `Err`, logs it, and drops everything. So a transient state store hiccup while trying to acquire one unknown job would also discard status updates for jobs this scheduler does own and has cached. Those lost task completions would stall their stages. On top of that, jobs already processed earlier in the loop have had `update_task_status` applied to their graphs, but their events get thrown away with the `Err`, so the graph moves forward and nothing revives. Would it make sense to handle the acquire error per job instead, logging a warning and falling through to `vec![]` for that one job so the rest of the batch still lands?

**The miss path becomes a remote round trip on the task status path.** Cache misses are not just a failover thing. `succeed_job` and `cancel_job` remove the graph from `active_job_cache`, and straggler statuses for finished or cancelled jobs keep arriving after that. Every one of those will now attempt an acquisition. That is free with the in memory backend, but with a distributed backend it is a store call, and probably a lock or a compare and swap, on the executor status update path. Is it worth gating the attempt on something the scheduler already knows, or adding some negative caching, so late statuses for a job that just finished here do not repeatedly hit the store?

**Nothing stops a terminal job from being re-cached.** The `JobState::try_acquire_job` doc says it returns the graph only if the job is still running, so a well behaved backend gates this. But `TaskManager` inserts whatever it gets back. If a backend ever hands back a terminal graph, say in a race where this scheduler just completed the job, it lands in `active_job_cache` and nothing will remove it again. `running_job_number()` would be inflated and the entry would stay forever. A status check before the insert would make this safe regardless of what a backend does.

**One test may not be testing what it intends.** In `get_or_acquire_or_insert_preserves_existing_cache_entry` the job is inserted into `active_job_cache` before the call, so the function returns at the first `get_active_execution_graph` and never reaches the `or_insert`. It asserts `try_acquire_calls() == 0`, which is the same thing `get_or_acquire_returns_cached_graph_without_acquire` already asserts. The concurrent acquisition path that `or_insert` exists to protect is still uncovered. Exercising it would need two concurrent calls against a mock that yields inside `try_acquire_job`. The other three tests cover the real branches nicely.

**Small things.** `cargo fmt --all --check` currently fails on the new test module, `accept_job`, `save_job`, `fail_unscheduled_job` and the `try_acquire_job` body all want reformatting. The branch also conflicts with current `main` in `task_manager.rs` and needs a rebase, which should also get the full CI suite running on it.

I did not benchmark this. The cache hit path is unchanged and the miss path is a no-op on the default in memory backend, so there is no TPC-H exposure here.

andygrove:
Thanks for the contribution @weastel. I did a quick AI review (see previous comment) but will look at this more closely this week. I am getting back up to speed with some topics in Ballista, and want to review how we're doing testing for things like scheduler HA, and retrying failed tasks etc.

I would feel more confident merging PRs like this if I know that I can rely on tests to catch regressions.

PR #1962, opened 2026-07-07
fix: Adaptive query planning fails with remote object stores: "No suitable object store found"

### LINKED ISSUE #1950: Adaptive query planning fails with remote object stores: "No suitable object store found"
## Describe the bug

Enabling adaptive query planning (`ballista.planner.adaptive.enabled=true`) on a query
over **remote object-store data** (S3 in this case) fails the job with:

```
Internal error: No suitable object store found for s3://tpch/sf100/lineitem/.
See `RuntimeEnv::register_object_store`.
```

The **same query over the same data succeeds with adaptive planning disabled** — the
non-adaptive execution path reads the S3 data fine. So the object store is available to
normal execution but not to the adaptive-planning path.

## To Reproduce

Cluster: standalone scheduler + 2 executors (Ballista `main`), TPC-H SF100 stored in S3
(MinIO). The client registers the S3 object store on its `SessionContext` (as the
`benchmarks` `tpch` binary does via its S3 support), and the query runs against
`s3://tpch/sf100`.

Run the TPC-H benchmark with adaptive planning enabled:

```
tpch benchmark ballista \
  --host <scheduler> --port 50050 \
  --path s3://tpch/sf100 --format parquet --partitions 32 \
  -c ballista.planner.coalesce.enabled=true
```

The job fails immediately (0 queries complete):

```
[ERROR ballista_core::execution_plans::distributed_query] Job <id> failed:
DataFusion error: Internal error: No suitable object store found for
s3://tpch/sf100/lineitem/. See `RuntimeEnv::register_object_store`.
```

Removing the two `-c` overrides (adaptive disabled) → the query plans and executes
normally and reads the S3 data.

## Expected behavior

Adaptive query planning should work with remote object stores. Whatever `RuntimeEnv` the
adaptive re-planning runs against should have the same object store registrations as the
normal execution path (or the adaptive rules should not require resolving the object
store at plan time without it being registered).

## Additional context

- Both `ballista.planner.adaptive.enabled` and `ballista.planner.coalesce.enabled` were
  set; I have not isolated which rule triggers the object-store lookup. The AQE
  coalesce/size-estimation path seems the likely candidate (it needs scan statistics),
  but the master `adaptive.enabled` switch may be sufficient on its own — worth
  confirming.
- The failure is logged by the client-side `DistributedQueryExec`, but the message is a
  job failure reported back from the server side, so the missing registration appears to
  be on the scheduler/adaptive-planning path rather than the client (the client session
  has the store registered).
- The scheduler and executor pods have the S3 env vars (`AWS_ENDPOINT`, `AWS_REGION`,
  `AWS_ALLOW_HTTP`, credentials) available, but nothing in the scheduler/executor
  registers an object store from them by default — the store is only built where a
  session explicitly registers it.
- Environment: Ballista `main`, DataFusion 54, standalone deployment on Kubernetes,
  MinIO-backed S3.

This is likely a blocker for using adaptive planning with any non-local object store.

milenkovicm:

Closes #1950.

# Rationale for this change

### DIFF
--- a/ballista/scheduler/src/state/aqe/planner.rs
+++ b/ballista/scheduler/src/state/aqe/planner.rs
@@ -31,6 +31,7 @@ use datafusion::common;
 use datafusion::common::{HashMap, exec_err};
 use datafusion::error::DataFusionError;
 use datafusion::execution::context::SessionContext;
+use datafusion::execution::runtime_env::RuntimeEnv;
 use datafusion::execution::{SessionState, SessionStateBuilder};
 use datafusion::logical_expr::LogicalPlan;
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
@@ -87,6 +88,7 @@ impl AdaptivePlanner {
     /// # Arguments:
     ///
     /// * `session_config` - The session configuration for the job.
+    /// * `runtime_env` - runtime environment
     /// * `plan` - The physical execution plan for the job.
     /// * `job_name` - The name of the job.
     /// * `physical_optimizer_rules` - A list of physical optimizer rules to apply.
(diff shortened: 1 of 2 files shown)


andygrove:
Verified this fix on a 2-executor cluster (2×8) reading TPC-H SF100 Parquet from MinIO over `s3://`, with the adaptive planner enabled (`-c ballista.planner.adaptive.enabled=true`).

Before this change, AQE jobs against a remote object store failed at planning with `No suitable object store found`. With this branch (built from `bug_1950`), the same queries plan and execute successfully:

| Query (AQE + S3) | Result |
|---|---|
| Q1 | 6.35 s, 4 rows |
| Q5 | 41.8 s, 5 rows |
| Q9 | 46.8 s, 175 rows |

Confirmed AQE was actually engaged (not a static-planner fallback): the scheduler logged the `Adaptive Query Planning is EXPERIMENTAL` warning per job and the plans contain `AdaptiveDatafusionExec` nodes. **Zero** `No suitable object store found` errors across the run, and row counts are correct.

LGTM 👍

andygrove (approved):
Thanks @milenkovicm

PR #1963, opened 2026-07-07
feat: improve session state creation for AQE planner

milenkovicm:

Closes #.

 # Rationale for this change

this is follow up on #1962 which fixed issue with custom object store but missed to identify issue when custom user defined functions. Previously session state with default features was created, user defined functions would not be added as the consequence 

- changed session state creation in AQE 

No

### DIFF
--- a/ballista/scheduler/src/state/aqe/planner.rs
+++ b/ballista/scheduler/src/state/aqe/planner.rs
@@ -30,14 +30,14 @@ use ballista_core::serde::scheduler::PartitionLocation;
 use datafusion::common;
 use datafusion::common::{HashMap, exec_err};
 use datafusion::error::DataFusionError;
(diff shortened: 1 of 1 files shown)


ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -30,14 +30,14 @@ use ballista_core::serde::scheduler::PartitionLocation;
 use datafusion::common;
 use datafusion::common::{HashMap, exec_err};
 use datafusion::error::DataFusionError;
+#[cfg(test)]
+use datafusion::execution::config::SessionConfig;
```
andygrove, line 34: nit: shouldn't the import be moved into the test module?

andygrove (approved):
Thanks @milenkovicm

PR #1968, opened 2026-07-08
Add shuffle-read fetch metrics and surface per-operator metrics in plan display

### LINKED ISSUE #1958: Add shuffle-read (fetch) metrics to ShuffleReaderExec and surface per-operator metrics in executor logs
**Is your feature request related to a problem or challenge?**

When diagnosing query performance (e.g. the Ballista-vs-Comet SF100 TPC-H gap), the executor's `LoggingMetricsCollector` logs each completed stage-partition's plan with metrics — but only the **shuffle-writer** wrapper metrics are actually populated: `write_time`, `repart_time`, `spill_time` / `spill_count` / `spill_bytes`, `input_rows`, `output_rows`.

There is no visibility into the two costs that dominate a shuffle-heavy query:

1. **Shuffle-read / fetch cost.** `ShuffleReaderExec` carries only `BaselineMetrics` (`elapsed_compute` / `output_rows`). There are no fetch-specific metrics: time spent fetching remote partitions, bytes fetched, number of fetch requests, time blocked acquiring the reduce-side in-flight governor permits (in-flight-bytes / in-flight-requests / per-address, added in #1951), or retry counts. The reader's fetch path is exactly where the shuffle *transport* cost lives, and it is currently unmeasurable.

2. **Inner-operator compute.** `SortExec` / `SortMergeJoinExec` / `DataSourceExec` `elapsed_compute` (and spill) is not surfaced in the executor metrics display, so per-operator time cannot be attributed — only the shuffle-writer wrapper shows up.

**Concrete example.** SF100 TPC-H Q9 runs in ~82.7 s and shuffles ~963 M rows across ~12 stages (2 executors × 8 slots). The writer metrics account for only ~72 slot-seconds (~6 % of slot capacity). The remaining ~90 % — shuffle read/fetch and sort/join compute — is invisible in the logs, so it is impossible to tell whether the bottleneck is the shuffle transport, sorting ~963 M rows for the sort-merge joins, or something else.

1. Add fetch metrics to `ShuffleReaderExec` (`send_fetch_partitions` / the fetch path): fetch wall-time, bytes fetched, request count, time blocked acquiring governor permits (in-flight-bytes / in-flight-requests / per-address), and retry count.
2. Surface inner-operator `MetricsSet` (`elapsed_compute`, spill, output rows) in the executor `LoggingMetricsCollector` display, so each operator's contribution is visible — not just the shuffle-writer wrapper.

External profiling (perf, flamegraphs) can attribute CPU but not the fetch/backpressure wait time, which is application-level and only the reader knows. `EXPLAIN ANALYZE` is affected by the same missing metrics.

**Additional context**

Needed to observe the effect of the reduce-side in-flight governor (#1943 / #1951) — that work bounds the shuffle-read path, but there is currently no metric to measure how much time is spent there. Related to #1319 (EPIC: shuffle file execs improvement).

andygrove:

Closes #1958.

# Rationale for this change

Diagnosing shuffle-heavy queries (for example the SF100 TPC-H Ballista-vs-Comet gap) is currently blind. `ShuffleReaderExec` exposes only `BaselineMetrics` (`elapsed_compute`/`output_rows`), so the shuffle-read/transport cost — the time spent fetching remote partitions, the bytes moved, and the time blocked on the reduce-side in-flight governor (#1951) — is unmeasurable. Separately, the executor's `LoggingMetricsCollector` renders only the shuffle-writer wrapper metrics, so inner-operator compute (`SortExec`/`SortMergeJoinExec`/`DataSourceExec` `elapsed_compute`/spill) never appears even though DataFusion already collects it. Together these make it impossible to attribute where a shuffle-heavy query spends its time.

1. **Fetch metrics on `ShuffleReaderExec`.** A `ShuffleReadMetrics` struct (mirroring the existing `ShuffleWriteMetrics`) recorded per output partition and threaded through `send_fetch_partitions` into the concurrent local/remote fetch tasks:
   - `fetch_time` — wall-time fetching remote partitions (Arrow-Flight fetch + buffering)
   - `local_read_time` — wall-time opening node-local shuffle files
   - `permit_wait_time` — a single combined timer over the three reduce-side governor semaphores (request + per-address + byte), i.e. backpressure wait (#1951)
   - `decoded_bytes` — decoded in-memory Arrow footprint of fetched remote partitions (not compressed wire bytes)
   - `fetch_requests` — total remote fetch attempts issued, including retries
   - `fetch_retries` — of those, how many were retries
   - `local_partitions` / `remote_partitions` — the local/remote read split

   All metrics go through the standard `MetricBuilder`/`MetricsSet` machinery, so they also appear in `EXPLAIN ANALYZE`. The handles are `Arc`-backed atomics cloned into the concurrent fetch tasks (no added locking). Doc comments note that `fetch_time`/`permit_wait_time` are summed across concurrent tasks (aggregate cost, not wall-clock) and that `fetch_time` is additive to — not a re-slice of — the reader's existing `elapsed_compute`.

2. **Per-operator metrics in the plan display.** `ShuffleWriterExec::Display` now renders the child plan with metrics (`DisplayableExecutionPlan::with_metrics`, `ShowMetrics::Aggregated`) in addition to statistics, so inner-operator `elapsed_compute`/spill/output-rows surface in `LoggingMetricsCollector` output and `EXPLAIN ANALYZE`. (The originally-considered `set_show_metrics` setter does not exist on `DisplayableExecutionPlan`; `with_metrics` is the equivalent constructor and is already used elsewhere in the repo.)

No API changes. The additional metrics are visible in `EXPLAIN ANALYZE` output and executor metrics logs.


milenkovicm (approved):
useful, thanks @andygrove

PR #1978, opened 2026-07-09
fix: satisfy clippy 1.97 useless_borrows_in_formatting

andygrove:

Closes #.

# Rationale for this change

Stable `clippy` 1.97 (released 2026-07-07) tightened the `useless_borrows_in_formatting` lint to flag `&expr` arguments passed to formatting macros (`format!`, `println!`, `log::trace!`, `info!`, `debug!`, `error!`) where the reference is redundant. As a result, the `Clippy` CI job (`ci/scripts/rust_clippy.sh`, which runs with `-D warnings`) now fails on `main` and on every open PR, on pre-existing code untouched by those PRs.

Drop the redundant `&` at each flagged site (21 arguments across 8 files):

- `ballista/core/src/object_store.rs`, `ballista/client/src/extension.rs`
- `ballista/scheduler/src/scheduler_server/grpc.rs`, `query_stage_scheduler.rs`
- `ballista/executor/src/executor_process.rs`, `executor_server.rs`
- `ballista-cli/src/tui/ui/main/jobs/stage_tasks_popup.rs`
- `benchmarks/src/bin/tpch.rs`

`ci/scripts/rust_clippy.sh` and `cargo fmt --all -- --check` both pass cleanly after the change.

No. The values were already `Display`/`Debug`-formatted through the reference, so the formatted output is byte-for-byte identical. No behavior or API change.

### DIFF
--- a/ballista-cli/src/tui/ui/main/jobs/stage_tasks_popup.rs
+++ b/ballista-cli/src/tui/ui/main/jobs/stage_tasks_popup.rs
@@ -77,7 +77,7 @@ pub(crate) fn render_stage_tasks_popup(f: &mut Frame, app: &App) {
         Block::default()
             .title(format!(
                 " Tasks for stage '{}' of job '{}' ",
-                stage.id, &popup.job_id
+                stage.id, popup.job_id
             ))
             .borders(Borders::ALL)
             .border_style(app.theme.popup_border_alt)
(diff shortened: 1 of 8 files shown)


milenkovicm (approved):
Thanks @andygrove

Issue #1980, opened 2026-07-09
Make shuffle-file compression codec configurable (add ZSTD option)


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Ballista compresses shuffle files with a hard-coded `LZ4_FRAME` codec. There is no way to change the codec, and no way to disable compression. The codec is fixed in two places:

- `ballista/core/src/execution_plans/shuffle_writer.rs` (the `ShuffleWriterExec` partition-file writer)
- `ballista/core/src/utils.rs` (`write_stream_to_disk`)

Both do:

```rust
let options = IpcWriteOptions::default()
    .try_with_compression(Some(CompressionType::LZ4_FRAME))?;
```

LZ4 favors throughput over ratio. On disk-constrained environments this matters — for example, the `TPC-H SF10` CI job intermittently fails with `No space left on device (os error 28)` while spilling/writing shuffle files on GitHub runners. A denser codec such as ZSTD would trade some CPU for a smaller shuffle footprint, which would help both CI and real deployments that are I/O- or storage-bound rather than CPU-bound. Different workloads want different points on that trade-off, so it should be configurable rather than one hard-coded choice.

**Describe the solution you'd like**

Make the shuffle-write compression codec configurable via `BallistaConfig` (`ballista/core/src/config.rs`), plumbed through to both write sites above. Concretely:

- A config key such as `ballista.shuffle.compression.codec` accepting `none` | `lz4` | `zstd`, defaulting to `lz4` to preserve current behavior.
- Apply the resulting `Option<CompressionType>` to `IpcWriteOptions::try_with_compression` in `shuffle_writer.rs` and `utils.rs` (a single shared helper would keep the two paths in sync).

**Describe alternatives you've considered**

- Keeping LZ4 hard-coded and just bumping CI runner disk / lowering scale factor — treats the symptom, not the general need for storage/CPU trade-off control.
- A configurable compression *level* as well: worth noting Arrow's `IpcWriteOptions::try_with_compression` only accepts a `CompressionType` (codec), not a level — ZSTD level is fixed inside `arrow-ipc`. So level tuning can't be exposed at the IPC layer today without upstream Arrow support; this issue therefore focuses on codec selection (incl. disabling). Level configurability can be a follow-up gated on Arrow exposing it.

**Additional context**

[...]

PR #1982, opened 2026-07-09
feat: delete intermediate shuffle files on job success

### LINKED ISSUE #1981: Flaky CI: TPC-H SF10 job intermittently fails with "No space left on device"
**Describe the bug**

The `TPC-H SF10 (all queries, AQE off + on)` CI job (`.github/workflows/tpch.yml`) intermittently fails with `No space left on device (os error 28)` while writing/spilling shuffle files. It is non-deterministic — the same commit and even `main` pass on some runs and fail on others — so it is a flaky, infra-level failure rather than a code regression. It shows up as red X's on unrelated PRs and erodes trust in CI signal.

The failure surfaces two ways depending on where the disk fills:

```
Task 27 failed: DataFusion error: Arrow error: Io error:
No space left on device (os error 28)
```
```
called `Result::unwrap()` on an `Err` value:
External(Os { code: 28, kind: StorageFull, message: "No space left on device" })
```

**To Reproduce**

Not deterministically reproducible (that's the problem), but recent examples on 2026-07-09:

- `main` @ `fd30f37` (PR #1973 merge commit) — SF10 failed, `StorageFull`.
- PR #1954 branch `reuse-exchange` @ `ebfba6a` — SF10 failed, `No space left on device`; the *same branch* passed SF10 on several earlier commits the same day.

Job configuration for context:
- `runs-on: ubuntu-latest` (standard GitHub-hosted runner, ~14 GB free disk) inside a container.
- Generates TPC-H **scale factor 10** data into `$RUNNER_TEMP/tpch-data`, then runs a scheduler + executor cluster and executes all 22 queries with AQE off and on.
- Shuffle files are written (LZ4-compressed) to the executor work dir on the same volume, on top of the SF10 source data and build artifacts.

So the disk holds: build artifacts + SF10 source parquet + shuffle spill for all 22 queries × 2 AQE modes — and occasionally that exceeds the runner's free space.

**Expected behavior**

SF10 CI should be a reliable signal — either it always has enough headroom to complete, or a genuine failure is distinguishable from "the runner ran out of disk."

**Additional context**

Filing this so solutions can be discussed rather than proposing one. Some options that have come up / are worth weighing:

- **Free up runner disk** before the job (e.g. remove preinstalled toolchains/SDKs, `docker image prune`) — cheap, common GitHub-Actions mitigation, buys several GB.
- **Reduce on-disk footprint** — clean up per-query shuffle output between queries, and/or point shuffle work dir and TPC-H data at different volumes if available.
- **Denser shuffle compression** — the shuffle codec is currently hard-coded to LZ4; a ZSTD option would shrink the shuffle footprint. Tracked separately in #1980.
- **Larger runner** — a runner/volume with more disk (if the project has access to larger hosted or self-hosted runners).
- **Lower the scale factor or split the matrix** — smaller SF, or splitting AQE-off / AQE-on into separate jobs so peak concurrent disk use is lower.
- **Surface disk usage** — a `df -h` step before/after to quantify actual headroom and confirm which artifact dominates.

Related: #1980 (configurable/denser shuffle compression codec).

andygrove:

Closes #1981.

Related to #1980 (denser shuffle compression codec is a complementary, separate lever for the same disk pressure).

# Rationale for this change

When a query job succeeds, Ballista keeps the entire job directory on the executors — intermediate inter-stage shuffle data *and* the final-stage output — and only deletes it after a delay of `finished_job_data_clean_up_interval_seconds` (default 300s). That delay exists for a good reason: on success the scheduler returns the final stage's output partition locations to the client, and in the default (pull) model the client fetches those results from the executor's job directory *after* the job is marked successful, so the final output must survive for the client to read it.

The cost of that delay is disk. The intermediate shuffle data — which is the bulk of a job's on-disk footprint — is dead the moment the job succeeds (every stage's output has already been consumed by its downstream stage), yet it lingers for the full interval. In back-to-back runs (for example the TPC-H SF10 CI job, which runs all 22 queries with AQE off and on), intermediate shuffle from many just-completed queries accumulates at once and intermittently exhausts the runner's disk (`No space left on device`, os error 28).

This change reclaims the intermediate shuffle data immediately on job success while retaining the final-stage output for the existing delayed cleanup, so the client can still fetch results.

- **Protobuf:** a new `repeated uint32 remove_stage_ids` field on `RemoveJobDataParams` (push path) and `CleanJobDataParams` (pull/poll path). Empty means "remove the whole job directory" — identical to today's behavior, so the change is backward compatible on the wire.
- **Executor:** the whole-job `remove_job_dir` becomes stage-aware `remove_job_data(work_dir, job_id, remove_stage_ids)` — empty removes the whole job dir; non-empty removes only the given `work_dir/{job_id}/{stage_id}` subdirectories (missing ones skipped), keeping the existing `is_subdirectory` safety guard per path.
- **Scheduler:** the cleanup plumbing carries `remove_stage_ids` through both the push (direct RPC) and pull (`pending_cleanup_jobs` → poll response) paths. On job success the scheduler computes the intermediate stage set (`all stages − final stages`, where a final stage has empty `output_links`) and fires an immediate reclaim via a new `clean_up_intermediate_job_data`, in addition to the existing delayed whole-job cleanup that now just removes the retained final-stage output.

Safety invariant: the immediate reclaim only ever deletes stages in `all − final`; a final-stage output directory can never be reached by it, and any unresolvable/failed job path degrades to an empty set (no reclaim, existing delayed cleanup only). Failed-job cleanup is unchanged.

Behavioral, not API. By default, intermediate shuffle files for a completed query are removed from executors immediately on job success rather than after `finished_job_data_clean_up_interval_seconds`. Final query results are unaffected — they are retained for the existing delayed-cleanup window exactly as before. No public API or configuration changes; no new configuration is required.

### DIFF
--- a/ballista/scheduler/src/state/execution_graph.rs
+++ b/ballista/scheduler/src/state/execution_graph.rs
@@ -240,6 +240,16 @@ pub trait ExecutionGraph: Debug {
     /// Exposes executions stages and stage id's
     fn stages(&self) -> &HashMap<usize, ExecutionStage>;
 
+    /// Stage ids of all non-final (intermediate) stages — those whose
+    /// `output_links` is non-empty. The final stage(s) are excluded.
+    fn intermediate_stage_ids(&self) -> Vec<u32> {
+        self.stages()
+            .iter()
+            .filter(|(_, stage)| !stage.output_links().is_empty())
+            .map(|(stage_id, _)| *stage_id as u32)
+            .collect()
+    }
+
     /// returns next task to run
     /// (used for testing only)
     #[cfg(test)]
@@ -1785,6 +1795,33 @@ mod test {
         test_coalesce_plan, test_join_plan, test_two_aggregations_plan,
         test_union_all_plan, test_union_plan,
     };
+
+    #[tokio::test]
+    async fn test_intermediate_stage_ids() {
+        // A simple aggregation produces a 2-stage graph: one intermediate
+        // stage (non-empty output_links) feeding one final stage (empty
+        // output_links).
+        let graph = test_aggregation_plan(4).await;
+
+        assert_eq!(graph.stages().len(), 2);
+
+        // Exactly one final stage.
+        let final_count = graph
+            .stages()
+            .values()
+            .filter(|s| s.output_links().is_empty())
+            .count();
+        assert_eq!(final_count, 1);
+
+        // Intermediate = all - final = exactly one stage, and none of the
+        // returned ids is a final stage.
+        let intermediate = graph.intermediate_stage_ids();
+        assert_eq!(intermediate.len(), 1);
+        for id in &intermediate {
+            let stage = graph.stages().get(&(*id as usize)).unwrap();
+            assert!(!stage.output_links().is_empty());
+        }
+    }
 
     #[tokio::test]
     async fn test_fail_job_sets_end_time_and_failed_metadata() -> Result<()> {
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -783,6 +783,9 @@ message CancelJobResult {
 
 message CleanJobDataParams {
   string job_id = 1;
+  // Specific stage ids to remove within the job dir.
+  // Empty means remove the whole job dir (legacy behavior).
+  repeated uint32 remove_stage_ids = 2;
 }
 
 message CleanJobDataResult {
@@ -820,6 +823,9 @@ message CancelTasksResult {
 
 message RemoveJobDataParams {
   string job_id = 1;
+  // Specific stage ids to remove within the job dir.
+  // Empty means remove the whole job dir (legacy behavior).
+  repeated uint32 remove_stage_ids = 2;
 }
 
 message RemoveJobDataResult {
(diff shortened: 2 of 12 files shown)


ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -240,6 +240,16 @@ pub trait ExecutionGraph: Debug {
     /// Exposes executions stages and stage id's
     fn stages(&self) -> &HashMap<usize, ExecutionStage>;
 
+    /// Stage ids of all non-final (intermediate) stages — those whose
+    /// `output_links` is non-empty. The final stage(s) are excluded.
+    fn intermediate_stage_ids(&self) -> Vec<u32> {
```
milenkovicm, line 245: this might not work with AQE as output links are not populated, have to check

milenkovicm (commented):
thanks @andygrove, this makes a lot of sense, just one comment, can we check this works with AQE as output links might not be populated by AQE

ballista/scheduler/src/state/execution_graph.rs
andygrove: You are correct. The current PR does no harm under AQE (and we run TPC-H in CI with AQE enabled). I filed a follow-on issue to implement this for AQE.
https://github.com/apache/datafusion-ballista/issues/1996

milenkovicm (approved):

andygrove:
Merged. Thanks @milenkovicm

PR #1983, opened 2026-07-09
perf: avoid redundant TaskStatus clones in scheduler update_task_status

andygrove:

None — small, self-contained hot-path allocation cleanup.

# Rationale for this change

`ExecutionGraph::update_task_status` runs on the scheduler's hot path — it is
invoked for every batch of task status updates from executors. Its running-stage
loop performed several redundant clones of the `TaskStatus` protobuf message per
task:

- `task_status.clone().partition_id` deep-cloned the entire `TaskStatus`
  (including its `metrics` vector and nested `status`) only to copy out a single
  `u32`, then dropped the clone.
- `task_status.metrics.clone()` cloned the metrics vector unconditionally, even
  on the failure path where it is unused.
- `RunningStage::update_task_info(partition_id, task_status.clone())` passed a
  full clone even though the method only reads a few `Copy` fields and needs to
  clone the inner status enum for storage in `TaskInfo`.

This applies to both `StaticExecutionGraph` and the AQE `AdaptiveExecutionGraph`,
which carry duplicate copies of this loop.

Two commits:

1. Read `partition_id` directly instead of via `task_status.clone().partition_id`.
2. Change `RunningStage::update_task_info` to borrow `&TaskStatus` instead of
   taking it by value, and destructure the owned `TaskStatus` once at the call
   site to move its `status` and `metrics` into the match arms. This removes the
   per-update metrics clone and the full-message clone on both hot paths, leaving
   only the single inner-status clone that `update_task_info` must persist.

No behavior change; the existing `update_task_info` and task-status-update tests
in `execution_graph.rs` / `execution_stage.rs` continue to pass.

No.

### DIFF
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -656,7 +656,7 @@ impl ExecutionGraph for AdaptiveExecutionGraph {
                             );
                             continue;
                         }
-                        let partition_id = task_status.clone().partition_id as usize;
+                        let partition_id = task_status.partition_id as usize;
                         let task_identity = format!(
                             "TID {} {}/{}.{}/{}",
                             task_status.task_id,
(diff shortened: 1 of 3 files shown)


milenkovicm (approved):
Makes sense, thanks @andygrove

PR #1991, opened 2026-07-10
chore: add script and CI check to sync vendored DataFusion protos

andygrove:

N/A — no tracking issue. Happy to file one if preferred.

# Rationale for this change

`ballista/core/proto/datafusion.proto` and `datafusion_common.proto` are vendored copies of DataFusion's protobuf definitions. They are **not** compiled into Rust here: `ballista/core/build.rs` maps their packages to the real `datafusion-proto` / `datafusion-proto-common` crates via `extern_path`, so the vendored files exist only so that `protoc` can resolve `ballista.proto`'s `import` statements at build time. Because of that they must stay compatible with the DataFusion version Ballista depends on.

Today they are re-synced by hand whenever DataFusion is upgraded (and occasionally in standalone catch-up commits). That drifts silently: nothing fails the build when the vendored stubs lag the pinned crate, because a stub only needs the definitions `ballista.proto` actually references. The recent DataFusion 54 upgrade is a live example — the crates were bumped to 54 but these two files were never re-synced.

- `dev/update_datafusion_proto.py` — copies the two `.proto` files out of the exact crate source resolved by `cargo metadata --locked` and rewrites the one import path that differs from our flat layout (`datafusion/proto-common/proto/datafusion_common.proto` → `datafusion_common.proto`). A `--check` mode prints a diff and exits non-zero on drift.
- `ci/scripts/rust_proto_check.sh` + a new `datafusion-proto-sync-check` job in `rust.yml` that runs `--check` on every PR.
- `dev/update_datafusion_versions.py` now prints a reminder to re-run the proto sync after a DataFusion bump.
- Re-sync of the two vendored proto files to the pinned `datafusion-proto` / `datafusion-proto-common` 54.0.0 crates (the drift the new check caught).

Note: `datafusion-proto-common` did not ship its `.proto` file until v54, so on older pins that file cannot be sourced from the crate. The script skips any proto a release does not ship (rather than failing), so the check stays green across future up/downgrades.

No.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -251,6 +251,21 @@ jobs:
               exit 1
           fi
 
+  datafusion-proto-sync-check:
+    name: Check vendored DataFusion proto is in sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7.0.0
+        with:
+          fetch-depth: 1
+      - name: Setup Rust toolchain
+        run: |
+          rustup toolchain install stable
+          rustup default stable
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
+      - name: Check vendored DataFusion proto matches pinned crate
+        run: ci/scripts/rust_proto_check.sh
+
 # Coverage job was failing. https://github.com/apache/arrow-datafusion/issues/590 tracks re-instating it
 
 # coverage:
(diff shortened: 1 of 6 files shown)


andygrove:
The first commit added he CI check, which failed as expected.

The second commit is the result of running the script, and now CI passes.

milenkovicm (approved):

milenkovicm:
Thanks @andygrove

andygrove:
Merged. Thanks @milenkovicm

Issue #1829, opened 2026-06-07
Add support for `DataFrame.cache()` to Ballista


JarroVGIT:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Single node DatafFusion creates a `MemTable` when you call [`DataFrame.cache()`](https://docs.rs/datafusion/53.1.0/src/datafusion/dataframe/mod.rs.html#2368). It essentially executes the plan, and stores the resulting partitions as a `MemTable` in the SessionState, returning a `DataFrame` that will just read from that `MemTable`.

This is currently not supported in Ballista, and supporting it presents some challenges due to the distributed nature of the execution. For example, where would the cache "live" and how would executioners get their hands on it during execution? Another problem (I think, could be wrong!) is that `CacheFactory::create()` is a synchronous function.

Ballista already implements some of the extension points to handle a call to `.cache()`. It has an implementation of `CacheProvider` (`BallistaCacheFactory` in the core crate), which either returns the `LogicalPlan` unchanged or wraps it in a `BallistaCacheNode`. The behaviour is steered by a configuration option (`ballista.cache.noop` which defaults to true), and if you were to enable this option today, the execution fails at physical planning because no `ExtensionPlanner` knows how to turn it into an `ExecutionPlan`. 

The basic wiring is in place, now we need to figure out what "caching" means in an executed context and what is stored where. I am doing some reading and research on what approaches would be possible and will post here if I find anything worth discussing.

**Describe alternatives you've considered**
I think we should be able to re-use the Shuffle mechanics to store partitions on executioners and leverage the same mechanics to get the data to the right executioner. The previous sentence is based fully on my (limited) mental model of Ballista, so if it doesn't make sense I will soon find out I guess.

**Additional context**
(copied from Datafusion example docs)
There is an example in Datafusion (`datafusion-examples`) which demonstrates how to leverage `CacheFactory` to implement custom caching strategies for dataframes in DataFusion. By default, `DataFrame::cache` in Datafusion is eager and creates an in-memory table. This example shows a basic alternative implementation for lazy caching. Specifically, it implements:
- A `CustomCacheFactory` that creates a logical node `CacheNode` representing the cache operation.
- A `CacheNodePlanner` (an `ExtensionPlanner`) that understands `CacheNode` and performs caching.
- A `CacheNodeQueryPlanner` that installs `CacheNodePlanner`.
[...]

PR #1993, opened 2026-07-11
feat: add extension to support `DataFrame::checkpoint()`

### LINKED ISSUE #1937: Add suport for `DataFrame.checkpoint()`
Add support for checkpointing the DataFrame - so essentially writing the intermediate representation of a plan to disk and breaking lineage.

Introducing a `DataFrame.checkpoint()` method

**Additional context**
Related to #1829, but split into two parts.

sandugood:

Closes #1937 .

 # Rationale for this change
Before this change, there was no `checkpoint()` method for the DataFrame in Ballista (where we break lineage of the plan and store intermediate results to specified location)

Introducing new `DataFrameExt` trait in the client, also introducing both `checkpoint()` and `checkpoint_lazy()` there. 

A new entry to the `BallistaConfig` was added as well. When trying to call `checkpoint()` or` checkpoint_lazy()` without the config entry set user will get an error.

Also added new logic to a separate file (`checkpoint.rs`) on the `Scheduler`'s side, which is being used during planning in the `submit_plan()`

New methods introduced.

`checkpoint() `method is eager. It breaks lineage and doesn't wait for an action (i.e: `show()`) => so it executes immediately
`checkpoint_lazy()` method is lazy. It breaks lineage but does wait for an action


sandugood:
Is that OK that we'll hold the `DataFrameExt` in the `client/src/extension.rs` or should we move it to a separate file?

ballista/client/src/extension.rs
```diff
@@ -234,3 +237,46 @@ impl Extension {
...
+
+        let path = format!(
+            "{}/{}/{}",
+            base_dir.trim_end_matches('/'),
+            ctx.state().session_id(),
+            Uuid::new_v4()
+        );
+
+        // Executes the original plan as a normal distributed job.
+        ctx.execute_logical_plan(plan)
```
milenkovicm, line 287: this is one way to implement this, but difference from spark that ballista checkpoint will materialize the dag (up to that point) instead of waiting standard actions, such as write. 

this will work for first version, but maybe we should consider adding new logical plan extension which would represent checkpoint, wait for write op and then split plan on scheduler side using planner. 


wdyt @sandugood

milenkovicm (commented):
sorry for very late review @sandugood I'm catching up with backlog after summer break. 
the implementation does make sense, but it diverges from spark implementation as ballista checkpoint will materialise dag. perhaps if it is not too complicated we should consider implementing same behaviour like spark. wdyt?

ballista/client/src/extension.rs
sandugood: Seems like a reasonable thing to do
Should we open a separate PR for this though?

milenkovicm: Let's keep this one, if possible

sandugood: I chose the path of introducing `BallistaCheckpointNode` (alongside the `BallistaCacheNode`), and also adding lazy materialization (to mirror Spark's functionality)

sandugood:
I think, same as `.cache()` this needs a physical operator (something like `CheckpointExec`), an extension planner, a physical codec entry
Right now it seems OK with logical planning. wdyt @milenkovicm

milenkovicm:
I believe you could get away without physical node if you split into two jobs at planning

sandugood:
> I believe you could get away without physical node if you split into two jobs at planning

Tried to introduce this concept and it worked out nicely
Added unit tests and an example as well. thanks @milenkovicm 

For me it seemed to introduce new `CheckpointMaterializer` trait as a reasonable thing

sandugood:
Also, I think it is good to have an example with remote-checkpoint, so user could see the output of it in the ballista-tui

image

During this planning I have:
1. Read .`parquet` and applied a simple filter. Then called `.checkpoint_lazy()`, which split the plan into two jobs
2. Then reused the same `DF`, which didn't read from the original `.parquet`, but rather from the checkpointed part


Hence having 3 separated jobs in the TUI

milenkovicm:
thanks @sandugood will have a look

ballista/client/tests/checkpoint.rs
```diff
@@ -0,0 +1,164 @@
...
+    let df = ctx.sql("select * from test").await?;
+
+    let err = df.checkpoint().await.unwrap_err();
+    assert!(err.to_string().contains("ballista.checkpoint.dir"));
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn should_insert_checkpoint_node_without_executing() -> datafusion::error::Result<()>
```
milenkovicm, line 101: `should_insert_checkpoint_lazy` as name?

ballista/client/tests/checkpoint.rs
```diff
@@ -0,0 +1,164 @@
...
+        Default::default(),
+    )
+    .await?;
+
+    let df = ctx
+        .sql("select string_col, timestamp_col from test where id > 4")
+        .await?;
+    let schema_before = df.schema().clone();
+
+    let checkpointed = df.checkpoint_lazy()?;
```
milenkovicm, line 121: should we have test to check execution of `checkpoint_lazy` or ive missed it

ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -331,6 +334,19 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T
...
         subscriber: Option<JobStatusSubscriber>,
     ) -> Result<JobId> {
+        // Lazy checkpoints become their own jobs; what reaches the planner is
+        // always checkpoint free. Physical plans cannot carry checkpoint nodes,
+        // since those are resolved before physical planning.
+        let resolved;
+        let plan = match plan {
+            SubmitPlan::Logical(logical) if contains_checkpoint(logical) => {
+                resolved =
+                    SubmitPlan::Logical(resolve_checkpoints(&ctx, logical, self).await?);
```
milenkovicm, line 344: if im not mistaken, this will block job submission (of reading part) until checkpoint creation finishes ? can we create two jobs and link them using job notification `subscriber`

milenkovicm (commented):
thanks @sandugood, my appologise for slow rewview. i have few comments, please have a look when you get chance.

I'm still not sure if we should provide `checkpoint_lazy` or `checkopint(lazy)` i guess second approach would give possibility to users to change from ine to other type withut method change, but lets think about it

ballista/scheduler/src/checkpoint.rs
```diff
@@ -0,0 +1,595 @@
...
+use futures::StreamExt;
+use futures::future::BoxFuture;
+use std::sync::Arc;
+
+/// Runs a plan to completion as its own job.
+///
+/// Implemented by `SchedulerServer`; the indirection keeps this module free of
+/// the event loop and lets tests substitute a local runner.
+#[async_trait]
+pub(crate) trait CheckpointMaterializer: Send + Sync {
```
akshaychitneni, line 48: I think .cache() (#1829) needs this same materialize-then-substitute approach. I think a generic name here like `Materializer`, and materialize() returning the output's partition locations (not just Result<()>) so cache can register/pin where the output landed, would let it reuse this instead of re-deriving. I can propose the generalization as a follow-up so it doesn't grow this PR's scope. cc @sandugood @milenkovicm

sandugood:
I will try and fix all of the comments asap. Then we can proceed with `.cache()`

PR #1995, opened 2026-07-11
feat: Share read-side runtime state across a session's tasks on the executor

### LINKED ISSUE #1994: Executor rebuilds a fresh RuntimeEnv per task instead of sharing runtime-scoped state across tasks

On the executor, every task builds a brand-new `RuntimeEnv` from scratch. The default runtime producer (`ballista/executor/src/executor_process.rs`, `run_executor`) is:

```rust
Arc::new(move |_| {
    let runtime_env = RuntimeEnvBuilder::new()
        .with_temp_file_path(wd.clone())
        .build()?;
    Ok(Arc::new(runtime_env))
})
```

This closure is invoked once per task (`get_task_definition` / `get_task_definition_vec` in `ballista/core/src/serde/scheduler/from_proto.rs`, and `execution_loop.rs`), so each `TaskContext` gets a fresh `DiskManager`, `CacheManager`, `ObjectStoreRegistry`, and memory pool. Nothing runtime-scoped is shared between concurrent tasks on the same executor — including tasks of the same job/stage.

The optional `--memory-pool-size` wrapper (`wrap_runtime_producer_with_memory_pool`, #1624) does not change this: it still calls the inner producer (fresh env) per task and layers a **fresh per-task `FairSpillPool`** of `total_bytes / concurrent_tasks` on top.

Consequences:

1. **No cross-task memory arbitration (the main one).** Memory is statically partitioned per task slot. A task never sees the executor's real aggregate memory pressure and cannot borrow budget from idle sibling slots. Under partition skew — the case Ballista's scalability work most cares about — a hot task that could stay in-memory with more headroom is instead capped at one slot's slice and forced to spill, while neighboring slots sit idle. The default (no `--memory-pool-size`) is an unbounded pool per task, i.e. no accounting at all.
2. **No object-store client reuse.** A fresh `ObjectStoreRegistry` per task means object-store clients (connection pools, credentials, TLS sessions) are rebuilt per task. Negligible for local Parquet, a real per-task tax against S3/GCS at scale.
3. **No file metadata / statistics / listing cache reuse.** A fresh `CacheManager` per task means Parquet footers, file listings, and statistics are re-read by every task and every stage, even when many tasks on the same executor scan the same table.

Allow runtime-scoped state to be shared executor-wide instead of rebuilt per task, at least for the pieces where sharing is safe, and evaluate sharing the memory pool:

- **Shared `ObjectStoreRegistry` and `CacheManager`**: build these once at executor startup and reuse them across tasks (preserve them through the runtime producer rather than reconstructing). Low risk; these are designed to be shared.
- **Executor-wide memory pool (evaluate)**: replace the per-task `FairSpillPool` slice with a single executor-wide pool shared by all concurrent tasks, so a hot task can expand into slack left by idle slots. This is a real trade-off — a shared pool gives better utilization under skew but weaker isolation (one greedy task can starve/OOM others), whereas static per-task slices give predictability. Worth measuring both on the SF10 harness with a skewed query, under AQE on and AQE off.

An escape hatch already exists: `ExecutorProcessConfig::override_runtime_producer` lets an embedder inject a producer closing over shared singletons, so (1)/(2)/(3) are achievable today without core changes — but the default does not do it, and the memory-pool wrapper deliberately makes a fresh per-task pool.

- Keep the per-task isolation model and instead make the static memory slice smarter (e.g. weight by stage). Simpler, but still can't reclaim idle slots' budget for a hot task.
- Share only the read-side caches (`ObjectStoreRegistry`, `CacheManager`) and leave the memory pool per-task. This captures the low-risk wins (2)/(3) without the isolation trade-off of a shared pool.

## Additional context

Nothing runtime-scoped currently persists across tasks; the executor's `RuntimeEnv` is fully task-local. Related, but distinct: #1624 (per-task `--memory-pool-size`), #1578 (cache ballista clients on executor — gRPC clients, not object store), #1836 (coordinator OOM at SF100), #1952 (spill large shuffle fetches).

andygrove:

Closes #1994.

# Rationale for this change

On the executor, every task builds a brand-new `RuntimeEnv` from scratch. The default runtime producer runs once per task, so each `TaskContext` gets a fresh `DiskManager`, `CacheManager`, `ObjectStoreRegistry`, and memory pool. Nothing runtime-scoped is shared between tasks on an executor — including tasks of the same session running the same tables.

This rebuilds read-side state per task and per query:

- **Object-store clients (connection pools, credentials, TLS) are reconstructed on every task.** This is the same class of problem Comet hit on EKS: creating a fresh object store per file read produced ~5,000 DNS queries/sec/pod (~500× vanilla Spark), exceeding the Route 53 Resolver per-ENI limit of 1,024 qps and causing intermittent connection failures. Comet fixed it with a process-wide object-store cache, which collapsed DNS volume back to Spark levels ([writeup](https://datafusion.apache.org/blog/2026/06/03/comet-eks/#excessive-dns-queries), apache/datafusion-comet#3802). Ballista rebuilds the object store even more aggressively — once per task rather than once per file — so it is exposed to the same failure mode against remote object stores at scale.
- **Parquet footers are re-parsed on every scan.** A benchmark session that runs all 22 TPC-H queries re-parses the same `lineitem`/`orders` footers on every query, because each task gets its own empty `CacheManager`.

This PR shares read-side runtime state across a session's tasks — and across queries within a session — without changing the per-task memory model.

A bounded, session-keyed cache of shared *base* `RuntimeEnv`s on the executor:

- **`SessionRuntimeCache` (new `ballista/executor/src/runtime_cache.rs`)** — an LRU (default 16 sessions) of base `RuntimeEnv`s keyed by `session_id`. A base env carries the shared read-side state: the `ObjectStoreRegistry`, the `CacheManager` (whose Parquet footer / file-metadata cache is thereby reused across the session), and the `DiskManager`. The base env is built outside the lock on a miss, so a miss never stalls other sessions' lookups.
- **Per-task memory pool is unchanged.** Each task's real runtime is derived from the shared base via `RuntimeEnvBuilder::from_runtime_env(base).with_memory_pool(fresh_pool)`. `from_runtime_env` reuses the base's inner file-metadata cache and object-store registry while installing a fresh per-task memory pool, so memory isolation is byte-for-byte identical to before. The existing `--memory-pool-size` `FairSpillPool` behavior is preserved (the memory-pool logic is refactored into a `MemoryPoolPolicy` applied on top of the shared base).
- **`Executor::produce_runtime_for_session(session_id, config)`** — used by both execution-time call sites (`execution_loop.rs` poll path, `executor_server.rs` push path). The decode-time throwaway env in `from_proto::get_task_definition` (which does no I/O) is left unchanged.
- **Escape hatch preserved.** Session caching applies only to the default runtime producer. When `override_runtime_producer` is set, the cache is left unset and tasks are served per-task exactly as before.
- **New config knob** `session_runtime_cache_capacity` (`ExecutorProcessConfig` + `--session-runtime-cache-capacity`, default 16). `0` disables caching and restores the previous fresh-per-task behavior.

Eviction is bounded and correctness-neutral: sessions are held in an LRU with a hard cap, and a wrongly-evicted-then-reused session simply rebuilds its base env. There is no reliable session-end signal at the executor (`RemoveJobData` is job-scoped, and a session spans many jobs), so the LRU cap — not a lifecycle hook — is what bounds memory.

Covered by unit tests in the executor crate: same-session sharing of the footer cache + registry through the per-task pool rebuild, cross-session isolation, LRU eviction, capacity-0 passthrough, the override fallback, and preservation of the `FairSpillPool` size.

One new, optional executor config: `--session-runtime-cache-capacity` (`ExecutorProcessConfig::session_runtime_cache_capacity`), default `16`; `0` restores the previous behavior of building a fresh runtime per task. No breaking API changes; no change to the per-task memory model or to query results.


ballista/executor/src/runtime_cache.rs
```diff
@@ -0,0 +1,194 @@
...
+/// `DiskManager` too; this is safe because spill temp files are uniquely
+/// named, matching the standard one-`RuntimeEnv`-per-`SessionContext` model.
+/// Each task's real runtime is produced by applying [`MemoryPoolPolicy`] to
+/// the shared base, which installs a fresh per-task memory pool — so memory
+/// isolation is unchanged.
+///
+/// The cache is bounded by an LRU of `capacity` sessions. A capacity of `0`
+/// disables caching entirely: every call builds a fresh base env, matching the
+/// behavior of building a runtime per task.
+pub struct SessionRuntimeCache {
```
milenkovicm: would it make sense to make `SessionRuntimeCache` a trait and provide `DefaultSessionRuntimeCache`? would users ever want to change this implementation ?

milenkovicm (approved):
thanks @andygrove 
I’m keen to see the performance difference with cache enabled.

andygrove:
> thanks @andygrove I’m keen to see the performance difference with cache enabled.

I will try a run with minio in my local cluster, but do not expect to see a difference in this env.

andygrove:
> > thanks @andygrove I’m keen to see the performance difference with cache enabled.
> 
> I will try a run with minio in my local cluster, but do not expect to see a difference in this env.

No difference, as I suspected.

ballista/executor/src/runtime_cache.rs
andygrove: Good idea. I made this change.

PR #1999, opened 2026-07-11
Log task execution duration in the executor "Finished task" message

### LINKED ISSUE #1998: Log task execution duration in the executor "Finished task" message

The executor logs `Execute task : [...]` when a task starts and `Finished task : [...]` when it ends, but neither line reports how long the task took. To find a slow task today you have to correlate the two log lines by timestamp, or inspect the task-status reports sent back to the scheduler. Per-task duration is the single most useful number when eyeballing executor logs for skew or long-pole stages, and it is not there.

Include the elapsed execution time in the `Finished task` log message, e.g.:

```
Finished task : [job/stage/partition] in 2.31s
```

This applies to both task-dispatch paths — the `executor_server` push path and the `execution_loop` pull path.

## Additional context

Small, self-contained logging improvement; no behavior or API change.

andygrove:

Closes #1998.

# Rationale for this change

The executor logs `Execute task : [...]` at the start of a task and `Finished task : [...]` at the end, but neither reports the task's duration. To spot a slow or skewed task from the logs today you have to correlate the two lines by timestamp or inspect the task-status reports. Per-task duration is the most useful number when scanning executor logs for long-pole stages, so this adds it directly to the finished-task line.

Capture an `Instant` immediately before `execute_query_stage` and format its `elapsed()` into the `Finished task` log message, at both task-dispatch sites:

- `ballista/executor/src/executor_server.rs` — the push-based path
- `ballista/executor/src/execution_loop.rs` — the pull-based path

The message becomes `Finished task : [<task>] in <duration>` (e.g. `in 2.31s`). A monotonic `Instant` is used rather than differencing the existing wall-clock timestamps so the duration is unaffected by clock adjustments.

Only the executor log line changes; the `Finished task` message now includes the elapsed execution time. No API or behavior changes.

### DIFF
--- a/ballista/executor/src/execution_loop.rs
+++ b/ballista/executor/src/execution_loop.rs
@@ -44,7 +44,7 @@ use std::any::Any;
 use std::convert::TryInto;
 use std::error::Error;
 use std::sync::mpsc::{Receiver, Sender, TryRecvError};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Instant, SystemTime, UNIX_EPOCH};
 use std::{sync::Arc, time::Duration};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tonic::codegen::{Body, Bytes, StdError};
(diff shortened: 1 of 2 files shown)


milenkovicm (approved):
makes sense thanks @andygrove

PR #2014, opened 2026-07-12
fix: render TUI banner with widely supported Block Elements glyphs

### LINKED ISSUE #2013: TUI banner renders as garbled glyphs on most fonts
## Describe the bug

The `ballista-cli` TUI banner renders as garbled characters / tofu boxes rather than the "DataFusion Ballista" wordmark. Reported on macOS, but it will reproduce almost anywhere.

Copying the garbled region out of the terminal yields characters like:

```
🬂██🬂🬎🬹🬭  🬭🬭🬭🬭    🬭🬹█🬭🬭   🬭🬭🬭🬭
```

Those are U+1FB02, U+1FB0E, U+1FB39, U+1FB2D — **sextants**, from Unicode's *Symbols for Legacy Computing* block. Most monospace fonts do not include that block, so they render as tofu.

## Root cause

`render_banner` in `ballista-cli/src/tui/ui/header/mod.rs` selects the `tui-big-text` `PixelSize` by width, and every branch it can pick draws from Symbols for Legacy Computing:

```rust
let banner_size = match area.width {
    0..70  => PixelSize::Octant,        // U+1CD00 octants (Unicode 16)
    70..80 => PixelSize::QuarterHeight, // also U+1CD00 octants
    _      => PixelSize::ThirdHeight,   // U+1FB00 sextants
};
```

`tui-big-text` documents exactly this hazard on all three variants:

> *Note: depending on how the used terminal renders characters, the generated text with this PixelSize might look very strange.*

The library's other pixel sizes (`Full`, `HalfHeight`, `HalfWidth`, `Quadrant`) draw only from **Block Elements** (U+2580..U+259F), which is universally supported — but none of them are reachable in the current code.

The reason the code ended up on the exotic sizes is the header height. `ui::render` gives the header `Constraint::Length(6)` and the banner is two lines, so each line gets 3 rows. Per `PixelSize::pixels_per_cell`, every size that fits two lines into 6 rows is a legacy-computing one:

| PixelSize | cols/char | rows/char | glyph block |
|---|---|---|---|
| Full | 8 | 8 | Block Elements |
| HalfHeight | 8 | 4 | Block Elements |
| HalfWidth | 4 | 8 | Block Elements |
| Quadrant | 4 | 4 | Block Elements |
| ThirdHeight | 8 | 3 | Sextants |
| Sextant | 4 | 3 | Sextants |
| QuarterHeight | 8 | 2 | Octants |
| Octant | 4 | 2 | Octants |

So the 6-row header is the real root cause; the glyph choice is downstream of it.

Note that picking a different font is not a viable workaround: the `Octant`/`QuarterHeight` branches need U+1CD00, added in Unicode 16 (Sept 2024), which essentially no shipping font carries.

## To Reproduce

Run `ballista-cli` with the TUI on any terminal whose font lacks Symbols for Legacy Computing (the common case) and look at the header.

## Expected behavior

The banner renders legibly using glyphs available in ordinary monospace fonts.

andygrove:

Closes #2013.

# Rationale for this change

The TUI banner renders as garbled tofu boxes for most users. `render_banner` picked its `tui-big-text` `PixelSize` by width, and all three reachable branches draw from Unicode's *Symbols for Legacy Computing* block, which most monospace fonts do not carry:

```rust
let banner_size = match area.width {
    0..70  => PixelSize::Octant,        // U+1CD00 octants (Unicode 16)
    70..80 => PixelSize::QuarterHeight, // also U+1CD00 octants
    _      => PixelSize::ThirdHeight,   // U+1FB00 sextants
};
```

`tui-big-text` warns about exactly this on all three variants ("depending on how the used terminal renders characters, the generated text with this PixelSize might look very strange"). Its remaining sizes — `Full`, `HalfHeight`, `HalfWidth`, `Quadrant` — use only Block Elements (U+2580..U+259F), which is universally supported, but none were reachable.

Switching fonts is not a workaround: `Octant`/`QuarterHeight` need U+1CD00, added in Unicode 16 (Sept 2024), which essentially no shipping font has.

The underlying cause is the header height. `ui::render` gave the header `Constraint::Length(6)` for a two-line banner, so each line got 3 rows — and per `PixelSize::pixels_per_cell`, every size that fits two lines into 6 rows is a legacy-computing one. The exotic glyphs were a consequence of the cramped header, so fixing the glyphs means giving the banner the vertical room it needs.

- Header grows from `Constraint::Length(6)` to `Length(8)`, costing 2 rows from the main view.
- Banner uses `PixelSize::Quadrant` (4 rows per line, Block Elements only). The width-based `match` is dropped: `Quadrant` is 4 columns per character, identical to the `Octant` it replaces, so narrow-terminal behaviour is unchanged.
- Two regression tests rendering through a `TestBackend`:
  - `banner_uses_only_widely_supported_block_element_glyphs` asserts every emitted glyph falls in U+2580..U+259F.
  - `banner_natural_height_matches_the_header_height` renders into an over-tall buffer and pins the banner's natural height to the header's, so a future `PixelSize` change can neither silently clip the banner nor leave the header underfilled.

The banner now renders as:

```
▜▛▙      ▟      ▜▛▀▌         ▀
▐▌▐▌▝▀▙ ▝█▀ ▝▀▙ ▐▙▌ █ █ ▟▀▀ ▝█  ▟▀▙ █▀▙
▐▌▟▘▟▀█  █▗ ▟▀█ ▐▌▘ █ █ ▝▀▙  █  █ █ █ █
▀▀▘ ▝▀▝▘ ▝▘ ▝▀▝▘▀▀  ▝▀▝▘▀▀▘ ▝▀▘ ▝▀▘ ▀ ▀
▜▛▜▖    ▝█  ▝█   ▀       ▟
▐▙▟▘▝▀▙  █   █  ▝█  ▟▀▀ ▝█▀ ▝▀▙
▐▌▐▌▟▀█  █   █   █  ▝▀▙  █▗ ▟▀█
▀▀▀ ▝▀▝▘▝▀▘ ▝▀▘ ▝▀▘ ▀▀▘  ▝▘ ▝▀▝▘
```

Both new tests were confirmed to fail on the code they guard against, not merely to pass on the fix: reverting to `ThirdHeight` fails the glyph test (reporting the offending sextants) and the height test (6 rows vs 8), while an over-tall `PixelSize::Full` passes the glyph test but is correctly caught by the height test (15 rows vs 8).

`cargo test -p ballista-cli` passes (246 tests), and clippy is clean on both feature combos CI runs for this crate (`cli,tui` and `web`).

Yes, visually. The banner is legible instead of garbled, and the TUI header is 2 rows taller, so the main view is 2 rows shorter. No public API changes.


andygrove:
@martin-g @milenkovicm I have no idea what any of this means, but it works for me now.

Screenshot 2026-07-12 at 1 50 22 PM

milenkovicm:
image

works for me in ghostty

martin-g:
> I have no idea what any of this means

Any of what ? :-) The explanation in the PR description ?!

I will test it soon but the idea of using different pixel sizes was to support resizing of the terminal application (Ghostty, Tilix, iTerm2, ...). Using a specific PixelSize (like Quadrant) will work well if the terminal application is "big enough", but if the user resizes it to a smaller size then the banner will overlap with the navigation buttons in the top-right area.

There is another problem with tui-big-text in the web rendered version too. There we use an `` with an absolute CSS position over the generated `<canvas>` HTML element for the logo. Maybe `PixelSize::Quadrant` fixes it for the web rendering too.

So, yes - the banner/logo is problematic...

andygrove:
> Any of what ? :-) The explanation in the PR description ?!

I don't have any experience with the topic covered by this PR. I just asked Claude to make it work for me.

I'll move this to draft since it likely doesn't meet the requirements of your original implementation.

I wonder if we should consider just not having the banner?

martin-g:
https://github.com/user-attachments/assets/ede82d89-b115-4ba0-89fb-989af858287b

I've pushed a small change that uses either one line for the banner text or two lines, depending on the available horizontal space.

martin-g (approved):

martin-g:
@andygrove With https://github.com/apache/datafusion-ballista/pull/2014/commits/62d4e0c87ac2ff2338ebcfb16834341dc1b8e4d9 I've changed the PixelSize from `Quadrant` to `Sextant` and reduced the vertical space for the header from `Length(8)` to `6`. This fixes the (visual) issue with the vertical alignment of the text for the navigation buttons ("Jobs", "Executors" and "Metrics"). 
In the video in my previous comment you can see that the buttons' height was too big and the text label was top aligned. Now the label is centered again.

Does the banner still render fine for you ?

martin-g:
The [comment](https://github.com/apache/datafusion-ballista/pull/2014/changes#diff-a6267de14153501859c940bc29f2e953a4b0b2dec9ef1526d3f7bcd7a71e1d4eR105) in the new tests says that `Sextant` is not supported :-/ 
```
Computing (sextants at U+1FB00.., octants at U+1CD00..), which most fonts lack and 
render as tofu.
```

martin-g:
> I wonder if we should consider just not having the banner?

I will drop the banner/logo

martin-g:
The banner/logo is removed in https://github.com/apache/datafusion-ballista/pull/2057

PR #2015, opened 2026-07-12
docs: remove outdated TPC-H benchmark results from READMEs

andygrove:

N/A

# Rationale for this change

The top-level `README.md` has a "Performance" section showing Ballista vs Apache Spark TPC-H comparison charts. Those numbers and images are very outdated and no longer reflect the current state of the project, so they are misleading to anyone landing on the repo page. Rather than leave stale claims in the most prominent place a new user looks, remove them.

The same four charts were also embedded in `ballista/client/README.md` (which is the crates.io README for the `ballista` crate) by linking to the images in this repo, so they are removed there too. Otherwise deleting the image files would leave that page rendering four broken images.

- Remove the "Performance" section from `README.md`
- Remove the "Performance" section from `ballista/client/README.md`
- Delete the four now-unreferenced chart images from `docs/source/_static/images/`:
  - `tpch_allqueries.png`
  - `tpch_queries_compare.png`
  - `tpch_queries_speedup_abs.png`
  - `tpch_queries_speedup_rel.png`

No other file in the repo references these images.

Documentation only. No API or behavior changes.


andygrove:
@milenkovicm I think it would be better to wait until benchmarks run without error at 1TB scale before publishing benchmarks. Hopefully we can get there in the next couple of weeks.

milenkovicm (approved):
I agree @andygrove

PR #2016, opened 2026-07-13
bug: Prevent exec hang when task decode fails

### LINKED ISSUE #1908: Scheduler marks executor dead on a deterministic task launch/decode failure, hanging the job
## Describe the bug

When an executor fails to **launch/decode** a task (a non-transient, deterministic error such as a plan deserialization failure), the scheduler treats it the same as losing the executor: it removes the executor, rolls back its stages, and — with a single executor — then reports "There are no alive executors to bind tasks" and the job hangs forever instead of failing.

A task that cannot be decoded will fail the same way on every executor, so this is a query-level error, not an executor-health problem. Marking the executor dead both hides the real error and (with few executors) wedges the whole job.

## Observed sequence

Submitting TPC-H q15 (an uncorrelated scalar subquery) on DataFusion 54 with 1 scheduler + 1 executor:

```
ERROR ballista_scheduler::state: Failed to launch new task: Internal Ballista error:
  Failed to connect to executor <id>: Status { code: InvalidArgument, message:
  "DataFusion error: Internal error: ScalarSubqueryExpr can only be deserialized as part
   of a surrounding ScalarSubqueryExec. ..." }
INFO  ballista_scheduler::state::executor_manager: Removing executor <id>: Some("Failed to launch new task: ...")
WARN  ballista_scheduler::state::execution_graph: Reset N tasks ... on lost Executor <id>
WARN  ballista_scheduler::state::executor_manager: There are no alive executors to bind tasks
DEBUG ballista_scheduler::state: No schedulable tasks found to be launched
```

The client then blocks indefinitely (both client and scheduler idle at ~0% CPU). The underlying `InvalidArgument` error is never surfaced to the client.

(The specific decode error above is a separate DataFusion-54 issue — filed separately — but it is a good example: any deterministic per-task decode/launch failure produces this hang.)

## Expected behavior

A deterministic task launch/decode failure (e.g. gRPC `InvalidArgument`, or an error that recurs across attempts) should **fail the query** and propagate the error to the client, rather than marking the executor dead. Executor removal should be reserved for genuine connectivity/liveness failures (transport errors, timeouts, heartbeat loss).

## Suggested direction

- Distinguish "executor unreachable / lost" from "executor rejected this task" at the task-launch site (`ballista/scheduler/src/state/` task launching). `InvalidArgument` (and similar deterministic statuses) should mark the **task/stage/job** failed, not the executor.
- Respect the task retry limit for launch failures so a deterministically-undecodable task fails the job after its attempts instead of looping or wedging.

## Impact

With a single executor this turns any undecodable task into an indefinite hang with no error reported. In CI this presents as a timeout rather than a clear failure, and one bad query can take down the whole suite.

coderfender:

Closes https://github.com/apache/datafusion-ballista/issues/1908

 # Rationale for this change

Previously, all errors were masked in a blanket error but we now handle task serialization / gRPC errors in a separate path and send an event to fail the corresponding job without marking the executor as dead essentially failing the query. This should help with cluster stability and determinism with regards to task serialization errors


andygrove:
Thanks @coderfender. The changes look reasonable, but is it possible to add tests to prevent regressions?

coderfender:
Added tests

ballista/scheduler/src/state/mod.rs
```diff
@@ -258,7 +274,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
     async fn launch_tasks(
```
phillipleblanc, line 282: `launch_tasks` can combine stages from multiple jobs into one executor RPC, and then would now treat any `InvalidArgument` as a failure for every job in that RPC. The executor will process each job sequentially and then fail the first one with invalid argument:

`ballista/executor/src/executor_server.rs`:L850
```rust
   for multi_task in multi_tasks {
       let multi_task: Vec<TaskDefinition> = get_task_definition_vec(
           multi_task,
           // ...
       )
       .map_err(|e| Status::invalid_argument(format!("{e}")))?;

       for task in multi_task {
           task_sender
               .send(CuratorTaskDefinition {
                   scheduler_id: scheduler_id.clone(),
                   task,
               })
               .await
               .unwrap();
       }
   }
```

So this could cause tasks that were processed before the failing one to have already been enqueued, while skipping any tasks after the bad one. Then this code would mark every job as failed.

I think a better approach would be to isolate the failed jobs in `launch_multi_task` and return it as a separate list and only mark those specific ones as failed.

phillipleblanc: You could test this by having a test that sends a launch_tasks call with multiple jobs, where only one is invalid and verify that only that is marked as failed.

coderfender: Thank you for the comment @phillipleblanc . Let me try and isolate the test assertion to a single job in the unit test.

coderfender:
@phillipleblanc  , thank you for the feedback. Please take a look whenever you can. One of the things I have observed while implementing review feedback and failing only invalid job(s) instead of all the jobs part of a batch RPC is to report the cores back t be reallocated as well. Now, the current logic assumes one core per task which should be true for most if not all cases . My plan is to create a followup on tackling the core count issue as a follow and keep this PR isolated to job scheduling aspects only

phillipleblanc (approved):
Thanks, this looks good to me now.

ballista/core/proto/ballista.proto
```diff
@@ -980,6 +980,7 @@ message LaunchTaskResult {
 
 message LaunchMultiTaskResult {
   bool success = 1;
+  repeated string failed_jobs = 2;
   // TODO when part of the task set are scheduled successfully
```
milenkovicm: i guess this comment is irrelevant now, could we remove it ?

ballista/executor/src/executor_server.rs
```diff
@@ -914,7 +923,10 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
                     .unwrap();
             }
         }
-        Ok(Response::new(LaunchMultiTaskResult { success: true }))
+        Ok(Response::new(LaunchMultiTaskResult {
+            success: true,
```
milenkovicm: its a bit confusing to have success: true but have a failing jobs, do we need success property ?

ballista/scheduler/src/state/executor_manager.rs
```diff
@@ -408,23 +408,19 @@ impl ExecutorManager {
         executor_id: &str,
         multi_tasks: Vec<MultiTaskDefinition>,
         scheduler_id: String,
-    ) -> Result<()> {
+    ) -> Result<Vec<JobId>> {
```
milenkovicm: this result is a bit confusing, we're return ok with list of failed jobs. Do we need to return error in this case, or some other indicator that some of the jobs failed

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -106,10 +106,10 @@ impl TaskLauncher for DefaultTaskLauncher {
                 executor.id, tasks_ids
             );
         }
-        executor_manager
+        let res = executor_manager
             .launch_multi_task(&executor.id, tasks, self.scheduler_id.clone())
             .await?;
-        Ok(())
+        Ok(res)
```
milenkovicm, line 114: same comment here, its a bit confusing that there Ok result with failing jobs, no documentation explains return type

ballista/scheduler/src/test_utils.rs
```diff
@@ -377,7 +377,34 @@ impl TaskLauncher for VirtualTaskLauncher {
...
+pub struct RejectingTaskLauncher {}
+
+#[async_trait::async_trait]
+impl TaskLauncher for RejectingTaskLauncher {
+    async fn launch_tasks(
+        &self,
+        _executor: &ExecutorMetadata,
+        tasks: Vec<MultiTaskDefinition>,
+        _executor_manager: &ExecutorManager,
+    ) -> Result<Vec<JobId>> {
```
milenkovicm: should we return set if we do not expect duplicates

milenkovicm (commented):
hey @coderfender sorry for late review, i think this makes sense. I have one comment regarding return types of method. i find it a bit cunfusing that we return Ok and then list of failed jobs, could we find a better way to indicate partial failure, wdyt ?

ballista/executor/src/executor_server.rs
```diff
@@ -889,8 +889,10 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> ExecutorGrpc
             scheduler_id,
         } = request.into_inner();
         let task_sender = self.executor_env.tx_task.clone();
+        let mut failed_jobs = vec![];
```
milenkovicm: set instead of list perhaps ?

coderfender:
Requesting another look @milenkovicm :)

milenkovicm:
apologies @coderfender will  have a look tomorrow

milenkovicm (approved):
thanks @coderfender and my apologies for such long delay, not acceptable from my side.

milenkovicm:
there is change of ballista proto, i belive we already increased proto version for this release

PR #2022, opened 2026-07-13
fix: resolve #1845 — Add TPC-DS workflow

### LINKED ISSUE #1845: Add TPC-DS workflow

#1688 adds TPCH workflow, it would be great if we could replicate same behaviour and get TPC-DS (SF1 or SF10) workflow. 

copy current approach in TPCH workflow 

**Additional context**

Nam0101:
## Summary

fix: resolve #1845 — Add TPC-DS workflow

## Problem

**Severity**: `High` | **File**: `benchmarks/tpcds/generate_data.sh (or same pattern as TPCH data gen)`

TPCH uses dbgen (or downloads). TPC-DS uses dsdgen from TPC tools. CI must produce parquet (or CSV then convert) under a stable path the workflow mounts.

## Solution

Copy TPCH generate script. Swap tool to dsdgen; SF=1; output parquet preferred (Ballista/DataFusion path). Document tool download URL/version pin. If building dsdgen in CI is heavy, cache tarball or use prebuilt like TPCH does. Tables: full TPC-DS schema for SF1. `ponytail: SF1 only; SF10 via input when CI budget allows.`

## Changes

- `benchmarks/tpcds/generate_data.sh` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
_Note: this change was drafted with AI assistance and reviewed locally before submission._

### DIFF
--- a/benchmarks/tpcds/generate_data.sh
+++ b/benchmarks/tpcds/generate_data.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+# Generate TPC-DS SF1 data as Parquet for Ballista benchmarks.
+# Requires dsdgen from TPC-DS tools (https://www.tpc.org/tpcds/).
+#
+# Usage:
+#   ./generate_data.sh [SCALE_FACTOR] [OUTPUT_DIR]
+# Defaults: SCALE_FACTOR=1, OUTPUT_DIR=./data
+#
+# Environment:
+#   DSDGEN_DIR  - path to built TPC-DS tools (must contain dsdgen)
+#   DSDGEN_URL  - optional tarball URL if tools not present
+#
+# ponytail: SF1 only; SF10 via input when CI budget allows.
+
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


benchmarks/tpcds/generate_data.sh
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env bash
+# Generate TPC-DS SF1 data as Parquet for Ballista benchmarks.
+# Requires dsdgen from TPC-DS tools (https://www.tpc.org/tpcds/).
```
andygrove, line 3: I'd recommend using https://github.com/clflushopt/tpchgen-rs instead. It supports generating both TPC-H and TPC-DS and is much faster than dsdgen

phillipleblanc: It can also output to parquet directly

andygrove:
Thanks for starting this @Nam0101 but we now have TPC-DS enabled in CI, so will close this.

PR #2037, opened 2026-07-14
fix(scheduler): persist terminal job state before cache eviction

### LINKED ISSUE #2036: Terminal jobs can appear as Running while their final state is persisted
**Describe the bug**
`TaskManager::succeed_job` and `TaskManager::abort_job` remove a job from the active cache before saving its terminal state.

During that window, `get_job_status` misses the active cache and reads shared state, which may still report `Running`. A completed, failed, or cancelled job can therefore appear to move back to `Running`.

The abort path has an additional failure mode. It collects the running tasks that need cancellation, then awaits `save_job`. If persistence fails, the method returns before those tasks reach `QueryStageScheduler`, so executor work is not cancelled.

**To Reproduce**
Use a `JobState` implementation whose `save_job` method can block or fail:

1. Add a running job to the active cache and shared state.
2. Mark the execution graph as successful or failed.
3. Call `succeed_job` or `abort_job`.
4. Block `save_job` before it updates shared state.
5. Call `get_job_status` while persistence is blocked.
6. Observe that the status falls back to stale shared state and reports `Running`.

For the abort failure:

1. Assign at least one task to an executor.
2. Call `abort_job`.
3. Make `save_job` return an error.
4. Observe that the captured running tasks are not dispatched for cancellation.

**Expected behavior**
Terminal job status should remain continuously visible while it is persisted. Status requests should not observe `Running` after the local graph becomes successful, failed, or cancelled.

Running executor tasks should be dispatched for cancellation even if terminal-state persistence fails. The scheduler should also avoid holding execution-graph locks across asynchronous persistence.

**Additional context**
The race is difficult to observe with the in-memory state backend because persistence is fast. It becomes more likely with a slower external state backend.

phillipleblanc:

Closes #2036

 # Rationale for this change

Terminal jobs were removed from the scheduler's active cache before their final state was saved. During that window, status requests could fall back to shared state and incorrectly report the job as `Running`.

Abort handling had a related failure mode. If terminal-state persistence failed, `abort_job` returned before dispatching cancellation for running executor tasks.

This PR keeps terminal jobs visible in the active cache until persistence finishes. It applies the same ordering to successful, failed, and cancelled jobs.

   - Persist terminal state before removing jobs from the active cache.
   - Count only cached `Running` jobs in the running-job metric.
   - Dispatch executor task cancellation before abort persistence, so save failures cannot suppress cancellation.

No

### DIFF
--- a/ballista/scheduler/src/state/task_manager.rs
+++ b/ballista/scheduler/src/state/task_manager.rs
@@ -52,6 +52,7 @@ use log::{debug, error, info, trace, warn};
 use rand::distr::Alphanumeric;
 use rand::{RngExt, rng};
 use std::collections::{HashMap, HashSet};
+use std::future::Future;
 use std::ops::Deref;
 use std::sync::Arc;
 use std::time::Duration;
@@ -341,7 +342,12 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
 
     /// Get the number of running jobs.
     pub fn running_job_number(&self) -> usize {
-        self.active_job_cache.len()
+        self.active_job_cache
+            .iter()
+            .filter(|job_info| {
+                matches!(job_info.status, Some(job_status::Status::Running(_)))
+            })
+            .count()
     }
 
     /// Generate an ExecutionGraph for the job and save it to the persistent state.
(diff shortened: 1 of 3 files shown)


lukekim (approved):

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -655,67 +664,117 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> TaskManager<T, U>
...
+        &self,
+        job_id: &JobId,
+        snapshot: &ExecutionGraphBox,
+    ) -> Result<()> {
+        if let Some(mut job_info) = self.active_job_cache.get_mut(job_id) {
+            job_info.status = snapshot.status().status.clone();
+        }
+
+        let mut attempt = 0;
+        let save_result = loop {
```
milenkovicm: would it make sense to have this retry loop be implemented as part of job state implementation? some job state implementations might be retriable some not, i believe its specific to implementation, and it might  not make sense to me to have retry logic here?

milenkovicm (commented):
thanks @phillipleblanc  i think this make sense, one comment, should we do retry as part of  job state implementation rather than part of task manager? to me it looks like retry ability is implementation specific ? wdyt ?

ballista/scheduler/src/state/task_manager.rs
phillipleblanc: Yeah, you are correct - this should really be an implementation detail of the job state implementation. I'll add a comment to the trait method that this is an expectation of the implementer to handle retries so its clear. Thanks!

phillipleblanc: I've updated this: https://github.com/apache/datafusion-ballista/compare/df258a882968ddbc1f6cc1412c2fd32647417fc1..610b4c1df6f78ae556a0d69d8660dd197983762b

milenkovicm (approved):

milenkovicm:
Thanks @phillipleblanc & @lukekim

PR #2038, opened 2026-07-14
Multi-partition tasks: partition_slice plumbing + SortShuffleWriter refactor

avantgardnerio:
## Summary

Migrates Ballista from a 1-task-per-partition model to a slice-per-task model, so that a 16-vcore executor can drive 16 partitions with one task rather than launching 16 single-partition tasks. Goal is to allow arbitrary re-partitioning strategies like the upcoming `OrderedRangeRepartition` and `UnorderedRangeRepartition` while sharing them with pure DataFusion (a follow-up will remove the `Hash` branch of `ShuffleWriter` and replace it with a DataFusion `RepartitionExec(Hash)`).

Opens the door to:
1. [parallel sort](https://webdocs.cs.ualberta.ca/~jonathan/publications/parrallel_computing_publications/psrs1.pdf)
2. parallel windows
3. parallel sort-merge joins
4. parallel aggregations
5. OOMs at cluster threshold, not process threshold

### Core changes

- **Coordinator + oneshot handoff** in both `ShuffleWriter` and `SortShuffleWriter`. A task's `global_output_partition_ids.len()` input partitions run concurrently, each producing its own file(s); the internal state fans results out through DataFusion's per-partition `execute(N)` contract. Sort writer's per-pipeline memory budget is divided across concurrent pipelines to preserve the configured per-task total. ASCII threading diagrams live on both writer doc comments.
- **`global_output_partition_ids`** on `TaskDefinition`/`TaskId` proto so the executor knows which global partition ids each task's restricted plan is covering. Encoded as `repeated uint32 ... = 3 [packed=false]` so a single-partition message is bit-identical on the wire to the legacy `uint32 partition_id = 3` scalar. `ShuffleWriter(None)` walks its child plan (`walk_child_partition_mapping`) and picks the right mapping: `SortPreservingMergeExec` -> all outputs go to partition 0, `RepartitionExec::Hash/RoundRobin` -> K-space is inherently global, otherwise -> `global = global_output_partition_ids[local]`. `ShuffleWriter(Hash)` and `SortShuffleWriter` use their K-space / input-partition space directly.
- **Scan restriction extended**: `MemorySourceConfig` was silently unrestricted at dispatch and over-reading in multi-partition tasks; `restrict_shuffle_reader` preserves the `Hash`/`RoundRobinBatch` partitioning kind after shrink so `InterleaveExec` above the reader keeps its "children share hash partitioning" invariant. Task builder routes each `UnionExec` parent partition to its child's local index.
- **Stage partition sizing**: read `partition_count` directly from the shuffle writer's immediate child's `output_partitioning()` — that's the number of independent output-partition polling contexts the writer needs to drive. Replaces the earlier leaf-sum walker, which undercounted after upstream #2062 (`EmptyExec` serde-safe rewrite) and overcounted through aligned-input joins.

### Scheduler bookkeeping

- `SuccessfulStage::to_running` walks `task_infos` newest-to-oldest so retried partitions aren't double-pushed to `pending`.
- `resubmit_successful_stages` / `reset_running_stages` handle the `map_partition_id`-is-`task_id` semantics correctly (bounds check against `task_infos.len()`, TODO for eventual partition-id semantics).
- Failure error messages name the *partition(s)* that hit `max_task_failures`, since retry attempts get fresh task ids.
- `task_id` and the previously separate `task_index` (which always moved in lockstep at bind time) are merged into a single per-stage append-order slot id.

### Ops knobs and observability

- **`ballista.scheduler.max_partitions_per_task`** — Spark-persona escape hatch. `0` (default) keeps this branch's multi-partition-tasks behavior; setting to `1` restores the pre-PR one-task-per-partition scheduling for non-collapse stages. Collapse stages ignore the cap since a split collapse would produce partial results downstream can't merge.
- **Per-partition metrics preserved**: added `optional uint32 partition = 16` to `OperatorMetric` on the wire so per-partition detail survives multi-partition tasks. `execution_stage` used to key metrics by the task's single partition id; under this branch that arg is the task slot, so without this field all N partitions in a slice would file into one bucket. Additive proto change; the UI's per-partition breakdown works again.
- **Docs**: `docs/developer/architecture.md` gains a section on the slice-per-task dispatch model, contrasting it with Spark's per-partition task unit, and showing how the model composes with in-flight DataFusion AQE PoCs (apache/datafusion#23026, apache/datafusion#23167).

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test --workspace`
- [x] `should_execute_submitted_physical_plan_across_shuffle_stages`
- [x] Full sort shuffle test suite
- [x] Manual TPC-H sweep (21 queries) end-to-end

### DIFF
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -302,7 +302,10 @@ message TaskId {
   uint32 task_attempt_num = 2;
   // Global partition ids covered by this task, in slice order. Position i in
   // the restricted plan corresponds to `global_output_partition_ids[i]` globally.
-  repeated uint32 global_output_partition_ids = 4;
+  // `[packed=false]` keeps the single-partition case wire-compatible with the
+  // legacy `uint32 partition_id = 3` field: one element writes as a single
+  // varint at tag 3, exactly matching the old scalar encoding.
+  repeated uint32 global_output_partition_ids = 3 [packed=false];
   // Vcores this task consumed from the executor's budget at bind time.
   // The executor uses this to size the task's memory pool proportionally
   // (`pool = total_memory * vcores_consumed / total_vcores`), so a
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -289,10 +289,26 @@ message PartitionId {
   uint32 partition_id = 4;
 }
 
+// Within-stage task identity (paired with job_id + stage_id on the parent
+// MultiTaskDefinition). One task processes a slice of partitions, so
+// `task_index` names the task within the stage; `global_output_partition_ids` gives
+// the concrete global partition ids the task's restricted plan is
+// covering. Writers use these to name shuffle files with global identity
+// (via `create_shuffle_path`) so downstream reads have a stable, canonical
+// address regardless of how the scheduler split the work.
 message TaskId {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
-  uint32 partition_id = 3;
+  uint32 task_index = 3;
+  // Global partition ids covered by this task, in slice order. Position i in
+  // the restricted plan corresponds to `global_output_partition_ids[i]` globally.
+  repeated uint32 global_output_partition_ids = 4;
+  // Vcores this task consumed from the executor's budget at bind time.
+  // The executor uses this to size the task's memory pool proportionally
+  // (`pool = total_memory * vcores_consumed / total_vcores`), so a
+  // multi-partition task claiming N vcores gets N/total of the executor's
+  // memory budget rather than a single per-vcore share.
+  uint32 vcores_consumed = 5;
 }
 
 message PartitionStats {
@@ -530,7 +546,10 @@ message TaskStatus {
   string job_id = 2;
   uint32 stage_id = 3;
   uint32 stage_attempt_num = 4;
-  uint32 partition_id = 5;
+  // Task index within the stage (was `partition_id` — under the
+  // multi-partition-task model one task processes a partition slice, not a
+  // single partition).
+  uint32 task_index = 5;
   uint64 launch_time = 6;
   uint64 start_exec_time = 7;
   uint64 end_exec_time = 8;
@@ -549,17 +568,30 @@ message PollWorkParams {
   repeated TaskStatus task_status = 3;
 }
 
+// Pull-based single-task dispatch. One task processes a slice of
+// partitions; `task_index` names the task within the stage,
+// `global_output_partition_ids` gives the concrete global partition ids it covers. The
+// task's plan is scheduler-side shrink-restricted so its leaves report
+// `slice.len()` partitions; the writer uses `global_output_partition_ids` to attach
+// global identity to shuffle files (except in cases where the plan itself
+// resets partitioning: the writer walks its child plan to detect
+// SortPreservingMergeExec or RepartitionExec::Hash and picks the right
+// global mapping).
 message TaskDefinition {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
   string job_id = 3;
   uint32 stage_id = 4;
   uint32 stage_attempt_num = 5;
-  uint32 partition_id = 6;
+  uint32 task_index = 6;
   bytes plan = 7;
   string session_id = 9;
   uint64 launch_time = 10;
   repeated KeyValuePair props = 11;
+  // Global partition ids covered by this task, in slice order.
+  repeated uint32 global_output_partition_ids = 12;
+  // Vcores this task consumed at bind time — see TaskId.vcores_consumed.
+  uint32 vcores_consumed = 13;
 }
 
 // A set of tasks in the same stage
@@ -843,7 +875,8 @@ message RunningTaskInfo {
   uint32 task_id = 1;
   string job_id = 2;
   uint32 stage_id = 3;
-  uint32 partition_id = 4;;
+  // Task slot within the stage (was `partition_id` — see TaskStatus).
+  uint32 task_index = 4;
 }
 
 service SchedulerGrpc {
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -378,6 +378,10 @@ message OperatorMetric {
     NamedRatio ratio = 14;
     uint64 output_batches = 15;
   }
+  // Local partition index within the reporting task's plan (0..slice_len-1).
+  // The scheduler maps this to a global partition id via the task's
+  // `global_input_partition_ids` when folding metrics into stage metrics.
+  optional uint32 partition = 16;
 }
 
 // Used by scheduler
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -132,7 +132,6 @@ message ExecutionGraph {
   uint64 output_partitions = 5;
   repeated PartitionLocation output_locations = 6;
   string scheduler_id = 7;
-  uint32 task_id_gen = 8;
   repeated StageAttempts failed_attempts = 9;
   string job_name = 10;
   uint64 start_time = 11;
@@ -290,16 +289,17 @@ message PartitionId {
 }
 
 // Within-stage task identity (paired with job_id + stage_id on the parent
-// MultiTaskDefinition). One task processes a slice of partitions, so
-// `task_index` names the task within the stage; `global_output_partition_ids` gives
-// the concrete global partition ids the task's restricted plan is
-// covering. Writers use these to name shuffle files with global identity
-// (via `create_shuffle_path`) so downstream reads have a stable, canonical
+// MultiTaskDefinition). One task processes a slice of partitions;
+// `task_id` names the task within the stage (it's the task's append-order
+// slot in `RunningStage.task_infos`, so (job_id, stage_id, task_id) is
+// globally unique). `global_output_partition_ids` gives the concrete
+// global partition ids the task's restricted plan is covering. Writers
+// use these to name shuffle files with global identity (via
+// `create_shuffle_path`) so downstream reads have a stable, canonical
 // address regardless of how the scheduler split the work.
 message TaskId {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
-  uint32 task_index = 3;
   // Global partition ids covered by this task, in slice order. Position i in
   // the restricted plan corresponds to `global_output_partition_ids[i]` globally.
   repeated uint32 global_output_partition_ids = 4;
@@ -542,14 +542,14 @@ message ShuffleWritePartition {
 }
 
 message TaskStatus {
+  // Task's append-order slot in `RunningStage.task_infos`; (job_id,
+  // stage_id, task_id) is globally unique. Under the multi-partition-task
+  // model one task processes a partition slice, so this names the task,
+  // not a single partition.
   uint32 task_id = 1;
   string job_id = 2;
   uint32 stage_id = 3;
   uint32 stage_attempt_num = 4;
-  // Task index within the stage (was `partition_id` — under the
-  // multi-partition-task model one task processes a partition slice, not a
-  // single partition).
-  uint32 task_index = 5;
   uint64 launch_time = 6;
   uint64 start_exec_time = 7;
   uint64 end_exec_time = 8;
@@ -569,10 +569,11 @@ message PollWorkParams {
 }
 
 // Pull-based single-task dispatch. One task processes a slice of
-// partitions; `task_index` names the task within the stage,
-// `global_output_partition_ids` gives the concrete global partition ids it covers. The
-// task's plan is scheduler-side shrink-restricted so its leaves report
-// `slice.len()` partitions; the writer uses `global_output_partition_ids` to attach
+// partitions; `task_id` names the task within the stage (its append-order
+// slot in `RunningStage.task_infos`), `global_output_partition_ids` gives
+// the concrete global partition ids it covers. The task's plan is
+// scheduler-side shrink-restricted so its leaves report `slice.len()`
+// partitions; the writer uses `global_output_partition_ids` to attach
 // global identity to shuffle files (except in cases where the plan itself
 // resets partitioning: the writer walks its child plan to detect
 // SortPreservingMergeExec or RepartitionExec::Hash and picks the right
@@ -583,7 +584,6 @@ message TaskDefinition {
   string job_id = 3;
   uint32 stage_id = 4;
   uint32 stage_attempt_num = 5;
-  uint32 task_index = 6;
   bytes plan = 7;
   string session_id = 9;
   uint64 launch_time = 10;
@@ -872,11 +872,11 @@ message RemoveJobDataResult {
 }
 
 message RunningTaskInfo {
+  // Per-stage append-order slot; (job_id, stage_id, task_id) is globally
+  // unique (see TaskStatus).
   uint32 task_id = 1;
   string job_id = 2;
   uint32 stage_id = 3;
-  // Task slot within the stage (was `partition_id` — see TaskStatus).
-  uint32 task_index = 4;
 }
 
 service SchedulerGrpc {
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -342,16 +342,16 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
         ConfigEntry::new(
             BALLISTA_SCHEDULER_MAX_PARTITIONS_PER_TASK.to_string(),
             "Upper bound on the number of input partitions packed into a single \
-             task's `partition_slice`. `0` (default) means unbounded — the scheduler \
-             fills each task up to the executor's free vcore count. Set to `1` to \
-             restore the pre-multi-partition-tasks execution model (one task per \
-             partition), matching the Spark user persona at the cost of the \
-             parallel-sort / parallel-join wins this branch enables. Does not apply \
-             to collapse stages, which must pack their full pending queue into a \
-             single task for correctness."
+             task's `partition_slice`. `1` (default) is the pre-multi-partition-tasks \
+             execution model — one task per partition, matching the Spark user \
+             persona and preserving master's behaviour on merge. Raise to enable \
+             multi-partition tasks (fewer tasks, parallel-sort / parallel-join wins); \
+             `0` means unbounded — the scheduler fills each task up to the executor's \
+             free vcore count. Does not apply to collapse stages, which must pack \
+             their full pending queue into a single task for correctness."
                 .to_string(),
             DataType::UInt64,
-            Some(0.to_string()),
+            Some(1.to_string()),
         ),
     ];
     entries
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -142,6 +142,10 @@ pub const BALLISTA_CHAOS_EXECUTION_SEED: &str = "ballista.testing.chaos_executio
 /// Valid values are: none, lz4, zstd
 pub const BALLISTA_SHUFFLE_COMPRESSION_CODEC: &str = "ballista.shuffle.compression.codec";
 
+/// Configuration key for the scheduler's per-task partition-slice cap.
+pub const BALLISTA_SCHEDULER_MAX_PARTITIONS_PER_TASK: &str =
+    "ballista.scheduler.max_partitions_per_task";
+
 /// Result type for configuration parsing operations.
 pub type ParseResult<T> = result::Result<T, String>;
 use std::sync::LazyLock;
@@ -334,6 +338,20 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
             none, lz4, zstd. Defaults to lz4 to preserve current behaviour".to_string(),
             DataType::Utf8,
             Some("lz4".to_string()),
+        ),
+        ConfigEntry::new(
+            BALLISTA_SCHEDULER_MAX_PARTITIONS_PER_TASK.to_string(),
+            "Upper bound on the number of input partitions packed into a single \
+             task's `partition_slice`. `0` (default) means unbounded — the scheduler \
+             fills each task up to the executor's free vcore count. Set to `1` to \
+             restore the pre-multi-partition-tasks execution model (one task per \
+             partition), matching the Spark user persona at the cost of the \
+             parallel-sort / parallel-join wins this branch enables. Does not apply \
+             to collapse stages, which must pack their full pending queue into a \
+             single task for correctness."
+                .to_string(),
+            DataType::UInt64,
+            Some(0.to_string()),
         ),
     ];
     entries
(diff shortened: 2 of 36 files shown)


andygrove:
Thanks for the contribution @avantgardnerio. It's going to take me a while to work through this and understand it all.

One initial question, though. This PR changes the default execution model from one task per partition to multiple partitions per task. Rather than making this a default change in Ballista, maybe we could consider a config around this? Something like "max partitions per task"? I am concerned that forcing this change on the [Spark user persona](https://datafusion.apache.org/ballista/contributors-guide/user-personas.html#persona-2-the-spark-user-wanting-the-same-execution-model) will cause confusion. If it is something they can opt into for better performance, that is fine.

andygrove:
TPC-H CI job is failing due to q15 returning incorrect data (returning too many rows).

avantgardnerio:
> "max partitions per task"

Yeah, absolutely, strictly speaking I think this is a superset of that, so "max of 1" should be easy - I'll add a config.

ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -195,182 +289,286 @@ impl SortShuffleWriterExec {
             .partition_count()
     }
 
-    /// Execute the sort-based shuffle write for a single input partition.
+    /// Executes the sort-based shuffle write for this task.
+    ///
+    /// One pipeline per input partition of the (scheduler-side-shrink-restricted)
```
andygrove: I want to check that I am understanding this correctly. The current model is one input partition per task, producing one output data file + index, and the new model ends up producing the same number of files, just moving the concurrency within a task rather than across tasks?

avantgardnerio: Correct. Say previously 16 input partitions = 16 datafusion session contexts (tasks), producing 16 output files

Proposed: 16 input partitions, 1 task, 16 output files

ballista/core/proto/ballista.proto
```diff
@@ -419,9 +429,14 @@ message ExecutorSpecification {
...
 }
 
+// Virtual cores this executor has been assigned at runtime — analogous to
+// YARN vcores (yarn.nodemanager.resource.cpu-vcores) or Spark's
+// spark.executor.cores. Not physical `nproc`: may over- or under-subscribe.
+// One task per executor drives this many DataFusion partitions in parallel
+// through one plan-Arc.
 message ExecutorResource {
   oneof resource {
-    uint32 task_slots = 1;
```
andygrove: The rename of `task_slots` to `vcores` could potentially be split out into a separate PR? I don't think it would cause any regression in current behavior?

avantgardnerio: Happy to do it. I just thought it didn't make sense in isolation. If we agree we're going in the direction of this PR, I can break it up in as many parts as make sense.

andygrove: Sure, we can hold off until I am farther along with the review

ballista/core/proto/ballista.proto
```diff
@@ -289,10 +289,20 @@ message PartitionId {
...
+// (via `create_shuffle_path`) so downstream reads have a stable, canonical
+// address regardless of how the scheduler split the work.
 message TaskId {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
-  uint32 partition_id = 3;
+  uint32 task_index = 3;
+  // Global partition ids covered by this task, in slice order. Position i in
+  // the restricted plan corresponds to `partition_slice[i]` globally.
+  repeated uint32 partition_slice = 4;
```
andygrove: So `partition_slice` / `i` is the `partition_id`?, or an index into some list of partition ids?

.github/workflows/tpch.yml
```diff
@@ -111,7 +111,7 @@ jobs:
             --bind-port 50051 \
             --scheduler-host 127.0.0.1 \
             --scheduler-connect-timeout-seconds 10 \
-            --concurrent-tasks 4 \
+            --vcores 4 \
```
Dandandan, line 114: Why change this to vcores?

ballista/core/proto/ballista.proto
Dandandan: Why call it slice? Partitions / partition_ids would be easier to understand?

milenkovicm:
Will review over weekend, until then i have one unreleased comment: great to see all of you here 😁

.github/workflows/tpch.yml
avantgardnerio: Because it only represented concurrent tasks in the old model, where 1 task consumed 1 partition, which is generally modeled to use 1 core (unless deliberately overridden to over/under subscribe).

In the new model, the scheduler will attempt to assign as many partitions as possible to a single task. So the `task_slot` / `concurrent-tasks` were no longer semantically correct.

I needed a new term. I felt `vcores` was the most clear / concise. It roughly represents number of cores, but not really due to over/under subscription, kubernetes / docker, hyperthreading - thus the `v`.

I'm happy to call it whatever we decide as long as the new term is semantically correct / unambiguous (the old `partitions` turned into 3 things in the new model).

ballista/core/proto/ballista.proto
```diff
@@ -289,10 +289,20 @@ message PartitionId {
...
+// MultiTaskDefinition). One task processes a slice of partitions, so
+// `task_index` names the task within the stage; `partition_slice` gives
+// the concrete global partition ids the task's restricted plan is
+// covering. Writers use these to name shuffle files with global identity
+// (via `create_shuffle_path`) so downstream reads have a stable, canonical
+// address regardless of how the scheduler split the work.
 message TaskId {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
-  uint32 partition_id = 3;
```
andygrove, line 295: Should we just make `partition_id` repeated? The current case of passing a single id would carry on working, and it would also support the multiple partitions per task approach?

ballista/core/proto/ballista.proto
avantgardnerio: Globally, partitions might be `[0,1,2,3,4,5,6,7]`. There might be 2 executors with 4 `vcores`. Barring other concurrent usage, the scheduler will now assign them to executors as:

```
exec A: [0,1,2,3]
exec B: [4,5,6,7]
```

However, exec B just gets 4 partitions, so without `partition_slice`, it just sees a zero-indexed `Vec` of 4 partitions `[0,1,2,3]`. So to enable it to write output shuffles with the correct name / location, we send it `partition_slice: [4,5,6,7]` so it knows its local `[0,1,2,3]` maps to the global `[4,5,6,7]`.

1. I tried sending empty partitions `[[],[],[],[],[file4],[file5],[file6],[file7]` but that had execution ramifications since it would have to drive empty streams among other issues.
2. I tried _not_ sending `partition_slice` so the local DF context knew nothing about the global situation, and have the scheduler map all the local results back to global ones but it was more complex.

I'm happy to re-evaluate anything.

edit: short answer `partition_slice[i]` is the global partition number.

ballista/core/proto/ballista.proto
```diff
@@ -536,22 +557,33 @@ message TaskStatus {
...
+// SortPreservingMergeExec or RepartitionExec::Hash and picks the right
+// global mapping).
 message TaskDefinition {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
   string job_id = 3;
   uint32 stage_id = 4;
   uint32 stage_attempt_num = 5;
-  uint32 partition_id = 6;
+  uint32 task_index = 6;
```
andygrove: What is the difference between `task_id` and `task_index`?

ballista/core/proto/ballista.proto
andygrove: Thanks for clarifying. To me, it would be simpler to rename `partition_slice` as `partition_id` or `partition_ids` since it is a slice of global partition IDs.

.github/workflows/tpch.yml
andygrove: Spark has `spark.executor.cores` and `spark.task.cpus` (number of cores per task, defaults to 1).

`vcores` seems consistent with that.

ballista/core/proto/ballista.proto
avantgardnerio: short answer: it's tech debt, I'm sorry I bothered you with it. I'll fix.

longer more revealing answer:

There's really no longer a concept of a stable ID for a task, because partitions are assigned dynamically to tasks based upon number of `vcores` available as tasks are dispatched. A task could be dispatched onto an executor with 4 free `vcores`, fail, and be regenerated as 2 new tasks on 2 executors with 2 `vcores`. `task_index` was added to address this. Every time a task (including failed and regenerated) gets dispatched it has an new `task_index`. It's an index into `RunningStage.task_infos: Vec<TaskInfo>`.

So truth be told, it's almost equivalent to `task_id` and we should merge them and fix the code properly. I'll add that to the PR.

Also `stage_attempt_num` needs to be refactored into `partition_attempt_nums[]`.

andygrove:
It looks like per-partition metrics break with this PR. Here is what my AI tells me ...

Right now OperatorMetricsSet on the wire is a flat list of metrics with no partition field. The per-partition granularity we show in the UI has always been derived from the fact that a task's index and its partition id were the same number. In execution_stage.rs we stamp every metric a task reports with Some(partition), and under this PR that value is now the task index rather than a partition id.

The effect is that a task covering a slice of N partitions files all N partitions' metrics under one task-index key, so they get merged into a single bucket. The read side in handlers.rs stays self-consistent so nothing errors, but the per-partition detail is gone before it is ever stored. In practice that means skew inside a slice becomes invisible. If one partition of sixteen carries most of the rows, the stage view shows one large task with no way to see which partition drove it, and that is exactly the view Spark-model users lean on to debug skew.

At one partition per task this all looks fine, which is what makes it easy to miss. The regression only shows up once a slice covers more than one partition.

I think the clean fix is to make this additive rather than inferred: add an optional partition field to OperatorMetric, have the executor label each metric with its local partition index, and have the scheduler map local to global through partition_slice when it folds task metrics into the stage. That keeps per-partition metrics working in the new model instead of collapsing them to per-task.

ballista/core/proto/ballista.proto
andygrove: If it's not too much trouble, I think it would be good to go ahead with a standalone PR to do this rename to `vcores`. I personally think it makes more sense than the current naming, because it allows for more flexibility in the execution model, including for the proposed approach in the current PR.

avantgardnerio: Great! I'll get right on it. Any incremental value we can deliver is a win in my book :smile:

ballista/core/proto/ballista.proto
avantgardnerio: Would `global_partition_ids` be good? Now that we have both, it's probably helpful to distinguish.

ballista/core/proto/ballista.proto
avantgardnerio: The PR is up! https://github.com/apache/datafusion-ballista/pull/2045

avantgardnerio:
> TPC-H CI job is failing due to q15 returning incorrect data (returning too many rows).

Fixed!

avantgardnerio:
This branch is now at rough performance parity with main (which is the goal, performance improvements will come as we're able to implement features on top of this work):

image

milenkovicm:
Would it be possible to have a paragraph or two explaining the approach and compare it with spark, somewhere in documentation?

ballista/core/proto/ballista.proto
```diff
@@ -289,10 +289,26 @@ message PartitionId {
...
+// (via `create_shuffle_path`) so downstream reads have a stable, canonical
+// address regardless of how the scheduler split the work.
 message TaskId {
   uint32 task_id = 1;
   uint32 task_attempt_num = 2;
-  uint32 partition_id = 3;
+  uint32 task_index = 3;
+  // Global partition ids covered by this task, in slice order. Position i in
+  // the restricted plan corresponds to `global_output_partition_ids[i]` globally.
+  repeated uint32 global_output_partition_ids = 4;
```
andygrove: It isn't clear to me from the current comments how the Spark persona case is handled, where each task processes a single input partition and potentially produces multiple output partitions. Where would the input partition id be specified?

andygrove: I posted this comment in the wrong place. I will repost.

ballista/core/proto/ballista.proto
```diff
@@ -549,17 +568,30 @@ message PollWorkParams {
...
   uint32 stage_id = 4;
   uint32 stage_attempt_num = 5;
-  uint32 partition_id = 6;
+  uint32 task_index = 6;
   bytes plan = 7;
   string session_id = 9;
   uint64 launch_time = 10;
   repeated KeyValuePair props = 11;
+  // Global partition ids covered by this task, in slice order.
+  repeated uint32 global_output_partition_ids = 12;
```
andygrove, line 599: It isn't clear to me from the current comments how the Spark persona case is handled, where each task processes a single input partition and potentially produces multiple output partitions. Where would the input partition id be specified, now that `partition_id` has been removed?

avantgardnerio: Can you help me understand the problem? IIUC, the only thing the input partition number was used for was to restrict the execution plan on the executor. That functionality has now moved to the scheduler, so plans arrive already restricted. So in the Spark case (max_paritions_per_task=1) , the only thing that changes is that this plan has a single input partition (and can still produce multiple outputs). It doesn't need to know it's input number (and technically not even really it's output numbers aside from file collisions with other tasks). If there's something I'm missing, I'm happy to add it back, but it was coming up as dead code.

andygrove:
Thanks for this — the slice-per-task direction is compelling. I ran this branch on my k3s TPC-H **SF1000** cluster to see how it moves the needle, and wanted to share the numbers plus a root-cause read while the run was going, since a clear pattern showed up early.

I stopped the run after Q7 (single iteration) once the signal was consistent, so this is a partial suite — but it's enough to surface a regression on the join-heavy queries that's worth discussing.

### Setup (identical to our published SF1000 baseline, only the image differs)

| | |
|---|---|
| Branch | `avantgardnerio/arrow-ballista@308f1495` (this PR) |
| Baseline | `apache/main` (the published [benchmarking-guide](https://datafusion.apache.org/ballista/contributors-guide/benchmarking.html) SF1000 AQE-off numbers, same cluster) |
| Scale | SF1000, node-local Parquet |
| Cluster | 2 executors, one per node, 8 cores / 56 GiB, `--memory-pool-size=48GB` |
| Config | `target_partitions=32`, `prefer_hash_join=false`, `enable_dynamic_filter_pushdown=false`, **AQE off** (`ballista.planner.adaptive.enabled=false`) |
| Iterations | 1 |

### Results (Q1–Q7, seconds; ratio = baseline / PR, >1 = PR faster)

| Query | PR #2038 | main (AQE off) | ratio | |
|---|---|---|---|---|
| 1  | 41.3  | 70.8  | **1.71×** | 🟢 |
| 2  | 183.2 | 155.5 | 0.85× | 🔴 |
| 3  | 256.5 | 206.2 | 0.80× | 🔴 |
| 4  | 67.2  | 76.8  | **1.14×** | 🟢 |
| 5  | 692.9 | 542.7 | 0.78× | 🔴 |
| 6  | 15.0  | 18.3  | **1.22×** | 🟢 |
| 7  | 748.7 | 575.5 | 0.77× | 🔴 |
| **Σ 1–7** | **2004.8** | **1645.8** | **0.82×** | |

The direction is consistent: **join-free queries improve (Q1, Q4, Q6), join-heavy queries regress ~20–25% (Q2, Q3, Q5, Q7)**.

### Root cause: coarser scheduling granularity → straggler tails on skewed join stages

The `DIAG bind_one` scheduler traces make the mechanism visible. On a light query, each stage binds cleanly as a few fat slices with no leftover single-partition tasks and a sub-second spread. On a join-heavy query, the same 32-partition stages fragment and dispatch over a long window:

| stage kind | tasks | 1-partition fragments | first→last bind spread |
|---|---|---|---|
| light query, stage | 4 | 0 | <1 s |
| join query, stage A | 6 | 2 | 100 s |
| join query, stage B | 9 | 5 | **178 s** |
| join query, worst | 18 | 15 | — |

A slice-task holds **all** of its vcore slots until its **slowest** partition finishes. On join stages, per-partition runtime is very uneven (join-key skew + spilling `SortExec` — one partition spilled 22 GB / 455 M rows in an SMJ with `join_time≈413 s`). So the binder places a couple of fat slices, then has to dribble the remaining partitions out one at a time as slots free (the `frag1` counts), and a whole stage's partitions end up spread across 100–180 s. During each slice's tail, the sibling vcores sit idle and the next partitions can't dispatch. The pre-PR 1-task-per-partition model freed each slot independently, so the 32 partitions load-balanced finely across the 16 vcores. Uniform stages don't hit the tail and actually gain from the lower per-task overhead — hence the wins on Q1/Q4/Q6.

A second-order cost: fragmented stages re-encode a restricted plan per task, so a 15-fragment stage pays ~15× the plan-serialization of a clean 4-task stage.

### What this does *not* appear to be: memory starvation

I initially suspected the shared per-task pool, but `memory_pool_policy` scales it correctly — an N-vcore slice gets `N × per_vcore`, so an 8-vcore task gets the full 48 GB `FairSpillPool` split across its 8 pipelines (≈6 GB each, same as the old per-task budget, and better at the tail as siblings deregister). The heavy spilling looks inherent to a 455 M-row SMJ sort at ~6 GB/partition, not amplified by this PR. So this reads as a scheduling/utilization regression rather than a memory one.

### Possible directions (just ideas)

- Cap or shrink slice size on stages whose child is a join / sort so partition granularity stays fine where runtimes are skewed.
- Let a slice release vcores as individual partitions complete, rather than holding all slots until the slowest finishes.

Happy to re-run the full 22-query suite (and with AQE on) on any follow-up, or share the raw scheduler/executor logs. Really nice work opening up the parallel-sort / parallel-join path — this is about tuning the scheduling granularity, not the idea.

---

*Benchmarking, log analysis, and this write-up were done with LLM assistance (Claude).*

ballista/core/proto/ballista.proto
andygrove: Ah, ok. So the plan arrives with the just the relevant scans needed for the input partition(s) and there is no need to add the number in this protobuf message. Makes sense.

avantgardnerio: BTW, "shrunk" not "restricted". That was a load bearing issue. Restricting just emptied out the partitions IIRC, but they still needed to be driven. Now there are only the necessary partitions.

docs/developer/architecture.md
```diff
@@ -48,6 +48,47 @@ processes.
 
 ![Query Execution Flow](images/query-execution.png)
 
+## Multi-Partition Tasks
```
avantgardnerio, line 51: @milenkovicm WDYT? Better, or worse?

ballista/core/src/config.rs
```diff
@@ -335,6 +339,20 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
...
+             task's `partition_slice`. `0` (default) means unbounded — the scheduler \
+             fills each task up to the executor's free vcore count. Set to `1` to \
+             restore the pre-multi-partition-tasks execution model (one task per \
+             partition), matching the Spark user persona at the cost of the \
+             parallel-sort / parallel-join wins this branch enables. Does not apply \
+             to collapse stages, which must pack their full pending queue into a \
+             single task for correctness."
+                .to_string(),
+            DataType::UInt64,
+            Some(0.to_string()),
```
andygrove: Could we default this to 1 for now, so that this is not a breaking change?

andygrove:
I'll benchmark this PR today, both with max of 1 partition per task (current model) and with config set to 0 to allow multiple partitions per task. If there is no regression with setting of 1, I think this is good to go.

@phillipleblanc @milenkovicm WDYT?

docs/developer/architecture.md
```diff
@@ -55,8 +55,9 @@ virtual cores (`vcores`), and the scheduler packs up to that many of a stage's o
 task. All partitions in the slice execute concurrently under one DataFusion plan invocation: scans and shuffle
 readers are rewritten to see only the assigned partition ids, and DataFusion's per-partition `execute(N)`
 contract fans the work across the executor's threads. Slice size is bounded by the executor's free vcore count
-and by `ballista.scheduler.max_partitions_per_task` (`0` = unbounded, `1` = one task per partition — the
-pre-multi-partition-tasks model).
+and by `ballista.scheduler.max_partitions_per_task` (`1` = one task per partition — the default and the
```
milenkovicm, line 58: would it make more sense to have this property  `ballista.scheduler.max_partitions_per_task_unbounded` ?

ballista/core/src/config.rs
```diff
@@ -342,16 +342,16 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
         ConfigEntry::new(
```
milenkovicm, line 342: is there going to be any other value than 0 and 1, if not, would it make sense to have it as boolean

milenkovicm (commented):
few comments regarding config property naming and type. if you're happy with performance please do merge it @andygrove

ballista/core/src/config.rs
avantgardnerio: Yes, I think it makes sense to allow users to choose their level of task-internal parallelism. I see it as a tuning knob.

milenkovicm: ok, sorry my bad, i thought its a 0-1 range but in practice is 0 ..

docs/developer/architecture.md
milenkovicm: makes sense, thanks @avantgardnerio

ballista/core/src/config.rs
avantgardnerio: I think in general it would be good to move towards 0, as it fully allows DF internal parallelism to optimize things. But unfortunately with the state the PR is in today, Andy saw a 20% perf regression on queries with very skewed partitioning. Setting this to 1 is the other extreme, and probably the safe step for existing Spark users.

What I'd like to do over time is to fix how this deals with skewed partitioning (or possibly allow tasks to release vcores before completion), but I think it's outside the scope of this PR.

So in the mean time the correct value is quite likely somewhere between 1 and infinity (0), which is why I think it's a knob not a boolean. Sorry, I should have explained my reasoning the first time.

ballista/core/src/config.rs
```diff
@@ -335,6 +339,20 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
             DataType::Utf8,
             Some("lz4".to_string()),
         ),
+        ConfigEntry::new(
+            BALLISTA_SCHEDULER_MAX_PARTITIONS_PER_TASK.to_string(),
+            "Upper bound on the number of input partitions packed into a single \
+             task's `partition_slice`. `1` (default) is the pre-multi-partition-tasks \
+             execution model — one task per partition, matching the Spark user \
+             persona and preserving master's behaviour on merge. Raise to enable \
```
andygrove, line 347: nit: comment talks about current state of PR "preserving master's behavior on merge". This comment won't mean much to users once this is merged.

andygrove:
I ran TPC-H @ sf1000 q1-7. q2 was slightly slower other queries were about the same. My benchmarking setup is quite new, so I am not even sure how stable performance numbers are yet. This LGTM.

andygrove (approved):
LGTM. Thanks @avantgardnerio! I only tested with the default setting for the new config, since that is the use case I am most interested in right now, but look forward to seeing performance improvements with the multi-partition-per-task config set.

avantgardnerio:
> LGTM. Thanks @avantgardnerio! I only tested with the default setting for the new config

Thanks @andygrove ! I'm going to go ahead and merge. I'll address the 1 nit in a follow up PR tomorrow. I think we can live with max_partitions_per_task=1 for now. Other settings improve some queries and hurt others, but at least it's up to users to tweak for their workloads.

Merging this will unblock me to submit a series of followups to make more small incremental improvements. I really appreciate all the reviews & feedback @milenkovicm @Dandandan & @andygrove !

ballista/core/src/config.rs
avantgardnerio: Please see https://github.com/apache/datafusion-ballista/pull/2093

PR #2039, opened 2026-07-14
feat: make compression codec configurable

### LINKED ISSUE #1980: Make shuffle-file compression codec configurable (add ZSTD option)

sandugood:

Closes #1980.

 # Rationale for this change
Beforehand, only LZ4 compression type was being used in the shuffle write process. Also, there was no way to disable compression at all.

Introduced new config entry: `ballista.shuffle.compression.codec` as well as a helper function, that would pick up compression codec during runtime shuffle write process.

Yes, but not breaking. Compression codec defaults to "lz4" if not provided.


ballista/core/src/utils.rs
```diff
@@ -200,8 +216,7 @@ pub async fn write_stream_to_disk(
             BallistaError::IoError(e)
         })?);
 
-        let options = IpcWriteOptions::default()
-            .try_with_compression(Some(CompressionType::LZ4_FRAME))?;
+        let options = create_new_write_options(config).unwrap();
```
milenkovicm: can `.unwrap()` be avoided this ?

milenkovicm (commented):
thanks @sandugood, one minor comment

ballista/core/src/config.rs
```diff
@@ -334,6 +290,14 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
             DataType::Utf8,
             Some("".to_string()),
         ),
+        ConfigEntry::new(
+            BALLISTA_SHUFFLE_COMPRESSION_CODEC.to_string(),
+            "Compression codec specification \
+            used in the shuffle process. Possible values: \
+            none, lz4, zstd. Default to lz4 to preserve current behaviour".to_string(),
```
phillipleblanc:
```suggestion
            none, lz4, zstd. Defaults to lz4 to preserve current behaviour".to_string(),
```

nit: grammar

ballista/core/src/config.rs
```diff
@@ -585,6 +517,11 @@ impl BallistaConfig {
         self.get_bool_setting(BALLISTA_PROPAGATE_EMPTY_ENABLED)
     }
 
+    /// Returns compression codec that will be used during write stage of shuffle
+    pub fn shuffle_compression_codec(&self) -> String {
+        self.get_string_setting(BALLISTA_SHUFFLE_COMPRESSION_CODEC)
```
phillipleblanc: Maybe we could return Result<CompressionType> here and do the parsing here instead of in `create_new_write_options`?

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -315,10 +311,10 @@ impl ShuffleWriterExec {
 
                                             debug!("Writing results to {p:?}");
 
-                                            let options = IpcWriteOptions::default()
-                                                .try_with_compression(Some(
-                                                    CompressionType::LZ4_FRAME,
-                                                ))?;
+                                            let options =
+                                                create_new_write_options(config.clone())
```
phillipleblanc: This should propagate the error instead of doing unwrap

phillipleblanc (commented):
Thanks! This looks great, just a few minor comments.

ballista/core/src/execution_plans/shuffle_writer.rs
sandugood: Fixed

milenkovicm (approved):
Thanks @sandugood & @phillipleblanc

ballista/core/src/config.rs
```diff
@@ -585,6 +597,21 @@ impl BallistaConfig {
...
+        &self,
+    ) -> std::result::Result<Option<CompressionType>, String> {
+        match self
+            .get_string_setting(BALLISTA_SHUFFLE_COMPRESSION_CODEC)
+            .as_str()
+        {
+            "lz4" => Ok(Some(CompressionType::LZ4_FRAME)),
+            "zstd" => Ok(Some(CompressionType::ZSTD)),
+            "none" => Ok(None),
+            other => Err(format!("Invalid compression codec: {}", other)),
```
```suggestion
            other => Err(format!("Invalid compression codec: {other}")),
```
nit: In my other projects I use more strict lint rules and this is reported as a warning. It would be nice to be prepared if this lint rule is enabled here too one day.

ballista/core/src/utils.rs
```diff
@@ -177,6 +177,23 @@ pub fn default_config_producer() -> SessionConfig {
...
+    config: Arc<BallistaConfig>,
+) -> std::result::Result<IpcWriteOptions, DataFusionError> {
+    let compression_codec = config
+        .shuffle_compression_codec()
+        .map_err(DataFusionError::Configuration)?;
+
+    let options = IpcWriteOptions::default()
+        .try_with_compression(compression_codec)
+        .map_err(|err| {
+            DataFusionError::Internal(format!("Failed to set compression codec: {}", err))
```
martin-g:
```suggestion
            DataFusionError::Internal(format!("Failed to set compression codec: {err}"))
```

ballista/core/src/config.rs
```diff
@@ -585,6 +597,21 @@ impl BallistaConfig {
         self.get_bool_setting(BALLISTA_PROPAGATE_EMPTY_ENABLED)
     }
 
+    /// Returns compression codec that will be used during write stage of shuffle
+    pub fn shuffle_compression_codec(
+        &self,
+    ) -> std::result::Result<Option<CompressionType>, String> {
+        match self
+            .get_string_setting(BALLISTA_SHUFFLE_COMPRESSION_CODEC)
+            .as_str()
```
martin-g: nit: For better UX - trim + to_lowercase the value ?!

ballista/core/src/config.rs
```diff
@@ -334,6 +338,14 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
...
             Some("".to_string()),
         ),
+        ConfigEntry::new(
+            BALLISTA_SHUFFLE_COMPRESSION_CODEC.to_string(),
+            "Compression codec specification \
+            used in the shuffle process. Possible values: \
+            none, lz4, zstd. Defaults to lz4 to preserve current behaviour".to_string(),
+            DataType::Utf8,
+            Some("lz4".to_string()),
+        )
```
```suggestion
        ),
```

ballista/core/src/config.rs
martin-g: Also, I think it would be nice to list the supported values in the error message.

martin-g (commented):
https://github.com/sandugood/datafusion-ballista/blob/ddc95039e3c9a03d922e2897479d0eda9e2925e3/ballista/core/src/execution_plans/sort_shuffle/config.rs#L42 still hardcode LZ4

sandugood:
> https://github.com/sandugood/datafusion-ballista/blob/ddc95039e3c9a03d922e2897479d0eda9e2925e3/ballista/core/src/execution_plans/sort_shuffle/config.rs#L42 still hardcode LZ4

Yeah, thank you for this catch.
Now changed in the `SortShuffle` config to accept `Option<CompressionType>`, which it gets from the `BallistaConfig`

andygrove (approved):
Nice improvement. Thanks @sandugood

PR #2045, opened 2026-07-15
Rename `task_slots` -> `vcores` (executor spec + CLI)

avantgardnerio:
## Summary

Follow-up to review feedback on #2038: split the `task_slots` → `vcores` rename into its own PR so it can land independently.

Mechanical rename, no behavior change. Follows the YARN precedent (`yarn.nodemanager.resource.cpu-vcores`) and Spark's `spark.executor.cores`. The `v` prefix is self-documenting — readers grepping in a year don't have to hunt for a doc comment to know these aren't physical `nproc`. It also leaves room for future execution models that decouple "how many tasks" from "how many cores per task."

## What changed

- `ExecutorSpecification.task_slots` → `vcores` (u32), with a YARN/Spark doc comment
- Proto: `ExecutorResource.task_slots` → `vcores`, `AvailableTaskSlots` → `AvailableVcores`, `ExecutorTaskSlots` → `ExecutorVcores`, `PollWorkParams.num_free_slots` → `num_free_vcores`. Field numbers preserved for wire compat.
- CLI: `--concurrent-tasks` → `--vcores` on both `ballista-executor` and `ballista-cli`
- Cluster bookkeeping: `InMemoryClusterState.task_slots` → `available_vcores`
- TUI: "Task Slots" column → "vCores"
- Docs, `docker-compose.yml`, tpch workflow, and CLI README updated so `--concurrent-tasks` invocations don't fail at boot under the new flag name
- Historical changelogs left untouched

## Test plan

- [x] \`cargo check --workspace --all-targets\`
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\`
- [x] \`cargo test -p ballista-scheduler --lib\` (224 tests)
- [x] \`cargo test -p ballista-core --lib\` (116 tests)
- [x] Proto codegen produces no diff after build

### DIFF
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -419,9 +419,14 @@ message ExecutorSpecification {
   repeated ExecutorResource resources = 1;
 }
 
+// Virtual cores this executor has been assigned at runtime — analogous to
+// YARN vcores (yarn.nodemanager.resource.cpu-vcores) or Spark's
+// spark.executor.cores. Not physical `nproc`: may over- or under-subscribe.
+// One task per executor drives this many DataFusion partitions in parallel
+// through one plan-Arc.
 message ExecutorResource {
   oneof resource {
-    uint32 task_slots = 1;
+    uint32 vcores = 1;
   }
 }
 
@@ -437,13 +442,16 @@ message ExecutorOperatingSystemSpecification {
   uint64 open_files_limit = 9;
 }
 
-message AvailableTaskSlots {
+// One instance per executor. Tracks free vcores for scheduler-side task
+// binding. See ExecutorResource for the vcore definition (YARN vcores or
+// Spark's spark.executor.cores).
+message AvailableVcores {
   string executor_id = 1;
-  uint32 slots = 2;
-    }
+  uint32 vcores = 2;
+}
 
-message ExecutorTaskSlots {
-  repeated AvailableTaskSlots task_slots = 1;
+message ExecutorVcores {
+  repeated AvailableVcores vcores = 1;
 }
 
 message ExecutorData {
@@ -536,7 +544,7 @@ message TaskStatus {
 
 message PollWorkParams {
   ExecutorRegistration metadata = 1;
-  uint32 num_free_slots = 2;
+  uint32 num_free_vcores = 2;
   // All tasks must be reported until they reach the failed or completed state
   repeated TaskStatus task_status = 3;
 }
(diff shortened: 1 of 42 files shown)


andygrove:
@lukekim fyi

andygrove:
This LGTM because it is a more generic approach to advertising executor compute capacity and it is not tied into any specific task/partition execution model.

I'd like to see feedback from other contributors/users before approving, since technically it is a breaking change.

ballista/core/proto/ballista.proto
```diff
@@ -437,13 +442,16 @@ message ExecutorOperatingSystemSpecification {
   uint64 open_files_limit = 9;
 }
 
-message AvailableTaskSlots {
+// One instance per executor. Tracks free vcores for scheduler-side task
+// binding. See ExecutorResource for the vcore definition (YARN vcores or
+// Spark's spark.executor.cores).
+message AvailableVcores {
```
phillipleblanc, line 448: could we leave the old messages in for at least one release and mark them as deprecated and silently convert them? Upgrading a cluster in place is not going to be possible if we don't. Then in the next next release we can remove it.

phillipleblanc (approved):
I think this makes sense, my one concern would be I think we should keep the ability to gracefully upgrade existing cluster by having this release understand both protobuf message types, and then in the release after we can remove the old messages.

ballista/core/proto/ballista.proto
avantgardnerio: Thanks for flagging — I looked into this and I think binary wire compatibility is actually preserved (numbers are preserved, names shouldn't go over the wire, IIUC). I'm scanning for other places where the PR could have made breaking changes and will address them one by one.

avantgardnerio:
@phillipleblanc , thanks for the approval! I think the wire protocol will actually be okay for this one, but I added fallbacks for the CLI & API which should help. They work for this PR, but I'm not sure if they will work for more complex ones like [multi-partition-tasks](https://github.com/apache/datafusion-ballista/pull/2038). 

So I've filed [an issue](https://github.com/apache/datafusion-ballista/issues/2071) where we can discuss what Ballista's live upgrade strategy might be.

docs/source/contributors-guide/user-personas.md
```diff
@@ -83,7 +83,7 @@ to make sure a change that delights one audience does not quietly break another.
 - **Coming from**: Apache Spark.
 - **Why Ballista**: Keep the Spark mental model — plans split into **stages** at
   shuffle boundaries, one **task** per partition, shuffle files, executors with
-  task slots, and adaptive query execution (**AQE**) for runtime adaptivity — on
+  vcores, and adaptive query execution (**AQE**) for runtime adaptivity — on
```
andygrove: I'm not sure that we should update the persona documentation for this change?

andygrove (approved):
LGTM - with one request to revert the change to the user persona documentation. I agree that the wire format should be unchanged. Thanks @avantgardnerio!

Issue #2047, opened 2026-07-16
Adaptive (AQE) execution panics with `EmptyExec invalid partition` on many TPC-DS queries


andygrove:
## Describe the bug

With adaptive execution enabled (`ballista.planner.adaptive.enabled=true`), a large set of TPC-DS queries fail with the same internal assertion:

```
DataFusionError(Internal("Assertion failed: partition < self.partitions:
  EmptyExec invalid partition N (expected less than 1)"))
```

The consumer requests partition `N` (observed 1–7) from an `EmptyExec` that was built with a single partition (`expected less than 1`). The **same queries succeed under the static planner** (AQE off), so this is specific to the adaptive planner's partitioning of an `EmptyExec` (or of the stage that consumes it).

## Affected queries (SF1)

15 queries fail with this exact assertion:

```
q1 q8 q16 q24 q29 q34 q37 q41 q43 q44 q46 q54 q61 q68 q73
```

(partition index in the message varies per query: q1→1, q8→3, q29→4, q37→5, q44→6, q46→7, …)

## To Reproduce

Generate SF1 TPC-DS Parquet, start a scheduler + executor, then:

```
tpcds --host localhost --port 50050 --path <data> --partitions 16 --verify \
  --query 1 -c datafusion.optimizer.prefer_hash_join=false \
  -c ballista.planner.adaptive.enabled=true
# -> Job failed due to stage N failed: ... EmptyExec invalid partition 1 (expected less than 1)
```

The same query with `ballista.planner.adaptive.enabled` unset (static planner) succeeds.

## Expected behavior

Adaptive execution runs these queries to completion, as the static planner does.

## Additional context

Reproduced at SF1 on a 1-scheduler / 1-executor native cluster (8 slots, 16 partitions), base `apache/main` @ f45996cf.

Two other AQE-only failures were seen in the same sweep and look **separate** from this assertion (not included above):
- **q72** — `ResourcesExhausted` on `HashJoinInput` (memory-pool exhaustion under the adaptive broadcast path).
- **q23, q77** — hang (>180s), consistent with a cross-stage dynamic-filter deadlock.

Part of the adaptive-execution epic #1359; distinct from #1826 (TPC-DS Q72 AQE `OffsetOverflowError`). Surfaced by the TPC-DS correctness harness proposed in the PR for #1845.

Issue #2046, opened 2026-07-16
Distributed execution returns incorrect results for several TPC-DS queries (static planner)


andygrove:
## Describe the bug

Under the **default (static) planner**, Ballista returns results for several TPC-DS queries that differ from single-process DataFusion executing the same SQL over the same data. The differences are stable across repeated runs (not query non-determinism), and single-process DataFusion returns the same answer under both `prefer_hash_join=true` and `prefer_hash_join=false`, so this is not a join-strategy artifact — it points to a distributed-execution correctness problem.

Surfaced by a TPC-DS correctness cross-check (SF1) that runs each query on a Ballista cluster and diffs the result against a single-process DataFusion oracle over the same Parquet data (the mechanism added by the TPC-H `--verify` gate, extended to TPC-DS).

## Confirmed cases (SF1, `target_partitions=16`, `prefer_hash_join=false`)

| Query | Shape | Single-process DataFusion | Ballista (distributed) |
|-------|-------|---------------------------|------------------------|
| q38 | `count(*)` over `INTERSECT` of 3 channel customer sets | **104** | **0** |
| q87 | `count(*)` over `EXCEPT` | **46753** | **47011** |
| q78 | `… ORDER BY … LIMIT 100` | **100 rows** | **73 rows** (27 dropped) |
| q4  | year-over-year CTE ratios, `LIMIT 100` | **8 rows** | **0 rows** |

- **Stable, not ties:** q38/q78/q87 reproduce identically across repeated Ballista runs, and single-process DataFusion is stable too. (Queries whose results genuinely vary run-to-run due to `LIMIT`/`ORDER BY` ties are a separate matter and are excluded from this report.)
- **Not a join-config artifact:** single-process DataFusion returns the correct answer under **both** `prefer_hash_join=true` and `false` at `target_partitions=16` (verified with `datafusion-cli`). Only distributed vs single-process differs.

The `INTERSECT`/`EXCEPT` (q38, q87) and `LIMIT` row-drop (q78) symptoms suggest the set-operation and limit lowering across shuffle stages as the area to look at.

## To Reproduce

Generate SF1 TPC-DS Parquet, start a scheduler + executor, then for the distributed result:

```
tpcds --host localhost --port 50050 --path <data> --partitions 16 --verify \
  --query 38 -c datafusion.optimizer.prefer_hash_join=false
# -> Query 38 VERIFY MISMATCH: expected 104, got 0
```

Single-process baseline (both join modes give 104):

```
[...]

PR #2048, opened 2026-07-16
Add TPC-DS correctness testing

### LINKED ISSUE #1845: Add TPC-DS workflow

andygrove:

Closes #1845.

# Rationale for this change

Ballista has a TPC-H correctness gate in CI (`--verify` cross-checks every query against single-process DataFusion over the same data), but no equivalent for TPC-DS. TPC-DS exercises a much wider surface — 24 tables, 99 queries with correlated subqueries, set operations, rollups, and window functions — so it reaches distributed-execution paths TPC-H does not.

Adding a self-validating TPC-DS gate extends coverage with no golden-answer files to maintain. It already earned its keep: the initial SF1 sweep surfaced real distributed-execution bugs, now filed as #2046 (results diverge from single-process DataFusion on q4/q38/q78/q87 under the static planner) and #2047 (the adaptive planner panics with an `EmptyExec invalid partition` assertion on many queries).

- **Shared harness library** (`benchmarks/src/lib.rs`): the benchmark-agnostic result-comparison/oracle helpers are extracted from `benchmarks/src/bin/tpch.rs` and reused by both binaries. TPC-H behavior is unchanged.
- **Data + queries**: `benchmarks/tpcds-gen.sh` generates SF1 TPC-DS Parquet via a git-pinned `tpcgen-cli`; `dev/vendor-tpcds-queries.sh` vendors the 99 DataFusion query files into `benchmarks/queries-tpcds/`.
- **Runner** (`benchmarks/src/bin/tpcds.rs`): runs each non-skipped query on a Ballista cluster and, with `--verify`, diffs the result against a single-process DataFusion oracle over the same data. It records every per-query failure and exits non-zero if any query fails.
- **CI** (`.github/workflows/tpcds.yml`): runs the gate at SF1 under the default (static) planner. Under the static planner the gate is green: 86/99 verified, 13 skipped.
- **Skip list**: `SKIP` in `tpcds.rs` documents the excluded queries by cause — the #2046 divergences, genuinely non-deterministic queries (LIMIT/ORDER BY ties), and 5 queries whose `tpcgen-cli` schema column names differ from the DataFusion query text (not a Ballista issue).
- **Docs**: a TPC-DS section in `benchmarks/README.md`.

The adaptive planner (AQE on) is intentionally not run in CI yet because of #2047; re-adding it is a follow-up once that is fixed.

Tested: `cargo fmt`/`clippy --all-targets --workspace -D warnings` clean; `ballista-benchmarks` tests pass (including the unchanged TPC-H `--verify` tests); the full SF1 gate runs green end-to-end on a native 1-scheduler/1-executor cluster (86 queries verified against DataFusion).

A new `tpcds` benchmark binary, a new CI workflow, and benchmark documentation. No changes to public Ballista APIs.


benchmarks/README.md
```diff
@@ -298,4 +298,84 @@ $SPARK_HOME/bin/spark-submit \
...
+## TPC-DS Correctness Tests
+
+Unlike the TPC-H suite above (which measures performance), the TPC-DS suite is
+a correctness gate: it runs each query on a Ballista cluster and compares the
+result, row-by-row, against a single-process DataFusion oracle running the
+same query.
+
+### Vendoring the queries
+
+The 99 TPC-DS queries are not stored in this repo; they are vendored from
```
phillipleblanc: They are stored in the repo :D

phillipleblanc: Maybe you meant to gitignore them?

phillipleblanc (approved):

benchmarks/README.md
andygrove: Claude created this PR while I was sleeping! I will review it soon and mark ready for review after that.

andygrove: Fixed. I appreciate the review @phillipleblanc! I will merge this PR when CI is green and then start work on fixing the issues this uncovered.

phillipleblanc (approved):
Thanks!

PR #2057, opened 2026-07-16
refactor(tui): Drop the header banner/logo

### LINKED ISSUE #2013: TUI banner renders as garbled glyphs on most fonts

martin-g:

Closes #2013.

# Rationale for this change

The banner was using tui-big-text crate that seems to use too modern UTF-8 symbols which are not yet available for many operating systems and because of this the banner didn't render well.

Remove the code related to the header banner (TUI) and logo (WebTUI)

Only UI.


martin-g:
https://github.com/user-attachments/assets/166cc729-d1e5-49cb-9648-d2040de53e34

phillipleblanc (approved):

andygrove (approved):
LGTM. Thanks @martin-g

PR #2059, opened 2026-07-16
feat(scheduler): Redirect '/' to nighlies.a.o WebTUI

martin-g:

N/A

# Rationale for this change

Opening http://localhost:50050 in a browser will now redirect to the convenience deployment of the WebTUI app at
https://nightlies.apache.org/datafusion/ballista/tui/...

Add a new GET handler for the root (`/`) that redirects to the convenience deployment of the WebTUI app at https://nightlies.a.o

No API changes!
New GET endpoint is added.


phillipleblanc (approved):

milenkovicm:
Can we override endpoint using configuration?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -227,6 +229,48 @@ pub enum PlanFormat {
...
+
+    for (k, v) in params.iter() {
+        query_string.push_str(&format!(
+            "&{}={}",
+            url_escape::encode_query(k),
+            url_escape::encode_query(v)
+        ));
+    }
+
+    let target = format!("{NIGHTLIES_URL}/{BALLISTA_VERSION}/?{query_string}");
```
milenkovicm, line 282: it would be great if we could make target configurable and forward to NIGHTLIES_URL only if there is no such configuration, this would allow users to have its own version of tui deployed somewhere else

ballista/scheduler/src/api/handlers.rs
```diff
@@ -227,6 +229,48 @@ pub enum PlanFormat {
...
+/// A handler for GET requests to the root (`/`).
+/// It redirects to `https://nightlies.apache.org/datafusion/ballista/tui/<BALLISTA_VERSION>/`
+/// forwarding any query parameters
+pub async fn get_root(
+    header_map: HeaderMap,
+    Query(mut params): Query<HashMap<String, String>>,
+) -> Result<Redirect, (StatusCode, String)> {
+    const NIGHTLIES_URL: &str = "https://nightlies.apache.org/datafusion/ballista/tui";
+
+    let ballista_scheduler_url =
```
milenkovicm, line 250: should we also take into account `external_host` from scheduler configuration g?

milenkovicm (commented):
thanks @martin-g, it looks good, but it does redirect but for some reason im getting scheduler down when i run with default configuration, not sure why

martin-g:
> Can we override endpoint using configuration?

Done with https://github.com/apache/datafusion-ballista/pull/2059/commits/327979a09825789503490eb5d3e76519c7567709

martin-g:
> thanks @martin-g, it looks good, but it does redirect but for some reason im getting scheduler down when i run with default configuration, not sure why

You can find the error in the browser Dev Tools > Console (or Network) tab.

By default the Scheduler web server allows Ajax requests from localhost:8080 and https://nightlies.apache.org:
https://github.com/apache/datafusion-ballista/blob/fbd8525bc6da6ffe2d6dc8f426678b246c7df522/ballista/scheduler/src/api/routes.rs#L88-L89

So, unless you overwrite this with `--cors-allowed-origins=...` when starting the Scheduler this shouldn't be the reason.

ballista/scheduler/src/api/handlers.rs
martin-g: If they have their own deployment then they do not need to use `http://localhost:50050/...` at all. They could just point to their deployment.

ballista/scheduler/src/api/handlers.rs
martin-g: Done with aed9569f

milenkovicm:
for some reason this does not work for me, i'm not sure what am i doing wrong. I'm starting scheduler with"

```bash
cargo run --bin  ballista-scheduler  --release --
```

i'm opening 

```
http:/localhost:50050/
```

which redirects me to


```
https://nightlies.apache.org/datafusion/ballista/tui/54.0.0/?ballista_scheduler_url=http://localhost:50050
```

image


i just dont get it

milenkovicm:
```
http://localhost:50050/api/state
```

is open, i can connect to it

milenkovicm:
I'm on macos, and safari

milenkovicm:
does it work for you @phillipleblanc ?

milenkovicm:
```bash
curl -v -X OPTIONS http://127.0.0.1:50050/api/state

*   Trying 127.0.0.1:50050...
* Connected to 127.0.0.1 (127.0.0.1) port 50050
> OPTIONS /api/state HTTP/1.1
> Host: 127.0.0.1:50050
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< access-control-allow-methods: GET,OPTIONS,PATCH
< access-control-allow-headers: *
< access-control-allow-origin: *
< allow: GET,HEAD
< content-length: 0
< date: Mon, 03 Aug 2026 17:24:48 GMT
<
* Connection #0 to host 127.0.0.1 left intact
```

i've added ` access-control-allow-origin: *` and still

phillipleblanc:
@milenkovicm this looks like it might be a Safari only restriction (https://bugs.webkit.org/show_bug.cgi?id=171934) - does it work for you in another browser?

It is working for me in Edge:
Screenshot 2026-08-05 at 12 06 30 PM

milenkovicm (approved):
thanks @martin-g and @phillipleblanc

PR #2061, opened 2026-07-16
fix: spill sort shuffle writer when the memory pool rejects a reservation grow

### LINKED ISSUE #2031: Executor can be OOM-killed because memory accounting misses untracked allocations
Ballista's executor memory accounting relies entirely on voluntary DataFusion `MemoryPool` reservations. `executor_process.rs` splits `--memory-pool-size` across `--concurrent-tasks` and gives each task a `FairSpillPool`, but that pool only sees bytes an operator explicitly reserves through a `MemoryReservation`.

It does not see Arrow buffer allocations made outside a reservation, join scratch space, expression-kernel temporaries, or shuffle read/write and gRPC framing buffers. Real RSS therefore runs well above what the pool believes is reserved.

Two problems follow.

**1. The executor process gets OOM-killed.** This is much worse in Ballista than a single failed task: the dead executor takes *all of its shuffle output files* with it, so every downstream stage that was going to read from it raises `FetchPartitionError`, which cascades into stage rollbacks and re-runs of the map stages — potentially across other concurrent jobs on the same cluster.

**2. The pool spills too late, or not at all,** because it is measuring the wrong quantity. The only remedy today is to hand-tune `--memory-pool-size` down to some workload-specific fraction of the container limit.

Separately, and independent of the above: a task that *does* exhaust its budget raises `DataFusionError::ResourcesExhausted`, which `From<BallistaError> for FailedTask` currently maps to `FailedReason::ExecutionError`. The scheduler fails the whole stage on that reason with no retry, so a transient memory spike kills the job outright.

### Proposed approach

Track the bytes the global allocator actually hands out, and use that signal to (a) reject memory-pool growth before real usage exceeds the budget, so DataFusion spills instead of OOMing, and (b) as a last resort, fail a single task rather than let the process die. Classify `ResourcesExhausted` as a retriable task failure so the scheduler reschedules the task.

Prior art: apache/datafusion-comet#4582 explores the same idea for Comet.

### LINKED ISSUE #2060: Sort shuffle writer ignores memory-pool rejection and never spills under pool pressure
`SortShuffleWriterExec` registers its `MemoryConsumer` with `.with_can_spill(true)`, but discards the result of the reservation grow ([`writer.rs`](https://github.com/apache/datafusion-ballista/blob/main/ballista/core/src/execution_plans/sort_shuffle/writer.rs)):

```rust
// Mirror the growth in the runtime pool reservation so the pool
// sees this writer's memory usage. Best-effort: if the pool is
// bounded and rejects the grow, that's fine — the absolute
// counter below still triggers a spill.
let _ = reservation.try_grow(growth);
buffered_bytes = buffered_bytes.saturating_add(growth);

if buffered_bytes >= memory_limit {
    // spill
}
```

The comment argues the rejection is safe to ignore because `buffered_bytes` still bounds RSS. That reasoning does not hold: `buffered_bytes` is a private per-task counter compared against `memory_limit_per_task_bytes`. It knows nothing about the state of the shared `MemoryPool`, so a rejected grow produces **no spill at all**.

Two consequences follow.

**1. The writer under-reports its usage to the shared pool.** The batch has already been pushed into `BufferedBatches` by this point. When the grow is rejected, the writer holds that memory while the pool has no record of it, so other consumers sharing the pool see bytes as available that are not. This is the same class of accounting gap as #2031, but self-inflicted rather than caused by an untracked allocator.

**2. Pool pressure produces no spill.** Registering with `can_spill(true)` is a contract: the pool rejects a grow to ask the consumer to spill. This writer declines to respond, so the only thing that can ever trigger a spill is its own per-task budget.

This directly undermines the approach proposed in #2031, whose remedy is to *\"reject memory-pool growth before real usage exceeds the budget, so DataFusion spills instead of OOMing\"*. That guard cannot work on this operator: the writer buffers straight through a rejection.

### Reproduction

Give the writer a per-task budget far above the payload (so the private counter can never fire) and a pool too small to admit a single batch, so every grow is rejected. The writer spills zero times and buffers the entire input.

### Proposed fix

Treat a rejected grow as a spill trigger alongside the existing per-task budget. Spilling flushes the buffered batches to disk and frees the reservation, so no retry is needed — the writer holds nothing afterwards.

andygrove:

Closes #2060.

Related to #2031 — see the rationale below.

 # Rationale for this change

`SortShuffleWriterExec` registers its `MemoryConsumer` with `.with_can_spill(true)` but then discards the result of the grow:

```rust
// Best-effort: if the pool is bounded and rejects the grow, that's fine —
// the absolute counter below still triggers a spill.
let _ = reservation.try_grow(growth);
```

The comment's reasoning does not hold. `buffered_bytes` is a private per-task counter compared against `memory_limit_per_task_bytes`; it knows nothing about the shared `MemoryPool`, so a rejected grow produces no spill at all. The batch has already been pushed into `BufferedBatches` at that point, so the writer holds memory the pool has no record of — other consumers on the same pool then see bytes as free that are not.

Registering with `can_spill(true)` is a contract: the pool rejects a grow to ask the consumer to spill. This writer declines to answer.

This also matters for #2031, whose proposed remedy is to *"reject memory-pool growth before real usage exceeds the budget, so DataFusion spills instead of OOMing"*. That guard cannot work on this operator while it buffers straight through a rejection. This PR does not close #2031 — it does not add allocator-level tracking or make `ResourcesExhausted` retriable — but it is a prerequisite for the rejection-based half of it to have any effect here.

Treat a rejected grow as a spill trigger alongside the existing per-task budget:

```rust
let pool_rejected_growth = reservation.try_grow(growth).is_err();
buffered_bytes = buffered_bytes.saturating_add(growth);

if pool_rejected_growth || buffered_bytes >= memory_limit {
```

No retry of the grow is needed after spilling. `push_batch` appends to `batches` unconditionally, so at the point of rejection the batch is always spillable; `spill_all_partitions` flushes it and calls `reservation.free()`, leaving the writer holding nothing.

Tests: adds `spills_when_memory_pool_rejects_growth`, which sets the per-task budget to 1 GiB (far above the payload, so the private counter can never fire) against a 64 KiB pool that cannot admit even one ~128 KiB batch — making pool rejection the only possible spill signal. It fails on `main` with `spill_count=0` and passes here. A `pool_bytes` parameter is threaded through the existing `run_round_trip` helper rather than duplicating it, so the new test also inherits its assertions that every input key survives the spill path and that `pool.reserved() == 0` at the end.

No API changes. Behaviourally, a sort shuffle writer running against a contended bounded pool now spills where it previously buffered without limit, which is the intended behaviour for a consumer registered with `can_spill(true)`.

### DIFF
--- a/ballista/core/src/execution_plans/sort_shuffle/writer.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/writer.rs
@@ -273,13 +273,16 @@ impl SortShuffleWriterExec {
                 growth += after.saturating_sub(before);
 
                 // Mirror the growth in the runtime pool reservation so the pool
-                // sees this writer's memory usage. Best-effort: if the pool is
-                // bounded and rejects the grow, that's fine — the absolute
-                // counter below still triggers a spill.
-                let _ = reservation.try_grow(growth);
+                // sees this writer's memory usage. A rejected grow means the
+                // pool is under pressure and has *not* accounted for the batch
+                // just buffered, so spill instead of holding memory the pool
+                // believes is free. Spilling flushes that batch to disk and
+                // frees the reservation, which is the response a spillable
+                // consumer owes the pool.
+                let pool_rejected_growth = reservation.try_grow(growth).is_err();
                 buffered_bytes = buffered_bytes.saturating_add(growth);
 
-                if buffered_bytes >= memory_limit {
+                if pool_rejected_growth || buffered_bytes >= memory_limit {
                     let spill_timer = metrics.spill_time.timer();
                     let (event_batches, event_bytes) = spill_all_partitions(
                         &mut buffered,
@@ -872,7 +875,8 @@ mod tests {
     /// `rows_per_batch` rows each with schema `(k: Int64, v: Int64)`, writes
     /// them through `SortShuffleWriterExec` into `num_partitions` output
     /// partitions with a per-task buffered-bytes budget of
-    /// `sort_shuffle_memory_limit_bytes`, reads every partition back via
+    /// `sort_shuffle_memory_limit_bytes` against a `FairSpillPool` of
+    /// `pool_bytes`, reads every partition back via
     /// `stream_sort_shuffle_partition`, and asserts:
     ///   - Every input key `0..total_rows` is present exactly once in the union
     ///     of partition outputs.
@@ -883,6 +887,7 @@ mod tests {
         rows_per_batch: usize,
         num_partitions: usize,
         sort_shuffle_memory_limit_bytes: usize,
+        pool_bytes: usize,
         expect_spills: bool,
     ) -> Result<()> {
         use super::super::reader::stream_sort_shuffle_partition;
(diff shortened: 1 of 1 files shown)


andygrove:
@phillipleblanc could you review?

phillipleblanc (approved):
This makes sense to me.

PR #2066, opened 2026-07-16
fix: restore input ordering a SortMergeJoinExec requires after lowering

### LINKED ISSUE #2065: Broadcast join promotion drops a sort order required by a parent SortMergeJoinExec, silently producing wrong results
## Describe the bug

Ballista's broadcast-join promotion rewrites a `SortMergeJoinExec` into a
`HashJoinExec(CollectLeft)`. A `HashJoinExec` does not preserve its input's
sort order, but nothing re-establishes that order afterwards. When the promoted
join feeds a **parent `SortMergeJoinExec`**, the parent then runs on unsorted
input.

A merge join on unsorted input does not error — it silently stops matching once
the merge cursors diverge, so **rows are dropped and the query returns a wrong
answer**.

This is distinct from #1055 (which is about which *join types* are safe to
broadcast, and is already guarded by `collect_left_broadcast_safe`). Here the
join type is fine; the *ordering contract* is what breaks.

## Evidence

In the distributed plan below (TPC-DS q78, SF1, `prefer_hash_join=false`), the
outer `SortMergeJoinExec` has a `SortExec` on its right input but **nothing on
its left** — the left arrives straight from a `HashJoinExec(CollectLeft)` that
replaced the inner SMJ:

```
SortMergeJoinExec: join_type=Left, on=[(ss_sold_year, cs_sold_year), (ss_item_sk, cs_item_sk), (ss_customer_sk, cs_customer_sk)]
  ProjectionExec: ...
    ProjectionExec: ...
      HashJoinExec: mode=CollectLeft, join_type=Right, on=[(ws_sold_year, ss_sold_year), ...]   <-- left input: no ordering
        UnresolvedShuffleExec: stage=8, broadcast=true, upstream_partitions: 16
        ProjectionExec: ...
          AggregateExec: mode=FinalPartitioned, gby=[d_year, ss_item_sk, ss_customer_sk], ...
            UnresolvedShuffleExec: stage=15, partitioning: Hash([d_year, ss_item_sk, ss_customer_sk], 16)
  SortExec: expr=[cs_item_sk ASC, cs_customer_sk ASC], preserve_partitioning=[true]              <-- right input: sorted
    ...
```

Before the rewrite the ordering came from a `SortExec` feeding the inner SMJ's
`ss` side; promoting that SMJ to a broadcast hash join removed the need for
that sort, and the parent SMJ's requirement went unmet.

## To Reproduce

SF1 TPC-DS, scheduler + executor, `target_partitions=16`. Using the q78 CTEs
(`ss`, `ws`, `cs`):

```sql
select ss_sold_year, ss_item_sk, ss_customer_sk, ss_qty from ss
left join ws on (ws_sold_year=ss_sold_year and ws_item_sk=ss_item_sk and ws_customer_sk=ss_customer_sk)
left join cs on (cs_sold_year=ss_sold_year and cs_item_sk=ss_item_sk and cs_customer_sk=ss_customer_sk)
where (coalesce(ws_qty,0)>0 or coalesce(cs_qty,0)>0) and ss_sold_year=2000;
-- Ballista: 73 rows.  DataFusion: 223 rows.
```

Ballista's own breakdown of that set is `total=223, ws_pos=73, cs_pos=150`: the
150 rows that match only via `cs` are exactly the ones lost, i.e. the outer
`ss ⋈ cs` join stops matching.

Disabling the promotion restores the correct answer, which isolates the cause:

| Config | Result |
|---|---|
| default (broadcast on) | 73 rows — **wrong** |
| `-c ballista.optimizer.broadcast_sort_merge_join_enabled=false` | 223 rows — correct |
| `-c ballista.optimizer.broadcast_join_threshold_bytes=0` | 223 rows — correct |
| full q78 + `broadcast_join_threshold_bytes=0` | 100 rows — **verifies OK** vs DataFusion |

Note the wrong answer does **not** depend on `ORDER BY` or `LIMIT`; it
reproduces with neither. (q78 was originally filed under #2046 as "distributed
LIMIT drops rows" — that framing is a red herring. Wrapping the query in
`count(*)` changes the plan enough to hide the bug, which is what made it look
limit-related.)

## Expected behavior

Promoting a join must not silently drop an ordering that a parent operator
requires. Either re-establish the ordering (insert the `SortExec` the parent
needs) after the rewrite, or decline to promote when the join's output ordering
is required upstream.

It would also be worth asserting, over the planned stages, that every
`SortMergeJoinExec` input satisfies the join's `required_input_ordering()` — an
unsorted SMJ input is always a bug, and today it fails silently and produces
wrong results rather than erroring.

## Additional context

- Reproduced at SF1 on a 1-scheduler / 1-executor native cluster (8 slots, 16
  partitions), on `upstream/main` (e843c8e8) under the default static planner.
- Surfaced by the TPC-DS correctness gate (#1845 / #2048); accounts for the q78
  half of #2046. q4 (also in #2046) is not yet diagnosed.
- Related: #1055 (broadcast join-*type* safety), #2006 (propagating ordering
  across pass-through shuffles).

andygrove:

Closes #2065.

Also resolves the q4 / q78 half of #2046 (the set-operation half, q38 / q87, is
a separate DataFusion-side bug handled in #2064).

 # Rationale for this change

Lowering a plan into stages rewrites nodes, and a rewrite can drop an ordering
its parent depends on. Ballista's broadcast promotion turns a
`SortMergeJoinExec` into a `HashJoinExec(CollectLeft)`, which does not preserve
input order — but nothing re-establishes that order, so a parent
`SortMergeJoinExec` ends up merging unsorted input.

A merge join given unsorted input does not error. It stops matching once the
merge cursors diverge, so rows are **silently dropped** and the query returns a
wrong answer.

In the q78 plan the outer `SortMergeJoinExec` had a `SortExec` on its right
input and nothing on its left, because the inner SMJ that used to supply that
ordering had been replaced by a broadcast hash join:

```
SortMergeJoinExec: join_type=Left, on=[(ss_sold_year, cs_sold_year), ...]
  ...
      HashJoinExec: mode=CollectLeft, join_type=Right, ...      <-- left input: no ordering
        UnresolvedShuffleExec: stage=8, broadcast=true, upstream_partitions: 16
        ...
  SortExec: expr=[cs_item_sk ASC, cs_customer_sk ASC], ...      <-- right input: sorted
```

Disabling the promotion isolated the cause — `broadcast_join_threshold_bytes=0`
or `broadcast_sort_merge_join_enabled=false` both restore the correct answer.

This is distinct from #1055, which concerns which *join types* are safe to
broadcast (already guarded by `collect_left_broadcast_safe`). Here the join type
is fine; the ordering contract is what breaks.

- `restore_required_input_ordering`: after lowering rewrites a node's children,
  re-establish any `Hard` input ordering a child no longer carries. Sorts are
  only added, never removed, so an input that already satisfies its requirement
  is untouched.
- `validate_sort_merge_join_orderings`: verify every `SortMergeJoinExec` in the
  planned stages has inputs satisfying the join's required ordering, so this
  failure mode surfaces as an error instead of wrong results.
- Un-skip TPC-DS q4 and q78 in the correctness gate (86 → 88 queries).

`Soft` requirements are deliberately ignored — they are opportunistic and an
operator works without them.

## Verification

SF1, 1 scheduler / 1 executor, `target_partitions=16`:

| | Before | After |
|---|---|---|
| q78 | 73 rows vs 100 | verified OK (100 rows) |
| q4 | 0 rows vs 8 | verified OK (8 rows) |
| minimal repro (q78 CTEs, no ORDER BY/LIMIT) | 73 rows vs 223 | 223 rows |

Full SF1 TPC-DS gate: **88/88 verified against single-process DataFusion**,
under both `prefer_hash_join=false` and `prefer_hash_join=true`. The new
validation reports **0 false positives** across all 88 queries under both
settings, which is why it is a hard error rather than a debug assertion.

`cargo test --workspace`, `cargo clippy --all-targets --workspace -- -D
warnings`, and `cargo fmt --all` are clean.

## Note on the q78 framing

q78 was originally reported under #2046 as "distributed LIMIT drops rows". That
framing is a red herring: the wrong answer does not depend on `ORDER BY` or
`LIMIT` and reproduces with neither. Wrapping the query in `count(*)` changes
the plan enough to hide the bug, which is what made it look limit-related.

Yes — queries where a broadcast-promoted join feeds a sort-merge join now
return correct results instead of silently dropping rows. A plan that would hit
this now fails with an explicit error rather than producing a wrong answer. No
API changes.

### DIFF
--- a/ballista/scheduler/src/planner.rs
+++ b/ballista/scheduler/src/planner.rs
@@ -34,6 +34,9 @@ use ballista_core::{
 };
 use datafusion::arrow::datatypes::DataType;
 use datafusion::config::ConfigOptions;
+use datafusion::physical_expr::{
+    LexOrdering, LexRequirement, OrderingRequirements, PhysicalSortRequirement,
+};
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
 use datafusion::physical_optimizer::enforce_sorting::EnforceSorting;
 use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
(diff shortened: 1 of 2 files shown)


andygrove:
Moving this to draft while I think about this more. Maybe the current broadcast optimization for SMJ is using the wrong approach.

andygrove:
> Moving this to draft while I think about this more. Maybe the current broadcast optimization for SMJ is using the wrong approach.

Let's go ahead and review this. It fixes the correctness issue. I'll keep investigating a better long term solution.

avantgardnerio (approved):
Would an extra pass of EnforceSorting be a robust alternative?

andygrove:
> Would an extra pass of EnforceSorting be a robust alternative?

Good idea... I will open a separate PR with this approach so we can compare

andygrove:
On second thoughts, the SMJ "broadcast" optimization was just a bad idea. I created https://github.com/apache/datafusion-ballista/pull/2072 to remove it.

PR #2069, opened 2026-07-16
fix: enable the executor client pool by default

### LINKED ISSUE #2068: Executor client pool is disabled by default, so shuffle fetches exhaust ephemeral ports on any multi-executor cluster
## Describe the bug

The executor's Ballista client pool is **disabled by default** (`--client-ttl`
defaults to `0`, "no cache, connection will be disposed"). With no pool, the
shuffle-read path opens and discards a **new TCP connection per fetch**. On any
multi-executor cluster a single query exhausts the machine's entire ephemeral
port range, and the query — plus every query after it — fails with
`AddrNotAvailable`.

With one executor this is invisible: shuffle reads are served from local files,
so there are no remote fetches. It appears as soon as a second executor exists,
which is every real deployment.

## To Reproduce

SF1 TPC-DS parquet, 1 scheduler + **2** executors (defaults, i.e. no
`--client-ttl`):

```
ballista-executor -c 4 -p 50052 --bind-grpc-port 50062 --work-dir /tmp/wd1
ballista-executor -c 4 -p 50053 --bind-grpc-port 50063 --work-dir /tmp/wd2

tpcds --host 127.0.0.1 --port 50050 --path <data> --partitions 16 --verify \
  -c datafusion.optimizer.prefer_hash_join=false
```

Result — the run collapses at q14 and never recovers:

```
Query 14 FAILED to run in 10.463s: ... FetchFailed(..., "Error connecting to
Ballista scheduler or executor at http://127.0.0.1:50052: ... ConnectError("tcp
connect error", 127.0.0.1:50052, Os { code: 49, kind: AddrNotAvailable,
message: "Can't assign requested address" })")
Query 15 FAILED ...
Query 16 FAILED ... (every subsequent query cannot connect)
```

Final tally: **11 OK / 77 FAILED**. The scheduler and both executors stay alive
the whole time — nothing crashes, the host is simply out of ephemeral ports.

### Measurements

Sockets sampled during a single q14 run (macOS ephemeral range 49152–65535 =
**16384 ports**):

| | default (`--client-ttl 0`) | `--client-ttl 30` |
|---|---:|---:|
| peak TCP sockets | 16563 | 1618 |
| peak TIME_WAIT | 16343 | 11 |
| q14 | FAILED (`AddrNotAvailable`) | verified OK |
| full TPC-DS gate (2 executors) | 11 OK / 77 FAILED | **88 OK / 0 FAILED** |

One query consumes the entire ephemeral port range. TIME_WAIT dominating the
count is the tell: connections are opened and immediately dropped rather than
reused.

The TIME_WAIT sockets are overwhelmingly against the **executor flight ports**,
which pins it to the shuffle-read path rather than scheduler traffic:

```
TIME_WAIT by destination port:
   8315 50052   <- executor 1 flight
   8023 50053   <- executor 2 flight
      1 50050   <- scheduler grpc
```

## Cause

`ballista/executor/src/executor_process.rs`:

```rust
if opt.client_ttl > 0 {
    let client_pool = Arc::new(DefaultBallistaClientPool::with_eviction_thread(
        Duration::from_secs(opt.client_ttl),
    ));
    Arc::new(DefaultExecutionEngine::with_client_pool(client_pool))
} else {
    Arc::new(DefaultExecutionEngine::new())   // no pool
}
```

and `ballista/executor/src/config.rs`:

```rust
#[arg(long, default_value_t = 0, help = "Number of seconds established client
connection should be cached if not used (0 means no cache, connection will be
disposed).")]
pub client_ttl: u64,
```

Without a pool, `DefaultExecutionEngine::new()` leaves `client_pool: None`, so
`ShuffleReaderExec` never gets one attached and each remote fetch builds a fresh
client. This is at odds with the shuffle reader's own design intent, which
assumes connection reuse:

> Because total in-flight bytes stay <= the sized h2 window, the governor — not
> the 64 KB transport window — is the binding backpressure, **which is what makes
> multiplexing over few connections safe.**

The pooling machinery is all present and works; it is simply off unless the
operator knows to pass `--client-ttl`.

## Expected behavior

Connection pooling should be **on by default**. The default should not be a
configuration that cannot survive a two-executor cluster. Opting out (via
`--client-ttl 0`) can remain available.

## Additional context

- Reproduced on `upstream/main` (e843c8e8), SF1, 1 scheduler + 2 executors
  (`-c 4` each), default static planner, `prefer_hash_join=false`.
- Workaround today: start executors with `--client-ttl <seconds>` (e.g. `30`).
- Not scale-dependent — SF1 is enough. It is driven by fetch count
  (stages x partitions), so larger scale factors and partition counts make it
  worse.

andygrove:

Closes #2068.

 # Rationale for this change

`client_ttl` defaults to `0`, which disables the executor's Ballista client pool
entirely. With no pool, the shuffle-read path opens and discards a **new TCP
connection per fetch**, so a single shuffle-heavy query exhausts the host's
ephemeral port range and fails with `AddrNotAvailable` — as does every query
after it, until those sockets leave `TIME_WAIT`.

A single executor never shows this: shuffle reads are served from local files, so
there are no remote fetches. It appears as soon as a second executor exists,
which is every real deployment.

Measured on the SF1 TPC-DS correctness gate, 1 scheduler + 2 executors (`-c 4`
each), `--partitions 16`, everything else default:

| | before | after |
|---|---:|---:|
| peak TCP sockets during one query | 16563 | 1618 |
| peak `TIME_WAIT` | 16343 | 12 |
| q14 | FAILED (`AddrNotAvailable`) | verified OK |
| full gate | **11 OK / 77 FAILED** | **88 OK / 0 FAILED** |

The test host's ephemeral range is 16384 ports (49152–65535), so a single query
consumed all of it. The `TIME_WAIT` sockets were overwhelmingly against the
executor **flight** ports (8315 on :50052, 8023 on :50053, versus 1 on the
scheduler's :50050), which pins it to the shuffle-read path.

The pooling machinery already exists and works, and the shuffle reader is written
assuming connection reuse:

> Because total in-flight bytes stay <= the sized h2 window, the governor — not
> the 64 KB transport window — is the binding backpressure, **which is what makes
> multiplexing over few connections safe.**

It was simply off unless the operator knew to pass `--client-ttl`.

- Default `client_ttl` to 30 seconds (was 0) so the pool is on unless explicitly
  disabled. `--client-ttl 0` still opts out.
- Update the flag help and the `ExecutorProcessConfig::client_ttl` doc to say
  what `0` actually costs.
- Add a test pinning the default to a pooling value, so it cannot regress to the
  disabled state unnoticed.

30s is chosen to comfortably outlive the gaps between fetches within a job while
still releasing idle connections; the existing eviction thread handles reclaim.
Happy to take a different number if maintainers prefer.

## Verification

With 2 executors and **no** `--client-ttl` flag, the full SF1 TPC-DS gate is
**88/88 verified** against single-process DataFusion (previously 11 OK / 77
FAILED in the same configuration). `cargo test --workspace`, `cargo clippy
--all-targets --workspace -- -D warnings`, and `cargo fmt --all` are clean.

Yes — executors now cache idle client connections for 30s by default instead of
disposing each one. This makes multi-executor clusters work out of the box.
Operators who want the old behaviour can pass `--client-ttl 0`. No API changes.

### DIFF
--- a/ballista/executor/src/config.rs
+++ b/ballista/executor/src/config.rs
@@ -180,11 +180,12 @@ pub struct Config {
         help = "Max number of sessions whose shared runtime state (object-store clients, Parquet footer cache) is retained on the executor (LRU). 0 disables caching."
     )]
     pub session_runtime_cache_capacity: usize,
-    /// Number of seconds established client connection should be cached if not used (0 means no cache)
+    /// Number of seconds an established client connection is cached while idle
+    /// (0 disables the cache, so every shuffle fetch opens a new connection)
     #[arg(
         long,
-        default_value_t = 0,
-        help = "Number of seconds established client connection should be cached if not used (0 means no cache, connection will be disposed)."
+        default_value_t = 30,
+        help = "Number of seconds established client connection should be cached if not used (0 means no cache, connection will be disposed; a shuffle-heavy query can then exhaust the host's ephemeral ports)."
     )]
     pub client_ttl: u64,
 }
@@ -255,5 +256,21 @@ mod tests {
     fn parse_rejects_invalid() {
         assert!(parse_memory_pool_size("banana").is_err());
         assert!(parse_memory_pool_size("").is_err());
+    }
+
+    /// The client pool is what keeps shuffle fetches from opening a fresh TCP
+    /// connection each time; with it off, one shuffle-heavy query can exhaust
+    /// the host's ephemeral ports and take the executor down with it. Pin the
+    /// default on so it cannot regress to the disabled state unnoticed.
+    #[cfg(feature = "build-binary")]
+    #[test]
+    fn client_pool_is_enabled_by_default() {
+        use clap::Parser;
+        let opt = super::Config::parse_from(["ballista-executor"]);
+        assert!(
+            opt.client_ttl > 0,
+            "client_ttl must default to a pooling value, got {}",
+            opt.client_ttl
+        );
     }
 }
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -187,7 +187,10 @@ pub struct ExecutorProcessConfig {
     pub override_arrow_flight_service: Option<Arc<ArrowFlightServerProvider>>,
     /// Override function for customizing gRPC client endpoints before they are used
     pub override_create_grpc_client_endpoint: Option<EndpointOverrideFn>,
-    /// Number of seconds established client connection should be cached (0 means no cache)
+    /// Number of seconds an established client connection is cached while idle.
+    /// `0` disables the pool, so every shuffle fetch opens and drops its own
+    /// connection and a shuffle-heavy query can exhaust the host's ephemeral
+    /// ports.
     pub client_ttl: u64,
 }
 
@@ -238,7 +241,7 @@ impl Default for ExecutorProcessConfig {
             override_physical_codec: None,
             override_arrow_flight_service: None,
             override_create_grpc_client_endpoint: None,
-            client_ttl: 0,
+            client_ttl: 30,
         }
     }
 }


andygrove:
I'm going to go ahead and merge this when CI is green, since it is a blocker for much of the testing I am doing.

PR #2070, opened 2026-07-16
fix: map a task's partition through UnionExec when restricting file scans

### LINKED ISSUE #2067: UNION ALL silently returns inflated results when the union stage exceeds one wave of executor task slots
## Describe the bug

A `UNION ALL` over two file scans returns **massively inflated results** whenever
the union's stage has more tasks than the executor has concurrent task slots.
The query does not error — it silently over-counts, so aggregates come back
several times too large.

## To Reproduce

SF1 TPC-DS parquet, 1 scheduler + 1 executor, release build:

```sql
select round(sum(p),2) from
 (select ws_ext_sales_price p from web_sales
  union all
  select cs_ext_sales_price p from catalog_sales);
```

With `--partitions 16` and an executor started `-c 8`:

```
expected 5492796521.50   (single-process DataFusion, same data)
got     42072988115.00   (~8x inflated; varies run to run)
```

Summing each table separately is correct, and the same aggregate over a single
table (no `UNION ALL`) is correct — only the union is wrong.

Note `select count(*)` over the union is **not** a reproducer: it is answered
from parquet statistics without reading data. The aggregate must touch a column.

## The rule: it breaks exactly when the stage exceeds one wave of task slots

A `UnionExec` has as many partitions as its children combined, so an N-partition
union stage runs `2N` tasks. The result is correct **iff those tasks fit in a
single concurrent wave** on the executor:

Executor `-c 8`:

| `--partitions` | union stage tasks | result |
|---:|---:|---|
| 3 | 6 | correct |
| 4 | 8 | correct |
| 5 | 10 | **wrong** (12808834840.20) |
| 6 | 12 | **wrong** (16466853999.55) |
| 8 | 16 | **wrong** (16466853999.55) |

Same binary, same data, executor `-c 32` — the threshold moves with the slot
count, which isolates the cause:

| `--partitions` | union stage tasks | result |
|---:|---:|---|
| 5 | 10 | correct |
| 8 | 16 | correct |
| 16 | 32 | correct |
| 20 | 40 | **wrong** (20124873158.90) |

So: **correct iff `2 * partitions <= executor slots`.**

## Suspected cause

`restrict_scan_to_partition` (`ballista/executor/src/execution_engine.rs`) exists
precisely because DF54's `DataSourceExec` hands file groups out from a shared
work-queue that is only divided correctly when all partitions of a plan instance
are polled concurrently (#1907). It pins each task's scan to its own file group —
but bails out early:

```rust
if partition_id >= config.file_groups.len() {
    return None;   // no restriction -> scan left free to drain the whole queue
}
```

Its own doc comment anticipates this case:

> Returns `None` ... for a `partition_id` outside the source's file groups (e.g.
> when an operator between the scan and the stage output changed the partition
> count).

A `UnionExec` is exactly such an operator: with `target_partitions=16` each scan
has 16 file groups but the union stage has 32 partitions, so tasks 16..31 skip
the restriction and their scans are left unrestricted. An unrestricted scan only
behaves when the whole stage happens to be polled in one concurrent wave, which
is precisely the boundary measured above.

Fixing it presumably means mapping the stage's output partition to each scan's
*local* partition index (a union offsets its right child's partitions by the left
child's count) rather than using the stage `partition_id` directly, so every
scan is pinned to the right file group.

## Impact

- **16 of the 99 TPC-DS queries use `UNION ALL`** (q2, q4, q5, q11, q14, q23,
  q33, q54, q56, q60, q66, q71, q74, q76, q77, q80), so this is broadly reachable
  on any realistic partitions/slots ratio.
- It accounts for the q2 and q5 failures currently skipped in the TPC-DS
  correctness gate as "non-deterministic (ORDER BY ties)". That label is wrong:
  with a total order imposed and `LIMIT` removed, both still mismatch, so they
  are real wrong results, not tie ambiguity. (q31 and q71 *are* genuinely
  ordering-only — both verify OK once a total order is imposed.)
- Silent: nothing errors, results are just too large.

## Expected behavior

`UNION ALL` returns the same result as single-process DataFusion, independent of
`target_partitions` and executor slot count.

## Additional context

- Reproduced at SF1 on a 1-scheduler / 1-executor native cluster, `upstream/main`
  (e843c8e8), default static planner, under both `prefer_hash_join=true` and
  `false`.
- Regression surface introduced with the DataFusion 54 upgrade; closely related to
  #1907, whose fix covers the non-union case (a plain scan verifies correctly at
  any partition/slot ratio) but leaves partitions beyond `file_groups.len()`
  unrestricted.

andygrove:

Closes #2067.

 # Rationale for this change

A `UNION ALL` over file scans silently returned **inflated** results — no error,
just several times too much data:

```sql
select round(sum(p),2) from
 (select ws_ext_sales_price p from web_sales
  union all
  select cs_ext_sales_price p from catalog_sales);
-- expected 5492796521.50
-- got     49389026433.70   (--partitions 16, executor -c 8; varied run to run)
```

A task executes one partition of its stage, so `restrict_scan_to_partition` pins
each file scan beneath it to the single file group that partition reads.
Without that, DataFusion 54's shared work-queue lets a lone `execute(p)` drain
the queue and scan the whole table (#1907).

The pinning applied the stage's partition number **directly to every scan**,
which is only correct when the partition passes straight through. A `UnionExec`
concatenates its children's partitions: with N-partition children, stage
partition `N + k` is the right child's local partition `k`. So:

- for partitions inside the left child, the right child's scan was pinned to the
  wrong group (harmless — it is never polled), and
- for partitions `>= N`, `partition_id >= config.file_groups.len()` bailed out
  and returned `None`, leaving that scan **unrestricted** and free to read its
  entire table.

`restrict_scan_to_partition`'s own doc comment anticipated exactly this case:

> Returns `None` ... for a `partition_id` outside the source's file groups (e.g.
> when an operator between the scan and the stage output changed the partition
> count).

Confirmed by disabling the trigger from the other side: the query was correct
whenever the union stage's tasks fitted in a single wave of executor slots
(so nothing was left half-polled), and wrong beyond it. Moving the executor from
`-c 8` to `-c 32` moved the failure threshold from 8 to 32 tasks exactly.

- Add `restrict_scans_for_partition`, which walks the plan from the stage root
  and mirrors `UnionExec::execute`'s own partition arithmetic, so each scan is
  pinned to the group its **local** partition reads. Children a partition never
  reaches are left untouched — this task does not execute them. Every other
  operator passes the partition straight through, as before.
- Split the scan restriction out of the `ShuffleReaderExec` rewrite, which is
  partition-independent.
- Un-skip TPC-DS q2 and q5 in the correctness gate.

## Verification

The minimal case above is now correct at **every** partition count tried — 4, 5,
6, 8, 16, 20, 32 — where previously anything past one wave of task slots was
wrong.

Full SF1 TPC-DS gate: **88/88 verified** against single-process DataFusion, both
on a single executor and across 2 executors. `cargo test --workspace`, `cargo
clippy --all-targets --workspace -- -D warnings`, and `cargo fmt --all` are
clean. New unit tests pin the union mapping, including the case that previously
left a scan unrestricted.

## Note on q2 / q5

Both were skipped as "non-deterministic (ORDER BY ties; varies run-to-run)".
That diagnosis was wrong: with a total order imposed and `LIMIT` removed they
still mismatched, so they were real wrong results, not tie ambiguity — they were
hitting this bug. Both now verify in their original form across repeated runs.
q31 and q71 *are* genuinely ordering-only (both verify once a total order is
imposed) and stay skipped.

Yes — `UNION ALL` over file scans now returns correct results instead of
silently inflated ones. No API changes.

### DIFF
--- a/ballista/executor/src/execution_engine.rs
+++ b/ballista/executor/src/execution_engine.rs
@@ -36,6 +36,7 @@ use datafusion::error::{DataFusionError, Result};
 use datafusion::execution::context::TaskContext;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::metrics::MetricsSet;
+use datafusion::physical_plan::union::UnionExec;
 use datafusion::prelude::SessionConfig;
 use log::warn;
 use std::any::Any;
@@ -165,6 +166,65 @@ fn restrict_scan_to_partition(
     Some(DataSourceExec::from_data_source(config))
 }
 
+/// Restrict every file scan in `plan` to the file group the task running
+/// `partition` will actually poll.
+///
+/// A task executes exactly one partition of its stage, so each scan beneath it
+/// must be pinned to the one file group that partition reads (see
+/// [`restrict_scan_to_partition`]). Which group that is depends on the path from
+/// the stage root: most operators pass a partition straight through to their
+/// children, but a `UnionExec` concatenates its children's partitions, so its
+/// child `i` sees `partition` minus the partition counts of the children before
+/// it. Applying the stage's partition number directly to a scan under a union
+/// therefore addresses the wrong group, and for partitions past the scan's group
+/// count it addresses none at all — leaving that scan unrestricted and free to
+/// read the whole table.
+///
+/// Children a partition never reaches are left untouched: this task will not
+/// execute them.
+fn restrict_scans_for_partition(
+    plan: &Arc<dyn ExecutionPlan>,
+    partition: usize,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    if plan.downcast_ref::<UnionExec>().is_some() {
+        // `UnionExec::execute(p)` walks its children subtracting each one's
+        // partition count until `p` lands inside a child, then executes that
+        // child alone. Mirror that walk so the scan under the serving child is
+        // pinned to the group that child's local partition reads.
+        let mut local = Some(partition);
+        let mut children = Vec::with_capacity(plan.children().len());
+        for child in plan.children() {
+            let count = child.properties().output_partitioning().partition_count();
+            match local {
+                Some(p) if p < count => {
+                    children.push(restrict_scans_for_partition(child, p)?);
+                    local = None;
+                }
+                Some(p) => {
+                    local = Some(p - count);
+                    children.push(Arc::clone(child));
+                }
+                None => children.push(Arc::clone(child)),
+            }
+        }
+        return Arc::clone(plan).with_new_children(children);
+    }
+
+    if let Some(rewritten) = restrict_scan_to_partition(plan, partition) {
+        return Ok(rewritten);
+    }
+
+    let children = plan
+        .children()
+        .into_iter()
+        .map(|child| restrict_scans_for_partition(child, partition))
+        .collect::<Result<Vec<_>>>()?;
+    if children.is_empty() {
+        return Ok(Arc::clone(plan));
+    }
+    Arc::clone(plan).with_new_children(children)
+}
+
 impl ExecutionEngine for DefaultExecutionEngine {
     fn create_query_stage_exec(
         &self,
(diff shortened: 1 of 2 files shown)


avantgardnerio (approved):
LGTM

andygrove:
Merged. Thanks @avantgardnerio

PR #2072, opened 2026-07-16
fix: remove the static planner SortMergeJoinExec broadcast conversion

andygrove:

Partially addresses #2046 (q4 and q78 only).

# Rationale for this change

TL;DR. The SMJ "broadcast join" optimization in the static planner was poorly thought out. TPC-DS tests exposed that it caused correctness issues. Let's remove it.

This PR does not make any changes to AQE join selection.

# AI description of changes

`54.0.0` added a static-planner rewrite that converted a `SortMergeJoinExec` with a small build side into a broadcast `CollectLeft` hash join, gated by `ballista.optimizer.broadcast_sort_merge_join_enabled` (default `true`). The rewrite produces incorrect results, so the only safe value for the key is `false`. A config knob whose correct setting is always "off" is better removed along with the code it gates.

This is not an edge-case path. Ballista sets `datafusion.optimizer.prefer_hash_join = false` by default (#1648), so joins are planned as `SortMergeJoinExec` unless a user opts out, and the conversion then fired on any join with a side under the threshold.

- Remove `convert_sort_merge_to_hash_join` and its `strip_sort` helper from `DefaultDistributedPlanner`. A `SortMergeJoinExec` is no longer a broadcast candidate; only a `HashJoinExec(Partitioned)` is.
- Remove the `ballista.optimizer.broadcast_sort_merge_join_enabled` config key and its `BallistaConfig` accessor.
- Drop the test asserting the conversion happens, and the test asserting the threshold gated it (now vacuous). `distributed_sort_merge_join_unchanged_when_flag_disabled` becomes `distributed_sort_merge_join_is_never_broadcast`, since that is now unconditional behavior.
- Re-enable TPC-DS q4 and q78 in the correctness gate.
- Add a `55.0.0` upgrade guide; drop the stale "not been released yet" note from the `54.0.0` guide.

Broadcast promotion of a `HashJoinExec` under `ballista.optimizer.broadcast_join_threshold_bytes` is unchanged, as is the correctness guard that demotes a DataFusion-promoted `CollectLeft` join with an unsafe join type.

Yes, two, both documented in `docs/source/upgrading/55.0.0.md`:

1. `ballista.optimizer.broadcast_sort_merge_join_enabled` no longer exists. Ballista rejects unknown configuration keys, so a session that still sets it now fails with ``configuration key `optimizer.broadcast_sort_merge_join_enabled` does not exist``. This is a breaking config change against `54.0.0`.
2. A `SortMergeJoinExec` is never converted to a broadcast join. Because sort-merge is Ballista's default join strategy, join plan shape — and therefore performance and shuffle behavior — changes on upgrade for any query with a join whose smaller side is under `10 MB`. Users who want the broadcast can set `datafusion.optimizer.prefer_hash_join = true`, at the cost of opting into the non-spilling hash join for all joins in the session.

## Verification

Locally: `cargo clippy --all-targets --workspace --all-features -- -D warnings` passes, and the `ballista-scheduler` planner, `ballista-core` config, and `tpcds` bin tests pass.

The q4/q78 fix itself is **not** verified locally — it needs SF1 TPC-DS data and a live cluster. The TPC-DS SF1 workflow runs on this PR and is the real check; if q4 or q78 still diverge, those skip entries should go back with an updated reason.

Given the default-plan-shape change, TPC-H benchmark numbers before/after would be worth having before this merges.

### DIFF
--- a/benchmarks/src/bin/tpcds.rs
+++ b/benchmarks/src/bin/tpcds.rs
@@ -66,16 +66,8 @@ const SKIP: &[(usize, &str)] = &[
     // data (confirmed reproducible; DataFusion is correct under both join modes).
     // See https://github.com/apache/datafusion-ballista/issues/2046
     (
-        4,
-        "distributed result diverges from DataFusion (issue #2046)",
-    ),
-    (
         38,
         "distributed INTERSECT diverges from DataFusion (issue #2046)",
-    ),
-    (
-        78,
-        "distributed LIMIT drops rows vs DataFusion (issue #2046)",
     ),
     (
         87,
--- a/docs/source/upgrading/55.0.0.md
+++ b/docs/source/upgrading/55.0.0.md
@@ -1,0 +1,73 @@
+<!---
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Upgrade Guides
+
+## Ballista 55.0.0
+
+**Note:** Ballista `55.0.0` has not been released yet. The information provided
+in this section pertains to features and changes that have already been merged
+to the main branch and are awaiting release in this version.
+
+### Planner and execution behavior changes
+
+#### `SortMergeJoinExec` is no longer converted to a broadcast join
+
+`54.0.0` added a static-planner rewrite that converted a `SortMergeJoinExec`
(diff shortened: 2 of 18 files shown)


benchmarks/src/bin/tpcds.rs
```diff
@@ -65,18 +65,10 @@ const SKIP: &[(usize, &str)] = &[
     // Distributed execution diverges from single-process DataFusion on the same
     // data (confirmed reproducible; DataFusion is correct under both join modes).
     // See https://github.com/apache/datafusion-ballista/issues/2046
-    (
```
phillipleblanc, line 68: 🥳

docs/source/upgrading/55.0.0.md
```diff
@@ -0,0 +1,73 @@
...
+
+**Note:** Ballista `55.0.0` has not been released yet. The information provided
+in this section pertains to features and changes that have already been merged
+to the main branch and are awaiting release in this version.
+
+### Planner and execution behavior changes
+
+#### `SortMergeJoinExec` is no longer converted to a broadcast join
+
+`54.0.0` added a static-planner rewrite that converted a `SortMergeJoinExec`
```
phillipleblanc, line 32: Do you think this is worth a patch release on 54 to fix?

phillipleblanc (approved):
Thanks! I'll be upgrading Spice to ballista 54 soon, so I can take this commit as well - but it might be worth backporting onto the 54 release since this produces incorrect results? I can volunteer to do that if its helpful.

docs/source/upgrading/55.0.0.md
andygrove: Yes, absolutely. I will backport this. Thanks for the review!

Issue #2063, opened 2026-07-16
Adaptive (AQE) execution hangs when a replan cancels an in-flight stage


andygrove:
## Describe the bug

Under adaptive execution, when a replan collapses the residual plan and removes an in-flight stage's `ExchangeExec`, the scheduler cancels that stage but its already-running tasks keep going. When those tasks complete, the late status update fails and the job never finishes — the client hangs indefinitely.

The scheduler logs:

```
WARN  ballista_scheduler::state::aqe: there are stages to be cancelled but its not implemented. stages to cancel: {2}
ERROR ballista_scheduler::scheduler_server::query_stage_scheduler: Failed to update 3 task statuses
      for Executor <id>: DataFusionError(Execution("Can't find active cache resolve"))
```

Mechanism:

1. `AdaptivePlanner::actionable_stages` finds stage 2's exchange is gone from the replanned plan and returns it in `stages_to_cancel`.
2. `AdaptiveExecutionGraph::update_stage_progress` calls `cancel_stage(2)`, which removes the stage from `runnable_stage_cache` (`aqe/planner.rs`, `cancel_stage` — "TODO consider marking cancelled stage").
3. Stage 2's tasks were already launched and run to completion regardless — cancellation is not implemented on the executor side.
4. Their status update reaches `finalise_stage(2)`, whose `runnable_stage_cache.get(&stage_id)` now misses and returns `Execution("Can't find active cache resolve")`.
5. The error propagates out of the task-status batch, so the job's completion event is never emitted and the query hangs.

Note the final stage does complete successfully before this happens — the result is computed, it just never gets delivered.

## To Reproduce

Generate SF1 TPC-DS Parquet, start a scheduler + executor, then, with #2047 fixed (on `main` today the `EmptyExec` assertion aborts these queries first and masks this):

```
tpcds --host localhost --port 50050 --path <data> \
  --query 1 -c datafusion.optimizer.prefer_hash_join=false \
  -c ballista.planner.adaptive.enabled=true
# -> hangs; scheduler logs "Can't find active cache resolve"
```

Reproduced on q1 and q44 at SF1 (1 scheduler / 1 executor, 8 slots). The other 13 queries listed in #2047 complete once the assertion is fixed. The same queries succeed under the static planner.

## Expected behavior

A completion update for a stage that a replan has cancelled is ignored rather than treated as an error, and the job runs to completion.

## Additional context

[...]

Issue #2025, opened 2026-07-13
Hash-join build side OOMs on TPC-H Q18 at SF1000 and does not shrink with target_partitions


andygrove:
## Describe the bug

With `datafusion.optimizer.prefer_hash_join=true`, TPC-H Q18 at SF1000 fails with `ResourcesExhausted` in a hash-join build side, and **the per-task memory requirement does not shrink when `target_partitions` is increased**. That invariance is the interesting part: it means the memory is not sitting in a key-partitioned build side, and repartitioning cannot fix it.

This matters because hash joins are otherwise a large win at this scale (see below), and this OOM is the one thing blocking them from completing the suite.

## Evidence that it is invariant to `target_partitions`

Same cluster, same query, identical apart from `--partitions`. Per-task pool is `--memory-pool-size / concurrent_tasks` = 28GB / 8 = **3.3 GB**:

| `target_partitions` | Failure |
|---|---|
| 32 | `Failed to allocate additional 464.0 KB for HashJoinInput[12] with 3.3 GB already allocated ... pool_size: 3.3 GB` |
| 64 | `Failed to allocate additional 544.0 MB for HashJoinInput[0] with 3.3 GB already allocated ... pool_size: 3.3 GB` |

Doubling the partition count did **not** halve the per-task build side — it still saturates the same 3.3 GB, and in fact asks for *more* headroom. A build side that was hash-partitioned on the join key should have roughly halved.

Raising memory instead also fails to fix it. At `--memory-pool-size=48GB` (→ 5.6 GB per task slot):

```
Failed to allocate additional 1088.0 MB for HashJoinInput[4] with 4.8 GB already
allocated for this reservation - 792.2 MB remain available for the total memory
pool: fair(pool_size: 5.6 GB)
```

So the build side wants ~6 GB **per task**, and that figure is independent of how finely the input is partitioned. This is the signature of a `CollectLeft` (collected/replicated) build side — which every task materializes in full — rather than a partitioned one. Severe skew on the join key is the other candidate.

Sort-merge joins are unaffected: the same query with `prefer_hash_join=false` completes in **587.9s**, because SMJ streams and spills instead of materializing a build side in memory.

## Why this is worth fixing

Same cluster (2 executors × 8 cores / 56Gi), SF1000, single iteration, Q1–17 (all that hash joins survive before this OOM):

| Config | Q1–17 |
|---|---:|
| Ballista, `prefer_hash_join=true` | **2683.6s** |
[...]

PR #2075, opened 2026-07-17
docs: add benchmarking page to the contributors guide

### LINKED ISSUE #2074: Add a benchmarking page to the Contributors Guide

There is no single place in the documentation that describes how Ballista is benchmarked at scale, or that records what the current numbers actually are.

`benchmarks/README.md` covers generating TPC-H data and running the benchmark locally at small scale factors, but the questions that come up in performance work are not answered anywhere:

- What cluster shape are the published numbers from, and why that shape?
- How do I reproduce a number without the ad-hoc tooling it was produced with?
- How were Spark and Comet configured for a cross-engine comparison, and why is that comparison meaningful?
- Which known issues should I be aware of before interpreting a slow result?

Without this, benchmark numbers circulate in issues and PR descriptions with their conditions attached informally, and it is hard to tell whether two numbers are comparable. In particular, results under the static planner and the adaptive (AQE) planner are easy to mix up, even though they are produced by different planners with materially different join behaviour.

A `Benchmarking` page in the Contributors Guide that covers:

- What is measured and why (work balance across executor tasks), and why AQE on and AQE off are always reported together.
- The reference cluster shape, and the reasoning behind the parts that matter (executor placement, node-local data, memory-pool headroom).
- Self-contained commands for running TPC-H against Ballista, vanilla Spark, and Comet, so results are reproducible without repo-specific tooling.
- Why Comet is a useful comparison for Ballista, with the caveats that stop it from being an operator-level A/B.
- A single current result set, pinned to the exact commit it came from.
- The known issues worth knowing about when reading a benchmark result.

Extending `benchmarks/README.md` instead. That file is oriented around running the benchmark crate locally; the multi-node setup, the cross-engine comparison rationale, and the recorded results are a different audience and would crowd it.

**Additional context**

Results are currently produced on a small homelab cluster; the intent is to move to AWS with data in S3 once the numbers there are good. The page should say so, so readers do not read the numbers as absolute cloud throughput.

andygrove:

Closes #2074.

 # Rationale for this change

Benchmark numbers currently circulate through issues and PR descriptions with their conditions attached informally, which makes it hard to tell whether any two numbers are comparable. Results from the static planner and the adaptive (AQE) planner are especially easy to conflate, even though they come from different planners with materially different join behaviour.

`benchmarks/README.md` explains how to generate TPC-H data and run the benchmark locally, but not what cluster the published numbers came from, how to reproduce one, how Spark and Comet were configured for a cross-engine comparison, or which known issues to keep in mind when a query looks slow.

This adds one place that answers those questions and records the current result set.

A new `docs/source/contributors-guide/benchmarking.md`, wired into the Contributors Guide nav in `docs/source/index.rst`:

- **What we measure, and why** — work balance across executor tasks, and why AQE on and AQE off are always reported together.
- **Environment** — the results currently come from a small homelab cluster with node-local data; the intent is to move to AWS with data in S3 once the numbers here are good. Stated explicitly so the numbers are read as relative comparisons on controlled hardware rather than absolute cloud throughput.
- **Reference cluster** — the shape behind the numbers, and the reasoning for the parts that are easy to get wrong: one executor per node, node-local data rather than object storage, and headroom between the memory pool and the container limit.
- **Queries** — all engines run the SQLBench-H query set from `apache/datafusion-benchmarks`, which is what makes the comparison apples-to-apples.
- **Running the benchmark** — self-contained commands for Ballista, vanilla Spark, and Comet, so a result can be reproduced without repo-specific tooling.
- **Why compare against Comet** — Comet and Ballista execute DataFusion physical plans using the same operators, so a gap points at distribution, scheduling, shuffle, and memory rather than operator speed. Includes the caveats that stop it being an operator-level A/B: Spark's optimizer and Spark's AQE produce the plan Comet accelerates, Comet falls back to the JVM for unsupported operators, and the two use different distribution models.
- **Results** — a single current result set pinned to the exact commit, with Spark, Comet, Ballista AQE off, and Ballista AQE on side by side, plus conventions for recording a result.
- **Known issues found by benchmarking** — brief orientation for reading a result, linking #1359, #2025, and #2063.

**The result table is now complete.** TPC-H SF1000, 2 executors x (8 cores / 56 GiB), node-local data, 1 iteration, full-suite runs. Ballista at `main` @ `26b29391`:

| Engine | Total | Status |
|---|---:|---|
| Comet 0.17.0 | 2738.5 s | 22/22 |
| vanilla Spark 3.5.3 | 4687.9 s | 22/22 |
| Ballista (AQE on) | 6320.7 s | 22/22 |
| Ballista (AQE off) | 6378.2 s | 22/22 |

**All 22 queries completed on all four configurations** — no FAIL or OOM entries to record. Row counts agreed across every engine except Q15.

Measurement details recorded on the page rather than smoothed over:

- **The Ballista commit is pinned, not "main".** `26b29391` removed the static planner's sort-merge-join broadcast conversion, which changes the AQE-off path specifically, so results either side of that commit are not comparable. The page says why the exact commit is recorded.
- **Every figure is a suite figure.** Suite and standalone runs are not interchangeable: Q18 on Spark measured 458.0 s standalone versus 492.8 s within the suite, a 7.6% spread on the same build. The page makes "take the figure from a full suite run" an explicit convention.
- **Q15 row counts disagree**: Spark 0, Comet and Ballista 1. Q15 is a multi-statement query and the reported count depends on which statement the harness treats as the result. Recorded as an open discrepancy, not resolved here.
- Q11 returns 0 rows on every engine — its threshold constant is tuned for SF1.
- Verified that the CometBench-H queries Spark/Comet run are textually identical to the SQLBench-H set Ballista runs (all 22 match once the licence header is ignored), so the comparison is on the same SQL.

Ready for review.

No code changes. Adds a documentation page to the Contributors Guide.

### DIFF
--- a/docs/source/contributors-guide/benchmarking.md
+++ b/docs/source/contributors-guide/benchmarking.md
@@ -1,0 +1,318 @@
+<!---
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Benchmarking
+
+This page describes how Ballista is benchmarked at scale, and records the current
+results. It is aimed at contributors who want to reproduce a number, understand why
+a comparison is set up the way it is, or add a result of their own.
+
+The benchmark used here is derived from TPC-H. For running TPC-H locally at small
+scale factors, see [`benchmarks/README.md`][bench-readme] in the repository; this
+page covers the multi-node, large-scale-factor setup and the cross-engine
+comparisons.
+
+[bench-readme]: https://github.com/apache/datafusion-ballista/blob/main/benchmarks/README.md
+
+## What we measure, and why
+
+Ballista's central performance question is **how evenly work is spread across
+executor tasks**. A distributed query is only as fast as its slowest stage, and a
+stage is only as fast as its slowest task, so a query whose work concentrates onto
+a few partitions will underperform no matter how fast the underlying operators are.
+Benchmarks are therefore run at a scale where that imbalance actually shows up.
+
+Two configurations are always measured, because they select **different planners**
+with materially different join behaviour:
+
+- **AQE off** (`ballista.planner.adaptive.enabled=false`, the default) — the static
+  `DefaultDistributedPlanner`.
+- **AQE on** (`ballista.planner.adaptive.enabled=true`, experimental) — the adaptive
+  planner, which can re-plan stages using runtime statistics.
+
+A change that helps one planner can be a no-op or a regression under the other, so
+results are reported for both rather than for whichever looks better.
+
+## Environment
+
+The results on this page currently come from a **small homelab cluster**: two
+bare-metal nodes running Kubernetes, with the TPC-H data staged on node-local disk.
+This is a deliberate starting point rather than a final destination. A two-node
+cluster with local disk removes as many confounds as possible — no object-store
+latency, no cloud network variance, no noisy neighbours — so that when a query is
+slow, the cause is Ballista and not the environment.
+
+The intent is to **move these benchmarks to AWS with data in S3** once the results
+here are good. That environment is the one users actually run, and it exercises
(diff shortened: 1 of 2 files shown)


docs/source/contributors-guide/benchmarking.md
```diff
@@ -0,0 +1,328 @@
...
+requirement — the commands below work on any cluster, but numbers are only
+comparable when the shape matches.
+
+|                     |                                            |
+| ------------------- | ------------------------------------------ |
+| Executors           | 2, one per physical node                   |
+| Per executor        | 8 cores, 56 GiB, `--memory-pool-size=48GB` |
+| Per task slot       | 8 concurrent tasks → 6 GB pool each        |
+| Scheduler           | 1                                          |
+| Data                | TPC-H SF1000 Parquet, node-local disk      |
```
phillipleblanc, line 85: May also be useful to specify how this data is generated (i.e. via the tpchgen-rs, partitioned, etc)

phillipleblanc (approved):

andygrove:
This is the first clean run of the whole benchmark suite at this scale factor, so will merge this as a baseline and move onto debugging why performance is ~50% slower than Spark.

PR #2084, opened 2026-07-17
feat: Improve TPC-H performance with AQE on

### LINKED ISSUE #2081: AQE broadcast decision falls back to a size-blind row threshold when total_byte_size is unknown
**Describe the bug**

The adaptive (AQE) planner decides whether to broadcast a join's build side in `DynamicJoinSelectionExec::supports_collect_by_thresholds`:

```rust
let Ok(stats) = plan.partition_statistics(None) else { return false; };

if let Some(byte_size) = stats.total_byte_size.get_value() {
    *byte_size != 0 && *byte_size < threshold_byte_size      // 10 MB
} else if let Some(num_rows) = stats.num_rows.get_value() {
    *num_rows != 0 && *num_rows < threshold_num_rows         // 1,000,000
} else {
    false
}
```

When `total_byte_size` is known this is a size check, which is what we want. But when it is `Absent` the decision falls back to a **row count** compared against `hash_join_single_partition_threshold_rows` (default 1,000,000). A row count says nothing about size, so a build side of up to 1M arbitrarily wide rows can be collected and replicated to every probe task while the 10 MB guard never gets a chance to apply.

This is not a rare corner. `total_byte_size` is `Absent` for exactly the intermediates whose size we cannot guess:

1. **Join outputs, always.** DataFusion's `estimate_join_statistics` hardcodes `total_byte_size: Precision::Absent` and keeps only an estimated row count, so every join result has unknown size.
2. **Anything with a variable-width column.** `Statistics::calculate_total_byte_size` sums `DataType::primitive_width()` across the schema; that returns `None` for `Utf8`/`LargeUtf8`/`Utf8View`/`Binary` (and `Boolean`), so a single string column makes it bail out to `total_byte_size.to_inexact()`. `to_inexact()` on `Absent` is still `Absent`, so once the size is lost a projection cannot recover it.

In TPC-H that covers most dimension-side join results, since they carry names, addresses, or comments.

**To Reproduce**

TPC-H SF1000, 2 executors x (8 cores / 56 GiB), `target_partitions=32`, `prefer_hash_join=false`, adaptive planner on, with `RUST_LOG=info,ballista_scheduler::state::aqe=debug`. Q2 alone produces:

```
to_actual_join - plan_id: 1, decision: CollectLeft left: (Inexact(799140)   | Absent), right: (Exact(10000000) | Absent)
to_actual_join - plan_id: 7, decision: CollectLeft left: (Inexact(639312)   | Absent), right: (Inexact(159969600) | Inexact(3839270400))
to_actual_join - plan_id: 5, decision: CollectLeft left: (Inexact(10000000) | Inexact(320000000)), right: (Exact(25) | Exact(400))
to_actual_join - plan_id: 2, decision: CollectLeft left: (Inexact(3196560)  | Absent), right: (Exact(25) | Absent)
```

`plan_id: 1` and `plan_id: 7` are broadcast on the row rule alone (799,140 and 639,312 rows, size unknown). These are `part`/`supplier`-derived rows carrying string columns, so the real payload replicated to all 32 tasks is plausibly hundreds of MB rather than the 10 MB the byte threshold is meant to cap.

The same run also shows how easily the size is lost. Two decisions over the same 25-row `nation` table:

- `plan_id: 5`, right side `Exact(25) | Exact(400)` — an all-numeric projection, so bytes are recomputed as rows x width.
- `plan_id: 2`, right side `Exact(25) | Absent` — the schema carries a string column, so the recompute gives up.

**Expected behavior**

The broadcast decision should be driven by an estimate of the build side's **size**. When `total_byte_size` is unavailable, Ballista should estimate it (rows x estimated row width, from the schema's fixed-width columns plus per-column statistics or a configurable default for variable-width columns) and test that against the byte threshold, rather than treating "fewer than N rows" as a proxy for "small".

For reference, Spark drives broadcast selection from `spark.sql.autoBroadcastJoinThreshold` in bytes, and estimates a per-row size from the schema (`EstimationUtils.getSizePerRow`) when it must; it has no row-count threshold standing in for size.

**Additional context**

Found while benchmarking TPC-H SF1000 under both planners (2x8 cores, node-local data, 1 iteration). AQE-on and AQE-off totalled 6320.7 s and 6378.2 s respectively — a wash overall, but the per-query results move in both directions, and the queries where AQE loses most (Q5 700.5 vs 542.7, Q8 801.2 vs 657.1, Q9 1025.2 vs 891.8, Q18 820.9 vs 620.7) are join-heavy ones where this fallback is most likely to be making the call. Whether over-eager broadcasts explain those regressions is not yet established — this issue is about the decision rule being size-blind, which is true regardless.

There are related gaps that are worth tracking separately and are not proposed here:

- `stats_for_partitions` (`ballista/core/src/execution_plans/shuffle_reader.rs`) reports shuffle `num_rows`/`total_byte_size` as `Exact` from measured output, but sets `column_statistics` to all-unknown (there is a `TODO stats: add column statistics to PartitionStats`), and the wire `ColumnStats` message has no `byte_size` field. So Ballista measures real sizes at every shuffle boundary and then cannot carry them into downstream planning.
- The two DataFusion behaviours above (joins discarding `total_byte_size`; `calculate_total_byte_size` ignoring the existing `ColumnStatistics::byte_size` field for variable-width columns) would be better fixed upstream.

andygrove:

Closes #2081.

Related: #2025 (Q18 hash-join build-side OOM at SF1000).

# Rationale for this change

Two AQE join behaviours concentrate work on a few tasks or fail outright at
SF1000:

1. The AQE broadcast-join decision was driven by **row count**, not bytes, so it
   could broadcast a build side that is actually large, or repartition one that
   is actually small. Sizing the decision by bytes keeps the broadcast fast path
   on genuinely small build sides.
2. A `Partitioned` hash join whose largest build partition does not fit one task
   slot's memory pool builds in memory and **OOM-kills the executor** —
   DataFusion's hash-join build side does not spill — notably Q18 (#2025). There
   was no way to keep such a join running.

This PR makes the broadcast decision size-aware and adds an opt-in per-partition
build-size check that lowers an oversized hash join to `SortMergeJoin` (which
spills) instead of aborting. It also adds INFO/WARN observability for these
runtime decisions so they are visible in executor logs.

**Size-aware broadcast join (AQE):**

- Add `ballista.optimizer.broadcast_join_threshold_bytes` and
  `ballista.optimizer.broadcast_join_threshold_rows`; the AQE planner sizes the
  broadcast decision by bytes, falling back to rows when byte statistics are
  unavailable.
- Demote DataFusion `CollectLeft` joins whose build side exceeds the Ballista
  threshold.

**Hash-join build-size fallback (opt-in):**

- Add `ballista.optimizer.hash_join_max_build_partition_bytes` (default `0` =
  off). When a `Partitioned` hash join's largest build partition exceeds the
  budget, the join is lowered to `SortMergeJoin`; the decision is logged at INFO.

**Observability:**

- Log AQE join decisions at INFO and shuffle spills at WARN; INFO shuffle
  write/spill activity; byte-size statistics for completed query stages.

**Docs:**

- Refresh the SF1000 AQE-on results and keep the hash-join + fallback
  demonstration in `docs/source/contributors-guide/benchmarking.md`.

## Benchmark results (TPC-H SF1000, 2×16-core reference cluster, AQE on, 1 iteration)

**Default config — SMJ-planned (`prefer_hash_join=false`, fallback off).**
Ballista total excl. Q18: **4817.8 → 4661.0 s**. Largest movers are the
join-heavy queries:

| Query | Prior `main` | This build | Δ    |
| ----: | -----------: | ---------: | ---: |
|     7 |        457.0 |      369.0 |  −88 |
|     8 |        731.0 |      540.3 | −191 |
|     9 |        934.3 |      764.9 | −169 |

Q18 still OOMs in this configuration — the build-size fallback is opt-in and off
here.

**Hash-join mode + fallback (`prefer_hash_join=true`,
`hash_join_max_build_partition_bytes=64 MiB`).** All 22 queries complete with no
OOM; Q18 finishes at **744.6 s**, where the same run with the fallback disabled
fails the job (#2025). Full tables are in `benchmarking.md`.

Caveats: single iteration. The "Prior" column is the previous `main` baseline;
this build also changes the sort-shuffle spill-cap default, so per-query deltas
reflect the combined change rather than broadcast sizing alone. The two result
sections use different `prefer_hash_join` settings and are not directly
comparable to each other.

New session config keys (defaults preserve prior behaviour):

- `ballista.optimizer.broadcast_join_threshold_bytes`
- `ballista.optimizer.broadcast_join_threshold_rows`
- `ballista.optimizer.hash_join_max_build_partition_bytes` (default `0`, off)

New INFO/WARN executor log lines for AQE join decisions and shuffle spill
activity. No breaking API changes.


milenkovicm:
One note, while trying AQE on TPCDS Q72 for some reason bite size is not available in all joins (not sure why) hence AQE makes wrong decisions.

andygrove:
> One note, while trying AQE on TPCDS Q72 for some reason bite size is not available in all joins (not sure why) hence AQE makes wrong decisions. 
> 
> 

Yes, that's exactly what I am seeing. I am getting up to speed with the statistics code now.

ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -567,7 +568,7 @@ impl SortShuffleWriterExec {
                     write_time,
                 );
             } else {
-                debug!(
+                info!(
```
phillipleblanc: Looks like this got flipped back to info

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -315,7 +309,36 @@ impl DynamicJoinSelectionExec {
...
+
+        let action_label = match &action {
+            JoinSelectionAction::Repartition(_) => "Repartition",
+            JoinSelectionAction::CollectLeft(_) => "CollectLeft(broadcast)",
+            JoinSelectionAction::LateCollectLeft(_) => "LateCollectLeft(broadcast)",
+            JoinSelectionAction::Hash(_) => "Hash(Partitioned)",
+            JoinSelectionAction::Sort(_) => "SortMerge(Partitioned)",
+        };
+
+        info!(
```
phillipleblanc:
```suggestion
        debug!(
```

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -365,12 +388,29 @@ impl DynamicJoinSelectionExec {
...
+        // have allowed, never introduce a new one.
+        let Some(num_rows) = stats.num_rows.get_value().copied() else {
+            return false;
+        };
+
+        if num_rows == 0 || num_rows >= threshold_num_rows {
+            return false;
+        }
+
+        estimate_output_byte_size(&plan.schema(), num_rows, &stats.column_statistics)
```
phillipleblanc, line 412: This new byte estimate is not used when choosing the build side: `to_actual_join` ORs both sides, then build-side selection can fall back to row count. If only the right side passes this estimate but the wider left has fewer rows, AQE still broadcasts the left. It might be better to preserve the per-side eligibility when selecting the build input?

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -228,6 +229,8 @@ impl DynamicJoinSelectionExec {
             .unwrap_or_default();
         let threshold_collect_left_join_bytes = bc.broadcast_join_threshold_bytes();
         let threshold_collect_left_join_rows = bc.broadcast_join_threshold_rows();
+        let max_build_bytes = bc.hash_join_max_build_partition_bytes();
+        let build_max_partition_bytes = max_per_partition_build_bytes(&self.left);
```
phillipleblanc, line 233: `self.left` is measured before SelectJoinRule may swap the hash inputs and before AQE coalescing groups partitions. The eventual build task can therefore consume a different or larger partition than this check validated. It might be better to apply the budget to the post-swap, post-coalesce input.

phillipleblanc (approved):
Some minor feedback below.

ballista/core/src/execution_plans/sort_shuffle/writer.rs
andygrove: Fixed in b563844b — demoted to `debug!`. It fires once per shuffle partition; the spill case stays at `warn!`.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Applied. b563844b also switches the `use log::info` import to `use log::debug` so the file still compiles after the change.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Confirmed — the per-side eligibility is lost once `under_threshold` collapses to an OR of both sides. Tracking as a follow-up in #2121; it would need a re-benchmark so keeping it out of this PR.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Confirmed — the check measures the pre-swap, pre-coalesce `self.left`, which can differ from the partition the build task consumes. Tracking in #2121 to apply the budget to the post-swap, post-coalesce build input.

andygrove:
Thanks for the review @phillipleblanc. I'm going to work on the points you raised next (filed as #2121)

PR #2088, opened 2026-07-17
feat: add BALLISTA_PROTOCOL_VERSION + k8s health probes

### LINKED ISSUE #2071: Live upgrade strategy

Ballista users are now running clusters 24/7, and can't afford downtime during upgrades. However, Ballista is still young & flexible, and we want to maintain velocity even as we make [architectural changes](https://github.com/apache/datafusion-ballista/pull/2038). It would be good to have an official upgrade strategy that protects both.

Add a ballista_protocol_version: u32 field to ExecutorRegistration / HeartbeatFromExecutor. The scheduler holds its own compiled-in BALLISTA_PROTOCOL_VERSION constant. On every
  heartbeat or registration:

  - If executor.protocol_version == scheduler.protocol_version → proceed normally.
  - Otherwise → drop the heartbeat with an info! log and never dispatch tasks to that executor.

  We bump BALLISTA_PROTOCOL_VERSION only on releases that break wire compat, not once per major version. Most releases won't bump it; some will.

  Rolling upgrade flow

  1. New scheduler comes up with bumped BALLISTA_PROTOCOL_VERSION.
  2. Old executors heartbeat in; scheduler drops them; they drain naturally as their existing tasks finish (or the operator kills them).
  3. New executors with the bumped version come up and register normally.
  4. No dual-implementation window, no code carrying old + new proto types.

Both scheduler versions are still up during the transition, so operators can roll executors and schedulers in either order — mismatched pairs just don't talk to each other, they don't corrupt state.

  Prior art

  DQE (Coralogix' Ballista fork) has been doing this for a while with reasonable success.

3 release upgades: add, change impl & deprecate, remove

**Additional context**

The downsides:

1. cluster capacity is at worst halved during upgrades - mitigation: upgrade off hours, custom k8s operator
2. more operational complexity

avantgardnerio:
## Summary

Closes #2071.

My notes:

1. caution, I do not run upstream ballista in production
2. however, this is the strategy that until recently Coralogix used with mostly success
3. I still feel confident submitting this, because it does not appear we have this functionality at all yet?
4. my motivation is purely (well, mostly) selfish: I would like to release [multi-partition-tasks](https://github.com/apache/datafusion-ballista/pull/2038) without a multi-major-version-upgrade cycle

With that out of the way, onto the Claude ramblings:

Adds a compile-time `BALLISTA_PROTOCOL_VERSION: u32 = 1` in
`ballista_core` that gates executor↔scheduler traffic, plus
Kubernetes-style `/healthz` and `/readyz` probes on both binaries so
live upgrades have something to observe.

**Protocol version**
- New `uint32 ballista_protocol_version = 7` on `ExecutorRegistration`.
- Executor populates it on registration, heartbeat, and `poll_work`.
- Scheduler rejects mismatches (or missing metadata on heartbeat) with
  `Status::failed_precondition`, an `info!` log, and increments
  `ballista_scheduler_rejected_by_protocol_version_total` (prometheus
  feature).
- Executor `/readyz` flips off on RPC failure, and after 5 consecutive
  heartbeat failures the executor self-terminates so a mismatched pod
  crash-loops until its image is bumped rather than sitting idle.

**Health probes**
- Scheduler: mounted on the existing axum router (same port as gRPC/REST).
  `/healthz` is always 200. `/readyz` returns 200 iff at least
  `--min-ready-executors` executors are registered (default 1; set to 0
  for dev/single-node).
- Executor: new HTTP server on `--bind-health-port` (default `50053`),
  since Arrow Flight isn't HTTP. `/healthz` is always 200 — deliberately
  independent of scheduler connectivity to prevent restart cascades.
  `/readyz` reflects the last heartbeat.

**Rolling upgrade semantics** (documented in `kubernetes.md`)

| scheduler ↔ executor | outcome |
|---|---|
| old ↔ old | works |
| old ↔ new | works (old scheduler ignores the unknown field) |
| new ↔ new | works |
| new ↔ old | rejected — the case the check exists for |

Bump both Deployments' image tags together in the same release; each
rolls independently under its own `maxSurge`, and the scheduler's
`/readyz` gate keeps `Service` traffic on the old scheduler until the
new one has matching-version executors registered.

**Bump policy**: bump `BALLISTA_PROTOCOL_VERSION` only on releases that
change the executor↔scheduler wire format. Most releases won't.

## Test plan

- [x] `cargo test -p ballista-scheduler --lib` (236 pass, incl. 8 new)
- [x] `cargo test -p ballista-executor --lib` (45 pass)
- [x] `cargo clippy --all-targets -p ballista-core -p ballista-executor -p ballista-scheduler` (clean)
- [x] `cargo fmt --all`
- [x] End-to-end smoke against a 2-executor local cluster: all four
      `/healthz` and `/readyz` endpoints return 200; scheduler `/readyz`
      shows `ready: 2 executors registered`.

New unit tests cover:
- Register rejected on mismatched protocol version
- Heartbeat rejected on mismatched protocol version
- Heartbeat rejected when metadata is missing entirely
- `poll_work` rejected on mismatched protocol version
- `/healthz` always OK
- `/readyz` fails with no executors
- `/readyz` OK when `min_ready_executors = 0`
- `/readyz` OK after an active heartbeat is recorded

## Notes for reviewers

- The zero default of the new proto field naturally means "old
  executor" and is rejected once we start at
  `BALLISTA_PROTOCOL_VERSION = 1`. This is intentional per the issue's
  "no dual-implementation window" design.
- The old scheduler is permissive because it doesn't know the field
  exists — so during a rolling upgrade, old-scheduler ↔ new-executor
  keeps working. Only new-scheduler ↔ old-executor is rejected.
- Health probes are unconditional: they mount regardless of the
  `rest-api` feature or `--disable-rest-api`, since they're
  operational, not part of the REST API surface.

### DIFF
--- a/ballista/executor/src/executor_process.rs
+++ b/ballista/executor/src/executor_process.rs
@@ -61,7 +61,9 @@ use ballista_core::utils::{
     GrpcClientConfig, GrpcServerConfig, create_grpc_client_endpoint, create_grpc_server,
     default_config_producer, get_time_before,
 };
-use ballista_core::{BALLISTA_VERSION, ConfigProducer, JobId, RuntimeProducer};
+use ballista_core::{
+    BALLISTA_PROTOCOL_VERSION, BALLISTA_VERSION, ConfigProducer, JobId, RuntimeProducer,
+};
 
 use crate::client_pool::DefaultBallistaClientPool;
 use crate::execution_engine::{DefaultExecutionEngine, ExecutionEngine};
@@ -126,6 +128,8 @@ pub struct ExecutorProcessConfig {
     pub port: u16,
     /// Port for the executor's gRPC service.
     pub grpc_port: u16,
+    /// Port for the executor's HTTP health server (`/healthz` and `/readyz`).
+    pub health_port: u16,
     /// Hostname of the scheduler to connect to.
     pub scheduler_host: String,
     /// Port of the scheduler's gRPC service.
@@ -214,6 +218,7 @@ impl Default for ExecutorProcessConfig {
             external_host: None,
             port: 50051,
             grpc_port: 50052,
+            health_port: 50053,
             scheduler_host: "localhost".into(),
             scheduler_port: 50050,
             scheduler_connect_timeout_seconds: 0,
@@ -519,6 +524,10 @@ pub async fn start_executor_process(
     // Channels used to receive stop requests from Executor grpc service.
     let (stop_send, mut stop_recv) = mpsc::channel::<bool>(10);
 
+    // Shared health/readiness signal. Flipped by the heartbeat/poll_work loop
+    // and observed by the /readyz probe served on `health_port`.
+    let health = crate::health::ExecutorHealth::new();
+
     // Starting main executor process based on the TaskSchedulingPolicy
     //
     // PushStaged => starting new executor_server that waits for tasks from the schedule
@@ -534,6 +543,7 @@ pub async fn start_executor_process(
                     default_codec,
                     stop_send,
                     &shutdown_notification,
+                    health.clone(),
                 )
                 .await?,
             );
@@ -543,9 +553,31 @@ pub async fn start_executor_process(
                 scheduler.clone(),
                 executor.clone(),
                 default_codec,
+                health.clone(),
             )));
         }
     };
+
+    // Start the health probe HTTP server. It listens for the shutdown
+    // broadcast and stops gracefully, so the pod is not held up by /healthz.
+    {
+        let health_addr: SocketAddr = format!("{}:{}", opt.bind_host, opt.health_port)
+            .parse()
+            .map_err(|e| {
+                BallistaError::General(format!("invalid health_port address: {e}"))
+            })?;
+        let (health_shutdown_tx, health_shutdown_rx) = tokio::sync::oneshot::channel();
+        let mut probe_shutdown = shutdown_notification.subscribe_for_shutdown();
+        tokio::spawn(async move {
+            probe_shutdown.recv().await;
+            let _ = health_shutdown_tx.send(());
+        });
+        service_handlers.push(crate::health::spawn_health_server(
+            health_addr,
+            health,
+            health_shutdown_rx,
+        ));
+    }
     let shutdown = shutdown_notification.subscribe_for_shutdown();
     let override_flight = opt.override_arrow_flight_service.clone();
 
@@ -939,6 +971,7 @@ pub fn structure_executor_metadata(
             total_available_disk_space,
             open_files_limit,
         }),
+        ballista_protocol_version: BALLISTA_PROTOCOL_VERSION,
     }
 }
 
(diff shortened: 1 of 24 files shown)


avantgardnerio:
@phillipleblanc FYI

ballista/executor/src/executor_process.rs
```diff
@@ -126,6 +128,8 @@ pub struct ExecutorProcessConfig {
     pub port: u16,
     /// Port for the executor's gRPC service.
     pub grpc_port: u16,
+    /// Port for the executor's HTTP health server (`/healthz` and `/readyz`).
+    pub health_port: u16,
```
phillipleblanc: In our project we embed Ballista in our process that already defines health/readiness probes that are specific to our implementation. This would add a second health surface that we can't turn off without removing this code in our fork.

Perhaps similar to the `rest-api` feature flag, we could add a `health-probes` flag that controls this? (or bundle this with the `rest-api` feature flag?) Alternatively a separate option to control whether ballista starts the health service at all that we could disable.

phillipleblanc (commented):
I think the protocol version makes sense. I have one comment about the health service specifically below, but in general I think this points to a tension in what Ballista is and how we expect consumers to use it.

We use Ballista as a library that we embed into our existing process (i.e. we call `scheduler_process::create_scheduler()` and serve this on our existing gRPC port rather than let Ballista create it).

I think it would be useful to have a split between the Ballista standalone processes (i.e. `ballista-scheduler`, `ballista-executor`) vs the libraries that those binaries (and other consumers) can use to build their own distributed query engines. So for the health probes specifically, I think it would make sense to have it live in the binary/process startup, not as part of the library component. That way its still usable out of the box and is a good reference implementation for anyone that is integrating with ballista, but doesn't force them to take this specific implementation.

ballista/executor/src/executor_process.rs
avantgardnerio: @phillipleblanc , I wasn't aware you (or anyone) was using it that way. Given your use case (should we add this to a persona @andygrove ?) I think you are absolutely correct about architecture, and I was negligent in the design. PTAL at the current state. I have:

1. removed axum from the executor library
2. exposed a boolean is_healthy() method in both scheduler and executor
3. moved axum & the health check server into the executor binary
4. not changed the scheduler (this is the one I'm not sure about)

My thought is that if schedulers are already starting a REST API, it is natural to put the health check endpoint there so as to not require another port binding. However, I can see how this might be a separate concern from the REST API.

Happy to entertain either option...

phillipleblanc: @avantgardnerio no problem! Originally Ballista was started as a standalone binary process, and it was relatively recently that @milenkovicm started to focus it to be more of a library (which is one of the reasons we decided to go with Ballista rather than something like [`datafusion-distributed`](https://github.com/datafusion-contrib/datafusion-distributed) which is also a library, but has different design goals)

I agree that putting the health APIs on the scheduler rest API make sense. Everything else looks good now, thanks for reworking this!

phillipleblanc (approved):
Thanks! This looks good to me now.

milenkovicm:
https://datafusion.apache.org/ballista/contributors-guide/user-personas.html 

Persona 3 should cover @phillipleblanc case if im not mistaken

PR #2091, opened 2026-07-18
feat: log sort shuffle spills at WARN, keep write completion at DEBUG

andygrove:

Part of the observability work tracked in #2081. This extracts the sort
shuffle write/spill logging from the WIP PR #2084 into a small, standalone
change.

# Rationale for this change

When a sort shuffle task spills to disk under memory pressure it produced no
log output at the default level, so spilling was effectively invisible.
Spilling is a key signal when diagnosing slow or memory-constrained TPC-H
stages, so it is worth surfacing even at the default log level.

Per reviewer feedback, the routine per-partition write completion message
stays at `debug!` to avoid adding day-to-day noise.

- When a task spilled to disk, emit a `warn!` that names the job, stage, and
  input partition and reports the bytes, batches, and events spilled under
  memory pressure.
- Keep the per-partition shuffle write completion log at `debug!`, and add
  the output row count to that message.

Executor logs now report sort shuffle spill activity at `WARN`. The routine
write completion message remains at `DEBUG`. There are no API or
configuration changes.


ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -350,18 +350,35 @@ impl SortShuffleWriterExec {
...
+                    job_id,
+                    stage_id,
+                    input_partition,
+                    total_bytes_spilled,
+                    total_spilled_batches,
+                    spill_events,
+                    elapsed_secs,
+                );
+            } else {
+                info!(
```
milenkovicm: can info be debug in this case? 
it might be important at this moment but in day-to-day operations probably would be just noise? wdyt ?

milenkovicm (approved):
thanks @andygrove, just one comment, but fell free to ignore it if you dont agree

ballista/core/src/execution_plans/sort_shuffle/writer.rs
phillipleblanc: I agree this should probably be debug

phillipleblanc (approved):

ballista/core/src/execution_plans/sort_shuffle/writer.rs
andygrove: Claude also wanted debug, so it's 3 against 1 at this point. Updated!

andygrove:
The timings in the log messages are very misleading because they include the time to execute the query that produces the shuffle data. Moving this to draft for now.

phillipleblanc (approved):

PR #2094, opened 2026-07-19
feat(core): RuntimeStatsExec — passthrough tap for row counts + quantile sketches

avantgardnerio:
## Summary

Adds `RuntimeStatsExec`: a passthrough physical operator that observes input batches as they stream past and accumulates runtime state, without touching the data. Two accessor families:

- **row count** — per input partition, `AtomicUsize::fetch_add` on the hot path (no cross-partition contention, no lock overhead). Always tracked.
- **quantile sketch** — per input partition, `Mutex<TDigest>` (writes never contend across partitions). Only allocated when the caller passes `Some(order_by)` at construction.

Both accessors are readable at any point, so callers get a mid-stream snapshot when the sample is accurate enough or a post-drain snapshot when they want the fully-observed state (typical after a blocking downstream like `SortExec`).

## Why this exists — supporting the forthcoming range-repartition operators

`RuntimeStatsExec` is the shared substrate for two upcoming ops (`OrderedRangeRepartitionExec` and `UnorderedRangeRepartitionExec`) that will land the distributed parallel-window path. Two modes matched to two use sites:

**Sketching mode (`Some(order_by)`) → drives `UnorderedRangeRepartitionExec`.**
Placed above the scan, `RuntimeStatsExec` streams a T-Digest over the routing column while the data is scanned. Each executor's sketch is merged (`TDigest::merge_digests`) to choose global cut points at a scheduler barrier, then `UnorderedRangeRepartitionExec` uses those cuts to route rows to sub-partitions by value range — no upfront sort required.

**Row-count-only mode (`None`) → drives `OrderedRangeRepartitionExec`.**
Placed above a `SortExec`, it exposes exact runtime row counts per sub-partition. DataFusion's `partition_statistics()` only returns plan-time estimates, which are too coarse for the ordered stage's bin-packer. The tap fixes that at operator level without touching DF upstream.

The full parallel-window path (Stage 1 sketch → cut selection → Stage 2 slice) is described in the `parallel-window-kll-adaptive` design doc on the source branch and will land alongside the range-repartition operators.

## Scope of this PR

Just the operator, its proto/serde plumbing, and tests. Nothing consumes the accumulated state yet — the executor still ships `SuccessfulTask` unchanged, and no rule inserts `RuntimeStatsExec` into any plan. Those pieces land with the range-repartition operators.

Included:

- `datafusion-functions-aggregate-common = \"54\"` workspace dep (for `TDigest`).
- Proto: `RuntimeStatsExecNode` on `BallistaPhysicalPlanNode.oneof` (tag 6), and `QuantileSketchState` for the wire T-Digest (fixed 6-element `Vec<ScalarValue>` per `TDigest::to_scalar_state`).
- `sketch_to_proto` / `sketch_from_proto` helpers with a shape guard against corrupted wire input (would otherwise panic in `from_scalar_state`).
- TODO in module docs: swap `TDigest` (Float64-only, single-column) for a generic-over-`Ord` KLL sketch that can cover the full `Vec<PhysicalSortExpr>`.

## Tests (9)

- Three wire round-trips: populated sketch, empty sketch, wrong-element-count rejection.
- Three streaming: `execute_populates_sketch_and_row_count`, `execute_skips_null_routing_keys_from_sketch`, `execute_row_count_only_no_sketch`.
- Three plan-node serde: sketching-mode roundtrip, row-count-only-mode roundtrip, empty-`order_by` rejection.

## Test plan

- [x] `cargo test -p ballista-core` (all 123 tests + 2 doctests + 1 compile-fail pass).
- [x] `cargo clippy --all-targets` clean.
- [x] `cargo fmt --all`.

### DIFF
--- a/ballista/core/src/execution_plans/runtime_stats.rs
+++ b/ballista/core/src/execution_plans/runtime_stats.rs
@@ -1,0 +1,687 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Passthrough operator that accumulates runtime statistics on the data
+//! flowing through it. Two accessor families:
+//!
+//! - **Row count** — always tracked, per input partition. Written via
+//!   `AtomicUsize::fetch_add` on the hot path (no cross-partition
+//!   contention, no lock overhead).
+//! - **Quantile sketch** — optional (only when `order_by` is set at
+//!   construction). Per-partition `Mutex<TDigest>` keeps writes off any
+//!   shared lock.
+//!
+//! Timing is decoupled from correctness: both accessors are readable at
+//! any point. Callers get whatever has flowed through so far — a
+//! mid-stream snapshot for callers that decide the sample is accurate
+//! enough, a post-drain snapshot for callers that want the full state
+//! (typical after a blocking downstream like `SortExec`).
+//!
+//! The `order_by` field accepts the full `Vec<PhysicalSortExpr>` so
+//! multi-key `ORDER BY` survives serde (tie-breakers get preserved for
+//! downstream `SortExec` / `BoundedWindowAggExec` even though only the
+//! first key drives the sketch today).
+//!
+//! TODO: swap `TDigest` for a generic-over-`Ord` KLL sketch. TDigest is
+//! `Float64`-only and single-column; a KLL implementation would sketch
+//! the full `Vec<PhysicalSortExpr>` (composite keys, non-numeric types)
+//! and let the operator drop the "first expression, `Float64` only"
+//! restriction on the routing key.
+//!
+//! This PR lands the tap in isolation: nothing wires it into a plan yet,
+//! and the executor doesn't yet ship the accumulated state back to the
+//! scheduler. Those pieces arrive with the range-repartition operator
+//! (which is the first consumer).
+
+use std::fmt::{self, Debug, Formatter};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+use datafusion::arrow::array::Float64Array;
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::common::{Result, internal_datafusion_err, internal_err};
+use datafusion::execution::TaskContext;
+use datafusion::physical_expr::PhysicalSortExpr;
+use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion::physical_plan::{
+    DisplayAs, DisplayFormatType, ExecutionPlan, ExecutionPlanProperties, PlanProperties,
+    SendableRecordBatchStream,
+};
+use datafusion_functions_aggregate_common::tdigest::TDigest;
+use futures::stream::StreamExt;
+use log::info;
+
+/// T-Digest centroid budget. 100 is DataFusion's default and gives ~1%
+/// quantile error, plenty of margin over the sub-partition counts we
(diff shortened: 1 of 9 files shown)


avantgardnerio:
@phillipleblanc FYI

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -0,0 +1,687 @@
...
+        let slot = sketches.get(partition).ok_or_else(|| {
+            internal_datafusion_err!(
+                "RuntimeStatsExec: partition {} out of range (have {})",
+                partition,
+                sketches.len()
+            )
+        })?;
+        Ok(Some(
+            slot.lock()
+                .expect("RuntimeStats sketch mutex poisoned")
```
phillipleblanc: instead of panicking, could we just return an err?

phillipleblanc: here and below

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -0,0 +1,687 @@
...
+            return;
+        };
+        let rows = counter.load(Ordering::Relaxed);
+        if rows == 0 {
+            return;
+        }
+        match self.sketches.as_ref().and_then(|s| s.get(self.partition)) {
+            Some(slot) => {
+                let sketch = slot.lock().expect("RuntimeStats sketch mutex poisoned");
+                info!(
```
phillipleblanc: debug level?

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -0,0 +1,687 @@
...
+                        "asc"
+                    }
+                )
+            }
+            None => write!(f, "RuntimeStatsExec: rows"),
+        }
+    }
+}
+
+impl ExecutionPlan for RuntimeStatsExec {
```
phillipleblanc, line 270: I believe this will inherit some incorrect defaults for a passthrough exec, I think we need to implement these as well:

```rust
   fn maintains_input_order(&self) -> Vec<bool> {
       vec![true]
   }

   fn cardinality_effect(&self) -> CardinalityEffect {
       CardinalityEffect::Equal
   }

   fn partition_statistics(
       &self,
       partition: Option<usize>,
   ) -> Result<Arc<Statistics>> {
       self.input.partition_statistics(partition)
   }
```

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -0,0 +1,687 @@
...
+    /// Wrap `input`. If `order_by` is provided, its first entry drives
+    /// the per-partition T-Digest; the full slice is preserved for serde
+    /// and for downstream operators (`SortExec`, `BoundedWindowAggExec`)
+    /// that need it. When `Some`, at least one expression is required —
+    /// nothing to sketch on with an empty slice.
+    pub fn try_new(
+        input: Arc<dyn ExecutionPlan>,
+        order_by: Option<Vec<PhysicalSortExpr>>,
+    ) -> Result<Self> {
+        if let Some(exprs) = &order_by {
```
phillipleblanc, line 107: This should also validate that the exprs return Float64

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -0,0 +1,687 @@
...
+        // Row count is unconditional and cheap.
+        let counter = self.row_counts.get(self.partition).ok_or_else(|| {
+            internal_datafusion_err!(
+                "RuntimeStatsExec: partition {} out of range (have {}) — \
+                 execute() should have validated this",
+                self.partition,
+                self.row_counts.len()
+            )
+        })?;
+        counter.fetch_add(batch.num_rows(), Ordering::Relaxed);
```
phillipleblanc: This will be minor if the validation above is added.

This increments the row count before the below check that is evaluated array is Float64, which would incorrectly report rows that weren't forwarded.

phillipleblanc (commented):
I have a few comments, but this is looking good!

ballista/core/src/execution_plans/runtime_stats.rs
avantgardnerio: I fixed this, and also added not-null checks until we can put the KLL in that I think will support generic `Vec<Option<dyn Ord>>`

avantgardnerio:
Thanks for your feedback @phillipleblanc . As usual it was very helpful :smile:

ballista/core/src/execution_plans/runtime_stats.rs
avantgardnerio: I implemented all 6 of the algebraic ones.

phillipleblanc (approved):

andygrove:
Thanks @avantgardnerio and @phillipleblanc

PR #2095, opened 2026-07-20
feat(core): BufferExec — generic flow-control op with a memory-pool Dam mode

avantgardnerio:
## Summary

Adds `BufferExec`: a passthrough physical operator that buffers upstream batches with a policy-configurable release trigger. Currently supports one mode:

- **`BufferMode::Dam`** — buffer batches while a `MemoryReservation` grows against DataFusion's `MemoryPool`. Break the dam when `MemoryPool::try_grow` refuses further allocation (any pressure — this operator, or contention from elsewhere in the query). After the dam breaks, buffered batches drain in order and the remainder of the input passes through unchanged. No user-tunable byte threshold; the pool decides pressure, mirroring the mechanism `SortExec` uses to trigger spill.

Growth path is documented on the mode enum: `PassThrough`, `BufferAll`, `SpillOnPressure`, `BufferAndSpill`, `StageBoundary`. Only `Dam` is implemented today; the other names are fixed so future variants slot in without renaming callers. Ultimately this operator should land upstream in DataFusion so every blocking op (`SortExec`, `HashAggregateExec`, joins, ...) can share the same materialise/spill primitive instead of each rolling its own.

## Why this exists — pairing with `RuntimeStatsExec` (#2094)

Sketch-based routing has a chicken-and-egg problem: a range-repartition operator wants to pick cut points from an upstream `RuntimeStatsExec` T-Digest, but a streaming pipeline pulls one batch through the tap and hands it straight to the router — the router runs with a one-batch sample.

`BufferMode::Dam` inserted between `RuntimeStatsExec` and the sketch-consuming router closes the gap **without a plan-time size guess**:

1. Batches accumulate above the router, feeding the tap's T-Digest as they arrive.
2. Growth continues until `MemoryPool::try_grow` refuses — the pool decides \"enough sample\".
3. The dam breaks; the sketch now reflects representative data.
4. The router picks cuts once, then buffered batches (plus the input remainder) drain through in order.

The pool — not a magic byte count in code — decides when the sample is large enough. Same mechanism `SortExec` uses to decide spill timing; no new tuning knob.

## Scope of this PR

Just the operator, its proto/serde plumbing, and tests. Nothing inserts `BufferExec` into a plan yet — no rewrite rule, no scheduler hook. It lands alongside the parallel-window detection rule and the range-repartition operators (`OrderedRangeRepartitionExec` / `UnorderedRangeRepartitionExec`) that will consume the sketch.

Included:

- Proto: `BufferExecNode { BufferMode mode }` on `BallistaPhysicalPlanNode.oneof` (tag 6) and `enum BufferMode { BUFFER_MODE_DAM = 0; }`.
- Codec encode/decode registration on `BallistaPhysicalExtensionCodec`. Unknown mode discriminants surface as a decode error rather than a panic.
- No new external deps.

## Tests (4)

- `dam_drains_full_buffer_when_pool_has_headroom` — input fits within the pool, everything buffers and drains at end-of-stream.
- `dam_breaks_on_pool_pressure_and_passes_remainder` — pool tight enough that `try_grow` refuses partway through the stream; buffered rows plus the remainder all pass through, none lost or duplicated.
- `dam_propagates_oom_when_first_batch_wont_fit` — pool so tight even the first batch's reservation fails; operator surfaces OOM (mirrors `SortExec` behavior with nothing already buffered).
- `test_buffer_exec_dam_roundtrip` — plan-node serde roundtrip through the codec.

All three streaming tests use `GreedyMemoryPool` with a per-test byte cap so the dam-break branch fires without hardcoded thresholds in the operator.

## Test plan

- [x] `cargo test -p ballista-core` (121 lib + 2 doc tests pass; 4 new).
- [x] `cargo check --workspace --all-targets --locked` clean (no lock churn — no new external deps).
- [x] `cargo clippy --all-targets` clean.
- [x] `cargo fmt --all`.

## Relation to #2094

Independent PR. If #2094 (RuntimeStatsExec) lands first, this rebases to tag 7 for `BufferExecNode`; if this lands first, RuntimeStatsExec rebases the same way. No wire compatibility concerns since neither has shipped.

### DIFF
--- a/ballista/core/src/execution_plans/buffer.rs
+++ b/ballista/core/src/execution_plans/buffer.rs
@@ -1,0 +1,515 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Generic flow-control operator that DataFusion doesn't have (yet).
+//! Buffers upstream batches with a policy-configurable release trigger,
+//! then drains buffered rows and passes remaining input through.
+//!
+//! Currently supports one mode:
+//!
+//! - **`BufferMode::Dam`** — buffer input while a `MemoryReservation`
+//!   grows against DataFusion's `MemoryPool`. Break the dam when
+//!   `try_grow` refuses further allocation (any memory pressure — this
+//!   operator, or contention from elsewhere in the query). No
+//!   hardcoded byte threshold: the pool decides pressure, same mechanism
+//!   [`SortExec`] uses to trigger spill (`sort.rs:887-907`).
+//!
+//! Growth path (intentionally not built today; documented so the shape
+//! and future mode names are visible to reviewers):
+//!
+//! - `BufferMode::PassThrough` — noop, present so upstream rules can
+//!   drop the operator without restructuring the plan.
+//! - `BufferMode::BufferAll` — buffer until end-of-stream, then drain.
+//! - `BufferMode::SpillOnPressure` — same as `Dam`, but writes to disk
+//!   when the pool refuses rather than releasing to downstream. Fills
+//!   the "generic materialise/spill primitive" gap in DataFusion.
+//! - `BufferMode::StageBoundary` — the buffer IS a shuffle write.
+//!
+//! Only `Dam` is implemented for now. The other variants are named so
+//! that when they land, they slot in without renaming or restructuring.
+//!
+//! ## How this pairs with `RuntimeStatsExec`
+//!
+//! Sketch-based routing has a chicken-and-egg problem: the router
+//! wants to pick cut points from a `RuntimeStatsExec` sketch, but a
+//! streaming pipeline pulls one batch through the tap and hands it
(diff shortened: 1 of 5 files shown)


avantgardnerio:
@phillipleblanc FYI

ballista/core/src/execution_plans/buffer.rs
```diff
@@ -0,0 +1,405 @@
...
+
+impl DisplayAs for BufferExec {
+    fn fmt_as(&self, _t: DisplayFormatType, f: &mut Formatter<'_>) -> fmt::Result {
+        match &self.mode {
+            BufferMode::Dam => write!(f, "BufferExec: mode=Dam"),
+        }
+    }
+}
+
+impl ExecutionPlan for BufferExec {
```
phillipleblanc, line 185: Are there other methods we might need to implement for this exec? In general I prefer to be explicit and always implement default trait methods, even if we keep the default behavior to make it clear that it was a choice that we made.

ballista/core/src/execution_plans/buffer.rs
```diff
@@ -0,0 +1,405 @@
...
+            futures::stream::iter(buffered.into_iter().map(Ok::<_, DataFusionError>));
+        let out: futures::stream::BoxStream<'static, Result<RecordBatch>> = if dam_broken
+        {
+            drained.chain(input).boxed()
+        } else {
+            drained.boxed()
+        };
+        // Drop the reservation explicitly at the end of the fill phase —
+        // downstream can allocate freely from here on.
+        drop(reservation);
```
phillipleblanc: This drops the reservation before the stream has actually drained any batches, which could allow memory usage beyond the pool limits. Could we decrement the memory reservation incrementally as the stream is drained?

avantgardnerio: Thanks,  that was a major, load-bearing catch :raised_hands:

phillipleblanc (approved):
This makes sense to me, thanks @avantgardnerio.

andygrove:
This one needs a rebase @avantgardnerio

avantgardnerio:
@andygrove rebase complete, if you have a moment for review :bow:

andygrove:
Thanks @avantgardnerio and @phillipleblanc

PR #2102, opened 2026-07-20
feat(benchmarks): persist TPC-H results resiliently and add a compare subcommand

### LINKED ISSUE #2101: Benchmarks: persist partial/failed TPC-H results and add a compare tool

The `tpch` benchmark binary already writes a machine-readable JSON summary with `--output`, but two gaps make it awkward to use for tracking performance across refs:

1. **Results are lost when a query fails.** Both the `ballista` and `datafusion` run loops only write the summary *after* the whole loop, and a failing query aborts the run (`?`/`.unwrap()` on `ctx.sql`/`collect`). So a single failure — or an out-of-memory `SIGKILL` at a large scale factor — discards every result collected so far. This is easy to hit at SF1000.

2. **There is no built-in way to compare two runs.** Comparing a baseline against a candidate (per-query delta, totals, row-count agreement) has to be done by hand or with an out-of-tree script.

- Persist the summary JSON after **every** query (atomically), so a killed run keeps completed results.
- Record a failed query with an `error` message plus whatever iterations completed, and continue to the remaining queries; still exit non-zero if any query failed.
- Add a `tpch compare <baseline.json> <candidate.json>` subcommand that diffs two runs query-by-query: fastest iteration per side, delta, delta %, suite totals, and row-count agreement.

### Additional context

Complements the benchmarking guide added in #2075 by making the numbers that page records reproducible and comparable. Related to #1630.

andygrove:

Closes #2101.

# Rationale for this change

The `tpch` benchmark binary already writes a JSON summary with `--output`, but it is fragile as a basis for tracking performance across refs:

- The summary is written only *after* the whole run, and a failing query aborts the loop (`?`/`.unwrap()` on `ctx.sql`/`collect`). One failure — or an OOM `SIGKILL` at a large scale factor — discards every result collected so far.
- There is no in-tree way to compare two runs.

- **Resilient output.** The run summary is written after every query, atomically (temp file + rename), so a suite killed part way through keeps the results collected so far.
- **Continue on error.** A query that fails is recorded with an `error` message and its completed iterations, and the run continues to the remaining queries. The process still exits non-zero if any query failed. `QueryRun` gains an `error: Option<String>` field (omitted on success); the run structs now also derive `Deserialize`.
- **`compare` subcommand.** `tpch compare <baseline.json> <candidate.json>` diffs two result files query-by-query — fastest iteration per side, delta, delta %, suite totals, and row-count agreement (flags mismatches).
- Refactors each run loop's per-query body into a helper so the error handling is shared and testable.
- Unit tests for the min-of-iterations metric, the failure/serialization round-trip, and the comparison logic. README documents `--output`, the resilience behaviour, and `compare`.

Yes, all additive:
- `tpch compare` is a new subcommand.
- The JSON summary is written incrementally and gains an optional per-query `error` field.
- A benchmark run no longer aborts on the first failing query; it records the failure and continues, exiting non-zero at the end if any query failed.

No breaking API changes.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -165,6 +165,28 @@ cd $ARROW_HOME/benchmarks
 cargo run --release --bin tpch benchmark ballista --host localhost --port 50050 --query 1 --path $(pwd)/data --format tbl
 ```
 
+## Recording and comparing results
+
+Pass `--output <dir>` (works for both the `ballista` and `datafusion` benchmark
+subcommands) to write a machine-readable summary to `<dir>/tpch-<start_time>.json`.
+Each run records the benchmark and DataFusion versions, CPU count, the full CLI
+arguments (so the configuration is self-describing), and per-query per-iteration
+elapsed times and row counts.
+
+The summary is rewritten after **every query**, atomically, so a run that is
+killed part way through (for example an out-of-memory `SIGKILL` at a large scale
+factor) still leaves the results collected so far on disk. A query that fails is
+recorded with an `error` message and its completed iterations, and the run
+continues to the remaining queries instead of aborting; the process still exits
+non-zero if any query failed.
+
+Compare two summary files query-by-query — fastest iteration per side, delta,
+delta percent, suite totals, and row-count agreement:
+
+```bash
+cargo run --release --bin tpch compare baseline.json candidate.json
+```
+
 ## Running the Ballista Benchmarks on docker-compose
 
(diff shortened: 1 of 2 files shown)


milenkovicm (approved):
LGTM thanks @andygrove

avantgardnerio (approved):
I assume this came out of all your benchmarking this weekend. Thanks for doing that. LGTM

andygrove:
Merged. Thanks for the review @milenkovicm @avantgardnerio

PR #2119, opened 2026-07-21
Retryable IO errors misclassified when wrapped in DataFusionError::Shared

### LINKED ISSUE #2028: Retryable IO errors are misclassified as non-retryable when wrapped in DataFusionError::Shared
## Describe the bug

A genuinely retryable IO error is classified as **non-retryable** — and so fails the job instead of being retried — whenever it reaches the classifier wrapped inside `DataFusionError::Shared`. This happens in practice for errors raised under a join's shared build side.

## Root cause

`ballista/core/src/error.rs` (~line 237) decides retryability with a **shallow** match on the inner error:

```rust
BallistaError::DataFusionError(e) if matches!(*e, DataFusionError::IoError(_)) => FailedTask {
    retryable: true,
    count_to_failures: true,
    failed_reason: Some(FailedReason::IoError(IoError {})),
}
```

`matches!` inspects only the outermost variant. DataFusion wraps errors in `DataFusionError::Shared` (an `Arc<DataFusionError>`) when an error is shared across multiple consumers — which is exactly what happens on a hash join's build side. A `DataFusionError::Shared(Arc(IoError(..)))` does not match `DataFusionError::IoError(_)`.

The error therefore falls through to the catch-all (~line 248):

```rust
other => FailedTask {
    retryable: false,
    count_to_failures: false,
    failed_reason: Some(FailedReason::ExecutionError(ExecutionError {})),
}
```

so the task is never retried and the job fails.

This affects **real Parquet/object-store IO errors**, not only synthetic ones: a transient S3 or local-IO failure on a join's build side is precisely the case Ballista's retry logic exists to absorb, and it is the case where it silently does not apply.

## To Reproduce

Reproduced end to end on a real multi-process cluster by the chaos harness in #2026 (scenario `retryable_fault_is_retried_and_result_is_correct`, AQE-on case).

The harness injects exactly one `DataFusionError::IoError` from a UDF on a join input, with a cluster-wide budget of one fault, so the retry is guaranteed to succeed if it happens at all:

- **AQE off:** the error reaches the classifier unwrapped → correctly marked retryable → task retried → **query succeeds with the correct result.** ✅
- **AQE on:** the same error arrives wrapped in `DataFusionError::Shared` (shared join build side) → misses the `matches!` arm → marked `retryable: false` → **the job fails.** ❌

The difference between the two cases isolates the wrapping as the cause.

## Expected behavior

Retryability should be determined by the *root* error, not the outermost wrapper. An `IoError` should be classified `retryable: true` regardless of whether it arrives bare, inside `Shared`, or inside any other wrapping variant.

DataFusion provides `DataFusionError::find_root()` for exactly this; classification in `error.rs` should unwrap before matching.

## Additional context

Sibling of the `FetchFailed` issue: **structured errors lose their type crossing the DataFusion boundary, and retryability is decided by shallow matches that cannot see through the wrapping.** Fixing both probably wants a single, well-tested `classify(&BallistaError) -> FailedTask` that unwraps first.

goingforstudying-ctrl:
Hit this while running queries with joins against flaky object storage: a transient IO error on the build side of a hash join kills the whole job instead of getting retried.

Turned out the retry classifier in `ballista/core/src/error.rs` does a shallow `matches!(*e, DataFusionError::IoError(_))` on the outermost variant. But errors coming off a join's shared build side get wrapped in `DataFusionError::Shared` (an `Arc` for sharing across consumers), so the `IoError` inside is never seen and the task falls through to the catch-all `retryable: false` arm. AQE off works fine because the error arrives unwrapped, which is how I narrowed it down to the wrapping.

DataFusion already has `find_root()` for exactly this, so the fix is to classify on the root error instead of the wrapper. One-line change plus unit tests covering bare, `Shared`-wrapped, `Context`-wrapped-`Shared`, and non-IO cases.

Closes #2028.

Tested with `cargo test -p ballista-core --lib error::` (5 new tests pass) and `cargo clippy -p ballista-core --lib` is clean. I couldn't run the full cluster chaos harness from #2026 locally, but the unit tests cover the classifier paths directly.

### DIFF
--- a/ballista/core/src/error.rs
+++ b/ballista/core/src/error.rs
@@ -235,7 +235,7 @@ impl From<BallistaError> for FailedTask {
                 }
             }
             BallistaError::DataFusionError(e)
-                if matches!(*e, DataFusionError::IoError(_)) =>
+                if matches!(e.find_root(), DataFusionError::IoError(_)) =>
             {
                 FailedTask {
                     error: format!("Task failed due to DataFusion IO error: {e:?}"),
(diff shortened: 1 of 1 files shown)


phillipleblanc (approved):
This makes sense to me, thanks!

martin-g (approved):

andygrove (approved):
Thanks @goingforstudying-ctrl!

PR #2120, opened 2026-07-21
ci(tpch): add AQE-on smoke pass with multi-partition tasks

avantgardnerio:
Reduces build time 16m -> 9m while adding MPT tests.

## Summary
- Existing CI runs at the default `ballista.scheduler.max_partitions_per_task=1` (one task per input partition), so the multi-partition-task path — currently has no end-to-end coverage.
- Add a third `run_suite` invocation to `tpch.yml` at `AQE on + max_partitions_per_task=0` (unbounded, budget-capped at `vcores=4`), reusing the same scheduler/executor/data

## Test plan
- [ ] `TPC-H SF10` workflow shows a new `[AQE on, multi-partition tasks] Query N` group for every non-q16 query and passes `--verify` on each.
- [ ] Existing `AQE off` and `AQE on` groups still run and still pass.

### DIFF
--- a/.github/workflows/tpch.yml
+++ b/.github/workflows/tpch.yml
@@ -179,6 +179,11 @@ jobs:
             --verify \
             -c datafusion.optimizer.prefer_hash_join=false \
             -c ballista.planner.adaptive.enabled=true
+          run_suite "AQE on, multi-partition tasks" \
+            --verify \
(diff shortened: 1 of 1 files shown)


phillipleblanc (approved):

andygrove (approved):
Thanks @avantgardnerio

PR #2123, opened 2026-07-21
feat(core): add UnorderedRangeRepartitionExec — value-range router over unordered inputs

avantgardnerio:
## Summary

Adds `UnorderedRangeRepartitionExec`: a physical operator that reads P input partitions with no sort assumption, evaluates the first ORDER BY expression per row, and routes each row to one of K output partitions under the half-open convention: partition `p` owns `[cut[p-1], cut[p])` with virtual `-∞`/`+∞` sentinels on the ends.

Also adds `range_repartition_common.rs` — the shared building blocks (`discover_cuts`, `find_runtime_stats`, `preserves_distribution`, `split_batch_by_range`, `broadcast_error`) that the pending `OrderedRangeRepartitionExec` follow-up will also consume.

## Why — closing the loop on #2094 + #2095

Sketch-based routing needs three pieces:

1. **`RuntimeStatsExec`** (#2094) — the tap that accumulates the T-Digest as batches flow past.
2. **`BufferExec`** (#2095) — the dam that holds batches back long enough for the tap to see representative data.
3. **This PR** — the router that reads the sketch and does the actual repartitioning.

Boundaries are discovered *at runtime*, not baked in at plan time. On the first batch to arrive from any input partition, the operator walks its own child subtree for a matching sibling `RuntimeStatsExec`, snapshots its merged T-Digest, and computes `K − 1` quantile cuts at `1/K, 2/K, …, (K−1)/K`. All batches (including the one that triggered the snapshot) then route through those cuts.

The walker only descends through distribution-preserving operators (`preserves_distribution` whitelist). `BufferExec` is on the whitelist so a Dam between the tap and this router doesn't hide the sketch. `FilterExec`, unions, etc. are *not* on it — they'd stale the distribution.

Every discovery failure path (no matching stats, no sketch, empty sketch, mutex poisoning) falls back to a single-bucket routing that lands every row in output partition 0. Runtime routing must never crash — degraded-but-alive beats the alternative.

## Scope

Just the operator, its shared common module, proto/serde plumbing, and tests. Nothing inserts `UnorderedRangeRepartitionExec` into a plan yet — no rewrite rule, no scheduler hook. Those land with the parallel-window detection rule alongside the `OrderedRangeRepartitionExec` sibling.

Included:

- `range_repartition_common.rs` — `pub(super)` helpers; imports `BufferExec` + `RuntimeStatsExec` from main.
- `unordered_range_repartition.rs` — the operator itself + 12 unit tests.
- Proto: `UnorderedRangeRepartitionExecNode { repeated PhysicalSortExprNode order_by = 1; uint32 output_partitions = 2; }` on `BallistaPhysicalPlanNode.oneof` (tag 8; tags 6 and 7 taken by `RuntimeStatsExec` and `BufferExec`).
- No new external deps.

## Drift from source branch

- `RuntimeStatsExec::merged_quantile_sketch()` returned `Option<TDigest>` when this code was authored; on main it returns `Result<Option<TDigest>>` (mutex-poisoning was made explicit during the #2094 upstreaming). Adapted `discover_cuts` to distinguish `Ok(None)` (sketch not allocated) from `Err(_)` (snapshot failed) in the warn message; both still degrade to single-bucket fallback.
- `RuntimeStatsExec` now rejects nullable routing expressions at construction (T-Digest has no NULL slot). The original test suite declared `v2` as nullable and typed the batch builder as `Vec<Option<f64>>` but never actually populated a `None`. Tightened the schema + helper types to non-nullable `f64`; will loosen again when we swap T-Digest for KLL and NULL becomes a first-class routing key.

## Tests (14 new)

12 operator unit tests covering constructor validation, output-partitioning shape, discovery walker's descend/refuse policy, end-to-end row conservation, range-disjointness of outputs, multi-input-partition handling, and every discovery-failure fallback path.

2 serde roundtrip tests (single-key + multi-key ORDER BY) — the wire format must not silently drop non-primary ORDER BY entries (they're the substrate for the ROWS-frame follow-up).

## Test plan

- [x] `cargo test -p ballista-core` — 158/158 pass (14 new).

## Next in the stack

`OrderedRangeRepartitionExec` — same discovery mechanism, adds per-output k-way merge to preserve sort order. Reuses this PR's `range_repartition_common.rs` verbatim.

### DIFF
--- a/ballista/core/src/execution_plans/range_repartition_common.rs
+++ b/ballista/core/src/execution_plans/range_repartition_common.rs
@@ -40,7 +40,9 @@ use std::sync::Arc;
 
 use datafusion::arrow::array::{Array, Float64Array, RecordBatch, UInt32Array};
 use datafusion::arrow::compute::take_arrays;
+use datafusion::common::runtime::SpawnedTask;
 use datafusion::common::{Result, internal_datafusion_err};
+use datafusion::error::DataFusionError;
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::sorts::sort::SortExec;
@@ -277,6 +279,42 @@ pub(super) async fn broadcast_error(
             None => Err(internal_datafusion_err!("{}", message)),
         };
         let _ = sender.send(payload).await;
+    }
+}
+
+/// Await a scatter task and turn every terminal state into something the K
+/// downstream consumers can observe. Modelled on DataFusion's
+/// `RepartitionExec::wait_for_task`.
+///
+/// Without this, a panic in a scatter task closes its senders with no data
+/// sent — every output partition sees a clean EOF and the overall task
+/// reports success while silently dropping all remaining rows. Here every
+/// terminal state (panic, DFError, clean end) becomes an explicit signal
+/// on every output channel.
+pub(super) async fn wait_for_task(
+    input_task: SpawnedTask<Result<()>>,
+    senders: Arc<[mpsc::Sender<Result<RecordBatch>>]>,
+) {
+    match input_task.join().await {
+        Err(join_err) => {
+            let ctx = if join_err.is_panic() {
+                "scatter task panicked"
+            } else {
+                "scatter task cancelled"
+            };
+            broadcast_error(
+                &senders,
+                DataFusionError::Context(
+                    ctx.to_string(),
+                    Box::new(DataFusionError::External(Box::new(join_err))),
+                ),
+            )
+            .await;
+        }
+        Ok(Err(err)) => broadcast_error(&senders, err).await,
+        Ok(Ok(())) => {
+            // Drop the senders — receivers see clean EOF.
+        }
     }
 }
 
--- a/ballista/core/src/execution_plans/range_repartition_common.rs
+++ b/ballista/core/src/execution_plans/range_repartition_common.rs
@@ -36,17 +36,19 @@
 //! [`UnorderedRangeRepartitionExec`]: super::UnorderedRangeRepartitionExec
 //! [`RuntimeStatsExec`]: super::RuntimeStatsExec
 
+use std::any::Any;
+use std::future::Future;
+use std::panic::AssertUnwindSafe;
 use std::sync::Arc;
 
 use datafusion::arrow::array::{Array, Float64Array, RecordBatch, UInt32Array};
 use datafusion::arrow::compute::take_arrays;
-use datafusion::common::runtime::SpawnedTask;
 use datafusion::common::{Result, internal_datafusion_err};
-use datafusion::error::DataFusionError;
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::sorts::sort::SortExec;
 use datafusion::physical_plan::windows::{BoundedWindowAggExec, WindowAggExec};
+use futures::FutureExt;
 use log::warn;
 use tokio::sync::mpsc;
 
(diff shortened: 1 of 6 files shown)


avantgardnerio:
@phillipleblanc we're finally getting into the interesting stuff :smile:

ballista/core/src/execution_plans/unordered_range_repartition.rs
```diff
@@ -0,0 +1,859 @@
...
+            let input_partitions = self.input.output_partitioning().partition_count();
+            let routing_expr = self.order_by[0].expr.clone(); // TODO: KLL for multi-column?
+            for input_partition in 0..input_partitions {
+                let child = self.input.clone();
+                let senders = senders.clone();
+                let cuts_cell = cuts_cell.clone();
+                let routing_expr = routing_expr.clone();
+                let ctx = ctx.clone();
+                let output_partitions = self.output_partitions;
+                tokio::spawn(async move {
```
phillipleblanc: This `JoinHandle` is being discarded, so if `scatter_input_partition` panics then the sender is silently dropped, which would result in a "successful" task that returned wrong results. I believe the DataFusion `RepartitionExec` has a good model we could follow: https://github.com/apache/datafusion/blob/45d943dfb8699dc9cb9ef2320e955b73e3e6c03b/datafusion/physical-plan/src/repartition/mod.rs#L528-L542

ballista/core/src/execution_plans/unordered_range_repartition.rs
```diff
@@ -0,0 +1,859 @@
...
+        }
+        // TODO: fixed by KLL — a NULL-aware sketch lifts this restriction and
+        // lets `split_batch_by_range` honor SortOptions::nulls_first properly.
+        if routing.expr.nullable(&schema)? {
+            return internal_err!(
+                "UnorderedRangeRepartitionExec: routing expression `{}` must be non-nullable",
+                routing.expr
+            );
+        }
+        let properties = Arc::new(PlanProperties::new(
```
phillipleblanc: `PlanProperties::new` defaults to `EvaluationType::Lazy` and `SchedulingType::NonCooperative`. Since this spawns producers that eagerly drive each input partition, we should set `EvaluationType::Eager` and `SchedulingType::Cooperative` (similar to what DataFusion RepartitionExec does)

ballista/core/src/execution_plans/range_repartition_common.rs
```diff
@@ -0,0 +1,375 @@
...
+/// on its way to a matching [`RuntimeStatsExec`]. Unlisted operators might
+/// drop rows, duplicate rows, or transform the routing key's value — any of
+/// which would make an upstream sketch stale by the time data reaches us.
+///
+/// Being conservative is the safety net: unrecognized node → walker gives
+/// up → single-bucket fallback. Extending this list requires positive
+/// verification that the operator is a distribution-preserving passthrough
+/// for the routing key. Absent an upstream `ExecutionPlan::affects_distribution()`
+/// method (nice-to-have that hopefully lands one day), we maintain this by hand.
+pub(super) fn preserves_distribution(plan: &dyn ExecutionPlan) -> bool {
```
phillipleblanc, line 168: We may want a way for library consumers to also register their own exec plans here that are passthrough nodes. But probably an upstream trait type is the correct approach.

phillipleblanc (commented):
Thanks! I think the only blocking item from my side is dropping the JoinHandle for `scatter_input_partition`. Everything else looks good to me.

ballista/core/src/execution_plans/unordered_range_repartition.rs
avantgardnerio: Thanks @phillipleblanc , you keep finding structural stuff and I'm grateful for your reviews. I'll make sure I handle this class of error in future PRs.

I agree the `RepartitionExec` model is good, and battle tested. But it seemed unnecessarily complex? It has a spawned task just to await a spawned task. Instead I went with `AssertUnwindSafe`. Let me know what you think. I can easily change it to just be like `RepartitionExec`.

ballista/core/src/execution_plans/range_repartition_common.rs
avantgardnerio: I would strongly prefer to upstream. Once this is all working and making a benchmark faster, I'd like to submit all of the upstreamable parts to DF for review.

phillipleblanc (approved):

ballista/core/src/execution_plans/unordered_range_repartition.rs
phillipleblanc: Makes sense to me. Agree with keeping things simple.

andygrove (approved):
LGTM. This is purely additive and doesn't impact existing users. I look forward to seeing it wired up in the future. Thanks @avantgardnerio

PR #2125, opened 2026-07-21
docs: correct scan partitioning description in tuning guide

andygrove:

N/A - documentation fix, no issue filed.

# Rationale for this change

The "Partitions and Parallelism" section of the tuning guide describes behavior that is no longer accurate, and the inaccuracy points users in the wrong direction when they size their clusters.

It states that "Ballista currently treats each file within a table as a single partition (in the future, Ballista will support splitting files into partitions but this is not implemented yet)", and that "Ballista will never decrease the number of partitions based on this setting". Neither is true today:

- File listing chunks the table's files into **at most** `target_partitions` file groups (`FileGroup::split_files`), so a 200 file table read with the default `target_partitions = 16` gives 16 partitions, not 200. The setting does decrease partition count.
- The physical optimizer then splits individual files by byte range (`datafusion.optimizer.repartition_file_scans`, gated on `datafusion.optimizer.repartition_file_min_size`), so a table consisting of one large Parquet file does scale across the cluster, contrary to what the guide warns.

The guide also never mentioned `ballista.scheduler.max_partitions_per_task`, which defaults to `1` and therefore determines the task count for every stage. Someone tuning `--vcores` without knowing about it cannot explain why a 16 partition stage produces 16 tasks.

- Rewrites the scan partitioning explanation to describe the two actual steps (file group chunking capped at `target_partitions`, then byte range splitting within files) and notes that Ballista disables round robin repartitioning, so file group splitting is the only source of scan parallelism.
- Adds a "Partitions and Tasks" subsection covering the partition to task binding, `ballista.scheduler.max_partitions_per_task` (default `1`, `0` for unbounded), and the collapse stage exemption.

Documentation only. No code or API changes.

### DIFF
--- a/docs/source/user-guide/tuning-guide.md
+++ b/docs/source/user-guide/tuning-guide.md
@@ -26,14 +26,20 @@ by adding more compute resource.
 
 The basic unit of concurrency and parallelism in Ballista is the concept of a partition. The leaf nodes of a query
 are typically table scans that read files from object storage, and the number of partitions such a scan produces is
-driven by `datafusion.execution.target_partitions`. This is currently a global setting for the entire context. The
-default value for this setting is 16.
+driven by `datafusion.execution.target_partitions`. This is currently a global setting for the entire context.
+
+DataFusion's own default for this setting is the number of CPU cores available to the process. Ballista overrides it
+to 16 in `SessionConfig::new_with_ballista()`, so a context created with `SessionContext::remote()` or
+`SessionContext::standalone()` starts at 16. If you instead build your own `SessionConfig` and pass it to
+`remote_with_state()` or `standalone_with_state()`, Ballista leaves the setting alone and you get whatever that
+config carries, which is the client machine's core count for a plain `SessionConfig::new()`. Set it explicitly if
+you care about the value.
 
 A scan is partitioned in two steps:
 
 1. When the table is listed, its files are sorted by path and distributed into **at most** `target_partitions`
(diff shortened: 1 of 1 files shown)


avantgardnerio (approved):

docs/source/user-guide/tuning-guide.md
```diff
@@ -25,18 +25,31 @@ The goal of any distributed compute engine is to parallelize work as much as pos
...
-The configuration setting `datafusion.execution.target_partitions`can be set to the desired number of partitions. This is
-currently a global setting for the entire context. The default value for this setting is 16.
-
-Note that Ballista will never decrease the number of partitions based on this setting and will only repartition if
-the source operation has fewer partitions than this setting.
+are typically table scans that read files from object storage, and the number of partitions such a scan produces is
+driven by `datafusion.execution.target_partitions`. This is currently a global setting for the entire context.
+
+DataFusion's own default for this setting is the number of CPU cores available to the process. Ballista overrides it
+to 16 in `SessionConfig::new_with_ballista()`, so a context created with `SessionContext::remote()` or
```
milenkovicm, line 32: Would it make sense to increase default value of 16?

milenkovicm (approved):

docs/source/user-guide/tuning-guide.md
andygrove: I filed https://github.com/apache/datafusion-ballista/issues/2128 to discuss this

PR #2143, opened 2026-07-22
feat(core): report the actual sort shuffle spill trigger in logs

andygrove:

N/A - no issue filed for this. It is a small observability fix found while reading spill logs.

# Rationale for this change

The sort shuffle writer spills for two independent reasons:

1. The runtime `MemoryPool` rejects a reservation grow, meaning the executor as a whole is out of memory budget.
2. The per-task buffered-bytes budget (`ballista.shuffle.sort_based.memory_limit_per_task_bytes`) is reached, meaning this one writer hit its own configured cap while the pool still had room.

Both cases logged the same thing:

```
Sort shuffle spill: job=... stage=... input_partition=... spilled N bytes in M batches (K events) under memory pressure; ...
```

"Under memory pressure" is only accurate for the first case, and the two have different remedies. Pool rejection means the executor needs more memory or fewer concurrent tasks. Budget exhaustion means `memory_limit_per_task_bytes` is set too low for this workload. Anyone tuning a spilling query had no way to tell which one they were looking at.

- Add a `SpillTrigger` enum (`MemoryPool`, `TaskBudget`, `Both`) with a `classify` constructor. The spill condition in the write loop is now driven by `SpillTrigger::classify(...)`, so the branch that decides to spill and the cause that gets logged come from the same value and cannot drift apart. Both triggers can fire on the same input batch, hence the `Both` variant rather than picking a winner.
- Add `SpillTriggerCounts` to tally events per trigger over the life of the task, rendering only the triggers that actually fired.
- The end-of-task WARN now reports the breakdown and the configured budget:

```
Sort shuffle spill: job=j1 stage=2 input_partition=0 spilled 812345 bytes in 24 batches (6 events); triggered by: memory pool rejection x4, per-task buffer budget x2 (per-task buffer budget: 268435456 bytes); repart_time=... spill_time=... write_time=...
```

  The budget renders as `disabled` when `memory_limit_per_task_bytes` is 0, which is the documented way to leave the pool as the sole trigger.
- The per-event DEBUG line also names its trigger.
- Correct the doc comments in `spill.rs` and on the `spill_count` metric that repeated the same inaccurate "memory pressure" claim.
- Two unit tests cover the classification matrix and the summary rendering, including an assertion that the two causes never render with identical wording.

Log output only. The WARN emitted when a sort shuffle task spills has a new format and names the trigger. No public API, configuration, or behavior changes. Anything parsing that log line for the literal string "under memory pressure" would need updating.

### DIFF
--- a/ballista/core/src/execution_plans/sort_shuffle/spill.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/spill.rs
@@ -17,7 +17,9 @@
 
 //! Spill manager for sort-based shuffle.
 //!
-//! Handles writing partition buffers to disk when memory pressure is high.
+//! Handles writing partition buffers to disk when the writer decides to flush
+//! them, either because the runtime `MemoryPool` rejected a reservation grow or
+//! because the per-task buffer budget was reached.
 //! At finalization, the spill bytes are concatenated verbatim into the
 //! consolidated output file alongside the in-memory remainder.
 
@@ -191,7 +193,7 @@ impl SpillManager {
 
     /// Returns the total number of batches written to spill files across all
     /// partitions. Note this counts batches, not spill *events*: a single
-    /// memory-pressure event in the writer typically produces one batch per
+    /// spill event in the writer typically produces one batch per
     /// non-empty output partition. Spill-event accounting lives at the writer
     /// layer because the spill manager only sees batch-level calls.
     pub fn total_spilled_batches(&self) -> u64 {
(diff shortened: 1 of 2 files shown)


andygrove:
I am testing in EKS now and I found the current messages misleading. I had forgotten that there is a default config limiting shuffle memory to 256 MB.

avantgardnerio (approved):

PR #2148, opened 2026-07-22
test(tpcds): bump TPC-DS test data to scale factor 10

andygrove:

N/A

 # Rationale for this change

The TPC-DS correctness gate has been running at SF1. At that size several
tables fit in a single batch, so the suite mostly exercises the planner and
barely touches the distributed execution paths that actually break: shuffle
writes/reads across multiple partitions, repartitioning, and spilling.

Running the same 99-query suite at SF10 keeps the exact same coverage in terms
of queries but puts real data volume behind it, which is where distributed
result divergence tends to show up.

- `.github/workflows/tpcds.yml`: generate the data with `--scale-factor 10`
  instead of `1`, and rename the workflow, job, and log artifact to match.
- `benchmarks/tpcds-gen.sh`: default `SCALE_FACTOR` to `10` so a local run
  reproduces CI without an env override. It is still overridable.
- `benchmarks/README.md`: update the SF references in the data generation and
  CI sections.

No change to the query set or the skip list.

No. This only affects CI and the benchmark data generation script.

### DIFF
--- a/.github/workflows/tpcds.yml
+++ b/.github/workflows/tpcds.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: TPC-DS SF1
+name: TPC-DS SF10
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


andygrove:
This takes far too long to run

Issue #2158, opened 2026-07-23
Epic: investigate and upstream Spice AI scheduler/executor reliability fixes


andygrove:
## Overview

The Spice AI fork of Ballista ([spiceai/datafusion-ballista](https://github.com/spiceai/datafusion-ballista)) has accumulated a set of scheduler and executor reliability fixes that are directly relevant to distributed execution stability, most notably the executor heartbeat/hang failures tracked in #2144. Phillip LeBlanc (Spice) noted on #2144 that they have fixes still to be upstreamed.

This epic tracks investigating each change, validating it against our current `main` (the fork sits on a diverged base, so nothing is directly cherry-pickable), and porting the ones we want.

## Screening results (checked against current `main`)

Most of the fork's fixes turned out to be **already present** in upstream, either ported independently or obsoleted by upstream's own evolution (the fork branches sit on an older base). Remaining real work is small.

| Fix | Status |
| --- | --- |
| Keep heartbeating while all task slots are busy | ✅ **Ported — #2159** |
| Ignore stale `TaskStatus` for reset partitions | ✅ Already upstream — append-slot `task_infos` model marks reset tasks `Failed(ResultLost)` and `update_task_info` already guards late updates (regression test `test_update_task_info_after_executor_lost`). The fork's `Option::unwrap` panic is structurally impossible here. |
| Pool shuffle-fetch clients per peer | ✅ Already upstream — `BallistaClientPool` / `ShuffleReaderExec::with_client_pool`. |
| Treat a missing partition file as an empty partition | ✅ Obsolete — the sort-shuffle writer now writes one consolidated `data.arrow` per input partition with a per-partition offset index and a leading schema header, so every partition (including 0-row) resolves; there is no missing-file case. The fork's fix targeted the old per-output-partition lazy-file writer that upstream removed (6540d085). |
| External semaphore for executor task concurrency | ✅ Already upstream — `poll_loop` takes `free_vcores: Option<Arc<Semaphore>>`. |
| Broadcast channel for job state events | ✅ Already upstream (equivalent) — `ClusterState::job_state_events()` / `JobStateEvent` stream. |
| Persist terminal job status before removing from active cache | ✅ Already upstream — `persist_terminal_and_evict` updates the cached status, saves, then evicts. |
[...]

Issue #2144, opened 2026-07-22
TPC-H q18 hangs with 16 executors x 8 cores in EKS


andygrove:
**Describe the bug**
TPC-H q18 hangs with 16 executors x 8 cores in EKS. Executors show no activity after some time. 

```
            "datafusion.execution.target_partitions": str(target_partitions),
            "datafusion.execution.listing_table_factory_infer_partitions": "false",
            "datafusion.catalog.information_schema": "true",
            "datafusion.execution.collect_statistics": "true",
            "ballista.planner.adaptive.enabled": "true",
            "ballista.shuffle.sort_based.memory_limit_per_task_bytes": "0",
```

**To Reproduce**
TBD

**Expected behavior**

**Additional context**

andygrove:
Same for q9.

Saw this in scheduler log:

```
│ scheduler 2026-07-22T16:26:33.516100Z  INFO tokio-rt-worker ThreadId(09) ballista_scheduler::state::executor_manager: Removing executor 0c71e449-aa94-4ed9-aae8-87bdae4a8a6f: Some("ACTIVE executor 0c71e449-aa94-4ed9-aae8-87bdae4a8a6f heartbeat timed out aft │
│ scheduler 2026-07-22T16:26:33.516101Z  INFO tokio-rt-worker ThreadId(05) ballista_scheduler::state::executor_manager: Removing executor 3eca16d6-82ef-4456-82ad-2e964c052f62: Some("ACTIVE executor 3eca16d6-82ef-4456-82ad-2e964c052f62 heartbeat timed out aft │
│ scheduler 2026-07-22T16:26:33.516085Z  INFO tokio-rt-worker ThreadId(06) ballista_scheduler::state::executor_manager: Removing executor 8e93fa02-1677-4871-8c12-6c1859bf5ef0: Some("ACTIVE executor 8e93fa02-1677-4871-8c12-6c1859bf5ef0 heartbeat timed out aft │
│ scheduler 2026-07-22T16:26:33.516094Z  INFO tokio-rt-worker ThreadId(03) ballista_scheduler::state::executor_manager: Removing executor 9d0bcab9-e662-4f67-a0a0-7de26db4aaf5: Some("ACTIVE executor 9d0bcab9-e662-4f67-a0a0-7de26db4aaf5 heartbeat timed out aft │
[...]

PR #2159, opened 2026-07-23
fix(executor): keep heartbeating while all task slots are busy

andygrove:

Part of #2158. Contributes toward #2144.

# Rationale for this change

Under pull-based scheduling the executor's only heartbeat to the scheduler is `poll_work` — the scheduler records a heartbeat on every poll. The poll loop, however, blocked on `free_vcores.acquire().await` *before* polling, waiting for a free vcore to bind new work.

When every vcore is held by a task running longer than the scheduler's `executor_timeout` (e.g. a long stage-1 scan on a wide TPC-H query such as q9/q18), the loop blocked on `acquire` indefinitely and never polled. The scheduler then stopped receiving heartbeats, marked the healthy-but-busy executor dead, and reset its in-flight (and completed) tasks. When a stage's tasks consistently exceed the timeout this livelocks: `reset -> re-dispatch -> reset`, never converging — the executor goes idle and the job hangs.

This is one of the root causes behind the fleet-wide heartbeat timeouts reported in #2144. It was originally found and fixed in the Spice AI fork; this ports the fix to current `main` (the loop has since been refactored to use `free_vcores`/`num_free_vcores`, so the fix is adapted accordingly).

- Cap the vcore wait at a new `HEARTBEAT_POLL_INTERVAL` (5s, well below the scheduler's executor timeout) using `tokio::time::timeout`.
- On timeout, poll the scheduler anyway, reporting `num_free_vcores: 0` (heartbeat only), so executor liveness no longer depends on vcore availability.
- A freed vcore still wakes the wait immediately, so scheduling responsiveness is unchanged; when fully busy the loop now polls roughly every 5s instead of never.
- Handle a closed semaphore (executor shutdown) explicitly.

No. Behavior only changes for fully-busy executors, which now continue to heartbeat instead of being wrongly declared dead. No public API changes.

### DIFF
--- a/ballista/executor/src/execution_loop.rs
+++ b/ballista/executor/src/execution_loop.rs
@@ -49,6 +49,13 @@ use std::{sync::Arc, time::Duration};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tonic::codegen::{Body, Bytes, StdError};
 
+/// Maximum time the poll loop waits for a free vcore before polling the
+/// scheduler anyway. `poll_work` doubles as the executor's heartbeat under
+/// pull-based scheduling, so a fully-busy executor must keep polling (reporting
+/// zero free vcores) or the scheduler times it out and resets its tasks. Kept
+/// well below the scheduler's executor timeout.
+const HEARTBEAT_POLL_INTERVAL: Duration = Duration::from_secs(5);
+
 /// Main execution loop that polls the scheduler for available tasks.
 ///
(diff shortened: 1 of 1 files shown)


andygrove:
@phillipleblanc thanks for fixing this in your fork

milenkovicm (approved):
makes sense, thanks @andygrove

avantgardnerio (approved):

phillipleblanc (approved):
Thanks!

PR #2161, opened 2026-07-23
docs: refresh benchmarking.md with current SF1000 results (Ballista main @ 696ca29b, Spark 3.4 baseline)

andygrove:
## Summary

Refresh `docs/source/contributors-guide/benchmarking.md` with a new TPC-H
**SF1000** result set captured on a 32-executor Kubernetes cluster with
Parquet data on S3.

- Pins the exact commit under test: `696ca29b` (`main`, 2026-07-23,
  Ballista `54.0.0`, DataFusion `54.1.0`).
- Documents the environment (cluster shape, executor sizing) and the full
  Ballista + Spark configurations used.
- Publishes Q1–Q19 mean times over 2 iterations; marks Q20–Q22 as `FAIL`
  on this configuration.
- Adds a vanilla **Spark 3.4** baseline column measured on the same
  cluster shape, alongside the Ballista column.
- Trims the surrounding narrative from the previous homelab-era version of
  the page so the document mainly states current performance.

## Notes

- Ballista's `Q18` now completes at this commit and sizing (399.72 s).
- Q20–Q22 did not produce results in this run; recorded as `FAIL`
  pending root-cause investigation.
- Spark row uses the harness's mean-of-3 (first iteration dropped);
  Ballista row is mean-of-2.

## Test plan

- [x] Doc renders as GitHub markdown (tables + code blocks).
- [ ] Reviewer sanity-check: numbers match the JSON captured for this run.

### DIFF
--- a/docs/source/contributors-guide/benchmarking.md
+++ b/docs/source/contributors-guide/benchmarking.md
@@ -19,22 +19,22 @@
 
 # Benchmarking
 
-Current TPC-H **SF1000** results for Ballista. This page records one result set
-from one commit on one cluster shape; refresh the commit, cluster, and table
-together.
+Current TPC-H **SF1000** results for Ballista, compared against a vanilla
+**Spark 3.4** baseline running on the same cluster shape.
 
-## Version under test
+## Versions under test
 
-| Component  | Value                                                                                                     |
-| ---------- | --------------------------------------------------------------------------------------------------------- |
-| Ballista   | [`696ca29b`](https://github.com/apache/datafusion-ballista/commit/696ca29b3c1fe3013238822b7f5d8cdb918fdaa3) (`main`, 2026-07-23) |
-| Cargo pkg  | `54.0.0`                                                                                                  |
-| DataFusion | `54.1.0`                                                                                                  |
+| Engine   | Version                                                                                                                          |
+| -------- | -------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


docs/source/contributors-guide/benchmarking.md
```diff
@@ -19,384 +19,165 @@
...
+mean of 3 iterations (cold iteration dropped by the harness). `FAIL` = query
+did not complete on this Ballista configuration.
+
+| Query | Ballista (s) | Spark 3.4 (s) |
+| ----: | -----------: | ------------: |
+|     1 |        19.55 |         67.58 |
+|     2 |        32.84 |         29.80 |
+|     3 |        40.52 |         25.13 |
+|     4 |        27.98 |         21.19 |
+|     5 |       202.84 |         54.12 |
```
Dandandan: Spark does enable prefetching by default for S3?

I found it to have a big effect:
 https://github.com/apache/datafusion/pull/23492

milenkovicm:
AQE is enabled, is it performing better than static ?

andygrove:
> AQE is enabled, is it performing better than static ?

I'm just focussing on AQE for now. It was getting too much to try and keep running both, tbh.

andygrove:
I got a clean run, so will merge this docs-only PR, and keep iterating ...

docs/source/contributors-guide/benchmarking.md
andygrove: I'll need to look into this. I filed https://github.com/apache/datafusion-ballista/issues/2162 so that I don't lose track of this

PR #2175, opened 2026-07-24
feat(core,scheduler): transport RuntimeStatsExec reports to scheduler; log merged quantile cuts per stage

avantgardnerio:
## Summary

Groundwork for the parallel-window / range-repartition flow ([design](https://github.com/apache/datafusion-ballista/pull/2094)). Ships `RuntimeStatsExec` observations (per-partition row counts and T-Digest quantile sketches) from executor back to scheduler, merges them per stage, and logs the K-1 quantile cuts a globally-informed router would use for K output partitions. No planner change yet — reports don't naturally appear in production plans until a follow-up rule inserts the Stage-1 chain under BWAG.

Three commits, each mergeable on its own:

1. **`proto: add RuntimeStatsReport wire format on SuccessfulTask`** — pure wire additions: `SuccessfulTask.runtime_stats`, `RuntimeStatsReport { order_by, partitions }`, `RuntimeStatsPartitionEntry { partition_id, row_count, optional sketch }`. `QuantileSketchState` and `sketch_to_proto` / `sketch_from_proto` already landed with #2094 and are reused unchanged. All existing `SuccessfulTask` construction sites populated with `runtime_stats: vec![]`. Behavior unchanged.

2. **`executor: ship RuntimeStatsReports to scheduler; log on arrival`** — `collect_reports(plan)` walks through the `preserves_distribution` whitelist from the shuffle writer's plan root and emits one report per reachable `RuntimeStatsExec`. Stats-taps below any distribution-changing operator (e.g. `UnorderedRangeRepartitionExec`) are excluded automatically. `QueryStageExecutor::collect_runtime_stats_reports()` (default empty) with `DefaultQueryStageExec` overriding. `as_task_status` gains the `Vec<RuntimeStatsReport>` param and threads it through both call-sites. Scheduler-side `TaskManager::update_task_statuses` logs a per-arrival summary at `debug!` — `RUST_LOG` promotes for local verification.

3. **`scheduler: accumulate runtime-stats reports per stage; log merged cuts`** — `RunningStage` carries a `runtime_stats_reports` accumulator; each successful task update appends. When the stage transitions to Successful, `log_merged_runtime_stats` groups by `order_by` wire tag (prost-encoded bytes as HashMap key), merges T-Digests with `TDigest::merge_digests`, and logs the K-1 quantile cuts at `debug!`. K is read from the reports (`report.partitions.len()`). Mismatched-K within a group and sketch decode failures both surface as errors from a fallible `merge_reports() -> Result<Vec<MergedRuntimeStats>>`; the log-side helper catches at `warn!`. `MergedRuntimeStats` is public so a future AQE consumer can call `merge_reports` directly without paying the log-formatting cost.

Plus a small review-fixup commit tightening `merge_reports` to be honestly fallible (dropped `warn!+continue` per-group silent fallbacks) and renaming `.k` → `.partition_count`.

## Test plan

- [x] `cargo test -p ballista-core --lib` — 176 pass (was 166 pre-PR).
- [x] `cargo test -p ballista-scheduler --lib` — 269 pass, 2 ignored.
- [x] `cargo test -p ballista-executor --lib` — 53 pass.
- [x] `cargo clippy --all-targets` — clean.
- [x] `cargo fmt --all` — clean.

Nine new unit tests total: four walker tests exercising the whitelist boundary behavior (positive: `BufferExec` doesn't block, negative: `SortExec` with `preserve_partitioning=false` does), five merge tests covering disjoint-range union, quartile cuts on a uniform sample, row-count-only groups, mismatched-partition-count errors, sketch-decode-error propagation, and empty input.

- [ ] End-to-end cluster verification is intentionally deferred to the follow-up rule PR — no production plan naturally contains a `RuntimeStatsExec` today. Once the parallel-window detection rule lands, real queries will produce reports and the merged cuts will show up in the scheduler log against actual data.

### DIFF
--- a/ballista/executor/src/execution_engine.rs
+++ b/ballista/executor/src/execution_engine.rs
@@ -86,6 +86,18 @@ pub trait QueryStageExecutor: Sync + Send + Debug + Display {
 
     /// Collects execution metrics from all operators in the plan.
     fn collect_plan_metrics(&self) -> Vec<MetricsSet>;
+
+    /// Collect runtime-stats reports for every `RuntimeStatsExec` still
+    /// valid at the plan's output (walked through the distribution-
+    /// preserving whitelist). Called at task completion; whatever comes
+    /// back rides along in the task's `SuccessfulTask` message to the
+    /// scheduler. Default returns empty — implementers with real plans
+    /// override to walk their operator tree.
+    fn collect_runtime_stats_reports(
+        &self,
+    ) -> Vec<ballista_core::serde::protobuf::RuntimeStatsReport> {
+        Vec::new()
+    }
 }
 
 /// Default execution engine using DataFusion's ShuffleWriterExec.
@@ -288,6 +300,31 @@ impl QueryStageExecutor for DefaultQueryStageExec {
                 utils::collect_plan_metrics(writer)
             }
             ShuffleWriterVariant::Sort(writer) => utils::collect_plan_metrics(writer),
+        }
+    }
+
+    fn collect_runtime_stats_reports(
+        &self,
+    ) -> Vec<ballista_core::serde::protobuf::RuntimeStatsReport> {
+        // Walk from the shuffle writer's plan through the whitelist. If
+        // no `RuntimeStatsExec` sits within reach, we return an empty
+        // Vec — the majority of plans (anything not on the parallel-
+        // window path today). Serialization errors are logged and the
+        // report dropped rather than failing the task; the task's data
+        // was already produced correctly, telemetry loss shouldn't tank
+        // the query.
+        let plan: Arc<dyn ExecutionPlan> = match &self.shuffle_writer {
+            ShuffleWriterVariant::Passthrough(writer) => Arc::new(writer.clone()),
+            ShuffleWriterVariant::Sort(writer) => Arc::new(writer.clone()),
+        };
+        match ballista_core::execution_plans::collect_runtime_stats_reports(&plan) {
+            Ok(reports) => reports,
+            Err(e) => {
+                log::warn!(
+                    "collect_runtime_stats_reports failed, task will report empty stats: {e}"
+                );
+                Vec::new()
+            }
         }
     }
 }
--- a/ballista/executor/src/lib.rs
+++ b/ballista/executor/src/lib.rs
@@ -63,8 +63,8 @@ use log::info;
 
 use crate::shutdown::Shutdown;
 use ballista_core::serde::protobuf::{
-    FailedTask, OperatorMetricsSet, ShuffleWritePartition, SuccessfulTask, TaskStatus,
-    task_status,
+    FailedTask, OperatorMetricsSet, RuntimeStatsReport, ShuffleWritePartition,
+    SuccessfulTask, TaskStatus, task_status,
 };
 use ballista_core::serde::scheduler::TaskKey;
 use ballista_core::utils::GrpcServerConfig;
@@ -98,6 +98,21 @@ pub struct TaskExecutionTimes {
     end_exec_time: u64,
 }
 
+/// Side-channel data harvested from a task's executed plan, attached to the
+/// [`TaskStatus`] reported to the scheduler on success.
+///
+/// Marked `#[non_exhaustive]` so future additions (e.g. tracing IDs, further
+/// runtime reports) are non-breaking for external callers that construct via
+/// `TaskCompletionExtras { operator_metrics: …, ..Default::default() }`.
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct TaskCompletionExtras {
+    /// Per-operator metrics collected from the executed plan.
+    pub operator_metrics: Option<Vec<OperatorMetricsSet>>,
+    /// Runtime-stats reports harvested from `RuntimeStatsExec` taps in the plan.
+    pub runtime_stats: Vec<RuntimeStatsReport>,
+}
+
 /// Converts a task execution result into a [`TaskStatus`] protobuf message.
 ///
 /// This function wraps the outcome of task execution (success or failure)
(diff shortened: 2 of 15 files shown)


avantgardnerio:
@phillipleblanc WDYT?

ballista/executor/src/lib.rs
```diff
@@ -103,21 +103,25 @@ pub struct TaskExecutionTimes {
...
 /// along with timing and metrics information into a status message that
 /// can be sent back to the scheduler.
+#[allow(clippy::too_many_arguments)]
 pub fn as_task_status(
     execution_result: ballista_core::error::Result<Vec<ShuffleWritePartition>>,
     executor_id: String,
     stage_attempt_num: usize,
     key: TaskKey,
     operator_metrics: Option<Vec<OperatorMetricsSet>>,
+    runtime_stats: Vec<RuntimeStatsReport>,
```
andygrove: This adds a positional parameter to a public `pub fn` in one of the `missing_docs` crates, so it is a breaking change for anyone building a custom executor loop that calls `as_task_status`. Could we make it additive instead? Given this signature already carries `#[allow(clippy::too_many_arguments)]`, folding `runtime_stats` into a small completion struct (or the existing `TaskExecutionTimes`-style param) would avoid the break. If we do decide to take the break, it needs the `api change` label and an entry under `docs/source/upgrading/` so library users get a heads up.

Worth noting the trait method you added right next to this is done the additive way already, with a default impl returning empty, so external implementors do not break. Same treatment here would be ideal.

ballista/scheduler/src/state/execution_stage.rs
```diff
@@ -218,6 +219,10 @@ pub struct RunningStage {
     pub stage_metrics: Option<Vec<MetricsSet>>,
     /// [SessionConfig] used for this stage
     pub session_config: Arc<SessionConfig>,
+    /// `RuntimeStatsReport`s collected from every successful task in
+    /// this stage attempt. Merged and logged once the stage finalizes;
+    /// dropped when the stage transitions to Successful.
+    pub runtime_stats_reports: Vec<RuntimeStatsReport>,
```
andygrove: Minor, lower stakes than the `as_task_status` one. Adding a public field to `RunningStage` breaks external struct-literal construction since it is not `#[non_exhaustive]`. This reads as internal scheduler state built via `RunningStage::new`, so the real blast radius is probably nil, but flagging it since it is on the public surface of a `missing_docs` crate. Fine to leave as is if we are confident nobody constructs this externally, just want it to be a conscious call.

ballista/executor/src/execution_engine.rs
```diff
@@ -290,6 +302,31 @@ impl QueryStageExecutor for DefaultQueryStageExec {
             ShuffleWriterVariant::Sort(writer) => utils::collect_plan_metrics(writer),
         }
     }
+
+    fn collect_runtime_stats_reports(
```
andygrove, line 306: This walker now runs once per task at completion on every query, not just parallel-window plans. I expect it is in the noise since it is a short walk off the data path, but since it touches the task completion path, could you drop a quick TPC-H before/after against main just to confirm no regression? A couple of representative queries is plenty, no need for the full suite:

```sh
cargo run --release --bin tpch -- benchmark ballista \
  --host localhost --port 50050 \
  --iterations 3 --path <tpch parquet dir> --format parquet --query 1
```

andygrove (commented):
Nice, clean plumbing and the test coverage on the walker and merge paths is good. One thing I want to sort out before this merges is the public API break on `as_task_status`, plus a couple of smaller notes inline. Details on the lines below.

ballista/executor/src/lib.rs
avantgardnerio: Thanks! I'll add `api-change` to my mental checklist henceforth. I think I resolved this in a way that matches your suggestion. Let me know if I missed something.

ballista/scheduler/src/state/execution_stage.rs
avantgardnerio: The only instance I could find of this being struct literal'ed was in `SuccessfulStage::to_running()`. I don't think we can guarantee no one external does it, but it seems like largely an internal concern to the scheduler, so downstream users literaling it seem unlikely. We could probably make it non-pub, but that seems more likely to break things. My bet is that the least disruptive thing to do is leave it as-is.

ballista/executor/src/execution_engine.rs
avantgardnerio: How's this? I think it indicates there is no performance regression:
image

avantgardnerio:
@andygrove I really appreciate your review. I think I've addressed your feedback, so PTAL again whenever you have a moment :raised_hands:

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -950,12 +953,15 @@ impl ExecutionGraph for StaticExecutionGraph {
                             running_stage
                                 .update_task_metrics(task_id, operator_metrics)?;
 
+                            let SuccessfulTask {
+                                partitions,
+                                runtime_stats,
+                                ..
+                            } = successful_task;
+                            running_stage.append_runtime_stats_reports(runtime_stats);
```
phillipleblanc: Reports are not associated with task attempts. When successful work is invalidated and rerun (i.e. lost executor, and the scheduler requeues the task), the old report cannot be removed.

i.e. for this scenario:

```text
   Task succeeds
       v
   Shuffle files remain on that executor
       v
   Stage becomes Successful
       v
   Executor is lost
       v
   Its shuffle files become unavailable
       v
   Tasks producing those files are marked ResultLost
       v
   Stage returns to Running and those partitions are recomputed
```

This is relatively minor because this affects the quality of the calculated partition boundaries, not row correctness.

phillipleblanc (approved):
Looks good to me, one minor comment about making this resilient to stage retries - but I like where this is going.

ballista/scheduler/src/state/execution_graph.rs
avantgardnerio: Thanks @phillipleblanc , you're correct it wouldn't break things badly, but it certainly isn't good. Fixed!

PR #2179, opened 2026-07-24
chore: use taplo for Cargo.toml formatting

villebro:
## Background

The `Check Cargo.toml formatting` CI job does not currently fail on formatting violations, for two reasons:

- `ci/scripts/rust_toml_fmt.sh` runs `cargo tomlfmt -k -p` via `find … -exec`. `find` does not propagate the tool's non-zero exit code, so `set -e` is not triggered.
- The workflow's failure guard checks for a `Cargo.toml.bak` file, but `cargo-tomlfmt` 0.2.1 writes `Cargo.toml.new` rather than `.bak`, so the guard does not match.

As a result the job passes regardless of manifest formatting, and the manifests have drifted from a consistent style. Separately, `cargo-tomlfmt` 0.2.1 has some limitations for this use case: it writes a `.new` file rather than formatting in place, and it does not preserve comments when reformatting (see [tbrand/cargo-tomlfmt#40](https://github.com/tbrand/cargo-tomlfmt/issues/40)).

## Change

Move Cargo.toml formatting to **taplo**, aligning with the parent [apache/datafusion](https://github.com/apache/datafusion) project:

- Add `taplo.toml` (copied from apache/datafusion) — Cargo.toml-only scope, 4-space indent to match the existing style, `reorder_keys = false` for `[package]` (preserves key order and comments), alphabetized dependency tables.
- Rewrite `ci/scripts/rust_toml_fmt.sh` to run `taplo format --check` (with a `--write` mode to fix in place). This returns a non-zero exit code on violations, so the check enforces formatting, and it leaves no scratch files.
- `.github/workflows/rust.yml`: install `taplo-cli@0.9` via `taiki-e/install-action` and remove the `.bak` guard, which is no longer needed.
- `dev/rust_lint.sh`: install `taplo` instead of `cargo-tomlfmt`.
- Reformat the manifests to satisfy the check (dependency reordering and array expansion; comments preserved).

## Consistency with apache/datafusion

The parent project uses the same setup:

| | apache/datafusion | this PR |
|---|---|---|
| Install | `taplo-cli@0.9` via `taiki-e/install-action` | same |
| Check command | `taplo format --check` | same (via `rust_toml_fmt.sh`) |
| Rules | root `taplo.toml` | verbatim copy |

## Verification

- `ci/scripts/rust_toml_fmt.sh` (check) exits `0` on the formatted tree.
- Introducing a formatting violation in a manifest causes the check to exit non-zero (verified locally).
- No `Cargo.toml.new` / `.bak` scratch files are left behind.
- `ci/scripts/rust_toml_fmt.sh --write` reformats in place.
- No comment lines were removed by the reformat.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -238,22 +238,16 @@ jobs:
           rustup toolchain install stable
           rustup default stable
       - uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
-      - name: Install cargo-tomlfmt
+      - name: Install taplo
         uses: taiki-e/install-action@41049aa56687c35e0afa74eed4f09cec4f9afabf # v2.85.2
         with:
-          tool: cargo-tomlfmt@0.2.1
+          tool: taplo-cli@0.9
 
       - name: Check Cargo.toml formatting
         run: |
-          # if you encounter error, try rerun the command below, finally run 'git diff' to
-          # check which Cargo.toml introduces formatting violation
-          #
-          # ignore ./Cargo.toml because putting workspaces in multi-line lists make it easy to read
+          # if you encounter an error, run 'ci/scripts/rust_toml_fmt.sh --write'
+          # (or 'taplo format') locally to fix the formatting, then commit the result.
           ci/scripts/rust_toml_fmt.sh
-          if test -f "./Cargo.toml.bak"; then
-              echo "cargo tomlfmt found format violations"
-              exit 1
-          fi
 
   datafusion-proto-sync-check:
     name: Check vendored DataFusion proto is in sync
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -241,7 +241,7 @@ jobs:
       - name: Install taplo
         uses: taiki-e/install-action@41049aa56687c35e0afa74eed4f09cec4f9afabf # v2.85.2
         with:
-          tool: taplo-cli@0.9
+          tool: taplo-cli@0.10.0
 
       - name: Check Cargo.toml formatting
         run: |
--- a/ballista-cli/Cargo.toml
+++ b/ballista-cli/Cargo.toml
@@ -84,8 +84,8 @@ cli = [
     "dep:ballista",
     "dep:datafusion",
     "dep:datafusion-cli",
-    "dep:rustyline",
     "dep:mimalloc",
+    "dep:rustyline",
 ]
 
 tui = [
--- a/ballista-cli/Cargo.toml
+++ b/ballista-cli/Cargo.toml
@@ -36,8 +36,8 @@ crate-type = ["cdylib", "rlib"]
 ballista = { path = "../ballista/client", version = "54.0.0", features = ["standalone"], optional = true }
 datafusion = { workspace = true, optional = true }
 datafusion-cli = { workspace = true, optional = true }
-rustyline = { version = "18.0.0", optional = true }
 mimalloc = { workspace = true, optional = true }
+rustyline = { version = "18.0.0", optional = true }
 
 # TUI/web shared deps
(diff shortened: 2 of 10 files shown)


andygrove (approved):
Thanks for cleaning this up @villebro. LGTM.

.github/workflows/rust.yml
```diff
@@ -238,22 +238,16 @@ jobs:
           rustup toolchain install stable
           rustup default stable
       - uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
-      - name: Install cargo-tomlfmt
+      - name: Install taplo
         uses: taiki-e/install-action@7572810d7dd469b651bb7793945692cf78da5dd7 # v2.85.0
         with:
-          tool: cargo-tomlfmt@0.2.1
+          tool: taplo-cli@0.9
```
martin-g: Why 0.9 ?
https://github.com/tamasfe/taplo/releases/tag/0.10.0 has been released 1 year ago

ballista-cli/Cargo.toml
```diff
@@ -81,25 +81,57 @@ tokio = { workspace = true, features = ["macros", "sync", "time"] }
 default = ["cli", "tui"]
 
 cli = [
-    "dep:ballista", "dep:datafusion", "dep:datafusion-cli",
-    "dep:rustyline", "dep:mimalloc",
+    "dep:ballista",
+    "dep:datafusion",
+    "dep:datafusion-cli",
+    "dep:rustyline",
+    "dep:mimalloc",
```
martin-g:
```suggestion
    "dep:mimalloc",
    "dep:rustyline",
```

ci/scripts/rust_toml_fmt.sh
```diff
@@ -17,5 +17,19 @@
...
-find . -mindepth 2 -name 'Cargo.toml' -exec cargo tomlfmt -k -p {} \;
+# Checks Cargo.toml formatting using taplo. The formatting rules live in
+# taplo.toml at the repository root.
+#
+# Usage:
+#   ci/scripts/rust_toml_fmt.sh            # check formatting (default, used in CI)
+#   ci/scripts/rust_toml_fmt.sh --write    # reformat files in place
+
+set -e
+
```
martin-g, line 28: IMO it would be nice to auto-install it if it is not available on the system.
https://taplo.tamasfe.dev/cli/installation/binary.html

dev/rust_lint.sh
```diff
@@ -17,9 +17,9 @@
 # specific language governing permissions and limitations
 # under the License.
 set -e
-if ! command -v cargo-tomlfmt &> /dev/null; then
-    echo "Installing cargo-tomlfmt using cargo"
-    cargo install cargo-tomlfmt
+if ! command -v taplo &> /dev/null; then
+    echo "Installing taplo using cargo"
+    cargo install taplo-cli --locked
```
martin-g: 👍🏻 
let's do the same in `rust_toml_fmt.sh` too !

.github/workflows/rust.yml
villebro: Good catch, I bumped to 0.10.0

ballista-cli/Cargo.toml
villebro: Applied, thanks. I did notice a few other ones, too, but applying Taplo's `reorder_arrays` made very extensive changes. So I would almost defer automated array reordering to a potential follow-up PR that makes sure we're not reordering arrays that were intentionally non-ABC ordered.

dev/rust_lint.sh
villebro: I moved installation logic to `rust_toml_fmt.sh` so it's now contained in only one place.

villebro:
@martin-g I've addressed the review comments, please review at your convenience.

martin-g (approved):

Issue #2185, opened 2026-07-26
Adaptive (AQE) execution drops all rows for TPC-DS q61 at target_partitions=1


andygrove:
## Describe the bug

With adaptive execution enabled (`ballista.planner.adaptive.enabled=true`) at `target_partitions=1`, TPC-DS q61 returns an empty result where single-process DataFusion returns one row:

```
Query 61 VERIFY MISMATCH: Execution error: result mismatch: expected 1 rows, got 0 rows
```

Surfaced by extending the TPC-DS SF1 correctness gate to cover the adaptive planner and single-partition execution (#2182).

## Scope

- Only reproduces with **AQE on**. The same query at `--partitions 1` with the static planner passes.
- Does not reproduce at `--partitions 16`.
- q61 contains no `UNION`, so this appears distinct from #2184 (which covers the UNION scan-restriction failures at 1 partition). Filing separately on that basis; if root-causing shows a shared cause they can be merged.

## To Reproduce

Generate TPC-DS SF1 Parquet with `tpcgen-cli`, start a scheduler and executor, then:

```
tpcds --host localhost --port 50050 --path <data> --partitions 1 --verify \
  --query 61 \
  -c datafusion.optimizer.prefer_hash_join=false \
  -c ballista.planner.adaptive.enabled=true
# -> Query 61 VERIFY MISMATCH: expected 1 rows, got 0 rows
```

Passes with `--partitions 16`, and with `--partitions 1` when the adaptive planner is off.

## Expected behavior

The adaptive planner should return the same single row as single-process DataFusion, at every partition count.

PR #2186, opened 2026-07-26
fix: stop every task running every union branch at low partition counts

### LINKED ISSUE #2184: Distributed execution returns inflated results and crashes on UNION queries at target_partitions=1
## Describe the bug

At `target_partitions=1`, tasks of a `UnionExec` stage each execute **every** union branch instead of only the branch they own. That shows up two ways: silently inflated results, and a hard failure.

Surfaced by extending the TPC-DS SF1 correctness gate to run at 1 partition as well as 16 (#2182).

## Confirmed cases (SF1, `--partitions 1`, `prefer_hash_join=false`, static planner)

| Query | Single-process DataFusion | Ballista (distributed) | Ratio |
|---|---|---|---|
| q5 | `73724.99` | `294899.96` | exactly **4x** |
| q80 | `11343.60` | `22687.20` | exactly **2x** |

q49 does not return a wrong answer, it fails outright:

```
Job failed due to stage 7 failed: Task failed due to runtime execution error:
Internal("FileStreamBuilder invalid partition index: 0")
```

q5, q80 and q49 are exactly the three queries in the affected set that contain `UNION`. All three pass at `--partitions 16`.

## Root cause

Confirmed by instrumenting `restrict_plan_to_partitions` and capturing the real plans, not by inspection.

**Restricting q49's 3-partition union stage to `[0]` returns a plan that still has 3 partitions.** The executor (`drive_shuffle_writer_stage`) runs `0..partition_count`, so every task runs all three branches.

Why the restriction fails to shrink it: a union branch that collapses internally reports **one output partition no matter what its leaves do**. In these queries the collapse is

- `CoalescePartitionsExec` (q5's store and web branches),
- a `CollectLeft` `HashJoinExec` (q5's catalog branch),
- `SortExec: preserve_partitioning=[false]` (q49's `rank() over` branches).

`restrict` allocates the parent's partitions across children, so an unowned branch is restricted to `[]` — but `under_collect` scoping then deliberately leaves every leaf below the collapse reading its full upstream, and broadcast `ShuffleReaderExec`s are exempted too. The branch keeps reporting a partition and keeps producing all of its data.

Hence N tasks x N branches: N-fold duplication (q5, q80), or a task executing a branch whose scan was emptied (q49).

## Why this is not simply a gap in #2070

#2070 fixed a different failure in the same area (a scan reading its whole table because the stage partition index was applied directly to every scan) and verified partition counts 4, 5, 6, 8, 16, 20, 32 — never 1. This bug is adjacent but distinct: the per-child index arithmetic is correct; the problem is that an unowned branch cannot be reduced to zero partitions.

## Why it cannot be fixed inside the restriction

Three constraints that are mutually incompatible for this shape. Two of them were hit as real runtime failures while attempting fixes:

1. **The restricted plan must expose exactly `indices.len()` partitions.** `compute_global_output_partition_ids` returns `PassThrough(global_input_partition_ids)` when the child chain branches, and the executor maps its local output slots onto those global ids.
2. **Operator count must be identical across a stage's tasks.** `ExecutionStage::update_task_metrics` matches metrics positionally and rejects a length mismatch, so a rewrite cannot add or remove plan nodes. Replacing an idle branch with a 0-partition stub produces `task metrics array size 9 does not equal with the stage metrics array size 12` and the job then stalls.
3. **Operators require at least one input partition.** Draining an idle branch's leaves to zero partitions instead produces `CoalescePartitionsExec requires at least one input partition` (q5) and `Invalid HashJoinExec, the output partition count of the left child must be 1 in CollectLeft mode` (q49).

So an unowned collapsing branch cannot be made 0-partition by restriction (1), cannot be structurally replaced (2), and cannot have its leaves zeroed (3).

## Proposed fix

Relax constraint 1: give the executor the task's assigned partition indices and have it execute exactly those, rather than every partition of the plan it is handed. The union then needs no structural rewrite at all, operator count stays identical across tasks, and no operator is handed zero partitions. Leaf restriction is still needed so a scan does not drain DataFusion's shared work queue.

## To Reproduce

```
  --query 5 -c datafusion.optimizer.prefer_hash_join=false
# -> Query 5 VERIFY MISMATCH: expected `73724.99`, got `294899.96`

  --query 49 -c datafusion.optimizer.prefer_hash_join=false
# -> Internal("FileStreamBuilder invalid partition index: 0")
```

Both pass with `--partitions 16`.

## Expected behavior

Results at `target_partitions=1` should match single-process DataFusion, as they do at every other partition count.

andygrove:

Closes #2184.

# Rationale for this change

Per-task restriction gives each branch of a `UnionExec` the partitions that task owns and an empty slice to the rest, relying on an unowned branch becoming 0-partition so the union's index arithmetic routes past it.

That fails when a branch's partition count does not come from restrictable leaves. A `CoalescePartitionsExec` or a non-preserving `SortExec` reports one partition whatever its leaves do, and a broadcast `ShuffleReaderExec` is deliberately never pruned. The branch keeps reporting a partition and keeps producing all of its data, so the executor — which runs every partition of the plan it is handed — runs every branch for every task.

At SF1 with `target_partitions=1`:

| Query | Expected | Got | |
|---|---|---|---|
| q5 | `73724.99` | `294899.96` | 4x |
| q80 | `11343.60` | `22687.20` | 2x |
| q49 | — | `Internal("FileStreamBuilder invalid partition index: 0")` | |

This does not show at higher partition counts because those union stages are capped by a hash repartition, where the writer's output is an intrinsic K-space and inner partition counts do not matter. Confirmed by instrumenting the rewriter: restricting q49's 3-partition union stage to `[0]` returned a plan that still had 3 partitions.

**`planner.rs` — cut a stage boundary beneath a union branch that cannot be restricted away.** The branch becomes a `ShuffleReaderExec`, which restricts to nothing cleanly, and the collapse moves into its own stage where it still sees its whole input. Branches that already restrict away stay inline, so a union of plain scans is unaffected and pays no extra shuffle.

Whether a branch can be restricted away is decided by *asking the rewriter* (`restrict_plan_to_partitions(branch, &[])`) rather than by matching operator types, so it stays correct as the rewriter evolves.

Branches holding a broadcast input or a `CollectLeft` join are treated as unrestrictable regardless. Those build and probe sides can still be swapped after stage planning, and the two orders restrict differently — the build side is read whole, a broadcast input is never pruned — so a branch that looks restrictable at planning time may not be by the time the task is built. That is exactly what left q5's catalog branch running in every task after the first version of this fix.

**`task_builder.rs` — take restriction scope from `required_input_distribution()` for every operator** instead of a hardcoded list of join types. `CrossJoinExec` collects its left input but was not on that list, so each task of a multi-partition stage rebuilt the cross join's left side from its own slice. This surfaced as an inflated q77 (whose catalog branch is `from cs, cr`) once that branch got a stage of its own — caught by the full-suite run, not by the original repro.

No API change. Union queries whose branches collapse internally now materialize those branches as their own stage, which costs one extra shuffle per affected branch and is what makes the result correct.

## Testing

Four unit tests: a plain branch needs no stage of its own, a collapsing branch and a broadcast-holding branch do, and a cross join's collected side is read whole.

End-to-end at SF1 against a single-process DataFusion oracle:

- `--partitions 16` — **the full suite passes**, so the currently-green configuration is unaffected.
- `--partitions 1` — q5, q49, q80, q61 and q77 all verify OK. Only q72 remains, and it exhausts its per-task memory share rather than returning a wrong answer.

## Not covered

The adaptive planner builds stages through its own path (`state/aqe/`) and is untouched here; q5, q80 and q61 still fail under `ballista.planner.adaptive.enabled=true` at one partition. #2185 stays open, and #2182 will still show the AQE legs red.

### DIFF
--- a/ballista/scheduler/src/planner.rs
+++ b/ballista/scheduler/src/planner.rs
@@ -43,6 +43,7 @@ use datafusion::physical_plan::joins::{HashJoinExec, PartitionMode};
 use datafusion::physical_plan::projection::ProjectionExec;
 use datafusion::physical_plan::repartition::RepartitionExec;
 use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+use datafusion::physical_plan::union::UnionExec;
 use datafusion::physical_plan::{
     ExecutionPlan, Partitioning, with_new_children_if_necessary,
 };
@@ -51,6 +52,7 @@ use log::debug;
 use crate::physical_optimizer::join_selection::{
     collect_left_broadcast_safe, should_swap_join_order,
 };
+use crate::state::task_builder::restrict_plan_to_partitions;
 
 type PartialQueryStageResult = (Arc<dyn ExecutionPlan>, Vec<Arc<dyn ShuffleWriter>>);
 
@@ -184,6 +186,55 @@ impl DefaultDistributedPlanner {
             let new_join =
                 execution_plan.with_new_children(vec![broadcast_left, probe])?;
             return Ok((new_join, stages));
+        }
+
+        // UnionExec: a branch that cannot be restricted away needs its own
+        // stage. Per-task restriction gives each branch of a union the
+        // partitions that task owns and an empty slice to the rest, relying on
+        // an unowned branch becoming 0-partition so `UnionExec`'s index
+        // arithmetic routes past it.
+        //
+        // That fails when the branch's partition count does not come from
+        // restrictable leaves — a `CoalescePartitionsExec` or non-preserving
+        // `SortExec` reports one partition whatever its leaves do, and a
+        // broadcast `ShuffleReaderExec` is deliberately never pruned. The
+        // branch keeps reporting a partition and keeps producing all of its
+        // data, so every task runs every branch: results come back inflated by
+        // the branch count, or a task executes a branch whose scan was emptied
+        // (#2184).
+        //
+        // Cutting a boundary beneath such a branch turns it into a
+        // `ShuffleReaderExec`, which restricts to nothing cleanly, and leaves
+        // the collapse in its own stage where it sees its whole input.
+        if execution_plan.is::<UnionExec>() {
+            let mut stages = vec![];
+            let mut children = vec![];
+            for child in execution_plan.children() {
+                let (new_child, mut child_stages) =
+                    self.plan_query_stages_internal(job_id, child.clone(), config)?;
+                stages.append(&mut child_stages);
+
+                if new_child.is::<UnresolvedShuffleExec>()
+                    || restricts_to_nothing(&new_child)
+                {
+                    children.push(new_child);
+                    continue;
+                }
+
+                let writer = create_shuffle_writer_with_config(
+                    job_id,
+                    self.next_stage_id(),
+                    new_child,
+                    None,
+                    config,
+                )?;
+                children.push(create_unresolved_shuffle(writer.as_ref()));
+                stages.push(writer);
+            }
+            return Ok((
+                with_new_children_if_necessary(execution_plan, children)?,
+                stages,
+            ));
         }
 
         let mut stages = vec![];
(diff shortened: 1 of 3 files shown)


andygrove:
@phillipleblanc could you review?

phillipleblanc (approved):
LGTM

avantgardnerio (approved):
Thanks Andy. I think I may have broke this one. I like the fix.

andygrove:
Thanks for the reviews @phillipleblanc @avantgardnerio

PR #2195, opened 2026-07-27
feat(core): add PerPartitionFilterExec — per-input-partition boolean predicate

avantgardnerio:
## Summary

Adds a standalone `PerPartitionFilterExec` operator plus its physical-plan codec. Carries `Vec<Arc<dyn PhysicalExpr>>` — one boolean predicate per input partition — and applies the k-th predicate to the k-th input stream on `execute(k)`.

Not wired into any planner or rule; this is the operator surface, unit-tested in isolation. Follow-ups land the injection at range-repartition–consuming stages.

## Motivation

Comes out of the adaptive-range-shuffle work (URRE / ORRE — merged in #2123, #2169). Once the scheduler picks stage-global cut points and each producer's local sub-parts are routed by sketch-overlap, the downstream consumers over-fetch: a consumer whose assigned range is `[cuts[k-1], cuts[k])` may pull an over-inclusive sub-part that straddles the cut. Those out-of-range rows need to be dropped before the consumer sees them, and the drop is per-partition (each downstream partition has a distinct predicate).

Doing this as a plain filter on the whole consumer stream would require one task per predicate — K tasks for K partitions — which breaks vcore packing and defeats the whole point of the range-repartition family (a single task can own many partitions). `PerPartitionFilterExec` preserves the packing: N tasks each own a slice of partitions, and each partition applies its own predicate.

## How it plays with oversampling

Oversampling (many small sub-parts relative to downstream partition count) is what makes this cheap. Under reasonable oversampling, each downstream partition's range overlaps at most its \"own\" sub-part plus its two immediate neighbors — call it ~3 files fetched per consumer. The straddling case is rare and shallow, so the filter's rejection rate stays low.

## This is the simplest thing that works

The point of this PR is to unblock the end-to-end range-partitioning demo with the smallest operator surface possible. A later PR should push the filtering down into the shuffle/Flight path so the consumer never materializes rows it's about to reject:

- **Where:** on the reader side of the Flight shuffle, before the batch is handed up the plan.
- **How, cheap version:** the same per-partition predicate, applied to the arriving batch. Same semantics as this PR, one hop earlier.
- **How, cheaper version for sorted producers:** binary-search the sorted batch on the routing column and slice the in-range prefix/suffix directly, avoiding the per-row predicate eval. Applies naturally to ORRE (ordered producers) and to any URRE stage whose upstream happens to be sorted on the routing key.

That optimization is out of scope here. Landing the operator first keeps the range-partitioning wiring straightforward to review and lets us measure filter overhead on real workloads before deciding how far to push the follow-up.

## Test plan

- [x] `cargo test -p ballista-core --lib per_partition_filter` — 5 operator tests + 1 codec round-trip pass locally.
- [x] `cargo clippy --all-targets -p ballista-core` clean.
- [x] `cargo fmt --all -- --check` clean.
- [ ] CI green.

### DIFF
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -55,6 +55,7 @@ message BallistaPhysicalPlanNode {
     BufferExecNode buffer = 7;
     UnorderedRangeRepartitionExecNode unordered_range_repartition = 8;
     OrderedRangeRepartitionExecNode ordered_range_repartition = 9;
+    PerPartitionFilterExecNode per_partition_filter = 10;
   }
 }
 
@@ -122,6 +123,14 @@ message UnorderedRangeRepartitionExecNode {
   repeated datafusion.PhysicalSortExprNode order_by = 1;
   // K — number of output partitions. Must be ≥ 2.
   uint32 output_partitions = 2;
+}
+
+// Filter with per-input-partition predicates. `predicates[k]` is the
+// boolean expression applied to input partition `k`. Requires
+// `predicates.len() == input_partition_count`. The child plan is
+// plumbed by the framework as `inputs[0]` during decode.
+message PerPartitionFilterExecNode {
+  repeated datafusion.PhysicalExprNode predicates = 1;
 }
 
 message ChaosExecNode {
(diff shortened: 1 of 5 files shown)


avantgardnerio:
@phillipleblanc I swear one day soon I'm about to click together all these legos and show something cool :slightly_smiling_face:

andygrove (approved):
Purely additive. Did not review in depth. Thanks @avantgardnerio

PR #2196, opened 2026-07-27
feat(scheduler): range-repartition rule primitives (batteries-included)

avantgardnerio:
## Summary

I've finally done enough passes over it where I don't think it's ugly any more.

**pros:**

Batteries-included AQE machinery so anyone can write a `UnorderedRangeRepartitionExec` / `OrderedRangeRepartitionExec`-inserting optimizer rule and get correct end-to-end routing for free.

A range-partitioning rule now needs to do only *one thing*: splice a `RuntimeStatsExec` + `URRE`/`ORRE` above whatever it wants range-repartitioned. Everything downstream is handled:

- Executor already ships per-sub-part quantile sketches back to the scheduler (#2094 / #2175).
- Scheduler recognizes the range repartition at stage-completion time, merges sketches into `K − 1` global cuts, and rewrites the downstream location map so partition `k` pulls from every producer file whose sketched `[min, max]` overlaps `k`'s assigned cut range.
- Cuts + routing expression are parked on the boundary `ExchangeExec`; at task-specialization time the adapter wraps the `ShuffleReader` in a `PerPartitionFilterExec` (#2195) that trims straddling sub-parts to their assigned slice — correctness closes without the rule ever knowing about sketches, filters, or the AQE hook.

**cons:**

This is lots of plumbing code to achieve that. I hope I've done it in the right places. Honestly piggy backing  `range_repartition_routing` on `ExchangeExec` feels the worst. However this is following the precedent set by the `coalesce` field, so maybe it's okay?

## Design note: the `ExchangeExec` routing slot

`range_repartition_routing: Arc<Mutex<Option<RangeRepartitionRouting>>>` is a direct copy of the existing `coalesce: Arc<Mutex<Option<Arc<CoalescePlan>>>>` slot pattern that landed with `CoalescePartitionsRule` — same lifecycle (AQE rule parks decision, adapter consumes at plan-transform time), same `with_new_children` carry-through, same idempotent-overwrite setter. That slot went through review iteration on main (`b839f036 refactor(scheduler): tighten the AQE coalesce rule after review`, `232d7611 let the AQE coalesce slot be cleared`); this PR follows the pattern rather than inventing a new one.

## Example consumers

Two range-repartitioning rules are being prepared as follow-ups against this infrastructure:

- **`AdaptiveRangeShuffleRule`** — rewrites Hash exchanges as URRE at join-side boundaries (avoids spilling). Proof-of-concept for the machinery.
- **`ParallelWindowDetectRule`** (port from `brent/h2o-window-benchmarks`) — wraps `BoundedWindowAggExec` in URRE + `RuntimeStatsExec` for parallel `RANGE`-frame windows. Motivating consumer for the h2o single-partition OOM case.

Neither rule is in this PR — they'd conflate two decisions (the machinery vs. the specific consumer). ParallelWindowDetectRule is ready to send in short order once this lands.

## Caveats

- Additive: Machinery is inert until a rule inserts a range repartition; the primitives table above is the API surface.
- `RangeRepartitionRouting` is scheduler-only state — no wire/codec changes.
- `plan_contains_range_repartition` stops at exchange boundaries by design (a nested range repartition lives in a different stage anyway).


avantgardnerio:
@phillipleblanc this one is the base upon which anyone can build a range-repartitioned optimizer rule. I intend to use it for parallel windows in the next PR, but I was also playing with it to prevent spilling in SortShuffle on TPC-H q20. It's interesting because now we can play around and try things out.

phillipleblanc:
I won't have time to review this in depth for a couple of days, but I will look at it sometime this week.

andygrove:
Really like the shape of this, and the "a rule only has to splice two operators" goal is the right one. I dug into one thing and it looks like a real gap, so I built a repro rather than just asserting it.

## The issue

`update_stage_progress` does the duplication first and attaches the filter second:

```rust
let remapped = cut_partitions(partitions, reports, &routing.cuts)?;   // straddlers duplicated
self.planner.set_repartition_routing(stage_id, routing)?;             // may silently no-op
```

`set_repartition_routing` returns `Ok(())` when the stage's boundary root is an `AdaptiveDatafusionExec` rather than an `ExchangeExec`. That is documented as intentional, but by then `cut_partitions` has already pushed each straddling file into every overlapping downstream partition. With no routing parked, `BallistaAdapter` never injects `PerPartitionFilterExec`, so nothing trims them back. In the final-stage case it goes straight to the user via:

```rust
self.output_locations = partitions.into_iter().flatten().collect();
```

## The reachable shape is the canonical one

I initially assumed you would need something exotic to hit this. You do not. It is the exact `RuntimeStatsExec -> URRE` splice the PR tells rule authors to emit, sitting at the root of the plan.

`DistributedExchangeRule::optimize` runs `transform_up` first and only wraps the result in `AdaptiveDatafusionExec` afterwards. So at transform time the chain top has no parent for the new branch to visit, and the branch inspects `execution_plan.children()`. The only node visited is the `RuntimeStatsExec` itself, whose child is the URRE, which is not a chain top.

## Repro

Three tests against `a92f0617`, all passing.

**Leg 1**, dropped into the `distributed_exchange.rs` test module next to your existing ones:

```rust
#[test]
fn repro_root_level_chain_top_gets_no_exchange() {
    let rule = DistributedExchangeRule::default();
    let root = stats_over_urre_over_leaf();   // your existing helper

    let result = rule.optimize(root, &config()).unwrap();

    let adaptive = result.downcast_ref::<AdaptiveDatafusionExec>().unwrap();
    assert!(adaptive.input().is::<RuntimeStatsExec>());
    assert_eq!(count_exchanges(result.as_ref()), 0);
}
```

```
AdaptiveDatafusionExec: is_final=false, plan_id=0, stage_id=pending, stage_resolved=false
  RuntimeStatsExec: rows + sketch(routing=v@0 asc)
    UnorderedRangeRepartitionExec: routing=v@0 asc → 4 partitions
      StatisticsExec: col_count=1, row_count=Absent

exchanges inserted: 0
```

**Leg 2**, composing `cut_partitions` and `set_repartition_routing` in the order `update_stage_progress` calls them, through a real `AdaptivePlanner`. One producer file, sketch over `[5, 15, 25]`, single cut at 15:

```
after cut_partitions: 2 downstream partitions, file counts = [1, 1]
set_repartition_routing returned: Ok(())
--- plan after set_repartition_routing ---
AdaptiveDatafusionExec: is_final=true, plan_id=0, stage_id=0, stage_resolved=false
  RuntimeStatsExec: rows + sketch(routing=v@0 asc)
    UnorderedRangeRepartitionExec: routing=v@0 asc → 2 partitions
      DataSourceExec: partitions=1, partition_sizes=[0]

output_locations file_ids = [7, 7]
```

No `range_repartition_cuts` on the plan, so no `PerPartitionFilterExec` downstream. File 7 appears twice. Three rows would be read as six.

**Leg 3**, positive control so this is not just me misreading the mechanism. Same splice, same calls, only a `FilterExec` parent added:

```
AdaptiveDatafusionExec: is_final=false, plan_id=1, stage_id=pending, stage_resolved=false
  FilterExec: v@0 > 0
    ExchangeExec: partitioning=None, plan_id=0, stage_id=0, stage_resolved=false, range_repartition_cuts=1
      RuntimeStatsExec: rows + sketch(routing=v@0 asc)
        UnorderedRangeRepartitionExec: routing=v@0 asc → 2 partitions
```

Exchange inserted, cuts parked, everything works. Position in the plan is the only variable between leg 2 and leg 3.

<details>
<summary>Full leg 2 and leg 3 test file</summary>

Add `datafusion-functions-aggregate-common = { workspace = true }` to `ballista/scheduler` dev-dependencies, drop this in as `ballista/scheduler/src/state/aqe/test/repro_range_repartition.rs`, and register `mod repro_range_repartition;` in that directory's `mod.rs`.

```rust
use crate::state::aqe::execution_plan::RangeRepartitionRouting;
use crate::state::aqe::planner::AdaptivePlanner;
use ballista_core::execution_plans::{
    RuntimeStatsExec, UnorderedRangeRepartitionExec, cut_partitions,
};
use ballista_core::extension::SessionConfigExt;
use ballista_core::serde::protobuf::{RuntimeStatsPartitionEntry, RuntimeStatsReport};
use ballista_core::serde::scheduler::{
    ExecutorMetadata, ExecutorOperatingSystemSpecification, ExecutorSpecification,
    PartitionId, PartitionLocation, PartitionStats,
};
use datafusion::arrow::compute::SortOptions;
use datafusion::arrow::datatypes::{DataType, Field, Schema};
use datafusion::datasource::memory::MemorySourceConfig;
use datafusion::datasource::source::DataSourceExec;
use datafusion::physical_expr::PhysicalSortExpr;
use datafusion::physical_plan::ExecutionPlan;
use datafusion::physical_plan::expressions::col;
use datafusion::prelude::SessionConfig;
use std::sync::Arc;

fn v_schema() -> Arc<Schema> {
    Arc::new(Schema::new(vec![Field::new("v", DataType::Float64, false)]))
}

/// The canonical splice a range-repartition rule is told to emit,
/// here at the root of the plan.
fn stats_over_urre_root() -> Arc<dyn ExecutionPlan> {
    let schema = v_schema();
    let source: Arc<dyn ExecutionPlan> = Arc::new(DataSourceExec::new(Arc::new(
        MemorySourceConfig::try_new(&[vec![]], schema.clone(), None).unwrap(),
    )));
    let sort_expr = PhysicalSortExpr {
        expr: col("v", schema.as_ref()).unwrap(),
        options: SortOptions { descending: false, nulls_first: false },
    };
    let urre: Arc<dyn ExecutionPlan> = Arc::new(
        UnorderedRangeRepartitionExec::try_new(source, vec![sort_expr.clone()], 2).unwrap(),
    );
    Arc::new(RuntimeStatsExec::try_new(urre, Some(vec![sort_expr])).unwrap())
}

/// Same splice with an ordinary parent above it. A FilterExec survives the
/// optimizer pipeline. RoundRobin RepartitionExec and CoalescePartitionsExec
/// do NOT, both get stripped.
fn stats_over_urre_with_parent() -> Arc<dyn ExecutionPlan> {
    let pred: Arc<dyn datafusion::physical_expr::PhysicalExpr> =
        Arc::new(datafusion::physical_expr::expressions::BinaryExpr::new(
            col("v", v_schema().as_ref()).unwrap(),
            datafusion::logical_expr::Operator::Gt,
            Arc::new(datafusion::physical_expr::expressions::Literal::new(
                datafusion::scalar::ScalarValue::Float64(Some(0.0)),
            )),
        ));
    Arc::new(
        datafusion::physical_plan::filter::FilterExec::try_new(pred, stats_over_urre_root())
            .unwrap(),
    )
}

fn location(sub_part_id: usize, producer_task_id: usize, rows: u64) -> PartitionLocation {
    PartitionLocation {
        map_partition_id: 0,
        partition_id: PartitionId {
            job_id: "repro-job".into(),
            stage_id: 0,
            partition_id: sub_part_id,
        },
        executor_meta: ExecutorMetadata {
            id: format!("exec-{producer_task_id}"),
            host: "".to_string(),
            port: 0,
            grpc_port: 0,
            specification: ExecutorSpecification::default().with_vcores(0),
            os_info: ExecutorOperatingSystemSpecification::default(),
        },
        partition_stats: PartitionStats::new(Some(rows), None, None),
        file_id: Some(producer_task_id as u64),
        is_sort_shuffle: false,
    }
}

#[tokio::test]
async fn repro_routing_dropped_after_partitions_already_duplicated()
-> datafusion::error::Result<()> {
    let config = SessionConfig::new_with_ballista();
    let mut planner =
        AdaptivePlanner::try_from_plan(&config, stats_over_urre_root(), "repro-job".into())?;

    let stages = planner.runnable_stages()?;
    let stage_id = stages
        .as_ref()
        .and_then(|s| s.first())
        .map(|e| e.plan.stage_id())
        .expect("a runnable stage must exist");

    // Step 1: the remap, exactly as update_stage_progress calls it.
    // Producer task 7, sub-part 0, sketched [5, 25], straddles the cut at 15.
    let reports = vec![ballista_core::execution_plans::TaskRuntimeStats {
        producer_task_id: 7,
        report: RuntimeStatsReport {
            order_by: vec![],
            partitions: vec![RuntimeStatsPartitionEntry {
                partition_id: 0,
                row_count: 3,
                sketch: Some(ballista_core::execution_plans::sketch_to_proto(
                    &datafusion_functions_aggregate_common::tdigest::TDigest::new(100)
                        .merge_unsorted_f64(vec![5.0, 15.0, 25.0]),
                )?),
            }],
        },
    }];
    let cuts = vec![15.0];
    let remapped = cut_partitions(vec![vec![location(0, 7, 3)]], &reports, &cuts)?;

    assert_eq!(remapped[0].len(), 1, "straddler duplicated into partition 0");
    assert_eq!(remapped[1].len(), 1, "straddler duplicated into partition 1");

    // Step 2: park the routing so a read-side filter can trim it.
    let routing = RangeRepartitionRouting {
        cuts: cuts.clone(),
        routing_expr: col("v", v_schema().as_ref()).unwrap(),
    };
    let parked = planner.set_repartition_routing(stage_id as usize, routing);
    assert!(parked.is_ok(), "silently succeeds even though nothing was parked");

    // Step 3: confirm nothing was actually parked.
    let plan_str = format!(
        "{}",
        datafusion::physical_plan::displayable(planner.current_plan()).indent(true)
    );
    assert!(
        !plan_str.contains("range_repartition_cuts"),
        "no ExchangeExec carries the cuts, so no PerPartitionFilterExec downstream"
    );

    // What update_stage_progress assigns to output_locations.
    let file_ids: Vec<u64> = remapped
        .into_iter()
        .flatten()
        .map(|l| l.file_id.unwrap())
        .collect();
    assert_eq!(file_ids, vec![7, 7], "same file twice, 3 rows read as 6");

    Ok(())
}

#[tokio::test]
async fn control_routing_is_parked_when_chain_top_has_a_parent()
-> datafusion::error::Result<()> {
    let config = SessionConfig::new_with_ballista();
    let mut planner = AdaptivePlanner::try_from_plan(
        &config,
        stats_over_urre_with_parent(),
        "control-job".into(),
    )?;

    let stages = planner.runnable_stages()?;
    let stage_id = stages
        .as_ref()
        .and_then(|s| s.first())
        .map(|e| e.plan.stage_id())
        .expect("a runnable stage must exist");

    let routing = RangeRepartitionRouting {
        cuts: vec![15.0],
        routing_expr: col("v", v_schema().as_ref()).unwrap(),
    };
    planner.set_repartition_routing(stage_id as usize, routing)?;

    let plan_str = format!(
        "{}",
        datafusion::physical_plan::displayable(planner.current_plan()).indent(true)
    );
    assert!(
        plan_str.contains("range_repartition_cuts=1"),
        "control: the cuts must be parked on the boundary ExchangeExec"
    );

    Ok(())
}
```

</details>

## A side finding that might matter more

I first wrote the control with a `RoundRobinBatch` `RepartitionExec` parent, then with a `CoalescePartitionsExec` parent. Both were stripped by the optimizer pipeline, collapsing back to the root-level shape and making my control fail before I switched to `FilterExec`.

That is worth flagging on its own. `range_repartition_chain_top_gets_exchange_inserted` builds a `RoundRobinBatch` parent and calls `rule.optimize()` in isolation, where the parent survives and the exchange gets inserted. Through the real `AdaptivePlanner` pipeline that same parent is gone by the time DER runs. So the test passes for a plan shape the pipeline does not actually produce, and a rule author could write something that looks protected and lose the protection in production. Might be worth driving those DER tests through the planner rather than the rule alone.

## What I did not prove

I composed `cut_partitions` and `set_repartition_routing` directly in the order `update_stage_progress` calls them, rather than driving a full query to a wrong answer. Since no rule inserts a range repartition yet that is not constructible on this branch. The link I reasoned about but did not execute is the graph-side gate, `stages.get(&stage_id)` being `Running` with non-empty `runtime_stats_reports`. Reading `revive()` it promotes every Resolved stage to Running including the final one, and the `RuntimeStatsExec` in that stage's plan is what `collect_reports` walks, so I think it holds. Happy to be told I have that wrong.

## Suggested fix

Make `set_repartition_routing` error when it cannot find an `ExchangeExec` to park on, or move it ahead of `cut_partitions` so a failure to attach the filter aborts before the duplication is committed. Everything else in this code errors hard on invariant breaks like `missing file_id`, `no usable sketch`, and `merge_reports returned N groups`, so the silent `Ok(())` is the outlier. Separately `is_range_repartition_chain_top` probably wants to fire when the chain top is itself the plan root.

Nothing else in the review turned up anything. No public API breaks, the `GlobalPartitionMap` and `AdaptivePlanner` renames are both behind private modules, and the machinery really is inert until a rule lands.

Investigation and repro here were LLM assisted, with the test output and plan dumps above coming from actual runs against the PR head.

avantgardnerio:
@andygrove thank you for your thorough review. I'm very happy with the review quality in this project. The feedback is always substantive and positive. I believe I addressed your points, but I did leave a TODO. Unfortunately we're waiting on Datafusion for that enum variant. (RangeRepartitioning)

ballista/scheduler/src/state/aqe/optimizer_rule/distributed_exchange.rs
```diff
@@ -90,11 +93,36 @@ impl DistributedExchangeRule {
                 );
                 return Ok(Transformed::yes(Arc::new(exchange_exec)));
             }
+        } else if !execution_plan.is::<ExchangeExec>()
+            && let [child] = execution_plan.children().as_slice()
```
andygrove: Thanks for the fixes on the root-level case. I re-ran my repro and it is closed, and the fail-loud `set_repartition_routing` is exactly the right call.

One thing the simplification took with it though. Going from the multi-child walk to `let [child] = execution_plan.children().as_slice()` narrows this arm to single-child parents. A `UnionExec` no longer matches, and neither would a sort merge join range-partitioning both of its sides, which I would have guessed is one of the main consumers you have in mind for this machinery.

I ran the same probe against both commits, a `UnionExec` over two `RuntimeStatsExec -> URRE` chains:

| commit | exchanges inserted |
|---|---|
| `a92f0617` (before the fix) | 2 |
| `ea450346` (head) | 0 |

At head the plan comes out with no stage boundary anywhere:

```
AdaptiveDatafusionExec: is_final=false, plan_id=0, stage_id=pending, stage_resolved=false
  UnionExec
    RuntimeStatsExec: rows + sketch(routing=v@0 asc)
      UnorderedRangeRepartitionExec: routing=v@0 asc -> 4 partitions
        StatisticsExec: col_count=1, row_count=Absent
    RuntimeStatsExec: rows + sketch(routing=v@0 asc)
      UnorderedRangeRepartitionExec: routing=v@0 asc -> 4 partitions
        StatisticsExec: col_count=1, row_count=Absent
```

The probe, dropped into this file's test module next to your existing ones:

```rust
#[test]
fn probe_range_repartition_under_multi_child_parent() {
    let rule = DistributedExchangeRule::default();
    let parent: Arc<dyn ExecutionPlan> = Arc::new(
        datafusion::physical_plan::union::UnionExec::new(vec![
            stats_over_urre_over_leaf(),
            stats_over_urre_over_leaf(),
        ]),
    );

    let result = rule.optimize(parent, &config()).unwrap();
    println!("{}", display_plan(&result));
    assert_eq!(count_exchanges(result.as_ref()), 2);
}
```

The good news is this does not go silently wrong the way the root-level case did. `collect_reachable_stats` only walks single-child chains, so the union stage collects no reports, and `repartition_routing` then errors with `range-repartition stage N: no runtime-stats reports`. The job fails rather than duplicating rows.

The catch is that the message points somewhere else entirely. A rule author who splices under a join input would read it as a stats collection bug rather than "your splice position is not supported yet". So either restore the multi-child walk here, or if single-child-only is a deliberate scope call for now, would you mind saying so on `is_range_repartitioned` so the contract is explicit? A test pinning whichever behavior you pick would be worth having too.

Investigation and repro here were LLM assisted, with the plan dumps above coming from actual runs against both commits.

avantgardnerio: Thanks @andygrove . The change was intentional, though I did a poor job explaining that. Rationale: if this does both sides of a join (for instance), the scheduler needs to merge the cuts _from both sides of the join_ instead of each leg independently. I have not written that code yet, so walking both sides (like it did before) would have introduced another silent bug.

SMJ is definitely an intended use-case, so I think we need to add it in a follow up PR, but it is not a blocker for windows, so this PR does not include it - to keep this PR from growing further.

So I took option B from your advice and documented the behavior. The PR title probably slightly over-claims now, but I hope now that the code is consistent we can merge.

WDYT?

avantgardnerio: Edit: now with explicit `cross-stage cut coordination is not yet implemented` until this is implemented.

ballista/core/src/execution_plans/plan_algebra.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Looser sibling of [`preserves_distribution`]: partitioning survives,
+/// but rows and values within a partition are fair game.
+pub fn preserves_partitioning(plan: &dyn ExecutionPlan) -> bool {
+    // Distribution-preserving is strictly stronger; compose to keep the
+    // whitelist deduplicated.
+    preserves_distribution(plan)
+        // Drops rows, but per-partition — no rows migrate.
+        || plan.downcast_ref::<FilterExec>().is_some()
+        // Rewrites columns; partition boundaries untouched.
+        || plan.downcast_ref::<ProjectionExec>().is_some()
```
andygrove, line 67: A `ProjectionExec` here means the read-side filter can silently route on the wrong column.

`is_range_repartitioned` gates on `preserves_partitioning`, so a `ProjectionExec` between the URRE and the boundary is accepted and DER splices the `ExchangeExec` above it. But `repartition_routing_expr` recovers `order_by[0].expr` from the URRE, which is a `Column` indexed into the pre-projection schema, while `PerPartitionFilterExec` evaluates it against the post-projection schema. The old `preserves_distribution` list excluded `ProjectionExec` for exactly this reason ("might compute a new column that shadows or replaces the routing key"), and that exclusion is still what the routing expression needs.

I ran a probe with a projection that swaps two `Float64` columns above the URRE:

```
is_range_repartitioned(proj) = true

AdaptiveDatafusionExec: is_final=false, plan_id=1
  ExchangeExec: partitioning=None, plan_id=0
    ProjectionExec: expr=[tag@1 as tag, k@0 as k]
      UnorderedRangeRepartitionExec: routing=k@0 asc -> 4 partitions
        StatisticsExec: col_count=2, row_count=Absent

recovered routing_expr = k@0
read-side schema       = [tag: Float64, k: Float64]
predicate[1]           = k@0 >= 10 AND k@0 < 20
PerPartitionFilterExec::try_new -> Ok
```

`try_new`'s Boolean check passes because both columns are `Float64`, so index 0 now resolves to `tag` and this lands as wrong answers rather than an error.

The probe, if it is useful, dropped into the test module in `distributed_exchange.rs`:

```rust
#[test]
fn probe_projection_reindexes_routing_column() {
    use ballista_core::execution_plans::{
        PerPartitionFilterExec, range_partition_predicates, repartition_routing_expr,
    };
    use datafusion::physical_expr::PhysicalExpr;
    use datafusion::physical_plan::projection::ProjectionExec;

    let schema = Schema::new(vec![
        Field::new("k", DataType::Float64, false),
        Field::new("tag", DataType::Float64, false),
    ]);
    let stats = Statistics {
        num_rows: Default::default(),
        total_byte_size: Default::default(),
        column_statistics: vec![
            ColumnStatistics::new_unknown(),
            ColumnStatistics::new_unknown(),
        ],
    };
    let leaf: Arc<dyn ExecutionPlan> = Arc::new(StatisticsExec::new(stats, schema));
    let sort_expr = PhysicalSortExpr {
        expr: Arc::new(Column::new("k", 0)),
        options: SortOptions { descending: false, nulls_first: false },
    };
    let urre: Arc<dyn ExecutionPlan> = Arc::new(
        UnorderedRangeRepartitionExec::try_new(leaf, vec![sort_expr], 4).unwrap(),
    );
    // Projection swaps the columns: output is (tag, k).
    let proj_exprs: Vec<(Arc<dyn PhysicalExpr>, String)> = vec![
        (Arc::new(Column::new("tag", 1)), "tag".to_string()),
        (Arc::new(Column::new("k", 0)), "k".to_string()),
    ];
    let proj: Arc<dyn ExecutionPlan> =
        Arc::new(ProjectionExec::try_new(proj_exprs, urre).unwrap());

    println!("is_range_repartitioned = {}", is_range_repartitioned(&proj));
    let optimized = DistributedExchangeRule::default()
        .optimize(proj.clone(), &config())
        .unwrap();
    println!("{}", display_plan(&optimized));

    let routing = repartition_routing_expr(proj.as_ref()).unwrap().unwrap();
    println!("routing_expr = {routing}");
    println!("read-side schema = {:?}", proj.schema());

    let preds = range_partition_predicates(routing, &[10.0, 20.0, 30.0]);
    println!("predicate[1] = {}", preds[1]);
    println!("{:?}", PerPartitionFilterExec::try_new(proj, preds).is_ok());
}
```

To be clear, `preserves_distribution` is fine as-is, and `collect_reachable_stats` using it is correct. It is only this looser sibling being used as the gate in `is_range_repartitioned` that opens the hole. Options I can see are gating `is_range_repartitioned` on the strict predicate instead, or rebasing the routing expression onto the boundary schema when the cuts get parked.

ballista/scheduler/src/state/aqe/adapter.rs
```diff
@@ -105,7 +108,24 @@ impl BallistaAdapter {
                 )?,
             };
 
-            Ok(Transformed::yes(Arc::new(reader)))
+            let reader: Arc<dyn ExecutionPlan> = Arc::new(reader);
+            // Without a per-partition filter, straddling sub-parts from a
+            // range-repartitioned upstream would feed multiple downstream
+            // partitions and `FinalPartitioned` would split their partial sums.
+            if let Some(routing) = exchange.range_repartition_routing() {
```
andygrove, line 115: The coalesce slot and the routing slot can both be set on the same `ExchangeExec`, and this branch does not account for it.

`CoalescePartitionsRule` collects every leaf `ExchangeExec` in the subtree. It bails for `broadcast`, but there is no equivalent bail for a range-repartitioned exchange, so both slots land on the same node:

```
coalesce groups = 2, cuts = 3
```

When that happens, the `(Some(cp), false)` arm above builds a reader with `cp.groups.len()` partitions, and then this block hands `cuts.len() + 1` predicates to `PerPartitionFilterExec::try_new`, which rejects the count mismatch. So it is a hard error at plan time rather than wrong data, which makes it the lesser of the two issues, but it does mean the AQE coalesce path and any range-repartition rule are mutually exclusive with nothing saying so.

Worth noting the coalesce rule groups neighbouring upstream partitions only, so contiguous-group coalescing is actually compatible with range partitioning. Merging the cuts alongside the groups looks like a real fix rather than just a bail, if you want to go that way. A bail matching the broadcast one would be fine for now too, as long as it is explicit.

Repro:

```rust
#[test]
fn probe_coalesce_and_routing_on_same_exchange() {
    use ballista_core::execution_plans::{CoalescePlan, PartitionGroup};
    use crate::state::aqe::execution_plan::RangeRepartitionRouting;

    let ex = ExchangeExec::new(stats_over_urre_over_leaf(), None, 0);
    ex.set_coalesce(Arc::new(CoalescePlan {
        groups: vec![
            PartitionGroup { upstream_indices: vec![0, 1] },
            PartitionGroup { upstream_indices: vec![2, 3] },
        ],
        upstream_partition_count: 4,
    }));
    ex.resolve_range_repartition_routing(RangeRepartitionRouting {
        cuts: vec![10.0, 20.0, 30.0],
        routing_expr: Arc::new(Column::new("v", 0)),
    });
    println!(
        "coalesce groups = {}, cuts = {}",
        ex.coalesce().unwrap().groups.len(),
        ex.range_repartition_routing().unwrap().cuts.len()
    );
}
```

ballista/scheduler/src/state/aqe/execution_plan/exchange.rs
```diff
@@ -32,6 +33,26 @@ use parking_lot::Mutex;
...
+/// of this exchange. Written after the range-repartition-producing stage
+/// completes and its runtime-stats sketches are merged; read at
+/// task-specialization time to build per-downstream-partition range filters
+/// (see `PerPartitionFilterExec`).
+///
+/// `cuts` are `K - 1` monotone `f64` boundaries expressed in the value space
+/// of `routing_expr`; downstream partition `k` owns `[cuts[k-1], cuts[k])`
+/// with virtual `-∞`/`+∞` sentinels on the ends (matching the range
+/// repartition's write-side convention). `routing_expr` is the same
+/// expression the range repartition routes on — a `CAST(order_by[0] AS
```
andygrove: Small doc nit. `repartition_routing_expr` returns `order_by[0].expr` verbatim, not a cast of it. Both range repartition `try_new`s already require the expression to be `Float64`, so the code is right and this comment is just stale.

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -1357,7 +1462,7 @@ mod merge_tests {
     /// getting silently dropped.
     #[test]
     fn merge_reports_propagates_sketch_decode_errors() {
-        use crate::serde::protobuf::QuantileSketchState;
+        use QuantileSketchState;
```
andygrove: Leftover from the import rewrite. This compiles under uniform paths, but `use super::*` at the top of the module already brings `QuantileSketchState` in, so the line can go.

andygrove (approved):
Approving. The root-level exchange gap from my last round is genuinely closed, `set_repartition_routing` failing loud is the right call, and rejecting the multi-child shape at plan time with a clear message is a much better place to land than silently mis-routing. CI is green across TPC-H SF10 (AQE on and off) and TPC-DS SF1, and everything new is inert until a rule inserts a range repartition, so I am happy for this to go in as infrastructure.

Two things I found while probing that I do not think should block this PR, but that will bite the follow-up rule rather than this one. I left them inline. Could you file an issue covering both so they are tracked against whichever range-repartition rule lands first? Neither is reachable today, but the whole selling point here is that a rule author only has to splice two operators and gets correct routing for free, and these are the two cases where that contract does not hold yet.

One small thing, the PR body ends with the Claude Code footer. CLAUDE.md asks for no AI attribution in PRs, so worth stripping before merge.

Thanks for the iteration on this one, the shape is good.

ballista/scheduler/src/state/aqe/adapter.rs
avantgardnerio: Thanks for catching this. I will go with the "bail" option for now, but I filed https://github.com/apache/datafusion-ballista/issues/2220 for once we have a real rule using this (soon).

phillipleblanc:
Looks like Andy got to this before I had time to review, I think this makes sense though!

Issue #2198, opened 2026-07-28
Avoid single-task execution for null-aware anti joins (NOT IN subqueries)


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

#2188 fixed the correctness bug in #2187 by executing null-aware anti joins (`NOT IN` subqueries) in a single task. This is the only correct lowering available today, but it has two significant costs:

1. **No parallelism.** The join stage collapses to one task regardless of cluster size.
2. **The size cap lands on the outer table.** DataFusion only supports `null_aware` on `LeftAnti` joins with a single-column key (`hash_join/exec.rs` validation), which means the build side is always the *outer* table, not the subquery. So `ballista.optimizer.broadcast_join_threshold_bytes` (default 10 MiB) now effectively caps the size of the fact table in `SELECT ... FROM big_fact WHERE key NOT IN (SELECT k FROM small_dim)`. Any large outer table with known statistics gets a hard planning error.

Why single-task is currently unavoidable: DataFusion's null-aware hash join coordinates three pieces of global state across probe partitions of one operator instance, using in-process shared memory (`probe_side_has_null: AtomicBool`, `probe_side_non_empty: AtomicBool`, and the visited-build-row bitmap). The last probe partition to finish emits the unmatched build rows, suppressing all output if any partition saw a NULL key. Split the probe across processes and each task gets private copies of all three, producing duplicated or incorrect rows. Hash-partitioning both sides on the key does not help either, because "probe side has a NULL" is a global property: a NULL key hashes to one partition and every other partition needs to know about it.

For comparison, Spark (since 3.1, [SPARK-32290](https://issues.apache.org/jira/browse/SPARK-32290)) plans single-column `NOT IN` with the opposite build side: it broadcasts the *subquery* as a special `HashedRelation` that carries "is empty" and "contains NULL" flags computed at build time. Each outer partition then applies purely local logic, so the outer table stays fully distributed and the broadcast constraint lands on the subquery, which is almost always the small side.

**Describe the solution you'd like**

Three complementary options, roughly in order of leverage:

[...]

Issue #2193, opened 2026-07-27
Default join planning loses NOT IN null semantics in SortMergeJoinExec


phillipleblanc:
## Bug

Ballista defaults `datafusion.optimizer.prefer_hash_join` to `false`. For a SQL `NOT IN` predicate, DataFusion's physical planner can therefore replace the logical null-aware anti join with a `SortMergeJoinExec` before Ballista's scheduler rules run.

`SortMergeJoinExec` does not carry the `null_aware` flag. The scheduler cannot recover the required `NOT IN` NULL semantics afterward.

## Reproduction

Use four CSV partitions for each table:

- `t1(a)`: values `0..19`
- `t2(b)`: values `0..4`, with one key replaced by `NULL`

Run Ballista with its default join preference:

```sql
select a
from t1
where a not in (select b from t2)
order by a;
```

Expected result: **0 rows**, because the subquery contains `NULL`.

Observed result: **16 rows**. Single-process DataFusion returns the correct 0 rows.

Setting `datafusion.optimizer.prefer_hash_join = true` retains a null-aware `HashJoinExec`, so the scheduler can recognize and lower it correctly. Apache #2188 addresses that hash-join and adaptive-planning path, but cannot fix a null-aware flag that was already discarded.

## Expected behavior

Physical planning should retain null-aware anti-join semantics regardless of the global hash-join preference. Possible approaches include forcing a null-aware `HashJoinExec` for this logical shape or adding equivalent null-aware support to the selected physical join.

Discovered while validating #2188 from the end-to-end reproduction in @andygrove's review.

amitvijapur:
This looks addressed on the DataFusion side by apache/datafusion#22810 (merged 2026-07-24, three days before this issue): the physical planner now never routes a null-aware anti join to `SortMergeJoinExec`, whatever `prefer_hash_join` says (`physical_planner.rs`, the `can_repartition_join` / sort-merge branch checks `!*null_aware`). That commit is in DataFusion 55.0.0, which `main` here depends on.

I have not re-run the four-partition CSV reproduction on a cluster, so this is a pointer rather than a confirmation. If the repro passes on `main` now, the issue can probably be closed.

PR #2199, opened 2026-07-28
feat(core): rewrite NOT IN subqueries into distributable anti join plans

andygrove:

Part of #2198.

# Rationale for this change

#2188 made null-aware anti joins correct by collapsing them to a single task, because DataFusion's null-aware hash join coordinates its visited-row bitmap and probe-side NULL state through in-process atomics. That lowering loses all parallelism and caps the build side (the outer table) at `ballista.optimizer.broadcast_join_threshold_bytes`.

This PR implements option 2 from #2198: rewrite uncorrelated `NOT IN (subquery)` filter predicates during logical optimization so that no null-aware join is planned in the first place. For

```sql
SELECT a FROM t1 WHERE a NOT IN (SELECT b FROM t2)
```

the rewrite produces

```text
Projection: t1.a
  Filter: __cnt = 0 OR (t1.a IS NOT NULL AND __cnt = __cnt_non_null)
    CrossJoin
      Aggregate: count(1) AS __cnt, count(b) AS __cnt_non_null   (over t2)
      LeftAnti Join: t1.a = t2.b
        TableScan: t1
        SubqueryAlias: t2
```

which is equivalent under SQL three-valued `NOT IN` semantics in a `WHERE` context: an empty subquery passes every row, a matching probe value is dropped by the anti join, and a NULL probe value or a NULL in the subquery drops the remaining rows via the one-row aggregate. Every operator in the rewritten plan is an ordinary distributable one, so the outer table keeps full parallelism and the broadcast threshold no longer applies.

A side effect is that `NOT IN` now also returns correct results under Ballista's default sort-merge-join configuration, where the null-aware flag was previously dropped before scheduler lowering (the practical impact of #2193).

- New `NotInSubqueryRewrite` logical optimizer rule in `ballista-core`, prepended to DataFusion's default rules so it runs before `DecorrelatePredicateSubquery`. The rewrite only fires where null-aware semantics would otherwise be needed: uncorrelated subqueries with a single output column and a nullable key on either side. Correlated, multi-column, volatile, and provably non-nullable predicates are left to DataFusion's own decorrelation, and the #2188 single-task lowering remains as the fallback for anything the rewrite does not cover.
- Registration in every Ballista session builder: `SessionState::new_ballista_state`, `upgrade_for_ballista`, and the scheduler's `default_session_builder`.
- New config key `ballista.optimizer.not_in_subquery_rewrite` (Boolean, default `true`); setting it to `false` restores the previous null-aware planning and single-task lowering.
- Unit tests for the rewrite shape and its skip conditions, and new `default` and `rewrite_disabled` variants in the `null_aware` integration test. The `default` variant fails without this change (wrong results on the sort-merge path) and passes with it.
- Config table and tuning guide documentation updates.

`NOT IN` queries with uncorrelated single-column subqueries now execute as fully distributed anti joins under static and adaptive planning, including under the default configuration, and are no longer subject to the broadcast threshold cap or the single-task lowering. The new `ballista.optimizer.not_in_subquery_rewrite` key (default `true`) disables the rewrite if needed. No public API changes.


andygrove:
cc @phillipleblanc

andygrove:
For reviewers, it may be useful to compare this with how Spark handles the same problem.

Spark plans single column `NOT IN` as a null-aware anti join (SPARK-32290). It broadcasts the subquery side as a special `HashedRelation` that carries two flags computed while building the broadcast: whether the subquery is empty and whether it contains a NULL key. Each streamed partition of the outer table then applies purely local logic. The subquery is empty, so emit the row. The subquery has a NULL, or my key is NULL, or my key matches, so drop the row. No cross partition coordination is needed and the outer table stays fully distributed.

This PR reaches the same place but at the plan level instead of inside the join operator. The one row `count(*)` / `count(b)` aggregate captures exactly the two global facts Spark bakes into its broadcast relation, and the plain anti join takes the place of Spark's special null-aware probe logic. The filter over the cross join is the same local decision Spark makes per streamed row.

A few practical differences:

- Spark makes one pass over the subquery, since the flags are computed while building the broadcast. This rewrite scans the subquery twice, once for the anti join and once for the aggregate. Both scans are cheap relative to the join itself and both distribute.
- Spark's null-aware anti join requires the subquery to be broadcastable and falls back to `BroadcastNestedLoopJoin` otherwise. The rewritten plan here has no broadcast requirement at all, since the anti join can shuffle on the key, so large subqueries are fine.
- Both approaches are limited to single column `NOT IN`. Multi column predicates fall through to the existing paths.

The closest structural match to Spark would be a null-aware join that builds on the subquery side, which is the upstream DataFusion improvement tracked as option 1 in #2198. If that lands, the special casing here can eventually go away and the swap optimizations would produce Spark's shape natively.

avantgardnerio (approved):
This is a big step in the right direction, ty.

PR #2202, opened 2026-07-28
fix: classify wrapped FetchFailed as fetch partition error

### LINKED ISSUE #2027: Shuffle-fetch failures lose their type, so the map-stage resubmit never fires and the job fails
## Describe the bug

When an executor dies after producing shuffle output, Ballista **fails the query instead of re-running the map stage**. The `FetchPartitionError` recovery path — arguably the most important HA path in the engine — never fires, because the typed error that triggers it is destroyed before the scheduler sees it.

## Root cause

`ballista/core/src/execution_plans/shuffle_reader.rs` (lines ~1756 and ~1773) converts a Ballista error into a DataFusion error by **Debug-formatting it into a string**:

```rust
.map_err(|e| DataFusionError::Execution(format!("{e:?}")))
```

A typed `BallistaError::FetchFailed(executor_id, map_stage_id, map_partition_id, desc)` is thereby flattened into an opaque `DataFusionError::Execution("FetchFailed(\"...\", 1, 2, \"...\")")`.

By the time the executor builds its task status, the error is `BallistaError::DataFusionError(Execution(..))`, **not** `BallistaError::FetchFailed(..)`. So this arm in `ballista/core/src/error.rs` (~line 208):

```rust
BallistaError::FetchFailed(executor_id, map_stage_id, map_partition_id, desc) => FailedTask {
    failed_reason: Some(FailedReason::FetchPartitionError(FetchPartitionError { .. })),
    ..
}
```

never matches. The error falls through to the catch-all (~line 248) and is reported as:

```rust
retryable: false,
failed_reason: Some(FailedReason::ExecutionError(..)),
```

The scheduler therefore never receives a `FetchPartitionError`, so the logic in `ballista/scheduler/src/state/execution_graph.rs` (~line 826) that removes the lost input partitions and **resubmits the map stage** is unreachable via this path. The stage is failed instead, and the job fails.

## To Reproduce

Reproduced end to end on a real multi-process cluster by the chaos harness in #2026 (scenario `executor_killed_mid_stage_is_recovered`, which fails under **both** AQE on and AQE off):

1. Start a scheduler and 2 executors.
2. Run a query whose plan has a shuffle (join + grouped aggregate).
3. `SIGKILL` one executor while stage 1 has running tasks.
4. The downstream stage tries to fetch shuffle partitions from the dead executor.

Observed job failure:

```
Job failed due to stage 2 failed: Task failed due to runtime execution error:
DataFusionError(Execution("FetchFailed(\"e40d5c70-cf84-466a-b390-91cc861401ff\", 1, 2,
  \"Error connecting to Ballista scheduler or executor at http://127.0.0.1:49643:
   tonic::transport::Error(Transport, ConnectError(... Connection refused ...))\")"))
```

Note `Task failed due to runtime execution error` — that is the **catch-all** message from `error.rs`, confirming the `FetchFailed` arm did not match.

## Expected behavior

The scheduler should classify this as a `FetchPartitionError`, remove the input partitions produced by the dead executor, resubmit the map stage, and complete the query with the correct result. Ballista already implements all of that; it simply never gets the chance.

## Additional context

This is one of two bugs with the same underlying shape: **structured Ballista errors lose their type when they cross the DataFusion error boundary, and retryability is then decided by shallow pattern matches that cannot see through the wrapping.** See the sibling issue on `DataFusionError::Shared` and retryable IO errors.

A fix likely needs `BallistaError` to survive the round-trip through `DataFusionError` (e.g. carried as a boxed external error rather than a formatted string) so that `error.rs` can recover the typed variant — plus a downcast/unwrap at the classification site rather than a shallow `matches!`.

villebro:
Closes #2027.

# Rationale for this change

Issue #2027 describes scheduler recovery failing after an executor dies after producing shuffle output. The shuffle fetch path starts with a structured `FetchFailed { executor_id, map_stage_id, map_partition_id, message }`, but that error could be flattened into a generic `DataFusionError::Execution("...")` string while crossing the shuffle-writer handoff. Once flattened, the scheduler saw a non-retryable `ExecutionError` instead of the `FetchPartitionError` signal it needs to invalidate and re-run the lost map output.

The exact location hints in the issue reflect the codebase when the issue was written. The production type loss fixed here is the wrapping/stringification path above.

The core fix keeps shuffle-fetch failures structural until task failure classification. The regular and sort-shuffle writers now share the original structural error with every output-partition handoff instead of formatting it into a string, and the executor uses the existing `BallistaError` conversion at the query-stage boundary.

Error behavior before and after:

| Before | After |
| --- | --- |
| A shuffle fetch failure could reach the scheduler as a generic `ExecutionError` string. | Bare or wrapped `FetchFailed` errors become `FetchPartitionError`, including when wrapped through `ArrowError::ExternalError`, `DataFusionError::ArrowError`, `DataFusionError::Shared`, or `DataFusionError::Context`. |
| Sort-shuffle retryable IO errors crossing the writer handoff could be flattened into generic `ExecutionError`s. | The sort-shuffle coordinator shares the original `DataFusionError`, preserving the `IoError` root so it stays retryable. |
| Executor-reported `Cancelled` from stale task cleanup could be treated as a fatal execution failure. | It is classified as retryable, non-counting `TaskKilled`; explicit job cancellation still marks the job terminal before executor tasks are cancelled. |

Enabling the chaos regression also exposed adjacent recovery gaps needed for that end-to-end scenario:

- Executor-loss handling now wakes push scheduling after reset work.
- AQE now preserves stage input location metadata so lost map output can be invalidated and resubmitted.
- AQE now drops stale active shuffle locations for a lost executor before constructing downstream stages.

The chaos harness changes are mostly regression coverage for this fix. Two harness-stability changes are slightly unrelated to the production fix but were needed to avoid CI flakiness while testing it: the chaos executor now enables the existing scheduler-connect retry with a 10s timeout, and the startup canary uses a longer executor timeout.

The `ArrowError::ExternalError` wrapping is intentional: it matches the shape produced by the shuffle fetch stream boundary, and Ballista's conversion/classification logic now recovers that shape through `Shared` and `Context` wrappers. This PR does not introduce a separate `DataFusionError::External(Box<BallistaError>)` fetch path.

No API or documentation changes.

Behavior change: shuffle fetch failures that were previously reported as non-retryable execution errors are now classified as fetch-partition errors, so the scheduler can resubmit lost map output instead of failing the job. Sort-shuffle IO errors crossing the same handoff keep their retryable IO classification. Executor-side cancellation of stale task attempts during recovery is treated as retryable cleanup, while explicit job cancellation still marks the job terminal before executor tasks are cancelled. AQE also drops stale active shuffle locations for a lost executor before constructing downstream stages.

### DIFF
--- a/ballista/core/src/execution_plans/shuffle_writer.rs
+++ b/ballista/core/src/execution_plans/shuffle_writer.rs
@@ -849,19 +849,10 @@ async fn run_coordinator(
             }
         }
         Err(e) => {
-            // The stage driver drains output partition 0 first. Move the
-            // structural error there; later streams only need to unblock.
-            let msg = e.to_string();
-            let mut first_error = Some(e);
-            for (i, slot) in senders.iter_mut().enumerate() {
+            let shared = Arc::new(e);
+            for slot in senders.iter_mut() {
                 if let Some(sender) = slot.take() {
-                    let err = if let Some(e) = first_error.take() {
-                        e
-                    } else {
-                        DataFusionError::Execution(format!(
-                            "shuffle writer failed for output partition {i}: {msg}"
-                        ))
-                    };
+                    let err = DataFusionError::from(&shared);
                     let _ = sender.send(Err(err));
                 }
             }
@@ -943,6 +934,8 @@ pub(crate) fn summaries_to_batch(
 #[allow(dead_code, unused_imports)] // clippy false positive with local imports
 mod tests {
     use super::*;
+    use crate::error::BallistaError;
+    use crate::execution_plans::ChaosExec;
     use datafusion::arrow::array::{StringArray, StructArray, UInt32Array, UInt64Array};
     use datafusion::datasource::memory::MemorySourceConfig;
     use datafusion::datasource::source::DataSourceExec;
@@ -980,6 +973,50 @@ mod tests {
             all.extend(batches);
         }
         Ok(all)
+    }
+
+    async fn drive_partition_results(
+        plan: Arc<ShuffleWriterExec>,
+        task_ctx: Arc<TaskContext>,
+    ) -> Vec<crate::error::Result<Vec<RecordBatch>>> {
+        let k = plan.properties().output_partitioning().partition_count();
+        let mut handles = Vec::with_capacity(k);
+        for n in 0..k {
+            let plan = plan.clone();
+            let ctx = task_ctx.clone();
+            handles.push(tokio::spawn(async move {
+                let mut stream = plan.execute(n, ctx).map_err(BallistaError::from)?;
+                utils::collect_stream(&mut stream).await
+            }));
+        }
+        let mut results = Vec::new();
+        for h in handles {
+            results.push(
+                h.await
+                    .expect("drive_partition_results task should not panic"),
+            );
+        }
+        results
+    }
+
+    fn assert_shared_structural_error(err: &BallistaError) {
+        let BallistaError::DataFusionError(e) = err else {
+            panic!("expected DataFusionError, got {err:?}");
+        };
+        assert!(
+            matches!(e.as_ref(), DataFusionError::Shared(_)),
+            "expected shared DataFusionError, got {e:?}"
+        );
+        let DataFusionError::ArrowError(arrow, _) = e.find_root() else {
+            panic!("expected ArrowError root, got {:?}", e.find_root());
+        };
+        let ArrowError::ExternalError(inner) = arrow.as_ref() else {
+            panic!("expected Arrow external error, got {arrow:?}");
+        };
+        assert!(
+            inner.downcast_ref::<BallistaError>().is_some(),
+            "expected Arrow external BallistaError, got {inner:?}"
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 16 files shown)


villebro:
I'm marking this as draft until #2026 is merged so we can enable the disabled chaos harness tests related to shuffle fetching as part of this PR.

andygrove:
> I'm marking this as draft until #2026 is merged so we can enable the disabled chaos harness tests related to shuffle fetching as part of this PR.

#2026 is now merged

villebro:
I rebased and I'm now enabling the new tests, will push and remove draft when I have everything working.

milenkovicm:
One maybe unreleased question, what do you think if we remove BallistaError from interfaces and return DataFusionError. We can keep simplified BallistaError but it would be packed in DFE:: External?

We have a mix of DF and Ballista errors across interfaces and sometimes it gets confusing

villebro:
> One maybe unreleased question, what do you think if we remove BallistaError from interfaces and return DataFusionError. We can keep simplified BallistaError but it would be packed in DFE:: External?
> 
> We have a mix of DF and Ballista errors across interfaces and sometimes it gets confusing

@milenkovicm I agree we should simplify/consolidate as much as possible. However, I think it's a good idea to do this as a follow-up, as the consolidation work will probably branch out into many unrelated areas (this PR is already getting quite loaded). If this sounds good I can take this on after this PR is merged?

milenkovicm:
I agree, we can take it as a follow up it you believe it makes sense.

Had a quick look at this pr, it makes sense, will do another pass later

villebro:
> Had a quick look at this pr, it makes sense, will do another pass later

Thanks for reviewing. Btw, I removed Draft from this PR yesterday as I thought it was blocking CI. Turns out GH was just borked all day. But let me make sure CI is green and do one last review pass and then ping you for a final review when I feel it's completely done.

ballista/core/src/execution_plans/shuffle_writer.rs
```diff
@@ -575,7 +575,12 @@ impl ShuffleWriterExec {
                         compression_type,
                     )
                     .await
-                    .map_err(|e| DataFusionError::Execution(format!("{e:?}")))?;
+                    .map_err(|e| {
+                        DataFusionError::ArrowError(
```
milenkovicm, line 579: why dont we use `DataFusionError::External` ?

ballista/scheduler/src/state/aqe/planner.rs
```diff
@@ -650,6 +661,5 @@ impl AdaptivePlanner {
 /// Wraps stage plan with addition of references to previous stages
 pub(crate) struct AdaptiveStageInfo {
     pub(crate) plan: Arc<dyn ShuffleWriter>,
-    #[allow(dead_code)] // TODO: still not sure if this is needed
```
milenkovicm, line 653: i like to call myself API visionist I had a hunch it'll be needed in the future 😀

ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -3227,6 +3218,25 @@ mod test {
     //     todo!()
     // }
 
+    fn wrapped_fetch_failed_task(
+        executor_id: &str,
+        map_stage_id: usize,
+        map_partition_id: usize,
+    ) -> FailedTask {
+        let err = BallistaError::DataFusionError(Box::new(DataFusionError::ArrowError(
```
milenkovicm, line 3226: same question here, cant we use DFE::External

milenkovicm (approved):
thanks @villebro  great addition, i have just one question/comment regarding errors used im not sure i understand 

> This stays targeted to the known shuffle fetch path, where FetchFailed travels through ArrowError::ExternalError. A direct DataFusionError::External(Box<BallistaError>) fetch path is not known to be produced here and is outside this change.

ballista/core/src/execution_plans/shuffle_writer.rs
villebro: Good point. I agree it would be a cleaner/shorter wrapper shape, but switching to it here would require adding an explicit downcast/unwrapping path and corresponding tests, which would probably increase the complexity of this PR. I think it could be an excellent follow-up PR focused specifically on harmonizing these error paths, and one I can take on after this is merged.

villebro:
@milenkovicm I left a comment: let me know if you're ok with following up with a targeted follow-up PR for consolidating/simplifying the error paths or if you prefer to have those changes in this PR.

milenkovicm:
Thanks @villebro we can split them if you prefer it, let's merge this

PR #2211, opened 2026-07-30
feat: parallel prefix scan for UNBOUNDED PRECEDING window aggregates

avantgardnerio:
## Summary

`BoundedWindowAggExec` declares `SinglePartition` when there is no PARTITION BY, because a window frame spans rows across the whole input. For an ever-expanding frame that means the entire window computation runs on one core, however wide the cluster.

This PR removes the collapse. Each task computes a partition-local running aggregate over a range-disjoint slice, and a downstream operator corrects it with the merged accumulator state of every prior partition.

## Shape

```
PrefixMergeExec                        state applied row-wise
  ExchangeExec (partitioning: None)    boundary 2, planted by the rule
    PartitionedBoundedWindowAggExec    window runs per partition
      RangeFilterExec                  trim to this task's cut range
        ExchangeExec                   boundary 1, inserted by DistributedExchangeRule
          RuntimeStatsExec
            OrderedRangeRepartitionExec
              SortExec
                RuntimeStatsExec
                  <source>
```

Boundary 2 carries no repartition. It exists so the scheduler has a point where every upstream task has published its state.

## State flow

DataFusion 55 exposes accumulator state through a callback (apache/datafusion#24035). A `WindowStateCollector` catches it, `ShuffleWriterExec` translates each capture's task-local partition index to a stage-global one, and it rides `SuccessfulTask` to the scheduler. The scheduler prefix-merges the reports into one carry-in per partition and binds them to the downstream `PrefixMergeExec`.

Merging goes through the aggregate's own `merge_batch`, so non-decomposable aggregates work: two `approx_distinct` HLL sketches combine correctly where two distinct counts could not.

One state's wire size is fixed by the sketch rather than by the rows it saw. An `approx_distinct` state encodes to 16393 bytes at one row and at 100k alike, because the HLL is a dense register array, so a task's payload is (partitions in its slice) x (aggregate window expressions) x 16 KiB with no term for input size. `sketch_state_wire_size_does_not_grow_with_rows` pins that.

## Benchmarks

h2o `window.sql` Q7, the rolling sum:

```sql
SELECT id1, id2, id3, v2,
       sum(v2) OVER (ORDER BY id3 ROWS BETWEEN UNBOUNDED PRECEDING
                                           AND CURRENT ROW) AS my_rolling_sum
FROM large;
```

1e9 rows read from S3. Four executors at 48 vcores each, one per `r6i.24xlarge`,
pinned one-per-node by anti-affinity; the client runs on a fifth node so pulling
the result back does not contend with an executor. Partitions per task is set to
K/4 throughout, so each executor takes exactly one task and the shuffle reader's
merge width stays fixed at 4 while K varies. Every figure is a single run.

image

**Speedup grows with width:**

| partitions | rule off | rule on | server-side | wall |
|---|---|---|---|---|
| 16 | 290.2s | 84.5s | **3.43x** | 2.30x |
| 32 | 286.1s | 54.1s | **5.29x** | 2.81x |
| 64 | 305.1s | 42.4s | **7.19x** | 3.23x |
| 128 | 318.5s | 33.5s | **9.50x** | 3.64x |

**Where the time goes**, `elapsed_compute` per stage:

| stage | K=16 | K=32 | K=64 | K=128 |
|---|---|---|---|---|
| 0, sort plus ORRE | 48.35s | 31.11s | 27.89s | 24.43s |
| 1, the window | 22.51s | 12.41s | 9.34s | 6.09s |
| 2, `PrefixMergeExec` correction | 13.62s | 10.50s | 5.19s | 2.92s |
| the same window, rule off, 1 task | 253.46s | 269.29s | 294.82s | 313.57s |

Taking stages 1 and 2 together against the serial window, the window work itself
goes 7.0x, 11.8x, 20.3x, then **34.8x**.

Three things the tables show:

- **Both ends widen.** The rewrite improves with width while the serial arm gets
  monotonically *worse* (253s to 314s), because its `SortPreservingMergeExec`
  collapses an ever-wider fan into one task. Adding cores actively hurts the
  status quo, which is the shape of a serial bottleneck rather than a slow
  operator.
- **The correction is cheaper than the window it corrects at every width**, and
  it kept scaling to the end, 13.62s down to 2.92s. The per-row `evaluate()`
  replay is the cheapest of the three stages by K=128.
- **Stage 0 is the wall, not the correction.** It flattens at roughly 24s and by
  K=128 is 73% of the parallel plan, so the asymptote here is about 13x. Further
  width buys progressively less until that read-sort-repartition preamble gets
  cheaper.

## What this measurement understates

- **`SUM` is the cheapest possible window aggregate.** The design exists for aggregates whose per-row output is not a valid partial state: `approx_distinct`, quantile sketches, stddev. There each row costs a register scan or a quantile computation in `evaluate()`, so the window dominates and the sort and repartition tax rounds to noise. Q7 is the weakest showcase in the suite.
- **`WindowApply::Scalar` is implemented but not yet selected.** For SUM, COUNT, MIN and MAX it replaces the correction's per-row accumulator replay with one arrow kernel per batch, which should remove most of stage 2 for exactly the query measured above.
- **The result set inflates both arms equally.** Q7 emits one row per input row, so the client collects 1e9 rows in both configurations. That is a constant added to each side, which compresses the ratio: 9.50x server-side becomes 3.64x wall.

## What flatters it

- **The serial baseline may be bandwidth-bound rather than CPU-bound.** Its collapse stage pulls all 1e9 rows from S3 through a single task, so part of the 253-314s is fetch rather than window computation. The rewrite spreads that read across four executors, and some of the ratio is that rather than the window itself.

## What works against it

- **The extra materialization makes the win contingent on shuffle throughput.** Boundary 2 writes and reads the whole window output. Where that write path is slow relative to compute, the extra round trip can eat the gain.
- **The ORRE preamble is not free, and it is now the floor.** Stage 0 pays for the sketch and the range repartition, and it stops improving with width: 48.35s at K=16 down to only 24.43s at K=128, by which point it is 73% of the total.
- **Three stages rather than two,** so one more scheduler round trip and the state transport described above.
- **The correction is still O(rows) with an `evaluate()` per row.** It measures well here and scales with cores, but it is linear work that the Scalar path avoids entirely where it applies.
- **Narrow trigger.** See "Known limits".
- Cost-model: for a user like Coralogix, using 16x the cores to get 4x the wallclock is a big win. For the Spark persona running a fixed-sized cluster of owned nodes and not latency sensitive, this rule might be a net loss - better for the job to take longer and free the cores for other jobs. It could solve OOMs for them though.

## Correctness

`ballista/client/tests/prefix_window.rs` runs a distributed running sum against an independently computed oracle through the real scheduler, shuffle and executor, with `max_partitions_per_task = 2` so tasks carry genuine multi-partition slices. Two routing keys are covered: a non-null `Float64` one, and a nullable `Int64` one where the NULL run has to reach the partition that holds it.

## One fix outside the feature flag

Writing that nullable test surfaced a pre-existing bug in `cut_partitions`, which assigns each producer's shuffle files to consumers by value range. The reported extremes are *value* extremes, so a file holding both values and NULLs was delivered only to the consumers its values overlap, and its NULL run went wherever its values went. Every producer routes its own NULLs to its own last slot, so a stage with more than one producer task silently dropped all but one producer's share of the run: a short result with every surviving row individually correct.

Fixed by delivering a file carrying NULLs to the run's partition as well, skipped when its value range already reaches there. This is shared range-repartition machinery, reached by any ORRE/URRE stage rather than only by the rewrite added here, so it is the one change in this PR that is not behind the flag.

## Safety

Gated behind `ballista.planner.parallel_window.enabled`, off by default. The rewrite fires only on: no PARTITION BY, a single ascending ORDER BY column whose type the sort-key sketch codec can encode, and an ever-expanding frame. Everything else falls through untouched.

## Known limits

- No PARTITION BY, single ORDER BY column, ASC only.
- Fixed-width routing keys. The sketch codec packs a key into a `u64`, so variable-length types such as `Utf8` fall through the gate. Nullable keys are supported.
- `WindowApply::Scalar` is implemented but not yet selected; every apply currently routes through the accumulator path.

### DIFF
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -803,12 +803,6 @@ message SuccessfulTask {
 // `BoundedWindowAggExec`.
 message WindowStateReport {
   // The stage's *global* output partition this state belongs to.
-  //
-  // DataFusion reports a task-local index, because a task's plan is
-  // restricted to a partition slice. The producing `ShuffleWriterExec`
-  // translates it through the `global_output_partition_ids` the scheduler
-  // stamped on it, so what crosses the wire is already global — the
-  // scheduler must not re-derive it.
   uint32 global_partition_id = 1;
   // Position in the window operator's `window_expr()` list. Indexes window
   // expressions, not partitions.
--- a/ballista/core/proto/ballista.proto
+++ b/ballista/core/proto/ballista.proto
@@ -59,6 +59,7 @@ message BallistaPhysicalPlanNode {
     PartitionedBoundedWindowAggExecNode partitioned_bounded_window_agg = 11;
     RangeShuffleReaderExecNode range_shuffle_reader = 12;
     RangeFilterExecNode range_filter = 13;
+    PrefixMergeExecNode prefix_merge = 14;
   }
 }
 
@@ -211,6 +212,72 @@ message SortOptions {
 message RangeBound {
   datafusion_common.ScalarValue lo = 1;
   datafusion_common.ScalarValue hi = 2;
+}
+
+// Applies scheduler-computed prefix state to a window-aggregate column, so
+// each range-disjoint task's local running aggregate becomes a global one.
+message PrefixMergeExecNode {
+  // One entry per output column needing correction. Empty makes the operator
+  // a passthrough.
+  repeated WindowApplyNode applies = 1;
+  // Resolved prefix state, one entry per input partition, `[k]` summarising
+  // every partition before `k`. Encoding refuses while unresolved — an
+  // over-the-wire plan always ships with state bound, since an executor has
+  // no way to obtain it.
+  repeated FinalizedPartitionStateNode per_partition_state = 2;
+}
+
+// How to correct one window-function output column.
+message WindowApplyNode {
+  oneof apply {
+    ScalarWindowApplyNode scalar = 1;
+    AggregateWindowApplyNode aggregate = 2;
+  }
+}
+
+// Fast path: combine each row's value with a per-partition scalar.
+message ScalarWindowApplyNode {
+  ScalarOpNode op = 1;
+  // One scalar per input partition.
+  repeated datafusion_common.ScalarValue offset = 2;
+  uint32 output_column = 3;
+}
+
+// How a scalar offset combines with a row's existing value.
+enum ScalarOpNode {
+  SCALAR_OP_NODE_ADD = 0;
+  SCALAR_OP_NODE_MIN = 1;
+  SCALAR_OP_NODE_MAX = 2;
+  // Ignores the row's value entirely.
+  SCALAR_OP_NODE_OVERWRITE = 3;
+}
+
+// Fallback path: seed a fresh accumulator from the prefix state and replay
+// each row through it. Needed where a row's output is not itself a valid
+// partial state — AVG, and sketch-backed aggregates like approx_distinct.
+message AggregateWindowApplyNode {
+  // Resolved from the executor's function registry on decode.
+  string udf_name = 1;
+  repeated datafusion.PhysicalExprNode args = 2;
+  uint32 output_column = 3;
+  // Position in the upstream window operator's `window_expr()` list, which
+  // is what indexes into each FinalizedPartitionStateNode.
+  uint32 window_expr_index = 4;
+}
+
+// Prefix state for one input partition: one slot per window expression.
+message FinalizedPartitionStateNode {
+  repeated AggregateStateSlotNode slots = 1;
+}
+
+message AggregateStateSlotNode {
+  // Unset when that window expression published no state — a non-aggregate
+  // window function. Distinct from a present-but-empty state.
+  AggregateStateNode state = 1;
+}
+
+message AggregateStateNode {
+  repeated datafusion_common.ScalarValue values = 1;
 }
 
 // Wrapper for `BoundedWindowAggExec` that overrides
@@ -713,6 +780,48 @@ message SuccessfulTask {
   // executed `RuntimeStatsExec`; the scheduler groups by `order_by` tag
   // to combine reports across tasks/executors.
   repeated RuntimeStatsReport runtime_stats = 3;
+  // Finalized window-aggregate state captured during this task, one entry
+  // per (output partition, window expression, PARTITION BY group) that
+  // closed. Empty unless the plan contains an ever-expanding-frame window
+  // (`UNBOUNDED PRECEDING`), which is the only shape DataFusion will
+  // publish accumulator state for. The scheduler prefix-merges these across
+  // tasks and bakes the result into a downstream `PrefixMergeExec`.
+  //
+  // TODO: watch the size of this. Task completion is a hot, frequent
+  // message, and sketch-backed aggregates make the payload unbounded in a
+  // way row counts and quantile sketches are not — an HLL or KLL state is
+  // kilobytes per window expression per partition, and a task covering a
+  // wide partition slice carries one of each. If it stops being small,
+  // write the state as a sidecar next to the shuffle files instead, the way
+  // sort-shuffle already writes `<data>.arrow.index` beside its data
+  // (`sort_shuffle::get_index_path`), and send only a reference here. That
+  // keeps the completion message fixed-size regardless of aggregate.
+  repeated WindowStateReport window_state = 4;
+}
+
+// One finalized window-aggregate state from a task's
+// `BoundedWindowAggExec`.
+message WindowStateReport {
+  // The stage's *global* output partition this state belongs to.
+  //
+  // DataFusion reports a task-local index, because a task's plan is
+  // restricted to a partition slice. The producing `ShuffleWriterExec`
+  // translates it through the `global_output_partition_ids` the scheduler
+  // stamped on it, so what crosses the wire is already global — the
+  // scheduler must not re-derive it.
+  uint32 global_partition_id = 1;
+  // Position in the window operator's `window_expr()` list. Indexes window
+  // expressions, not partitions.
+  uint32 window_expr_index = 2;
+  // The PARTITION BY tuple that closed. Empty for a window with no
+  // PARTITION BY, which is the only shape the prefix rewrite plants today.
+  repeated datafusion_common.ScalarValue partition_key = 3;
+  // `Accumulator::state` for the closed group: one element for SUM / COUNT /
+  // MIN / MAX, two for AVG's (sum, count), one opaque Binary for
+  // sketch-backed aggregates like approx_distinct. Carried as ScalarValue
+  // rather than a numeric field so non-decomposable aggregates work
+  // unchanged.
+  repeated datafusion_common.ScalarValue state = 4;
 }
 
 // One report per `RuntimeStatsExec` in the executed plan.
--- a/ballista/core/src/execution_plans/prefix_merge.rs
+++ b/ballista/core/src/execution_plans/prefix_merge.rs
@@ -1,0 +1,1481 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Cross-partition state merge for windowed aggregates in an AQE range-shuffle
+//! pipeline.
+//!
+//! Range-shuffle produces `N` ordered disjoint partitions of a stream sorted
+//! by the window's ORDER BY key. Each executor task then runs a windowed
+//! aggregate over its slice, producing correct per-row values *within* the
+//! slice but not across slices — task `k`'s running SUM starts at zero, not
+//! at the sum of everything in partitions `[0..k)`.
+//!
+//! # Division of labor
+//!
+//! The prefix-merge splits cleanly across the scheduler/executor boundary:
+//!
+//! - **Scheduler (global step):** collects each upstream task's finalized
+//!   [`Accumulator::state`] via task-status transport, then computes the
+//!   prefix-merge — for each partition `k`, combining the individual states
+//!   of partitions `[0..k)` into a single already-merged state per window
+//!   expression. This is the step that requires global visibility across
+//!   tasks; only the scheduler has it.
+//! - **Executor (local step, this operator):** receives the *already-merged*
+//!   state for its partition in its constructor and folds it row-wise into
+//!   the window-aggregate columns. Aggregate-agnostic by construction — the
+//!   fold is the same [`Accumulator::merge_batch`]-shaped composition the
+//!   group-by `Partial → Final` protocol uses, so SUM adds, MIN/MAX take the
+//!   extreme, sketches (KLL, TDigest, HLL) merge as sketches, without this
+//!   operator knowing which aggregate is which.
+//!
+//! # Apply descriptors
+//!
+//! `try_new` takes a `Vec<`[`WindowApply`]`>` — one entry per output column
+//! that needs cross-partition correction, telling the operator *how* to
+//! rewrite that column. Two shapes:
+//!
+//! - [`WindowApply::Scalar`] — fast path. Combines each row's existing value
+//!   with a scheduler-provided scalar via [`ScalarOp`] (`Add`/`Min`/`Max` for
+//!   SUM/COUNT/MIN/MAX and `row_number`; `Overwrite` for `first_value` /
+//!   `last_value`). No `Accumulator` constructed.
+//! - [`WindowApply::Aggregate`] — fallback. Constructs a fresh `Accumulator`
+//!   seeded from the pre-merged state, feeds `args` per row, overwrites the
+//!   column with `evaluate()`. Fits AVG (without decomposition), sketch-backed
+//!   windows (APPROX_DISTINCT, APPROX_QUANTILE), and statistical aggregates.
+//!
+//! Non-corrected window functions don't appear in `applies` at all. `lead` /
+//! `lag` / `nth_value` are solved by halo rows in the shuffle layer.
+//! `rank` / `dense_rank` / `percent_rank` / `cume_dist` / `ntile` need a
+//! separate segment-tree-plus-broadcast design and are out of scope here.
+//!
+//! # Prefix-state input
+//!
+//! [`FinalizedPartitionState`] — one entry per input partition, holding
+//! the *pre-merged* [`Accumulator::state`] for each aggregate window
+//! expression (indexed by position in `BoundedWindowAggExec::window_expr()`).
+//! Only consumed by [`WindowApply::Aggregate`] entries;
+//! [`WindowApply::Scalar`] carries its own offsets inline. No PARTITION BY
+//! dimension is exposed: DataFusion publishes state per closed group, but the
+//! rewrite that plants this operator only fires on windows without a
+//! PARTITION BY, and the scheduler rejects any report carrying a key rather
+//! than flattening two groups together. A window that does have a PARTITION
+//! BY needs no help from this operator, since `BoundedWindowAggExec` asks for
+//! `KeyPartitioned` input and each partition's window is already independent.
+//! The scheduler bakes the state when it constructs the downstream stage
+//! after the upstream stage's tasks complete.
+//!
+//! **Status.** Both apply paths are implemented.
+//!
+//! - [`WindowApply::Aggregate`] builds a fresh `Accumulator` per partition,
+//!   seeds it via `merge_batch` from the offset state, and replays each row
+//!   through `update_batch` + `evaluate` to overwrite the output column.
+//! - [`WindowApply::Scalar`] applies the [`ScalarOp`] batch-at-a-time via
+//!   arrow kernels — `numeric::add` for `Add`, `cmp::lt_eq`/`gt_eq` + `zip`
+//!   for `Min`/`Max`, and a constant-fill for `Overwrite`.
+//!
+//! Inherits the DF-side getter's at-most-one-PARTITION-BY-group scoping
+//! (matches the AQE synthetic-PARTITION-BY pattern) — multi-key queries
+//! are handled by DataFusion's normal key-partitioned distribution and
+//! don't route through this operator.
+//!
+//! # Relation to DataFusion
+//!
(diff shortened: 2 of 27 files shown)


andygrove:
CI is red on two jobs here, but only one of them is yours.

`cargo doc` is real:

```
error: public documentation for `FinalizedPartitionState` links to private item `self`
error: public documentation for `PrefixMergeExec` links to private item `self`
```

The `[module-level docs][self]` links resolve to `mod prefix_merge`, which is private, so rustdoc rejects them under `-D warnings`. Making it `pub mod prefix_merge;` in `execution_plans/mod.rs` is the smallest fix and matches what `plan_algebra` and `sort_shuffle` already do. Dropping the two intra doc links works too if you'd rather keep the module private.

`test linux crates` is not this PR. It's `ballista-chaos::ha exhausted_retries_fail_the_job_and_leave_the_cluster_healthy::case_1_aqe_off` failing at cluster startup with executor registration `ConnectionRefused`, so an infrastructure flake. Should clear on a rerun.

Design feedback coming in a separate comment.

andygrove:
Design feedback, separate from the CI note above. The overall shape reads well to me, and the module header is genuinely good documentation. Splitting the global prefix merge onto the scheduler and leaving a row wise apply on the executor is the right decomposition, and the APPROX_DISTINCT case makes a convincing argument for why the `Aggregate` path has to exist. A few things I'd want settled before this grows more code on top of it.

**The per row accumulator replay looks like a performance trap.** `AggregateApply::apply` calls `update_batch` on a one row slice and then `evaluate()` once per row. For SUM that's just wasteful, but the motivating cases are sketches, and that's where it gets expensive. `evaluate()` on an HLL scans every register to produce a cardinality estimate, and on TDigest or KLL it runs a quantile computation. Doing that once per row turns a linear pass into something quite a bit worse, and it re derives work the upstream BWAG already did. The APPROX_DISTINCT test proves correctness on three rows, which is exactly the size that won't surface this. Could you run it over a realistic partition before we commit to the shape? If the numbers are bad there may be a middle path where the upstream emits partial state columns and the correction stays batch at a time.

**Serde is deferred, and it's the hard part.** #2255 landed its proto message and codec arm in the same PR, and I'd like this one to end up there too, since `PrefixMergeExec` can't reach an executor without them. No objection to a scaffold that defers it, I just want to flag that the remaining work isn't mechanical. `Arc<AggregateUDF>`, `Vec<Arc<dyn PhysicalExpr>>`, and `Vec<ScalarValue>` sketch state all have to cross the wire.

The related question is the transport the design picks. The description says upstream state reaches the scheduler over task status. That's a hot and frequent message, and HLL, KLL or TDigest state per task per window expression is not small. Since you describe the division of labor as fixed at design time, I'd rather pressure test that choice now than after the scheduler side is built on top of it.

**Partition index coupling has no guard.** Both `per_partition_state[k]` and `Scalar.offset[k]` are keyed by partition index, but the operator declares `UnspecifiedDistribution`, no required input ordering, and `maintains_input_order: true`, and `with_new_children` only re validates counts. So any rule that repartitions the input to the same partition count would silently attach each partition's offsets to the wrong rows. Wrong answers, no error. In practice the scheduler hands over a finished plan so it may never happen, but this is the "correct on one node, silently wrong once split across stages" shape that [user-personas.md](https://github.com/apache/datafusion-ballista/blob/main/docs/source/contributors-guide/user-personas.md) calls out for Persona 1, and I'd want at least a loud invariant comment on it.

**Smaller things:**

- `ScalarOp::Overwrite` is documented as fitting `first_value` and `last_value`. `first_value` I follow. For the cumulative frame `last_value` is just the current row's value and needs no correction at all, so overwriting every row with a single scalar would be wrong. Which frame is that aimed at?
- No metrics. `PrefixMergeExec` doesn't implement `metrics()` and `ApplyStream` has no `BaselineMetrics`. Given the first point above this is the operator you'd most want timings from, and Spark shaped users lean on per operator timings for skew debugging.
- Type drift is only caught by accident. If `numeric::add` promotes, or `evaluate()` returns a different type than the column it replaces, `RecordBatch::try_new` fails with an opaque arrow error rather than something naming the offending `applies[i]`.
- `ScalarOp` and `WindowApply` are public enums that you say will grow. Marking both `#[non_exhaustive]` now costs nothing and saves a breaking change on the first new variant. Similarly, `FinalizedPartitionState` as a transparent `pub type` alias means swapping it for the real DataFusion type once apache/datafusion#24007 lands is a silent public API change. A newtype now would keep that swap internal.
- Minor housekeeping, the PR description has the "Generated with Claude Code" footer, and `CLAUDE.md` in the repo asks us to keep that out of PRs.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1259 @@
...
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        let mut new_values: Vec<ScalarValue> = Vec::with_capacity(num_rows);
+        for i in 0..num_rows {
+            let row_args: Vec<ArrayRef> =
+                arg_arrays.iter().map(|a| a.slice(i, 1)).collect();
+            self.accumulator.update_batch(&row_args)?;
+            new_values.push(self.accumulator.evaluate()?);
+        }
```
avantgardnerio: > it re derives work the upstream BWAG already did

Yes, and that is exactly what the second pass is doing: the upstream window already advanced an accumulator through every row, and this walks it again from the carry-in.

The reason it is not avoidable today is where the state is exposed. apache/datafusion#24035 publishes accumulator state at PARTITION BY group close, one state per group, so there is no per row state to carry into a batch at a time correction.

> there may be a middle path where the upstream emits partial state columns

I think that is worth pursuing, but it needs more than the API. DataFusion's dense HLL is 16 KiB per sketch ([`approx_distinct.rs#L257`](https://github.com/apache/datafusion/blob/55.0.0-rc2/datafusion/functions-aggregate/src/approx_distinct.rs#L257)), so materializing per row state for a sketch is 16 KiB times row count as a column. A workable version probably needs a cheaper per row representation than raw accumulator state, which makes it a real piece of design rather than a switch to flip.

Worth being precise about the current state: this PR does not select the scalar path yet. `WindowApply::Scalar` exists on the operator and applies batch at a time via arrow kernels, but the rule currently builds an `Aggregate` apply for every window expression, including SUM. Selecting `Scalar` where it applies will land on this PR before merge, and for SUM, COUNT, MIN and MAX it removes this cost rather than halving it.

> Could you run it over a realistic partition

Yes. A three row test does not surface any of this. The trade we are making is 2x local work divided across cores (theoretically 16x on 32, to be measured shortly). Benchmarks will land in this thread before anything else is built on the shape.

ballista/core/proto/ballista.proto
```diff
@@ -656,6 +723,48 @@ message SuccessfulTask {
...
+  //
+  // TODO: watch the size of this. Task completion is a hot, frequent
+  // message, and sketch-backed aggregates make the payload unbounded in a
+  // way row counts and quantile sketches are not — an HLL or KLL state is
+  // kilobytes per window expression per partition, and a task covering a
+  // wide partition slice carries one of each. If it stops being small,
+  // write the state as a sidecar next to the shuffle files instead, the way
+  // sort-shuffle already writes `<data>.arrow.index` beside its data
+  // (`sort_shuffle::get_index_path`), and send only a reference here. That
+  // keeps the completion message fixed-size regardless of aggregate.
```
avantgardnerio, line 798: > HLL, KLL or TDigest state per task per window expression is not small

Agreed, and it is worth sizing. DataFusion's dense HLL is 16 KiB per sketch ([`approx_distinct.rs#L257`](https://github.com/apache/datafusion/blob/55.0.0-rc2/datafusion/functions-aggregate/src/approx_distinct.rs#L257)). A task carries one state per (partition in its slice, aggregate window expression), so `max_partitions_per_task` multiplies it: a 32 partition slice with a single sketch aggregate puts 512 KiB on that task's completion message.

Two things make me think the current shape is defensible as a starting point rather than a commitment. `RuntimeStatsExec` already ships quantile sketches over this exact path, so this adds a second payload of a class task status already carries rather than introducing one. And the escape hatch is cheap: the state can be written as a sidecar beside the shuffle files, the way sort shuffle already writes `<data>.arrow.index` next to its data, with only a reference on the message. That keeps completion fixed size regardless of aggregate, and it is a change to this one field rather than to the design around it.

What I have not done is measure it. The e2e only exercises a Float64 SUM, where the payload is a handful of bytes. If you would rather see a real number for `approx_distinct` before this merges, I can add a test that reports the encoded size and post it here.

ballista/scheduler/src/state/task_builder.rs
```diff
@@ -87,41 +86,20 @@ fn restrict(
...
-            .collect::<datafusion::common::Result<_>>()?;
-        return Ok(Arc::new(RangeFilterExec::try_new_resolved(
-            new_child,
-            rf.routing_expr().clone(),
-            rf.halo_lo().clone(),
-            rf.halo_hi().clone(),
-            sliced_bounds,
-        )?));
+        return sliceable.slice_to_partitions(new_child, partitions);
     }
```
avantgardnerio, line 103: > any rule that repartitions the input to the same partition count would silently attach each partition's offsets to the wrong rows

This was real and it fired on the first end to end run, though not by the route you predicted. Task specialization already restricts a stage plan to one task's partition slice, so `PrefixMergeExec` was rebuilt over a narrower input while still holding state keyed by global partition index. It failed loudly only because `try_new` validates the length: `per_partition_state.len() 4 does not match input partition count 1`. Without that check it is exactly the silent misattachment you describe, since a task's `execute(k)` numbers its own slice from zero.

Fixed by slicing state and `Scalar.offset` parallel to the input restriction, following the precedent `RangeFilterExec` already set for its bounds. It is behind a `PartitionSliceable` trait so each operator implements the slicing next to the fields being sliced, rather than the task builder knowing their internals: add a partition indexed field and the code that has to slice it is in the same file.

The e2e runs with `max_partitions_per_task = 2`, so tasks carry a genuine multi partition slice rather than the degenerate one partition case.

ballista/core/src/serde/mod.rs
```diff
@@ -704,6 +891,20 @@ impl PhysicalExtensionCodec for BallistaPhysicalExtensionCodec {
...
+                        "PrefixMergeExec expects exactly 1 input, got {}",
+                        inputs.len()
+                    )));
+                };
+                let schema = input.schema();
+                Ok(Arc::new(PrefixMergeExec::try_new_resolved(
+                    input.clone(),
+                    decode_window_applies(&node.applies, ctx, schema.as_ref(), self)?,
+                    decode_prefix_state(&node.per_partition_state)?,
+                )?))
```
avantgardnerio, line 910: > `Arc<AggregateUDF>`, `Vec<Arc<dyn PhysicalExpr>>`, and `Vec<ScalarValue>` sketch state all have to cross the wire

Done on this PR. The UDAF crosses by name and is resolved from the executor's `FunctionRegistry` on decode, args as `PhysicalExprNode`, and state as `ScalarValue` so sketch state works unchanged. A round trip test covers both `WindowApply` shapes, the registry lookup, and the distinction between a slot with no state and a slot with empty state.

Encoding refuses while the prefix state is unresolved, matching `RangeFilterExec`'s refusal on unresolved bounds. An executor has no route to that state, so a plan reaching the wire without it could only emit partition-local aggregates.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1341 @@
...
+    /// `output := max(row_value, offset)`. Fits MAX.
+    Max,
+    /// `output := offset`. Ignores `row_value`. Fits `first_value` over an
+    /// ever-expanding frame: every row's answer is the same global first
+    /// value, so the scheduler picks it once and every row gets a copy.
+    ///
+    /// Not `last_value`. Over an ever-expanding frame that is the current
+    /// row's own value and needs no correction at all, so overwriting every
+    /// row with one scalar would be wrong.
+    Overwrite,
```
avantgardnerio, line 164: > Which frame is that aimed at?

None, you are right. Over an ever-expanding frame `last_value` is the current row's own value and needs no correction, so overwriting every row with a single scalar would be wrong. The doc now claims `first_value` only and says why `last_value` is excluded, so the next reader does not have to re-derive it.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1341 @@
...
+///
+/// Two shapes are covered by construction; anything else is out of scope
+/// (`lead`/`lag`/`nth_value` are solved by halo rows in the shuffle layer, and
+/// ranking-family functions like `rank`/`percent_rank`/`ntile` want a separate
+/// segment-tree-plus-broadcast infrastructure).
+/// `#[non_exhaustive]`: the two shapes here cover what the prefix rewrite
+/// plants today, and a third (a segment-tree broadcast for the ranking
+/// family) is anticipated.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
```
avantgardnerio: > Marking both `#[non_exhaustive]` now costs nothing and saves a breaking change on the first new variant

Done, on both. Construction is unaffected so the rewrite rule still builds an `Aggregate` apply directly; it is exhaustive matching that now needs a wildcard, which is the part that would have broken.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1341 @@
...
+/// * `columns` - columns with `output_column` already replaced
+/// * `apply_index` - position in the exec's `applies` list, for the message
+/// * `output_column` - the column that was replaced, for the message
+fn rebuild_batch(
+    schema: SchemaRef,
+    columns: Vec<ArrayRef>,
+    apply_index: usize,
+    output_column: usize,
+) -> Result<RecordBatch> {
+    let replaced = columns[output_column].data_type().clone();
```
avantgardnerio, line 869: > `RecordBatch::try_new` fails with an opaque arrow error rather than something naming the offending `applies[i]`

Fixed. Both apply paths now rebuild through this helper, which reports `applies[3] produced Int64 for column 1, which the schema declares as Float64`. `AggregateApply` already carried `apply_index` for exactly this and was not using it at the failure site. There is a test on the message, since it is otherwise an untested branch.

avantgardnerio:
Heads up that this is functional now rather than a scaffold. It runs end to end and `ballista/client/tests/prefix_window.rs` asserts a distributed running sum against the serial answer through a real cluster. I have replied to each of your points inline on the relevant lines so they can be resolved individually.

Leaving it in draft: benchmarks for the per-row replay are not in yet, and the scalar path selection lands before merge. Both are tracked in the threads.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1433 @@
...
+    rows_corrected: Count,
+}
+
+struct ApplyStream {
+    input: SendableRecordBatchStream,
+    appliers: Vec<PreparedApply>,
+    schema: SchemaRef,
+    baseline: BaselineMetrics,
+    path_metrics: PathMetrics,
+}
```
avantgardnerio, line 808: > `PrefixMergeExec` doesn't implement `metrics()` and `ApplyStream` has no `BaselineMetrics`

Added. `BaselineMetrics` for `elapsed_compute` and `output_rows`, per partition since `execute(partition)` builds a stream each, plus a `rows_corrected` counter.

The two apply paths are timed separately rather than together. `WindowApply::Scalar` is an arrow kernel over a whole batch and `WindowApply::Aggregate` seeds an accumulator and replays every row, so a single number would cover two quite different costs for a query that mixes them. It also makes the replay you flagged something the operator reports in production rather than something only a benchmark can see.

I will post these numbers along with the benchmarking results.

ballista/core/src/execution_plans/prefix_merge.rs
```diff
@@ -0,0 +1,1481 @@
...
+/// readable nor searchable, and it spells "no state for this window
+/// expression" two ways — a missing index and a `None` — which every caller
+/// then has to handle. [`Self::slot`] collapses both into one answer. Any
+/// later change to the representation also stays internal rather than
+/// breaking whoever wrote the concrete type.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct FinalizedPartitionState {
+    /// Indexed by position in the upstream operator's `window_expr()` list.
+    per_window_expr: Vec<Option<Vec<ScalarValue>>>,
+}
```
avantgardnerio, line 113: > A newtype now would keep that swap internal

Done, though not for the reason you gave, and the difference seems worth recording.

apache/datafusion#24007 did not land. What shipped in DataFusion 55 is apache/datafusion#24035, a `WindowStateObserver` callback rather than a getter, and it publishes no type equivalent to this one. So there is no upstream type to swap to and the break you were guarding against cannot happen.

The newtype earns its place anyway. `Vec<Option<Vec<ScalarValue>>>` appears in this operator's public signatures, where it is neither readable nor searchable, and it spells "no state for this window expression" two ways: a missing index, and a `None` at a present index. Every caller had to handle both. `slot(window_expr_index)` now collapses them into one answer, and a later change to the representation stays internal.

While looking at this I removed two comments claiming the DataFusion side guarantees at most one PARTITION BY group per partition. It does not, and never did — the callback is keyed by group, so that invariant is ours. The scheduler now enforces it by rejecting any report that carries a key rather than flattening two groups together. Worth noting a window that does have a PARTITION BY needs nothing from this operator: `BoundedWindowAggExec` asks for `KeyPartitioned` input, so each partition's window is already independent and there is no serial bottleneck to remove.

avantgardnerio:
@phillipleblanc FYI

ballista/core/src/execution_plans/prefix_merge.rs
avantgardnerio: Benchmarks, as promised in this thread.

h2o Q7 at 1e9 rows, 2 executors, 16 partitions, release build, local NVMe:

| stage | rule off | rule on |
|---|---|---|
| sort (plus ORRE when on) | 40.82s | 54.56s |
| window | 122.71s, 1 task | 12.56s |
| `PrefixMergeExec` correction | n/a | 8.25s |
| wall | 179.5s | 91.3s |

The correction costs less than the window it corrects: 8.25s against 12.56s.

Both passes also scale with cores. Doubling from 8 to 16 partitions at fixed data: window 21.86s to 12.56s (1.74x), correction 14.84s to 8.25s (1.80x), while the serial window is flat at 118.53s and 122.71s because it is one task either way. End to end that moves 1.34x to 1.97x, so the advantage grows with cluster width rather than being a fixed constant.

One thing I said here needs correcting:

> Selecting `Scalar` where it applies will land on this PR before merge

That no longer follows from the numbers. Scalar selection is worth doing and still removes the replay for SUM, COUNT, MIN and MAX, but it is an optimization on top of a win rather than the thing that makes this a win. I would rather land it separately than hold this PR for it.

Three caveats. These are single runs. SUM is the cheapest window aggregate there is, so this understates the sketch cases the design exists for, where each row costs a register scan in `evaluate()`. And both executors share one disk, which penalises the arm doing more shuffle IO, namely this one.

ballista/core/proto/ballista.proto
avantgardnerio: I measured it and the above statement is correct.

milenkovicm:
Thanks @avantgardnerio will have a look tomorrow and over the weekend

phillipleblanc (approved):
Thanks for all of the detailed tests - between this and Andy's review I don't think there is anything else to address.

PR #2215, opened 2026-08-01
fix(scheduler): preserve stages on transition errors

### LINKED ISSUE #2214: StaticExecutionGraph drops stages when plan transformations fail
# StaticExecutionGraph drops stages when plan transformations fail

**Describe the bug**

Several `StaticExecutionGraph` state transitions remove a stage from `self.stages` before performing a fallible physical-plan transformation.

The affected transitions are:

- `resolve_stage`
- `rollback_running_stage`
- `rollback_resolved_stage`

For example, `resolve_stage` currently follows this pattern:

```rust
if let Some(ExecutionStage::UnResolved(stage)) =
    self.stages.remove(&stage_id)
{
    self.stages.insert(
        stage_id,
        ExecutionStage::Resolved(
            stage.to_resolved(self.session_config.options())?,
        ),
    );
}
```

The stage is removed before `to_resolved` is evaluated. If the transformation returns an error, `?` returns from the transition before `insert` is called, so the original stage is dropped and the stage ID is left absent from the execution graph.

The two rollback transitions have the same ordering around `to_unresolved(...)`.

These transformations can fail while replacing shuffle nodes, rebuilding physical-plan nodes, constructing shuffle readers, or applying physical optimizer rules.

The error is propagated through the scheduler, but callers operate directly on the active execution graph and do not restore a previous snapshot. A failed transition can therefore mutate the graph even though no replacement stage was successfully produced.

**To Reproduce**

A deterministic unit-level reproduction can use a test execution-plan node whose `with_new_children` implementation returns an error when the shuffle child is rewritten.

For each affected transition:

1. Construct a `StaticExecutionGraph` containing a stage in the required initial state.
2. Record the stage count and the stage's state.
3. Invoke the transition with a plan transformation that returns an error.
4. Verify that the transition returns `Err`.
5. Inspect `graph.stages()` for the original stage ID.

With the current implementation, the stage ID is no longer present because the stage was removed before the transformation failed.

The same outcome follows directly from the current control flow: `remove` completes before the fallible transformation is evaluated, while `insert` is skipped by the early return.

**Expected behavior**

Stage transitions should be failure-atomic.

If a physical-plan transformation fails:

- the transition should return the original error;
- the original stage should remain in the execution graph;
- its state, plan, task metadata, and attempt information should remain as they were at the start of the transition; and
- no replacement stage should be committed.

A successful transformation should retain the existing transition behavior.

**Additional context**

This issue was found by inspecting the transition control flow. The exact user-visible outcome depends on which physical-plan transformation fails, so this report does not claim a specific production failure mode or occurrence frequency.

The proposed scope is limited to the three `StaticExecutionGraph` transitions in `ballista/scheduler/src/state/execution_graph.rs`:

- compute the replacement stage before committing it to `self.stages`;
- commit the replacement only after the fallible transformation succeeds; and
- add regression coverage for the error path of each transition.

This should not require changes to public APIs, scheduling policy, retry policy, stage-attempt behavior, or stage dependency semantics. It also does not propose making the surrounding task-update or executor-loss operations fully transactional.

QuakeWang:

Closes #2214 

# Rationale for this change

`StaticExecutionGraph` removed a stage before running fallible plan transformations. If a transformation failed, the error returned before the stage was reinserted, leaving the execution graph without the original stage.

Build replacement stages before updating the stage map, so failed resolve and rollback transitions preserve the original stage. Add regression tests for all three affected transitions.

No API changes. On a failed running-stage rollback, the original stage and its running-task tracking are preserved, but the error still propagates before a `CancelTasks` event can be produced.

# Testing

- `ci/scripts/rust_fmt.sh`
- Scheduler all-targets/all-features Clippy
- Scheduler library tests

### DIFF
--- a/ballista/scheduler/src/state/execution_graph.rs
+++ b/ballista/scheduler/src/state/execution_graph.rs
@@ -1295,13 +1295,10 @@ impl ExecutionGraph for StaticExecutionGraph {
 
     /// Convert unresolved stage to be resolved
     fn resolve_stage(&mut self, stage_id: usize) -> Result<bool> {
-        if let Some(ExecutionStage::UnResolved(stage)) = self.stages.remove(&stage_id) {
-            self.stages.insert(
-                stage_id,
-                ExecutionStage::Resolved(
-                    stage.to_resolved(self.session_config.options())?,
-                ),
-            );
+        if let Some(ExecutionStage::UnResolved(stage)) = self.stages.get(&stage_id) {
+            let resolved_stage = stage.to_resolved(self.session_config.options())?;
+            self.stages
+                .insert(stage_id, ExecutionStage::Resolved(resolved_stage));
             Ok(true)
         } else {
             warn!(
@@ -1353,7 +1350,7 @@ impl ExecutionGraph for StaticExecutionGraph {
         stage_id: usize,
         failure_reasons: HashSet<String>,
     ) -> Result<Vec<RunningTaskInfo>> {
-        if let Some(ExecutionStage::Running(stage)) = self.stages.remove(&stage_id) {
+        if let Some(ExecutionStage::Running(stage)) = self.stages.get(&stage_id) {
             let running_tasks = stage
                 .running_tasks()
                 .into_iter()
@@ -1364,10 +1361,9 @@ impl ExecutionGraph for StaticExecutionGraph {
                     executor_id,
                 })
                 .collect();
-            self.stages.insert(
-                stage_id,
-                ExecutionStage::UnResolved(stage.to_unresolved(failure_reasons)?),
-            );
+            let unresolved_stage = stage.to_unresolved(failure_reasons)?;
+            self.stages
+                .insert(stage_id, ExecutionStage::UnResolved(unresolved_stage));
             Ok(running_tasks)
         } else {
             warn!(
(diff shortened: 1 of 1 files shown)


ballista/scheduler/src/state/execution_graph.rs
```diff
@@ -1364,10 +1361,9 @@ impl ExecutionGraph for StaticExecutionGraph {
                     executor_id,
                 })
                 .collect();
-            self.stages.insert(
-                stage_id,
-                ExecutionStage::UnResolved(stage.to_unresolved(failure_reasons)?),
-            );
+            let unresolved_stage = stage.to_unresolved(failure_reasons)?;
```
sandugood, line 1364: When this errors, the `running_tasks` was already collected, but it is now discarded. Better than a vanished stage, but it's a distinct bad state. Maybe add a line to "Are there any user facing changes"?

sandugood (commented):
Nicely done, thanks @QuakeWang. Added a comment

milenkovicm (approved):
thanks @QuakeWang this make sense to me

ballista/scheduler/src/state/execution_graph.rs
QuakeWang: Good point. The running-task list was also discarded on this error path before this PR, but preserving the stage makes the resulting behavior worth stating explicitly: the stage remains `Running`, its task tracking is retained, and the error propagates before a `CancelTasks` event can be emitted. I have clarified this in the user-facing changes section. Changing cancellation or recovery policy would be outside the failure-atomic transition scope of this PR.

PR #2216, opened 2026-08-01
perf(kll): batch-oriented ingest KLL

avantgardnerio:
## Summary

Performance improvements for KLL to make it viable for collecting RuntimeStats.

- `absorb(&[T])` (iterator) + `absorb_slice(&[T])` — batch-oriented ingest, amortizes `compact_all`.
- Per-level sortedness invariant — `compact_level` skips re-sort when the level was left sorted by an earlier promotion; promotions merge-extend in linear time when the destination is sorted.
- `absorb_sorted_slice(&[T])` — for callers that can guarantee sorted input, keeps level 0 permanently sorted and skips *every* compaction sort.
- Statistical test coverage against the DataSketches empirical bound (uniform, clustered, heavy-ties).

The actual `RuntimeStatsExec` integration, wire format for KLL sketches across the scheduler boundary, and the ORRE plan-shape follow-up are all out of scope here — separate PRs.

## The two arms

**ORRE arm (ordered input, requires a follow-up plan-shape move):**

```
Scan ──▶ RuntimeStats ──▶ Sort ──▶  ORRE ──▶ ShuffleWrite
```

`SortExec` is already required by ORRE. Moving it above `RuntimeStats` costs nothing plan-wise — the same rows get sorted the same way. 

**URRE arm (unordered input, ships as-is with this PR's shape):**

```
Scan ──▶ RuntimeStats ──▶ BufferExec(Dam) ──▶ URRE ──▶ ShuffleWrite
```

## Why not always use TDigest?

TDigest can only sketch a single non-null `f64` column. KLL is generic over `T: Ord + Clone`, and composes with `arrow::row::OwnedRow` to cover the full ORDER BY grammar for partition-boundary picking: multi-column keys, nullable, per-column ASC/DESC, configurable NULLS FIRST/LAST. That grammar is what parallel window (and later merge-join, order-by execution) actually needs from the boundaries.

## Numbers (uniform f64, 1M rows, `k=800` for KLL, `TDIGEST_MAX_SIZE=100`)

Baseline = TDigest sorted (`merge_sorted_f64`), the fastest arm. Higher × is worse:

| variant | time | throughput | ×baseline |
|---|---:|---:|---:|
| TDigest sorted (baseline) | 2.19 ms | 457.6 M/s | 1.00× |
| KLL<OrderedFloat<f64>> `absorb_sorted_slice` (this PR + planner follow-up, ORRE arm) | 7.22 ms | 138.6 M/s | 3.30× |
| TDigest unsorted (`merge_unsorted_f64`) | 13.89 ms | 72.0 M/s | 6.36× |
| KLL<OrderedFloat<f64>> `absorb_slice` (this PR, URRE arm) | 24.78 ms | 40.4 M/s | 11.34× |
| KLL<OrderedFloat<f64>> `insert` (per-row) | 47.65 ms | 21.0 M/s | 21.80× |
| KLL<OwnedRow> `absorb` (multi-column path, this PR) | 80.22 ms | 12.5 M/s | 36.71× |
| KLL<OwnedRow> `insert` (per-row) | 92.78 ms | 10.8 M/s | 42.45× |

Accuracy parity between the sketches — both hold worst-case rank error under 0.25% on the same 1M stream:

| sketch                | worst rank err (9 deciles) |
|-----------------------|---------------------------:|
| TDigest max_size=100  |                     0.0021 |
| KLL k=800             |                     0.0016 |

Rerun via `KLL_PARITY_CHECK=1 cargo bench --bench quantile_sketch`.

The `OwnedRow` path is the real regression — the compaction cost for heap-allocated rows is dominated by cache misses on the `OwnedRow` storage, not sort-algorithm efficiency. Closing that gap is deferred-optimization item (2) in the module doc (batch-oriented level-0-borrowed-row ingest), noted as a TODO on the interim iterator API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,6 +1038,7 @@ dependencies = [
  "env_logger",
  "futures",
  "mimalloc",
+ "ordered-float 4.6.0",
  "rand 0.10.2",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1033,6 +1033,7 @@ dependencies = [
  "clap 4.6.3",
  "criterion",
  "datafusion",
+ "datafusion-functions-aggregate-common",
  "datafusion-proto",
  "env_logger",
  "futures",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ insta = "1.47"
 itertools = "0.15"
 mimalloc = { version = "0.1" }
 object_store = "0.13.2"
+ordered-float = "4"
 prost = "0.14"
 prost-types = "0.14"
 rstest = { version = "0.26" }
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -291,6 +291,13 @@ impl<T: Ord + Clone> KllSketch<T> {
     /// Output is bit-identical to `for x in xs { self.insert(x) }`: the
     /// compaction trigger points, coin flips, and min/max tracking are all
     /// preserved.
+    ///
+    /// TODO: retire once **Deferred optimizations** item (2) lands. The
+    /// borrowed-row ingest path there takes an arrow `Rows` directly and
+    /// holds `Row<'_>` at level 0 — a specialized API, not a generalization
+    /// of this one. This function exists as a modest speedup for the
+    /// interim OwnedRow path (~14% over `insert`-in-loop on 1M rows) and
+    /// has no other production caller.
     pub fn absorb<I: IntoIterator<Item = T>>(&mut self, xs: I) {
         // Cache level 0's capacity; recompute whenever compaction may have
         // grown the stack (which shrinks per-level capacities).
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -702,55 +702,114 @@ mod tests {
         2.296 / (k as f64).powf(0.9723)
     }
 
-    /// One trial: insert a shuffled `1..=n` stream, probe rank at 9
-    /// interior quantiles, return the max normalized rank error observed.
-    fn worst_rank_error_uniform(k: usize, n: u32, seed: u64) -> f64 {
-        use rand::seq::SliceRandom;
-        let mut shuffle_rng = StdRng::seed_from_u64(seed);
-        // Sketch RNG derived from a different sub-seed so the shuffle and
-        // the compaction coin flips aren't drawn from correlated streams.
-        let mut sketch: KllSketch<u32> =
-            KllSketch::with_seed(k, seed.wrapping_add(0x9E37_79B9));
-        let mut stream: Vec<u32> = (1..=n).collect();
-        stream.shuffle(&mut shuffle_rng);
-        for x in &stream {
+    /// Insert `stream` into a KLL sketch at capacity `k`, probe rank at 9
+    /// interior quantiles picked from a sorted copy of the stream, and
+    /// return the worst normalized rank error observed.
+    fn worst_rank_error(stream: &[u32], k: usize, sketch_seed: u64) -> f64 {
+        let mut sketch: KllSketch<u32> = KllSketch::with_seed(k, sketch_seed);
+        for x in stream {
             sketch.insert(*x);
         }
+        let mut sorted = stream.to_vec();
+        sorted.sort_unstable();
+        let n = stream.len();
         (1..10)
             .map(|i| {
-                let q = i as f64 / 10.0;
-                let probe = ((q * n as f64) as u32).max(1);
-                // For a shuffled stream of `1..=n`, the true count strictly
-                // less than `probe` is `probe - 1`.
-                let true_rank = (probe as f64 - 1.0) / n as f64;
-                let est_rank = sketch.rank(&probe) as f64 / n as f64;
-                (est_rank - true_rank).abs()
+                let idx = (i as f64 / 10.0 * n as f64) as usize;
+                let probe = sorted[idx.min(n - 1)];
+                // True rank: count of items strictly less than probe.
+                // partition_point on the sorted stream gives it in O(log n).
+                let true_rank_frac =
+                    sorted.partition_point(|x| *x < probe) as f64 / n as f64;
+                let est_rank_frac = sketch.rank(&probe) as f64 / n as f64;
+                (est_rank_frac - true_rank_frac).abs()
             })
             .fold(0.0_f64, f64::max)
     }
 
-    #[test]
-    fn rank_error_meets_datasketches_bound_across_seeds() {
-        // The DataSketches epsilon is empirically fit to the P99 max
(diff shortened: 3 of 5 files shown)


avantgardnerio:
@phillipleblanc FYI

phillipleblanc:
I wonder if it would make sense to support both and let the user choose what they want to use?

i.e. something like `ballista.planner.range_partitioning.sketch = auto | tdigest | kll`
  - `auto` - recommended default. Use TDigest for simple numeric data when it is faster; use KLL when multiple columns, nulls, or other types require it.
  - `tdigest` - force the existing implementation. Reject queries it cannot support.
  - `kll` - force the new implementation for testing or predictable behavior.

I wouldn't add that to this PR, but for when we eventually get to wiring itup.

Other than that, I think this makes sense - some small comments:

- **Compare equal accuracy levels**. The PR compares `KLL` and `TDigest` using inconsistent accuracy calculations. It is like comparing two compressors using different image-quality settings. First choose settings that produce roughly equal accuracy, then compare speed and memory.
- **Give both algorithms equally sorted input**. The fastest KLL test receives already-sorted data, while TDigest receives unsorted data and must sort it itself. That is not an apples-to-apples race. Add a TDigest test that receives the same sorted data before claiming KLL is 1.9x faster.

avantgardnerio:
> Give both algorithms equally sorted input.

Sorry, I didn't realize TDigest also benefited from sorted input. It was not my intent to make inaccurate claims or skew benchmarks. I'll address that before I undraft this PR.

Of course I'll address the rest of the feedback, but I just wanted to apologize first.

phillipleblanc:
> > Give both algorithms equally sorted input.
> 
> Sorry, I didn't realize TDigest also benefited from sorted input. It was not my intent to make inaccurate claims or skew benchmarks. I'll address that before I undraft this PR.
> 
> Of course I'll address the rest of the feedback, but I just wanted to apologize first.

No need to apologize! Nuance is hard to convey over text, its more about minimizing variables when doing comparison benchmarks.

andygrove:
This is still a draft so please feel free to disregard any or all of this, I know you're mid iteration. I had it open anyway and found one thing I thought was worth passing along early rather than sitting on.

First the good parts. The statistical tests against the DataSketches empirical rank error bound are a real upgrade over what was in `kll.rs` before, and they're all seeded so they won't flake. The deferred optimizations section with the struck through items and measured numbers next to each is unusually easy to follow. And reporting the 5.7x `OwnedRow` regression plainly instead of burying it is the right call.

## CI

Three jobs are red. `test linux ballista` is the `ballista-chaos` cluster startup flake and has nothing to do with this PR. The other two are real:

```
test kll::tests::absorb_sorted_slice_debug_asserts_on_unsorted - should panic ... FAILED
```

The cause is that `[profile.ci]` in the root `Cargo.toml` sets `debug-assertions = false`, and CI runs `cargo test --profile ci`. So `debug_assert!` never fires under CI and a `#[should_panic]` test on one can't pass.

## The thing I'd actually push on

That failure isn't just a broken test, I think it's the API telling on itself. `absorb_sorted_slice` is a safe `pub fn` whose own docs say that violating the precondition silently corrupts the sketch, and the only guard is a `debug_assert!` that the project's test profile compiles out. So the protection doesn't exist in the configuration we test and ship.

What I'd suggest is verifying sortedness unconditionally and falling back to `absorb_slice` when the input isn't sorted. Checking costs O(n) comparisons against the O(n log n) sort you're skipping, so on the 1M row f64 case that's about a million `<=` comparisons against a 7.3 ms budget. Most of the win survives, the footgun goes away entirely, and the `should_panic` test turns into a "falls back correctly" test that passes under any profile.

I care about this more than I normally would for an unchecked precondition because of who the caller is. The invariant is a plan shape one, that a `SortExec` sits above `RuntimeStatsExec`. If some later optimizer change moves or drops that sort, the sketch quietly returns wrong quantiles, and those become wrong partition boundaries. Silent wrong answers from a plan shape assumption is a bad failure mode to design in.

## The sortedness invariant itself looks right

I went through this carefully since it's the risky part of the change. Every `levels[...]` mutation site on main is covered, `merge`, `insert`, and the three inside `compact_level`. `compact_all` only compacts levels at or over capacity and consumes no RNG when it no ops, so the "bit identical to insert in a loop" claim for `absorb` holds up. The merge branch in `compact_level` maintains the flag correctly, and the conservative `sorted[h] = false` in cross sketch `merge` is safe.

Two things I'd tighten:

- `absorb_sorted_slice` degrades silently when `sorted[0]` is false on entry. If you mix it with `insert` or `absorb_slice` on the same sketch, `merge_sorted` interleaves into an unsorted level 0. It stays correct, since the flag stays false and compaction sorts, and it self heals after the first compaction. But the doc's claim that level 0 stays permanently sorted only holds for a sketch used exclusively through this one method. A sort and set at the top would make the doc true.
- This PR changes the output of the existing `insert` path. `compact_level` now merge extends where it used to extend then `sort_unstable`. With ties a different set of duplicate copies survives. Values and weights are unchanged so quantiles are identical for any `T` where equal items are interchangeable, which covers every current use, but the PR reads as purely additive and this part isn't. Worth a line in the description.

## Smaller

- `absorb_sorted_slice` allocates twice per chunk, once for `to_vec()` and once for `merge_sorted`'s output, and drops `existing`. At level 0 chunk sizes across 1M rows that's a lot of small allocations. Already fast at 7.3 ms so not urgent, but an in place merge would probably widen the gap further.
- `datafusion-functions-aggregate-common` went into `[dependencies]` but is only used by the bench. Looks like it belongs in `[dev-dependencies]`.
- `ordered-float = "4"` is a raw version rather than `{ workspace = true }`. It isn't in `[workspace.dependencies]` yet, and we manage versions there.

No concerns on the persona side. The `absorb*` methods are additive on an existing public type and nothing is wired into `RuntimeStatsExec` yet.

avantgardnerio:
@phillipleblanc please re-read the PR body, I've updated it with revised framing where it no longer tries to justify a "swap". TDigest is strictly faster in apples-to-apples scenarios. The only remaining argument (which I am not making) for a "swap" would be code-complexity vs the ~1% time I saw in h2o q8 that was spent sketching at all. I'm fine with 2 modes for `RuntimeStatsExec`, but I would suggest perhaps it just picks the best one automatically? I'm not sure what it would do if you chose TDigest on a nullable or non-f64 column?

> Compare equal accuracy levels

Claude did some math with the first iteration of this PR to guess at what that was (100/800), but this time around we ran some empirical tests. It turns out those numbers are pretty close. Perhaps I misunderstood you, but I think the test is fair.

> Give both algorithms equally sorted input

Done, please see the new table in the PR body.

avantgardnerio:
@andygrove I think all your (great) feedback has been addressed as well. I'm undrafting this - I think it's the next part of the parallel windows puzzle.

phillipleblanc (approved):
This makes sense to me now, thanks @avantgardnerio!

Issue #2174, opened 2026-07-24
Add Iceberg integration


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
There is a PR to add Ballista integration to iceberg-rust https://github.com/apache/iceberg-rust/pull/2613 and there is a discussion about moving it to a new location.

**Describe the solution you'd like**
Let's discuss bringing this into Ballista repo.

**Describe alternatives you've considered**

**Additional context**

andygrove:
@milenkovicm I was thinking we could add a separate crate (not part of main workspace) and have independent release cycle so that it doesn't impact current Ballista release cycle

milenkovicm:
Same like python support, good idea

NoahKusaba:
Sounds good to me. I can close my PR on iceberg-rust and re-open it here if that works with everyone?
We can merge it in once I get the necessary iceberg-datafusion changes in. 
** Given that the changes are tightly coupled to the Iceberg-Datafusion changes, I'm going to migrate the PR once those get merged in as it is more convenient.

martin-g:
+1

sandugood:
+1
Would like to contribute and then test Ballista on my current workloads :)

phillipleblanc:
This makes sense to me!

NoahKusaba:
Opened the PR at
https://github.com/apache/datafusion-ballista/pull/2217
It's slow going getting the iceberg-datafusion changes merged in, so currently it's pointed to my iceberg-rust branch.

NoahKusaba:
@andygrove can we also look to moving the iceberg-datafusion crate out of iceberg-rust? It seems their team is interested in doing so as well. Currently getting PR's merged in is an extremely slow process due to lack of reviewers. 
Not sure where it would end up living.

NoahKusaba:
@andygrove 
datafusion-distributed is also looking to get iceberg-datafusion out of iceberg-rust, or to fork and maintain their own copy.
https://github.com/datafusion-contrib/datafusion-distributed/issues/533
I think it's extremely important that we move it to datafusion-contrib, so we don't end up with a bunch of forks of this library.

milenkovicm:
https://github.com/apache/iceberg-rust/discussions/2992

Discussion you might be interested in

PR #2217, opened 2026-08-01
feat: Add Iceberg Support

### LINKED ISSUE #1241: [DISCUSS] Add open table format support.
Currently popular open table formats such as iceberg, deltalake all have rust implementation, should we consider adding native support for them in ballista so that users could access these otfs out of box?

NoahKusaba:

Closes https://github.com/apache/datafusion-ballista/issues/1241

(ONLY INCLUDING DISTRIBUTED READS/WRITES integration for now)
(Note that READS are still not properly distributed, as output-partitioning is not supported in iceberg-datafusion yet https://github.com/apache/iceberg-rust/pull/2671)
Adds a new iceberg-ballista crate, which provides a distributed-query driver for Apache Iceberg for a distributed datafusion engine Apache Datafusion-Ballista + the targeted changes to iceberg-datafusion that make Iceberg's existing plan nodes serializable so they can cross node boundaries.

The core problem it solves

Iceberg's DataFusion integration already produces complete physical read and write plans, but every Iceberg plan node holds live, non-serializable state (Arc<dyn Catalog>, an open Table/FileIO). Ballista ships logical and physical plans to remote schedulers/executors, so those nodes couldn't travel. This branch closes that gap with one consistent idea: serialize a minimal self-contained recipe (IcebergCatalogConfig + identifiers), rebuild the live objects on the receiving node.

- IcebergLogicalCodec:  serializes the catalog-backed table provider (config + table ident, plus snapshot/metadata variants) so the scheduler can rebuild it and do physical planning, including INSERT.

- IcebergPhysicalCodec:  serializes the four Iceberg execution nodes (IcebergTableScan, IcebergWriteExec, IcebergCommitExec, IcebergMetadataScan) and the PartitionExpr physical expression.

- Tagged-envelope wire framing (TAG_ICEBERG / TAG_DELEGATED):every blob carries a leading tag; non-Iceberg nodes are delegated to Ballista's own codec, so shuffles/sorts/etc. keep working and an unknown tag is a hard error--> Based off comments from https://github.com/milenkovicm/ballista_delta 

- bridge.rs runtime bridge: Each executor node needs to build an HTTP client with the iceberg catalog which requires an async-call, but PhysicalExtensionCodec from datafusion_proto try_decode is synchronous. The block_on function is a workaround to make this async function call blocking. Each try_decode also performs a load_table catalog round-trip per plan node to resolve the table's current metadata pointer 

- Snapshot pinning at encode time on the scheduler for reads.

# Tests

- Distributed reads / writes tested against dockerized minio iceberg catalog ( Standalone + multi-executor cluster). Also tests partitioned files writes + iceberg table registration.
- Codec serialization / deserialization changes.
- Snapshot Read tests.
(will document better later)

# Other notes
There is a dependency on the changes to the iceberg-datafusion crate, which are not yet merged in.
Currently referencing my branch found at https://github.com/apache/iceberg-rust/pull/2613


milenkovicm:
Thanks @NoahKusaba will have a look tomorrow

milenkovicm (commented):
instead of using docker compose for tests, would it make sense to use testcontainers? we already use it, you could find example at https://github.com/apache/datafusion-ballista/blob/187a650630b418f4aa15ff9635f19b77728b7628/examples/tests/object_store.rs#L42

NoahKusaba:
> instead of using docker compose for tests, would it make sense to use testcontainers? we already use it, you could find example at
> 
> https://github.com/apache/datafusion-ballista/blob/187a650630b418f4aa15ff9635f19b77728b7628/examples/tests/object_store.rs#L42

Good call out, I made the change.

andygrove:
Thanks @NoahKusaba. I'm just getting up to speed with the contribution. My understanding is that this initial PR only works for in-process standalone mode and cannot be used in a real cluster yet, which is fine. It may be worth noting that in the PR description.

Perhaps you could add some minimal documentation in the PR as well, although I'd be fine with adding those as a follow on.

integrations/iceberg/tests/distributed_read_write.rs
```diff
@@ -0,0 +1,712 @@
...
+/// the physical plan, so this exercises that expression's serialization through
+/// the codec on top of the plan-node serialization — and fans the rows out to one
+/// writer per region across the executors.
+///
+/// The assertions target the correctness properties of a distributed, multi-writer
+/// write:
+///   1. the write commits exactly **one** atomic snapshot (not one per task),
+///   2. the parallel writers contributed multiple data files (one per region), and
+///   3. every input row lands exactly once (no loss, no duplication).
+#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
```
NoahKusaba: @andygrove This PR does cover a true multi-executor cluster deployment. 
This is the test that covers that, by making a standalone scheduler + executors.
I'll add in validation to ensure both executors did work (not just 1), and add an example for use.

integrations/iceberg/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Installs the Iceberg logical and physical extension codecs onto a
+/// [`SessionConfig`].
+///
+/// In a standalone cluster the scheduler and executor both derive their codecs
+/// from this config, so one call suffices. For a separately deployed scheduler
+/// and executor, set the same codecs on their process configs
+/// (`override_logical_codec` / `override_physical_codec`).
+pub fn register_iceberg_codecs(config: SessionConfig) -> SessionConfig {
+    config
+        .with_ballista_logical_extension_codec(Arc::new(IcebergLogicalCodec::default()))
```
akshaychitneni: Would this replace existing codecs when used with other ballista extentions? Could this compose instead - read the config's current logical/physical codec and wrap that as inner (fall back to Ballista only
  if unset)? cc @milenkovicm @andygrove

akshaychitneni (changes requested):

integrations/iceberg/src/lib.rs
NoahKusaba: Good catch! Hadn't considered other codec's being used besides Ballista's default. 
Will add it in a follow up PR.

I should also note that currently logical_codec's try_encode falls back to the set inner codec if it's not a icebergTableProvider, so this should get us the desired behavior.

andygrove:
@NoahKusaba I saw a notification a few days ago on a conversation about also moving the iceberg-datafusion integration out of the iceberg-rust repo, but now I cannot find that conversation. Would you mind sharing a link here?

NoahKusaba:
> @NoahKusaba I saw a notification a few days ago on a conversation about also moving the iceberg-datafusion integration out of the iceberg-rust repo, but now I cannot find that conversation. Would you mind sharing a link here?

https://github.com/apache/iceberg-rust/pull/2613#issuecomment-4830211788

Most recent conversation:

https://github.com/apache/datafusion-ballista/issues/2174#issuecomment-5072944934

milenkovicm:
hey @NoahKusaba whats the status of this PR? who do we wait for ?

NoahKusaba:
> hey @NoahKusaba whats the status of this PR? who do we wait for ?

Basically all iceberg-datafusion PR's seem to be blocked over on iceberg-rust until the migration to the new repo is finally done (vote is pending until next week). After that  I need to get all of my changes merged in from https://github.com/apache/iceberg-rust/pull/2613, which I'll split up into 3-4 PR's. 

There is one hitch, which is that iceberg-datafusion maintains two table providers: StaticTableProviders for reads and IcebergTableProvider for everything else. Currently when we do a try_decode_table_provider for the LogicalCodec in the scheduler, datafusion-proto encodes DML write targets as synthetic TableScan nodes, so we don't have a way to distinguish between reads and writes and being able to build the appropriate TableProvider accordingly. 

mbutrovich suggested making a PR to datafusion to address distinguishing reads from writes in the codec. I put together a prototype that "works", but given that the iceberg-datafusion's situation was in flux, I tabled working on it until I could get the bulk of the other iceberg-datafusion PR's in. I'm also not completely satisfied with the solution I came up with either, and it needs more work. 

You can see the bulk of that discussion over at : https://github.com/apache/iceberg-rust/pull/2862

milenkovicm:
Thanks @NoahKusaba

PR #2225, opened 2026-08-04
fix(executor): bind the gRPC listener before registering with the scheduler

### LINKED ISSUE #2224: Executor registers before its gRPC server is bound, so a lost race kills the executor at startup
**Describe the bug**

In push-based scheduling, an executor registers with the scheduler before its own gRPC server has bound its listening socket. The scheduler's registration handler dials back to the executor to verify connectivity, so if it wins the race it gets `ConnectionRefused`, registration fails, and the executor process exits.

The ordering is already flagged by a `TODO` in the code:

- `executor_server::startup` (`ballista/executor/src/executor_server.rs`) logs `"... Rust Executor Grpc Server listening on ..."` and then spawns the tonic server with `tokio::spawn(...serve_with_shutdown(addr, ...))`. The bind happens inside that spawned task, so the log line is emitted before the socket exists.
- Immediately afterwards, step 2 calls `register_executor`, with the standing comment: `// TODO the executor registration should happen only after the executor grpc server started.`
- On the scheduler side, `ExecutorManager::register_executor` (`ballista/scheduler/src/state/executor_manager.rs`) calls `test_connectivity`, which is a single `ExecutorGrpcClient::connect(...)` with no retry and no backoff.
- If that connect loses the race, registration returns an error, and `executor_process` treats it as fatal ("If there is executor registration error during startup, return the error and stop early"), so the executor exits rather than retrying.

**To Reproduce**

Observed in CI on the new HA chaos harness (`cargo test -p ballista-chaos --test ha`), in `baseline_matches_local_datafusion::case_2_aqe_on`. Two executors were started, one registered fine, the other lost the race and died, so the harness timed out:

```
timed out waiting for 2 executors to register
```

executor-1 log:

```
INFO  ballista_executor::executor_server] Ballista v54.0.0 Rust Executor Grpc Server listening on 127.0.0.1:42635
ERROR ballista_executor::executor_server] Executor registration failed due to: Grpc error: code: 'Internal error',
  message: "Fail to do executor registration due to: Internal Ballista error: Failed to register executor at
  127.0.0.1:42635, could not connect: tonic::transport::Error(Transport, ConnectError(ConnectError("tcp connect error",
  127.0.0.1:42635, Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })))"
```

scheduler log, same second:

```
INFO  ballista_scheduler::scheduler_server::grpc] Received register executor request for ExecutorRegistration { id: "09332458-...", grpc_port: 42635, ... }
ERROR ballista_scheduler::scheduler_server::grpc] Fail to do executor registration due to: Internal Ballista error:
  Failed to register executor at 127.0.0.1:42635, could not connect: ... ConnectionRefused ...
```

It is timing-dependent, so it reproduces rarely and mostly on loaded machines. It is not specific to the chaos harness: any push-staged deployment where the executor is slow to reach the bind can hit it, and the consequence there is an executor that exits at startup instead of joining the cluster.

**Expected behavior**

An executor that is otherwise healthy should always succeed in registering. Its gRPC listener should be accepting connections before it tells the scheduler about itself.

**Additional context**

Two candidate fixes, not mutually exclusive:

1. Bind synchronously in `executor_server::startup` before registering, i.e. create the `TcpListener` (or `TcpIncoming`) on the current task, then hand it to `serve_with_incoming_shutdown` in the spawned task. This closes the race at the source and lets the "listening on" log line tell the truth. It also removes the existing `TODO`.
2. Give `ExecutorManager::test_connectivity` a short bounded retry (a few hundred milliseconds) so a marginally slow executor is tolerated rather than rejected outright.

(1) alone fixes the local case. (2) additionally helps where the executor is on another host and is still coming up.

andygrove:

Closes #2224.

# Rationale for this change

In push-based scheduling the executor tells the scheduler about itself before its own gRPC server is listening, and the scheduler dials that port back to verify connectivity. When the callback wins the race the executor dies at startup.

The ordering in `executor_server::startup` was:

1. `tokio::spawn` the tonic server. `Server::serve` binds *inside* the future, so the socket does not exist yet when `spawn` returns — even though the `"... Grpc Server listening on ..."` line is already in the log.
2. `register_executor`, immediately.

On the scheduler side `ExecutorManager::register_executor` calls `test_connectivity`, a single `ExecutorGrpcClient::connect(...)` with no retry and no backoff. If the spawned task has not reached its `bind` yet, that connect gets `ECONNREFUSED`, registration returns an error, and `executor_process` treats a registration error as fatal, so the executor exits instead of joining the cluster.

The opposite direction is already tolerant: the executor retries its connection *to* the scheduler in a loop. Only the scheduler's callback to the executor is one-shot, which is why the executor's own startup ordering has to be right.

It is timing-dependent and so shows up rarely and on loaded machines. The instance that prompted this was CI, where one of two executors lost the race and the HA chaos harness timed out waiting for it:

```
timed out waiting for 2 executors to register

# executor-1
INFO  Ballista v54.0.0 Rust Executor Grpc Server listening on 127.0.0.1:42635
ERROR Executor registration failed due to: ... Failed to register executor at 127.0.0.1:42635,
      could not connect: ... Os { code: 111, kind: ConnectionRefused }

# scheduler, same second
ERROR Fail to do executor registration due to: ... ConnectionRefused ...
```

The consequence outside of tests is the same shape: an executor that is otherwise healthy fails to join, and in a restart loop it can keep failing to join.

- `ballista/core/src/utils.rs` — new `create_grpc_server_incoming(addr, &GrpcServerConfig)`, which binds the listening socket eagerly and returns tonic's `TcpIncoming`. tonic ignores the server builder's `tcp_nodelay` / `tcp_keepalive` when serving from a pre-bound listener, so the helper applies the same values `create_grpc_server` sets and keeps the two in one place. The remaining settings (timeout, HTTP/2 keep-alive) still come from the builder as before, so the socket is configured exactly as it was.
- `ballista/executor/src/executor_server.rs` — `startup` binds via that helper on the current task, before spawning, and the spawned task now serves with `serve_with_incoming_shutdown`. Registration therefore cannot run before the port is accepting connections. This retires the standing `// TODO the executor registration should happen only after the executor grpc server started.` The `"listening on"` log line is now true when it is printed.

One incidental improvement falls out of binding eagerly: a port conflict now fails `startup` directly, with the bind error, rather than being discovered later through the spawned server task.

No change to the scheduler's `test_connectivity`. The issue also floats giving it a bounded retry; that is worth doing on its own merits for genuinely remote executors, but it is a separate resilience change and is not needed to fix this race, which is entirely local to the executor's startup ordering.

New tests in `ballista-core`:

- `test_create_grpc_server_incoming_binds_eagerly` — connects to the bound address while nothing is serving on it. This is the property the fix depends on, and it fails against a lazily-bound socket.
- `test_create_grpc_server_incoming_port_in_use` — binding an address twice is an error rather than a panic, so a port conflict still surfaces as a normal startup failure.

No behaviour changes and no API breakage. `create_grpc_server_incoming` is a new public function in `ballista-core`, additive alongside `create_grpc_server`. Note that it must be called from within a Tokio runtime, since the listener registers with the reactor; this is documented on the function.

### DIFF
--- a/ballista/core/src/utils.rs
+++ b/ballista/core/src/utils.rs
@@ -359,18 +359,15 @@ pub fn create_grpc_server(config: &GrpcServerConfig) -> Server {
 }
 
 /// Binds a gRPC server's listening socket, for use with tonic's
-/// `serve_with_incoming` / `serve_with_incoming_shutdown`.
-///
-/// tonic's `serve` binds lazily, inside the future it returns, so a caller that
-/// spawns that future has no way to know when the port started accepting
-/// connections. Use this instead whenever something else must be able to reach
-/// the port as soon as the server is started: the socket is listening by the
-/// time this returns.
+/// `serve_with_incoming` / `serve_with_incoming_shutdown`. Unlike tonic's
+/// `serve`, which binds lazily inside the future it returns, the socket is
+/// listening by the time this returns — so use this whenever a peer may be
+/// told to connect as soon as the server is started.
 ///
 /// tonic ignores the builder's `tcp_nodelay` and `tcp_keepalive` when serving
 /// from a pre-bound listener, so this applies the same values that
-/// [`create_grpc_server`] sets. The remaining settings still come from the
-/// builder.
+/// [`create_grpc_server`] sets, for the same reasons. The remaining settings
+/// still come from the builder.
 ///
 /// # Panics
 ///
@@ -381,7 +378,6 @@ pub fn create_grpc_server_incoming(
     config: &GrpcServerConfig,
 ) -> Result<TcpIncoming> {
     Ok(TcpIncoming::bind(addr)?
-        // Disable Nagle's Algorithm since we don't want packets to wait
         .with_nodelay(Some(true))
         .with_keepalive(Some(Duration::from_secs(config.tcp_keepalive_seconds))))
 }
@@ -471,11 +467,11 @@ mod tests {
         assert!(result.is_err());
     }
 
-    /// The whole point of binding up front is that the port is reachable before
+    /// The point of binding up front is that the port is reachable before
     /// anything is served on it, so a peer told to connect back cannot arrive
     /// too early.
     #[tokio::test]
-    async fn grpc_server_incoming_accepts_connections_before_it_is_served() {
+    async fn test_create_grpc_server_incoming_binds_eagerly() {
         let incoming = create_grpc_server_incoming(
             "127.0.0.1:0".parse().unwrap(),
             &GrpcServerConfig::default(),
@@ -487,18 +483,5 @@ mod tests {
         tokio::net::TcpStream::connect(addr)
             .await
             .expect("port is already listening");
-    }
-
-    #[tokio::test]
-    async fn test_create_grpc_server_incoming_port_in_use() {
-        let first = create_grpc_server_incoming(
-            "127.0.0.1:0".parse().unwrap(),
-            &GrpcServerConfig::default(),
-        )
-        .expect("bind");
-        let addr = first.local_addr().expect("local addr");
-
-        let result = create_grpc_server_incoming(addr, &GrpcServerConfig::default());
-        assert!(result.is_err());
     }
 }
--- a/ballista/core/src/utils.rs
+++ b/ballista/core/src/utils.rs
@@ -484,4 +484,19 @@ mod tests {
             .await
             .expect("port is already listening");
     }
+
+    /// Binding eagerly means a port conflict surfaces here, as an error, rather
+    /// than later inside the spawned server task.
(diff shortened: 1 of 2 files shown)


andygrove:
@villebro @akshaychitneni could you review?

avantgardnerio (approved):
Thank you for fixing this!

Nit: the body claims `test_create_grpc_server_incoming_port_in_use` but I don't see that in the code?

andygrove:
Good catch, thanks. That test was dropped by a cleanup commit rather than renamed, so it really wasn't there. Restored it in d76c435, and fixed the body which was also still using the old name for the other test.

PR #2229, opened 2026-08-05
fix: update mTLS example for poll_loop arguments

phillipleblanc:
## What changed

Pass the new `free_vcores` argument (`None`) when the mTLS cluster example starts the executor poll loop.

## Why

PR #1893 added the `free_vcores` parameter to `poll_loop`, but `examples/examples/mtls-cluster.rs` retained the old five-argument call. Current `main` therefore fails the CI Clippy gate while compiling `ballista-examples` with `E0061`.

## Validation

- `./ci/scripts/rust_clippy.sh`

The Clippy failure was reproduced on Apache `main` at `10d4d087` before this change; the complete gate passes after it.


phillipleblanc:
cc @milenkovicm this fixes a clippy issue that landed on main with https://github.com/apache/datafusion-ballista/pull/1893

milenkovicm (approved):
thanks @phillipleblanc

PR #2240, opened 2026-08-06
fix(scheduler): route launch-failure executor removal through ExecutorLost

### LINKED ISSUE #2226: Flaky test: killing_every_executor_terminates_the_job::case_2_aqe_on hangs under AQE, surviving the #2029 fix
**Describe the bug**

`killing_every_executor_terminates_the_job::case_2_aqe_on` (`chaos-testing/tests/ha.rs`) intermittently fails on CI: the query never terminates, and the test's 120s guard fires.

This is the same hang that #2029 describes, surviving the fix in #2212 — but only on the AQE-on case, and only sometimes. The AQE-off case passed in the same run, in about 6 seconds.

Observed on the `macos test` job of https://github.com/apache/datafusion-ballista/actions/runs/30954222353/job/92143349419 (a PR run whose changes are unrelated to scheduling):

```
---- killing_every_executor_terminates_the_job::case_2_aqe_on stdout ----
thread 'killing_every_executor_terminates_the_job::case_2_aqe_on' panicked at chaos-testing/tests/ha.rs:410:10:
job must terminate, not hang, after every executor is killed: Elapsed(())

test result: FAILED. 9 passed; 1 failed; 6 ignored; 0 measured; 0 filtered out; finished in 150.45s
```

`Elapsed(())` is the `tokio::time::timeout(Duration::from_secs(120), query)` in the test, so the client future was still pending 120s after both executors were `SIGKILL`ed, with the harness grace period turned down to about a second.

This appears to be the first occurrence — I checked the last 25 `rust.yml` runs and every other failure was a different job or a different test. So the failure is intermittent rather than a straightforward breakage.

**To Reproduce**

```sh
cargo test -p ballista-chaos --test ha killing_every_executor_terminates_the_job
```

Expect to need many iterations to hit it. It reproduced on a macOS runner, where the executors are slower to start and reap, so the timing is likely load-dependent.

**Expected behavior**

Once the last executor is reaped and the grace period elapses, the scheduler fails the job with an error naming the executor loss, under AQE on and AQE off alike. The test should pass deterministically.

**Additional context**

One hypothesis worth checking first, from reading the `ExecutorLost` handler in `query_stage_scheduler.rs`: the no-executors check is single-shot per burst. `no_executor_check_pending` collapses the storm of `ExecutorLost` events into one timer, and after the grace period the timer walks `get_running_job_cache()` and skips (`continue`) any job whose live status is not `Running` at that instant, then clears the flag.

If AQE is mid-replan when the timer fires, so the job is momentarily not reported as `Running`, the job is skipped — and because both executors are already gone, no further `ExecutorLost` event will ever arrive to schedule a second check. The job would then hang forever. That is consistent with the failure being AQE-on-only and timing-dependent, but it is a hypothesis, not a diagnosis.

Related:

- #2029 is the original hang. It is still open because #2212's body has a malformed `Closes #.` reference, so GitHub never linked or auto-closed it. Worth deciding whether this issue supersedes it or whether #2029 should be closed and this tracked separately.
- #2092 (AQE by default) — this is a blocker for trusting the AQE-on path under executor loss.

andygrove:

Closes #2226.

# Rationale for this change

`killing_every_executor_terminates_the_job::case_2_aqe_on` has been failing
intermittently on CI: the query never terminates and the test's 120s guard
fires. The grace-period fix for #2029 (#2212) did not cover it.

The scheduler discovers that an executor is gone in two different places, and
only one of them arms the empty-cluster grace timer:

- The heartbeat reaper calls `SchedulerServer::remove_executor`, which posts
  `QueryStageSchedulerEvent::ExecutorLost`. The handler for that event rolls
  back the affected execution graphs, cancels their running tasks and, if no
  executors remain, arms the timer that fails the jobs left behind.
- A failing task launch calls `SchedulerState::remove_executor`, which did the
  rollback and cancellation inline and posted no event, so the timer was never
  armed.

The second path also deletes the executor's heartbeat, so the reaper can never
rediscover the executor and post the event later. When a whole cluster dies
while tasks are being launched onto it, no executor-loss event ever reaches the
handler: the job stays `Running` forever while the client polls its status
every 50ms. Whether a kill is noticed by a failing launch or by heartbeat
expiry is a timing race, which is why the scenario failed only sometimes and
only under some plans.

The scheduler log from a reproduced hang shows the whole story: both executors
removed with reason `Failed to launch new task: ... broken pipe`, both stages
reset, then nothing. No `heartbeat timed out`, because there was no heartbeat
left to expire.

- `SchedulerState::remove_executor` now posts `ExecutorLost` instead of
  duplicating the loss handling, so both removal paths converge on the single
  handler in `query_stage_scheduler.rs`. The event sender is threaded in from
  `revive_offers` through `launch_tasks`. Ordering is unchanged: the event is
  posted after the executor is removed from `ExecutorManager` and before the
  `ReviveOffers` that follows the launch batch.
- New unit test `test_running_job_fails_when_launch_failure_loses_last_executor`
  covering the launch-failure path deterministically, plus a
  `lose_executor_on_launch_failure` harness helper alongside the existing
  `lose_executor`, which mirrors the reaper. The new test fails on `main`
  (the job is still `Running` when the assertion runs) and passes here; the
  existing reaper-path test passes either way, which is why this gap went
  unnoticed.
- A note in `chaos-testing/README.md` under Finding 3 recording the second
  path.

Verification: `cargo test -p ballista-scheduler` (325 passed), the full
`cargo test -p ballista-chaos --test ha -- --test-threads=1` suite (10 passed,
6 pre-existing ignores), `cargo fmt --all -- --check`, `cargo clippy
-p ballista-scheduler --all-targets --all-features -- -D warnings`, and
`cargo check -p ballista-scheduler --no-default-features`.

No API changes. A job whose executors are all lost during task launch now fails
with the same "all executors were lost ..." error that the heartbeat-expiry
path already produced, instead of hanging.

### DIFF
--- a/ballista/scheduler/src/state/mod.rs
+++ b/ballista/scheduler/src/state/mod.rs
@@ -190,7 +190,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
         let state = self.clone();
         tokio::spawn(async move {
             let mut if_revive = false;
-            match state.launch_tasks(schedulable_tasks, sender.clone()).await {
+            match state.launch_tasks(schedulable_tasks, &sender).await {
                 Ok(unassigned_executor_slots) => {
                     if !unassigned_executor_slots.is_empty() {
                         if let Err(e) = state
@@ -227,11 +227,9 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
     ///    tasks, and — when this was the last executor — arms the grace timer
     ///    that fails the jobs left behind on an empty cluster.
     ///
-    /// The loss goes through the event rather than being handled inline so that
-    /// every removal path converges on one handler. Step 1 also drops the
-    /// executor's heartbeat, so the reaper can never rediscover the executor and
-    /// post the event itself: rolling the graphs back here without posting would
-    /// leave a job whose whole cluster died during a task launch running forever.
+    /// Every removal path must go through here, because step 1 also drops the
+    /// executor's heartbeat: once it is gone, nothing else can notice the
+    /// executor is missing and post the event later.
     /// See <https://github.com/apache/datafusion-ballista/issues/2226>
     pub(crate) async fn remove_executor(
         &self,
@@ -266,7 +264,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
     async fn launch_tasks(
         &self,
         bound_tasks: Vec<BoundTask>,
-        sender: EventSender<QueryStageSchedulerEvent>,
+        sender: &EventSender<QueryStageSchedulerEvent>,
     ) -> Result<Vec<ExecutorSlot>> {
         // Put tasks to the same executor together
         // And put tasks belonging to the same stage together for creating MultiTaskDefinition
--- a/ballista/scheduler/src/state/mod.rs
+++ b/ballista/scheduler/src/state/mod.rs
@@ -190,7 +190,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
         let state = self.clone();
         tokio::spawn(async move {
             let mut if_revive = false;
-            match state.launch_tasks(schedulable_tasks).await {
+            match state.launch_tasks(schedulable_tasks, sender.clone()).await {
                 Ok(unassigned_executor_slots) => {
                     if !unassigned_executor_slots.is_empty() {
                         if let Err(e) = state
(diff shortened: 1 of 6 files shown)


ballista/scheduler/src/state/mod.rs
```diff
@@ -309,7 +315,9 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
 
                             // It's OK to remove executor aggressively,
                             // since if the executor is in healthy state, it will be registered again.
-                            state.remove_executor(&executor_id, Some(err_msg)).await;
+                            state
```
akshaychitneni, line 318: I see we let the healthy executor register back and the default `executor_heartbeat_interval_seconds` is 60s. As with this change, the removal of executor can trigger `no_executors_grace_period_seconds` which default 30s, I think we should enforce  `no_executors_grace_period_seconds` > `executor_heartbeat_interval_seconds`

milenkovicm (approved):
thanks @andygrove

ballista/scheduler/src/state/mod.rs
akshaychitneni: @andygrove @milenkovicm  probably this was missed. Do you agree to update default value and add validation?

milenkovicm: Makes sense

PR #2241, opened 2026-08-06
chore(deps): Track DataFusion git, pinned to the 55.0.0-rc2 tag

### LINKED ISSUE #1920: Track DataFusion git main via a pinned rev and bump it periodically

andygrove:

Closes #1920.

Supersedes #1921, which carried the same work on a personal topic branch.

# Rationale for this change

Ballista depends on published DataFusion crates.io releases (today, `54`). Because a full release cycle passes between DataFusion versions, breaking changes only surface in Ballista at upgrade time, all at once, making each upgrade large and risky.

This PR switches `main` to tracking a git ref of `apache/datafusion` instead. Rather than keeping a long-running `datafusion-55` branch open and merging it at release time, the porting work lands on `main` now, and the pinned ref is bumped periodically (weekly) from then on. Breaking API and behaviour changes then arrive in small increments, close to when they land upstream, instead of accumulating until the next release.

The ref this PR lands on is the **`55.0.0-rc2` tag** — a fixed, reviewable point rather than a moving `main` commit. Subsequent bumps advance it to newer DataFusion revs, and it moves to the crates.io release once DataFusion 55.0.0 ships.

Each bump is expected to:

- advance the pinned ref in the workspace `Cargo.toml`,
- port any DataFusion API changes,
- refresh any snapshots/goldens affected by planner/metric changes,
- keep `cargo test` and `ci/scripts/rust_clippy.sh` (`-D warnings`) green.

## Dependency pin

Pin the `datafusion*` crates to the `55.0.0-rc2` tag of `apache/datafusion` and bump `arrow`/`arrow-flight` from `58.4` to `59.2` (rc2 requires `arrow` `^59.2`). Sync the vendored `datafusion_common.proto` for the new `max_in_list_size` field.

## Port to the DataFusion 55 APIs

- `TableReference` moved from `datafusion::sql` to `datafusion::common`.
- `PhysicalExtensionCodec::try_decode`/`try_encode` gained a `PhysicalProtoConverterExtension` argument.
- `QueryPlanner::create_physical_plan` takes `&dyn Session` instead of `&SessionState`.
- `MetricValue` gained a `PeakMemoryUsage` variant; serialize it as a named gauge.
- `Partitioning` gained a `Range` variant; format it in the DOT graph and treat it as a fresh K-space in the shuffle writer's partition mapping.
- `EnforceSorting` and `EnforceDistribution` were unified into `EnsureRequirements`; use it for post-stage-split re-optimization.
- `ExecutionPlan::partition_statistics` is deprecated in favour of `StatisticsContext::compute`; migrate all call sites. This is a **correctness fix**: DataFusion operators that no longer override the deprecated method return unknown statistics through it, which would silently disable Ballista's broadcast-threshold checks in `BallistaPlanner` and in the AQE join resolver.
- `ExecutionPlan::required_input_distribution` is deprecated in favour of `input_distribution_requirements`; migrate `child_scopes` in the task builder and its test.
- `ListingOptions` dropped its `target_partitions`/`collect_stat` fields; build it via the builder (`target_partitions` now comes from the session config).
- `Distribution::HashPartitioned` is deprecated; use `KeyPartitioned`.
- `datafusion_substrait::serializer::deserialize_bytes` is now synchronous and takes `&[u8]`.
- DataFusion 55 declares `rust-version = 1.94.0`, so the workspace MSRV moves from `1.88.0` to `1.94.0`. The `msrv dependency check` job passed either way, but the old value could no longer be built against.

### `ExecutionPlan::apply_expressions`

New **required** trait method: each node reports the root expressions it owns and evaluates, shallowly. Implemented for all 20 Ballista `ExecutionPlan`s:

| Operator | Reported expressions |
| --- | --- |
| `PerPartitionFilterExec` | all per-partition predicates |
| `RangeFilterExec` | routing expr (the bounds are `ScalarValue`s, not expressions) |
| `RuntimeStatsExec` | ORDER BY (first entry drives the T-Digest) |
| `OrderedRangeRepartitionExec`, `UnorderedRangeRepartitionExec` | ORDER BY (evaluated scatter-side to route by value range) |
| `RangeShuffleReaderExec` | merge key, as `SortPreservingMergeExec` does |
| `SortShuffleWriterExec` | hash partitioning keys (it hashes rows itself) |
| `DynamicJoinSelectionExec` | join keys + filter, matching `HashJoinExec` |
| `ExchangeExec` | shuffle partitioning (`Hash` keys / `Range` ordering) |
| `PartitionedBoundedWindowAggExec` | delegates to the inner BWAG, which is not a plan-tree child |
| pass-throughs and placeholders | none |

`ShuffleWriterExec` reports none: it preserves its input partitioning, so those expressions belong to the child plan.

### `with_new_children` → `replace_children`

`ExecutionPlan::with_new_children` is deprecated in favour of `replace_children` with `ReplaceChildrenOptions`, and the free function `with_new_children_if_necessary` is renamed to `replace_children_if_necessary`. Since CI runs clippy with `-D warnings`, all call sites are migrated.

Call sites use the explicit `ChildrenPropertiesMode::Recompute` form, which is a 1:1 translation of the old behaviour, rather than routing through `replace_children_if_necessary` — that helper adds pointer-equality and properties-cache short-circuits, and adopting them inside a dependency bump would be a behaviour change in the planner rules.

`with_new_children` is still a required trait method upstream, so the operator implementations keep it; `replace_children`'s default implementation delegates to it. Migrating the implementations is left as a follow-up.

One exception: `reset_metrics_for_execution_plan` rebuilt each node around its own children to drop decoded metrics. That is exactly what `reset_state` is for upstream, so it now calls that — its default implementation is `replace_children(.., Keep)`, which is what the old code did with the children unchanged.

## Test and snapshot updates

Disable dynamic filter pushdown in the AQE mock session config so it matches the Ballista production default. Ballista turns this off because dynamic filters cannot cross stage boundaries, and without it the mock config let AQE re-optimization perform join-input swaps that are illegal in the real configuration.

Refresh the EXPLAIN ANALYZE expectations, the AQE plan snapshots, and the TPC-H plan-stability golden files for these upstream changes:

- `AggregateExec` no longer reports a `peak_mem_used` metric.
- `HashJoinExec` now prints a `null_aware` marker in its display.
- Decimal literals now render as their scaled value (`0.05`) rather than as raw `Some(5),15,2` internals.
- Semi-join lowering changed: several TPC-H subqueries that previously produced `SortMergeJoinExec: join_type=Inner` now produce `LeftSemi`/`RightSemi`, which also changes the shuffle keys of the affected stages.
- Projections are now placed above sorts rather than below them in several plans.
- Filter conjunct ordering changed in q16.

None of these golden-file diffs change query results; they are plan-shape and formatting changes from upstream. The goldens refreshed against the earlier `main` rev still match rc2 unchanged.

# Verification

- `cargo build --workspace --all-targets` clean.
- `ci/scripts/rust_clippy.sh` (clippy `--all-targets -D warnings`, all packages) clean.
- `cargo test --workspace`: all tests pass except those that bind a local socket (the standalone-cluster client integration tests, the chaos cluster tests, and two `create_grpc_server_incoming` unit tests), which the local sandbox blocks. Notably the TPC-H plan-stability suite (25 tests) and the scheduler unit tests including the AQE snapshots (350 tests) pass.

Ballista builds against the `55.0.0-rc2` tag of DataFusion rather than a crates.io release, and picks up the DataFusion 55 behaviour changes listed above. There are no changes to Ballista's own public API.

Note on the Python bindings: `python/` is excluded from the workspace and continues to depend on the published `ballista`/`datafusion` `54.0.0` crates, so it is unaffected by this PR and its CI jobs are unchanged. Advancing it needs a `datafusion-python` release built on arrow 59 / pyo3 0.29, which does not exist yet; that is tracked separately from this bump.


andygrove:
@milenkovicm @avantgardnerio @phillipleblanc This PR upgrades DF from 54 to 55 (rc2) and as you can see, it is pretty huge PR. If we move to the new approach of pinning to DF main and upgrading weekly we will have lots of smaller PRs instead rather than one huge one at the end. It will also allow more time for fixing things upstream if we run into problems.

I have not run any benchmarks yet and probably cannot until next week, and DF 55 will likely have already shipped by then.

avantgardnerio (approved):
I had my Claude review this, and I think it's mergable as-is. I had him turn his notes into a [stacked PR](https://github.com/apache/datafusion-ballista/pull/2295) you can diff over any or all of it as you see fit.

We also ran:
1. h2o query 8 to ensure the ParallelWindows rule didn't break
2. TPC-H SF100 locally to completion, but didn't verify performance or results
3. I held off on running TPC-H 1000 due to the [outstanding issue](https://github.com/apache/datafusion-ballista/issues/2290) preventing benchmarking on main

I'll have Claude add his nits as comments as well. I'm looking forward to the new method of pinning DF commits in main :partying_face:

ballista/scheduler/src/state/task_builder.rs
```diff
@@ -292,6 +292,13 @@ fn select_output_partitions(
             Partitioning::UnknownPartitioning(_) => {
                 Partitioning::UnknownPartitioning(kept.len())
             }
+            Partitioning::Range(_) => {
```
avantgardnerio, line 303: This arm rewrites the partitioning to `UnknownPartitioning` while still slicing the partition list, so the reader handed back under-reports its own partitioning and its split points are gone.

It may also not be strictly unreachable: DataFusion 55 constructs `Partitioning::Range` in `physical_planner.rs` when lowering a logical range `Repartition`, and in `ListingTable`. Nothing routes one into a `ShuffleReaderExec` today, since `DistributedExchangeRule` only ever builds from `RepartitionExec::Hash`, so it is latent rather than live.

#2295 turns it into an `internal_err!` with a TODO describing what support would need (slicing the split points alongside the partition slice, the way the `RangeFilterExec` arm slices `raw_bounds`).

ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -871,6 +873,19 @@ impl ExecutionPlan for SortShuffleWriterExec {
...
+    /// This writer hashes rows itself rather than relying on an upstream
+    /// `RepartitionExec`, so the partitioning expressions are its own.
+    /// `try_new` rejects anything but `Hash`.
+    fn apply_expressions(
+        &self,
+        f: &mut dyn FnMut(&Arc<dyn PhysicalExpr>) -> Result<TreeNodeRecursion>,
+    ) -> Result<TreeNodeRecursion> {
+        match &self.shuffle_output_partitioning {
+            Partitioning::Hash(exprs, _) => apply_expression_roots(exprs, f),
+            _ => Ok(TreeNodeRecursion::Continue),
```
avantgardnerio: The doc comment above is correct that `try_new` rejects anything but `Hash`, which makes this arm unreachable. As written though, a writer that somehow held another scheme would silently report no expressions rather than surfacing the broken invariant.

#2295 swaps it for a let-else returning the same error `try_new` raises.

ballista/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -2226,10 +2247,12 @@ mod tests {
...
         let schema = Arc::new(Schema::new(vec![Field::new("c", DataType::Int32, false)]));
         let reader = ShuffleReaderExec::try_new_broadcast(7, vec![], schema, 3).unwrap();
-        let err = reader.partition_statistics(Some(1)).unwrap_err();
+        let err = StatisticsContext::new()
+            .compute(&reader, &StatisticsArgs::new().with_partition(Some(1)))
+            .unwrap_err();
         let msg = err.to_string();
         assert!(
-            msg.contains("invalid partition index 1"),
+            msg.to_lowercase().contains("invalid partition index"),
```
avantgardnerio: Might be worth keeping the index in this assertion. Dropping it hides that the test quietly changed target: DataFusion bounds-checks the partition index inside `StatisticsContext::compute` before dispatching to the operator, so the error is now upstream's (`Invalid partition index: 1, the partition count is 1`) rather than the broadcast guard in `partition_statistics` that this test was written for.

#2295 asserts on the message that actually fires, index included.

Related, and happy to leave alone: that broadcast guard now looks unreachable, and `ShuffleReaderExec` is one of the few operators here still implementing the deprecated `partition_statistics` rather than `statistics_from_inputs` / `child_stats_requests`. Glad to send a follow-up if you would like it migrated with the others.

ballista/core/src/execution_plans/runtime_stats.rs
```diff
@@ -296,6 +298,22 @@ impl ExecutionPlan for RuntimeStatsExec {
...
+    /// feed the T-Digest. The whole slice is reported: it is carried for serde
+    /// and for the downstream operators that consume it.
+    fn apply_expressions(
+        &self,
+        f: &mut dyn FnMut(&Arc<dyn PhysicalExpr>) -> Result<TreeNodeRecursion>,
+    ) -> Result<TreeNodeRecursion> {
+        apply_expression_roots(
+            self.order_by
+                .iter()
+                .flatten()
```
avantgardnerio: `execute` evaluates only the first key here: `routing_expr` is `order_by.as_ref().and_then(|exprs| exprs.first())`. The remaining keys are carried so multi-key ORDER BY survives serde for downstream operators, which makes them ordering metadata.

The trait doc excludes that case: an expression must not be visited solely because it describes an input or output property such as cached ordering, partitioning, or equivalence metadata.

No consumer impact today, since `plan_contains_expression_id` is the only caller in 55 and Ballista disables dynamic filter pushdown anyway. #2295 narrows it to the first key, derived the same way `execute` derives it so the two cannot drift.

benchmarks/src/bin/tpch.rs
```diff
@@ -1203,7 +1203,7 @@ async fn get_table(
     path: &str,
     table: &str,
     table_format: &str,
-    target_partitions: usize,
+    _target_partitions: usize,
```
avantgardnerio: Small one: now that `ListingOptions` has dropped the field, this parameter is dead rather than temporarily unused, and the callers already set `with_target_partitions` on the session config.

#2295 removes it. `register_datafusion_tables` existed only to forward it, so its `partitions` argument goes too.

avantgardnerio (commented):
Went through this one carefully and ran it locally: TPC-H SF100 passes 21/21 on a 2-executor cluster, and h2o Q8 keeps the full parallel-window shape (ORRE -> RangeShuffleReader -> RangeFilter(halo=[3,0]) -> PBWAG -> RangeFilter(halo=[0,0])) with cuts resolved from the runtime sketch. Nothing below is a blocker.

A few small things, all of them latent rather than live. Fixes are stacked in #2295 against this branch, so they can land with it or be ignored, whichever you prefer.

andygrove:
> A few small things, all of them latent rather than live. Fixes are stacked in #2295 against this branch, so they can land with it or be ignored, whichever you prefer.

Thanks @avantgardnerio I went ahead and merged your PR into this branch

andygrove:
I guess let's go ahead and merge this and bump again for the next RC as per the new dependency policy. Thanks for the review & fixes @avantgardnerio

PR #2242, opened 2026-08-06
fix(scheduler): Emit ExecutorLost after failed launch_tasks()

### LINKED ISSUE #2226: Flaky test: killing_every_executor_terminates_the_job::case_2_aqe_on hangs under AQE, surviving the #2029 fix

martin-g:

Closes #2226

# Rationale for this change

Emit `ExecutorLost` event after failed `state.launch_tasks()` call.

`launch_tasks` removes failed executors via
`SchedulerState::remove_executor`, which resets affected stages but does NOT post an `ExecutorLost` event. The grace-period timer that fails jobs on an empty cluster is only armed inside the `ExecutorLost` event handler. If `launch_tasks` is the path that removes the last executor (e.g. AQE plans a new stage right as executors die), no `ExecutorLost` event is ever posted, the grace timer never fires, and the job hangs forever.

Post `ExecutorLost` for each failed executor so the event handler's empty-cluster check runs. Calling
`task_manager.executor_lost()` twice for the same executor is safe (idempotent: the second call finds nothing to reset).

No

### DIFF
--- a/ballista/scheduler/src/state/mod.rs
+++ b/ballista/scheduler/src/state/mod.rs
@@ -190,9 +190,16 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
         let state = self.clone();
         tokio::spawn(async move {
             let mut if_revive = false;
+            // Collect executor IDs that failed during launch so we can
+            // post `ExecutorLost` events for them below.
+            let mut failed_executor_ids: Vec<String> = vec![];
             match state.launch_tasks(schedulable_tasks).await {
                 Ok(unassigned_executor_slots) => {
                     if !unassigned_executor_slots.is_empty() {
+                        failed_executor_ids = unassigned_executor_slots
+                            .iter()
+                            .map(|(id, _)| id.clone())
+                            .collect();
                         if let Err(e) = state
                             .executor_manager
                             .unbind_tasks(unassigned_executor_slots)
@@ -208,6 +215,21 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
                     if_revive = true;
                 }
             }
+
+            for executor_id in failed_executor_ids {
+                if let Err(e) = sender
+                    .post_event(QueryStageSchedulerEvent::ExecutorLost(
+                        executor_id.clone(),
+                        Some(format!(
+                            "Executor {executor_id} removed after launch failure"
+                        )),
+                    ))
+                    .await
+                {
+                    error!("Fail to send ExecutorLost event for {executor_id}: {e:?}");
+                }
+            }
+
             if if_revive
                 && let Err(e) = sender
                     .post_event(QueryStageSchedulerEvent::ReviveOffers)


martin-g:
This is a duplicate/alternative of https://github.com/apache/datafusion-ballista/pull/2240

andygrove:
Thanks for tracking this down, the diagnosis matches what I found in #2240.

One thing to watch out for: `unassigned_executor_slots` is not quite the same set as "executors we removed". `launch_tasks` also pushes a slot when `get_executor_metadata` fails, and that branch never calls `remove_executor`, so the executor is still registered and probably still healthy. Posting `ExecutorLost` for it makes the handler run `reset_stages_on_lost_executor`, which rolls back completed stages whose shuffle output lives there and cancels its running tasks. A transient cluster-state read error would then throw away finished work, and because the rollback triggers `ReviveOffers` it can loop.

I think it is safer to post the event at the point where the executor is actually removed rather than inferring it from the slot list. That also covers the case where a launch task panics and `join_all` returns `Err` after the removal already happened.

Second thing: could you add a test? The failure is an intermittent hang so there is nothing keeping it fixed otherwise. I added `lose_executor_on_launch_failure` to the scheduler test harness in #2240 which makes this deterministic, feel free to take it.

There is some overlap between this and #2240, happy either way on which one we land.

martin-g:
Closing in favour of https://github.com/apache/datafusion-ballista/pull/2240

PR #2244, opened 2026-08-06
test(chaos): add a Kubernetes (kind) backend for the chaos harness

akshaychitneni:
Rationale

 The chaos harness (#2026) exercises Ballista's HA paths on a real multi-process cluster, but only as OS processes on one host. Several failure modes are k8s-specific — graceful SIGTERM/decommission, pod rescheduling, /readyz→Service gating, autoscaling — and can't be reached that way. This adds a Kubernetes backend so the same scenarios can run on a real kind cluster.
  
  What's included

  - K8sCluster backend (kubectl-driven): scheduler as a Deployment behind a ClusterIP Service, executors as a labelled Deployment; reaches the scheduler via kubectl port-forward.
  - One baseline scenario (tests/k8s.rs) asserting the cluster result matches local DataFusion — a full client → scheduler → executors → shuffle → result smoke test.

  Are there user-facing changes?

  No. Gated behind the k8s cargo feature + CHAOS_BACKEND=kind; the default cargo test and the process harness are unchanged.

  Key decisions

  - Executors as a Deployment, not bare pods — Ballista executors self-register with a fresh UUID and are fungible, so the Deployment model fits and enables scale/delete pod for kill scenarios.
  - Results via the scheduler's embedded flight proxy (advertise_flight_sql_endpoint) — the out-of-cluster client can't reach executor pod IPs; the proxy relays through the port-forwarded scheduler. Exposing executors directly was rejected: their advertised address is shared with in-cluster peers, so it can't be both pod-internal and
  host-reachable.
  - Gated + opt-in CI — separate job (paths + nightly), off the per-PR critical path, since the process harness already covers HA on every PR.

  Testing

  Baseline scenario passes end-to-end on a local kind cluster via dev/chaos-kind.sh; process harness and default build/clippy unaffected.

  Follow-ups

  Kill / graceful-decommission / total-loss (#2029) scenarios

### DIFF
--- a/.github/workflows/k8s-chaos.yml
+++ b/.github/workflows/k8s-chaos.yml
@@ -16,9 +16,9 @@
 # under the License.
 
 # Kubernetes (kind) chaos scenarios. Gated: runs only when the chaos harness,
-# the engine, or the docker artifacts change, plus nightly and on demand — it is
-# deliberately off the per-PR critical path, since the process-based chaos
-# harness already covers HA on every PR.
+# the engine, or the docker artifacts change, plus a nightly cron schedule and
+# on demand — it is deliberately off the per-PR critical path, since the
+# process-based chaos harness already covers HA on every PR.
 name: k8s-chaos
 
 on:
@@ -35,12 +35,25 @@ on:
       # docs-only changes cannot affect this job; exclusions are applied last
       - "!**/*.md"
   schedule:
+    # Nightly cron (not the Rust nightly toolchain — this job builds with stable).
     - cron: "0 6 * * *"
   workflow_dispatch:
 
 concurrency:
   group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
   cancel-in-progress: true
+
+# Minimal token: this job only reads the repository.
+permissions:
+  contents: read
+
+env:
+  # Pinned tool versions, installed via curl to avoid marketplace actions that
+  # must be on the ASF allowlist.
+  KIND_VERSION: v0.30.0
+  KUBECTL_VERSION: v1.31.4
+  CLUSTER_NAME: ballista-chaos
+  CHAOS_FIXTURE_DIR: /tmp/ballista-chaos-fixtures
 
 jobs:
   k8s-chaos:
@@ -66,31 +79,30 @@ jobs:
 
       - uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
 
-      - name: Create the fixture mount directory
-        run: mkdir -p /tmp/ballista-chaos-fixtures
+      - name: Install kind and kubectl
+        run: |
+          curl -fsSLo ./kind "https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-linux-amd64"
+          curl -fsSLo ./kubectl "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+          chmod +x ./kind ./kubectl
+          sudo mv ./kind ./kubectl /usr/local/bin/
+
+      - name: Create kind cluster
+        run: |
+          mkdir -p "${CHAOS_FIXTURE_DIR}"
+          kind create cluster --name "${CLUSTER_NAME}" --config chaos-testing/k8s/kind-config.yaml
 
       - name: Build the chaos image
         run: ./dev/build-chaos-docker.sh
 
-      - name: Install kubectl
-        uses: azure/setup-kubectl@v4
-
-      - name: Create kind cluster
-        uses: helm/kind-action@v1
-        with:
-          cluster_name: ballista-chaos
-          config: chaos-testing/k8s/kind-config.yaml
-
       - name: Load the chaos image into kind
-        run: kind load docker-image ballista-chaos:test --name ballista-chaos
+        run: kind load docker-image ballista-chaos:test --name "${CLUSTER_NAME}"
 
       - name: Run the kind chaos scenarios
         env:
           CHAOS_BACKEND: kind
-          # Match the extraMounts path in chaos-testing/k8s/kind-config.yaml. On
-          # Linux runners the kind node binds host paths directly, so /tmp is
-          # fine (the $HOME default is only needed for macOS Docker Desktop).
-          CHAOS_FIXTURE_DIR: /tmp/ballista-chaos-fixtures
+          # Keep namespaces on failure so the diagnostics step below can collect
+          # logs — teardown would otherwise delete them. The runner is ephemeral.
+          CHAOS_KEEP_NS: "1"
         run: |
           cargo test -p ballista-chaos --features k8s --test k8s -- --test-threads=1 --nocapture
 
--- a/.github/workflows/k8s-chaos.yml
+++ b/.github/workflows/k8s-chaos.yml
@@ -1,0 +1,104 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Kubernetes (kind) chaos scenarios. Gated: runs only when the chaos harness,
+# the engine, or the docker artifacts change, plus nightly and on demand — it is
+# deliberately off the per-PR critical path, since the process-based chaos
+# harness already covers HA on every PR.
+name: k8s-chaos
+
+on:
+  pull_request:
+    paths:
+      - "chaos-testing/**"
+      - "ballista/**"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "rust-toolchain.toml"
+      - "dev/docker/chaos.Dockerfile"
+      - "dev/build-chaos-docker.sh"
+      - ".github/workflows/k8s-chaos.yml"
+      # docs-only changes cannot affect this job; exclusions are applied last
+      - "!**/*.md"
+  schedule:
+    - cron: "0 6 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  k8s-chaos:
+    name: kind chaos scenarios
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7.0.1
+        with:
+          fetch-depth: 1
+
+      - name: Install protoc
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+
+      - name: Setup Rust toolchain
+        run: |
+          rustup update stable
+          rustup toolchain install stable
+          rustup default stable
+
+      - uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
+
+      - name: Create the fixture mount directory
+        run: mkdir -p /tmp/ballista-chaos-fixtures
+
+      - name: Build the chaos image
+        run: ./dev/build-chaos-docker.sh
+
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v4
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1
+        with:
+          cluster_name: ballista-chaos
+          config: chaos-testing/k8s/kind-config.yaml
+
+      - name: Load the chaos image into kind
+        run: kind load docker-image ballista-chaos:test --name ballista-chaos
+
+      - name: Run the kind chaos scenarios
+        env:
+          CHAOS_BACKEND: kind
+          # Match the extraMounts path in chaos-testing/k8s/kind-config.yaml. On
+          # Linux runners the kind node binds host paths directly, so /tmp is
+          # fine (the $HOME default is only needed for macOS Docker Desktop).
+          CHAOS_FIXTURE_DIR: /tmp/ballista-chaos-fixtures
+        run: |
+          cargo test -p ballista-chaos --features k8s --test k8s -- --test-threads=1 --nocapture
+
+      - name: Dump cluster state on failure
+        if: failure()
+        run: |
+          kubectl get pods -A -o wide || true
+          for ns in $(kubectl get ns -o name | grep chaos- || true); do
+            kubectl logs -n "${ns#namespace/}" -l app=ballista-scheduler --tail=-1 || true
+            kubectl logs -n "${ns#namespace/}" -l app=ballista-executor --all-containers --tail=-1 || true
+          done
--- a/chaos-testing/src/k8s.rs
+++ b/chaos-testing/src/k8s.rs
@@ -1,0 +1,526 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! A Kubernetes (kind) backend for the chaos harness.
+//!
+//! Where [`crate::cluster::TestCluster`] spawns the scheduler and executors as
+//! local OS processes, [`K8sCluster`] runs them as pods in a `kind` cluster: the
+//! scheduler as a Deployment behind a `ClusterIP` Service, and the executors as
+//! a labelled Deployment (so a later scenario can `scale`/`delete pod` them).
+//!
+//! The fixture is shared through a `hostPath` volume mounted into both pods.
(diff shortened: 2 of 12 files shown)


martin-g:
For some reason no CI checks have ran for this PR.

.github/workflows/k8s-chaos.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Install protoc
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+
+      - name: Setup Rust toolchain
+        run: |
+          rustup update stable
+          rustup toolchain install stable
+          rustup default stable
```
martin-g, line 78:
```suggestion
      - name: Setup Rust toolchain
        uses: ./.github/actions/setup-builder
```

chaos-testing/tests/k8s.rs
```diff
@@ -0,0 +1,113 @@
...
+//! Kubernetes (kind) backend scenarios.
+//!
+//! Gated behind the `k8s` feature and `CHAOS_BACKEND=kind`, so the default
+//! `cargo test` never touches a cluster. Run with the kind runbook in the crate
+//! README:
+//!
+//! ```sh
+//! dev/build-chaos-docker.sh
+//! kind create cluster --config chaos-testing/k8s/kind-config.yaml
+//! kind load docker-image ballista-chaos-scheduler:test ballista-chaos-executor:test
```
martin-g:
```suggestion
//! kind load docker-image ballista-chaos:test
```
there is just one Docker image

dev/chaos-kind.sh
```diff
@@ -0,0 +1,170 @@
...
+#
+# Commands:
+#   up     Create the kind cluster (if missing), build + load the chaos images.
+#   test   Run the kind chaos scenarios against the current cluster.
+#   all    up + test           (default)
+#   down   Delete the kind cluster.
+#
+# Env:
+#   CLUSTER_NAME   kind cluster name           (default: ballista-chaos)
+#   KEEP_CLUSTER   if "0", `all` deletes the cluster when done (default: 1, keep)
```
martin-g: What is `all` here ? An alternative value for `"0"` ?!

dev/chaos-kind.sh
```diff
@@ -0,0 +1,170 @@
...
+# Examples:
+#   dev/chaos-kind.sh                       # build, (re)create, load, run everything
+#   dev/chaos-kind.sh up                    # just stand the cluster up
+#   dev/chaos-kind.sh test -- --nocapture   # re-run tests on an existing cluster
+#   dev/chaos-kind.sh down                  # tear it down
+
+set -euo pipefail
+
+CLUSTER_NAME=${CLUSTER_NAME:-ballista-chaos}
+KEEP_CLUSTER=${KEEP_CLUSTER:-1}
```
martin-g, line 44: Should `CHAOS_CONCURRENT_TASKS` be exported too ?!

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+        // One cluster per process; --test-threads=1 keeps it to one at a time.
+        let namespace = format!("chaos-{}", std::process::id());
+        let shared_dir = PathBuf::from(fixture_dir());
+
+        // Ensure the shared dir exists. Do NOT remove/recreate it: it is the
+        // bind-mount root, and deleting it can sever the mount so pod writes no
+        // longer reach the node. `Fixture::write` overwrites the parquet in
+        // place, so a stale deterministic fixture is harmless.
+        std::fs::create_dir_all(&shared_dir)
+            .map_err(|e| format!("create shared dir {}: {e}", shared_dir.display()))?;
```
martin-g, line 122: Should sub-folders of `shared_dir` be deleted here ? E.g. there might be files from a previous run (e.g. in local/non-ephemeral execution).

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+    metadata:
+      labels:
+        app: ballista-executor
+    spec:
+      terminationGracePeriodSeconds: 30
+      containers:
+        - name: executor
+          image: {CHAOS_IMAGE}
+          imagePullPolicy: IfNotPresent
+          command: ["/root/chaos-executor"]
```
martin-g, line 579: It would be good to make use of `readinessProbe` and `livenessProbe` - https://github.com/akshaychitneni/datafusion-ballista/blob/15f18b6f7504c6fac5c164076293f5a5267b1a19/ballista/executor/src/health.rs#L101-L102

chaos-testing/tests/k8s.rs
```diff
@@ -0,0 +1,113 @@
...
+use ballista::prelude::{SessionConfigExt, SessionContextExt};
+use chaos_testing::fixture::Fixture;
+use chaos_testing::k8s::K8sCluster;
+use datafusion::arrow::util::pretty::pretty_format_batches;
+use datafusion::execution::session_state::SessionStateBuilder;
+use datafusion::prelude::{SessionConfig, SessionContext};
+
+/// The k8s scenarios need a running kind cluster with the chaos images loaded;
+/// they are opt-in via `CHAOS_BACKEND=kind` so a plain `cargo test` skips them.
+fn kind_backend_selected() -> bool {
```
martin-g, line 41: Is this really needed ? The whole test is feature gated to `k8s` anyway

dev/build-chaos-docker.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
martin-g:
```suggestion
#!/usr/bin/env bash
```

dev/chaos-kind.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
martin-g:
```suggestion
#!/usr/bin/env bash
```

.github/workflows/k8s-chaos.yml
```diff
@@ -0,0 +1,104 @@
...
+      - uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 #2.9.1
+
+      - name: Create the fixture mount directory
+        run: mkdir -p /tmp/ballista-chaos-fixtures
+
+      - name: Build the chaos image
+        run: ./dev/build-chaos-docker.sh
+
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v4
```
martin-g: Is this action needed ? helm/kind-action also installs kubectl

.github/workflows/k8s-chaos.yml
```diff
@@ -0,0 +1,104 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Kubernetes (kind) chaos scenarios. Gated: runs only when the chaos harness,
+# the engine, or the docker artifacts change, plus nightly and on demand — it is
```
martin-g: `nightly` ? Is this the Rust toolchain ? I see nothing about nightly below

chaos-testing/src/k8s.rs
martin-g: Same would be useful for the Scheduler too but I see no such endpoints in its routes.

.github/workflows/k8s-chaos.yml
```diff
@@ -0,0 +1,104 @@
...
+          # fine (the $HOME default is only needed for macOS Docker Desktop).
+          CHAOS_FIXTURE_DIR: /tmp/ballista-chaos-fixtures
+        run: |
+          cargo test -p ballista-chaos --features k8s --test k8s -- --test-threads=1 --nocapture
+
+      - name: Dump cluster state on failure
+        if: failure()
+        run: |
+          kubectl get pods -A -o wide || true
+          for ns in $(kubectl get ns -o name | grep chaos- || true); do
```
martin-g, line 113: https://github.com/apache/datafusion-ballista/pull/2244/changes#diff-9002dd159592b148879ea04fc805a1a1fc2bb46f8249664c27b1a87346d85405R303 deletes the namespace at cluster drop. I think this won't collect any logs.

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+                .map_err(|e| e.to_string())?
+                .json()
+                .await
+                .map_err(|e| e.to_string())?;
+        Ok(body.as_array().map(|a| a.len()).unwrap_or(0))
+    }
+
+    /// Scale the executor Deployment. `0` is a total loss that stays lost (the
+    /// controller does not recreate the pods); scaling back up recovers.
+    pub async fn scale_executors(&self, replicas: usize) -> Result<(), String> {
```
martin-g, line 252: This method is not used anywhere so far.

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+}
+
+impl K8sCluster {
+    /// Deploy a scheduler + `executors` executor pods, wait until all executors
+    /// have registered, and open a port-forward to the scheduler.
+    pub async fn start(executors: usize) -> Result<Self, String> {
+        require_kubectl()?;
+
+        // One cluster per process; --test-threads=1 keeps it to one at a time.
+        let namespace = format!("chaos-{}", std::process::id());
```
martin-g: This may fail due to racing if there are two `tokio::test`s due to `--test-threads=1`.

dev/docker/chaos.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+FROM rust:1-bookworm AS builder
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends protobuf-compiler \
+    && rm -rf /var/lib/apt/lists/*
+WORKDIR /src
+COPY . .
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/src/target \
+    cargo build -p ballista-chaos --bin chaos-scheduler --bin chaos-executor \
```
martin-g, line 44: This will not use Swattinem:rust-cache action and it will re-build all dependencies for every build of the Docker image.
You may want to do it like https://github.com/apache/datafusion-ballista/blob/main/dev/build-ballista-docker.sh#L24

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+        ])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()
+        .map_err(|e| format!("spawn kubectl port-forward: {e}"))
+}
+
+/// Run `kubectl` with the given args, returning stdout on success.
+async fn kubectl(args: &[&str]) -> Result<String, String> {
+    let output = tokio::process::Command::new("kubectl")
```
martin-g, line 418: Should this check that the K8S context is the one from Kind ?

chaos-testing/src/k8s.rs
```diff
@@ -0,0 +1,526 @@
...
+                Ok(out) => eprintln!("$ kubectl {}\n{out}", args.join(" ")),
+                Err(e) => eprintln!("$ kubectl {} -> {e}", args.join(" ")),
+            }
+        }
+    }
+
+    /// How many executors the scheduler currently considers registered.
+    pub async fn registered_executors(&self) -> Result<usize, String> {
+        let body: serde_json::Value =
+            reqwest::get(format!("{}/api/executors", self.rest_url()))
```
martin-g: It would be good to configure timeouts here.

andygrove:
This is good work, thanks for putting it together. RAII teardown on both `K8sCluster` and a separate `NamespaceGuard`, `mem::forget` to hand over ownership only once startup fully succeeds, deadline based polling instead of fixed sleeps, `free_port()` rather than a hardcoded port, a `CHAOS_KEEP_NS` escape hatch, and diagnostics dumped before teardown. The rejected alternatives note about executor pod IPs versus the flight proxy is exactly the kind of reasoning that makes a PR easy to review.

I checked the "no user facing changes" claim rather than taking it on faith, and it holds. `SchedulerConfig::default()` already sets `external_host: "localhost"`, so `env_or("CHAOS_EXTERNAL_HOST", "localhost")` preserves the process harness behavior exactly, and every other new env var falls back to the old hardcoded value.

## CI is blocked on ASF Infra, and I think it's worth routing around

`asf-allowlist-check` is failing because `helm/kind-action@v1` isn't on the ASF allowlist. The check prints the remediation, which is to open a PR against `apache/infrastructure-actions` adding it. That does work, but it's an external dependency with unbounded lead time, and we've been bitten by exactly this on the release workflow before.

The cheaper path is to not need the action at all. This workflow already installs protoc with `apt-get` and Rust with `rustup` in plain `run:` steps rather than reaching for actions, so installing kind the same way would be consistent with what's already there:

```yaml
- name: Install kind
  run: |
    curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64
    chmod +x ./kind && sudo mv ./kind /usr/local/bin/kind
    kind create cluster --name ballista-chaos --config chaos-testing/k8s/kind-config.yaml
```

That has the side benefit of pinning the kind version explicitly, which `@v1` doesn't. And `azure/setup-kubectl@v4` may be droppable outright, since I believe kubectl ships on the GitHub hosted ubuntu runner images, though that's worth confirming rather than taking from me.

## One thing that will bite the follow-ups

`delete_namespace` passes `--wait=false`, so teardown is fire and forget and the namespace sits in `Terminating` for a while afterwards. Meanwhile the namespace name is `format!("chaos-{}", std::process::id())`, which is constant for the life of the test binary.

With a single scenario that's fine. But #2029 adds kill, graceful decommission and total loss scenarios, and as soon as a second scenario in the same binary calls `K8sCluster::start()`, the `kubectl apply` lands in a namespace that's still terminating and fails with "unable to create new content in namespace ... because it is being terminated". An atomic counter or a random suffix on top of the PID would head that off now, while it's a one line change rather than a flaky test investigation later.

## Smaller things

- Action pinning is inconsistent with the rest of the repo. `Swatinem/rust-cache` is pinned to the same SHA we use elsewhere, which is great, but `actions/checkout@v7.0.1`, `azure/setup-kubectl@v4` and `helm/kind-action@v1` are floating tags where most workflows here pin to a full SHA. Also `swatinem` is lowercase here and `Swatinem` everywhere else, which is purely cosmetic but easy to fix while you're in there.
- `free_port()` has the usual bind then release race before `kubectl port-forward` claims the port. Standard practice and low risk, just noting it exists.
- The empty string sentinel for `advertise_flight_sql_endpoint` is genuinely confusing, since `Some(proxy) if proxy.is_empty()` at `scheduler_process.rs:103` is what actually enables the proxy. That's a pre existing Ballista wart rather than anything you introduced, and your inline comment explaining it is a real service to the next reader. Might be worth a separate issue to turn it into an explicit bool.

## Contract checks

No persona red flags. Nothing here touches `ballista-core`, the scheduler or the executor as shipped code, the `k8s` feature is off by default and carries no dependencies, and the test is gated behind both the feature and `--test k8s` so the default `cargo test` is untouched. No public API changes, and no plan shape or execution changes.

.github/workflows/k8s-chaos.yml
akshaychitneni: Good catch. It's the cron schedule, not the Rust nightly toolchain. The job builds with stable

.github/workflows/k8s-chaos.yml
akshaychitneni: Dropped both marketplace actions entirely - helm/kind-action@v1 and azure/setup-kubectl@v4 aren't on the ASF allowlist. kind and kubectl are now installed via pinned curl in run steps

.github/workflows/k8s-chaos.yml
akshaychitneni: this job would need to run on ubuntu vm as it using docker, kind

dev/chaos-kind.sh
akshaychitneni: Not needed here. CHAOS_CONCURRENT_TASKS is consumed by the executor pods via their manifest env, not by the harness process

chaos-testing/src/k8s.rs
akshaychitneni: Good point for local, non-ephemeral runs. Added clear_dir_contents() which empties the mount on start. It deletes the entries, not the directory itself

chaos-testing/tests/k8s.rs
akshaychitneni: This is needed as the CHAOS_BACKEND=kind gate controls whether we actually reach a cluster

chaos-testing/src/k8s.rs
akshaychitneni: Thanks. Included probes

.github/workflows/k8s-chaos.yml
akshaychitneni: The CI test step now sets CHAOS_KEEP_NS=1, and dump_diagnostics runs before teardown, so the "Dump cluster state on failure" step has pods/logs to collect.

chaos-testing/src/k8s.rs
akshaychitneni: it's the k8s primitive the executor kill/loss scenarios (the #2029 follow-ups) will drive; the baseline scenario only needs a healthy cluster. Added a doc comment saying so, so it doesn't read as dead code

chaos-testing/src/k8s.rs
akshaychitneni: Fixed. The name is now chaos-{pid}-{counter} via an atomic NS_SEQ, so each K8sCluster::start gets a distinct namespace

dev/docker/chaos.Dockerfile
akshaychitneni: I am using the approach of building inside image instead of copying a host binary to make sure chaos harness with kind also works on local machine (macbook)

chaos-testing/src/k8s.rs
akshaychitneni: Good idea. Added require_kubectl now reads kubectl config current-context and refuses to run unless it starts with kind-

chaos-testing/src/k8s.rs
akshaychitneni: the executor-count poll now uses a reqwest::Client with an explicit 5s timeout so a stalled port-forward surfaces as a retryable error in the polling loop

akshaychitneni:
Thanks for the reviews @martin-g @andygrove. I have addressed all comments. Created a followup issue https://github.com/apache/datafusion-ballista/issues/2281 for the flight endpoint config improvement. I will add more failure scenarios with kind cluster in separate prs

dev/chaos-kind.sh
akshaychitneni: all isn't a value but it is the script's default command (up/test/down/all), and KEEP_CLUSTER only has an effect during all

andygrove (approved):
Thanks @akshaychitneni. Adding k8s testing is really important, IMO. Let's merge and keep iterating on this. I filed https://github.com/apache/datafusion-ballista/issues/2302 for some suggested follow ons.

PR #2230, opened 2026-08-05
ci: add h2o window suite CI job

avantgardnerio:
## Summary

Adds the h2o `db-benchmark` runner and a first CI job that exercises the h2o window suite against a Ballista cluster. Scoped to be **defensible and running** — not the complete benchmark story.

- **Port `h2o.rs`** — runner ported from `arrow-datafusion/benchmarks/src/h2o.rs`, with `datafusion` and `ballista` subcommands sharing one query loader / table registration. Query files are copied verbatim from arrow-datafusion so indices align 1:1.
- **`--explain` flag** — prints the physical plan per query without executing; useful when iterating on planner rules against the window suite.
- **New CI job (`.github/workflows/h2o.yml`)** — single job (not a matrix), mirrors the TPC-H CI shape:
  - one scheduler + one executor, `tpch-ci` build profile, `--vcores 4 --memory-pool-size 2GB`
  - data generated via `uv run falsa join --size SMALL --data-format PARQUET` (same generator arrow-datafusion uses)
  - runs only the window suite; groupby/join are already covered by TPC-H
  - only the `AQE on, multi-partition tasks` config leg; can add others later
  - **--verify** - just like TPC-H

## Test plan

- [x] CI: workflow runs to green on this PR
- [x] Note wall-clock time for the h2o window job

PR #2251, opened 2026-08-07
chore(docker): ca-certificates + h2o binary in benchmarks image

avantgardnerio:
## Summary

Additive Docker image changes for S3-backed EKS benchmark runs, plus wiring the
`h2o` binary (added in #2230) into the benchmarks image.

- **`ballista-scheduler` + `ballista-executor`**: install `ca-certificates`.
  Without them, DataFusion's `object_store` S3 client fails TLS verification with
  `InvalidCertificate(UnknownIssuer)`, and STS `AssumeRoleWithWebIdentity`
  (IRSA-authenticated S3 reads on EKS) fails the same way. Ubuntu's minimal
  image doesn't populate `/etc/ssl/certs`.
- **`ballista-benchmarks`**: same `ca-certificates` rationale, plus
  `COPY target/release/h2o /root/h2o` so the image bundles both the tpch and
  h2o binaries. The h2o binary shipped in #2230 but was never wired into the
  Dockerfile.
- **`.dockerignore`**: whitelist `target/release{,-nonlto}/h2o` so the
  benchmarks build context sees the h2o binary alongside tpch/etc. Missing
  whitelist entry means the `COPY` line above errors out during build.

TPC-H bench workflow is unaffected — every change is additive.

## Test plan

- [x] `docker build -f dev/docker/ballista-scheduler.Dockerfile .` succeeds
- [x] `docker build -f dev/docker/ballista-executor.Dockerfile .` succeeds
- [x] `docker build -f dev/docker/ballista-benchmarks.Dockerfile .` succeeds
- [x] `docker run --rm <benchmarks-image> ls /root/h2o` prints the binary path
- [x] `docker run --rm <executor-image> ls /etc/ssl/certs | wc -l` > 0
- [x] EKS run: h2o Q1 reads from an `s3://...` path without TLS errors

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,7 +13,9 @@ CHANGELOG.md
 !target/release/ballista-executor
 !target/release/ballista-cli
 !target/release/tpch
+!target/release/h2o
 !target/release-nonlto/ballista-scheduler
 !target/release-nonlto/ballista-executor
 !target/release-nonlto/ballista-cli
 !target/release-nonlto/tpch
+!target/release-nonlto/h2o
--- a/dev/docker/ballista-benchmarks.Dockerfile
+++ b/dev/docker/ballista-benchmarks.Dockerfile
@@ -30,6 +30,7 @@ ENV RUST_BACKTRACE=full
 COPY target/${RELEASE_FLAG}/ballista-scheduler /root/ballista-scheduler
 COPY target/${RELEASE_FLAG}/ballista-executor /root/ballista-executor
 COPY target/${RELEASE_FLAG}/tpch /root/tpch
+COPY target/${RELEASE_FLAG}/h2o /root/h2o
 
 COPY benchmarks/run.sh /root/run.sh
 COPY benchmarks/queries/ /root/benchmarks/queries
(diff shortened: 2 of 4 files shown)


milenkovicm (approved):
LGTM, thanks @avantgardnerio

avantgardnerio:
> LGTM, thanks @avantgardnerio

Damn, you're fast today :smile:

milenkovicm:
I want to be seen as helpful while avoiding other major PR changes that require deep engagement and focus. 😀
I finally got some time to spend on ballista, hope to help more, and im trying to catch up with all the changes i've missed

avantgardnerio:
> I want to be seen as helpful while avoiding other major PR changes that require deep engagement and focus. 😀 I finally got some time to spend on ballista, hope to help more, and im trying to catch up with all the changes i've missed

Well, [this is the big one](https://github.com/apache/datafusion-ballista/pull/2223) - it gets parallel windows working end-to-end with decent benchmark results.

So I'll just keep extracting smaller ones out of there, as long as you're willing to review, I can keep you busy :smile: . Not required, but very much appreciated!

milenkovicm:
will have a look over the weekend hopefully

PR #2252, opened 2026-08-07
feat(python): integrate Ballista through DataFusion FFI

timsaucer:
## Summary

- expose Ballista logical and physical extension codecs and an FFI query planner to datafusion-python
- add a reusable `BallistaExtension` bundle that installs its components atomically with `SessionContext.with_extensions`
- pin the Python and Rust dependencies to the latest commit from apache/datafusion-python#1679
- update the getting-started notebook to use the new extension API
- preserve context ownership semantics and cover atomic installation and context lifetime with tests

## Testing

- `cargo check`
- `cargo fmt -- --check`
- `uv run pytest python/tests/test_context.py -q` (21 passed)


python/examples/getting_started.ipynb
```diff
@@ -77,10 +82,25 @@
     "# ctx = BallistaSessionContext(\"df://your-scheduler:50050\")\n",
     "\n",
     "host, port = setup_test_cluster()\n",
-    "ctx = BallistaSessionContext(f\"df://{host}:{port}\")\n",
+    "address = f\"df://{host}:{port}\"\n",
+    "\n",
+    "ballista_config = ballista_datafusion_config_defaults()\n",
+    "ctx = SessionContext(SessionConfig(ballista_config)).with_extensions(\n",
```
milenkovicm, line 88: Can we use ballista context to hide next few lines and simplify session context creation ?

python/python/ballista/extension.py
milenkovicm, line 51: Can we remove this once we get FFI planner ?

python/python/ballista/extension.py
```diff
@@ -66,7 +68,12 @@ def method_wrapper(*args, **kwargs):
                 address = args[0].address
                 session_id = args[0].session_id
                 df = func(*args, **kwargs)
-                return DistributedDataFrame(df, session_id, address)
+                return DistributedDataFrame(
```
milenkovicm, line 71: DistributedDataFrame, should be removed with FFI, should it ?

ballista/core/src/serde/mod.rs
```diff
@@ -349,22 +358,98 @@ impl LogicalExtensionCodec for BallistaLogicalExtensionCodec {
             .encode(buf)
             .map_err(|e| DataFusionError::Internal(e.to_string()))
     }
+
+    fn try_decode_udf(
+        &self,
+        name: &str,
+        buf: &[u8],
+    ) -> Result<Arc<datafusion::logical_expr::ScalarUDF>> {
+        self.default_codec.try_decode_udf(name, buf)
```
milenkovicm, line 367: Do we need to do something extra to get python udf serde?

milenkovicm (commented):
After many years! Thanks @timsaucer awesome as always!

python/examples/getting_started.ipynb
```diff
@@ -77,10 +82,25 @@
     "# ctx = BallistaSessionContext(\"df://your-scheduler:50050\")\n",
```
martin-g, line 82: This should be removed too, right ?
```suggestion
```

milenkovicm (commented):
We might have an issue, I've continued to work on top of this PR, removing `DistributedDataFrame`. 

There are few test failures, but the one that might be the show stopper is 

```
FAILED python/tests/test_context.py::test_write_csv - Exception: DataFusion error: FFI error: This feature is not implemented: FFI does not support decode_file_format
FAILED python/tests/test_context.py::test_write_parquet - Exception: DataFusion error: FFI error: This feature is not implemented: FFI does not support decode_file_format
FAILED python/tests/test_context.py::test_write_json - Exception: DataFusion error: FFI error: This feature is not implemented: FFI does not support decode_file_format
```

looking at datafusion code `encode|decode_file_format` is not implemented 

if we ignore test failing for features added to distributed data frame, only df.write_* methods are affected with this issue

python/examples/getting_started.ipynb
milenkovicm: we can

python/python/ballista/extension.py
milenkovicm: we can remove it, tested

milenkovicm:
`FileFormatFactory` looks like a rabbit hole for FFI at least for me

milenkovicm:
proposal for workaround in https://github.com/apache/datafusion-python/pull/1679#issuecomment-5375229711

PR #2256, opened 2026-08-08
refactor(scheduler): extract REST API wire types into ballista-api-types

andygrove:

Part of #1923. This is the first of four PRs splitting up #1925, which is too large to review as one change.

# Rationale for this change

Ballista's live TUI reads the scheduler's REST API, but that state is ephemeral: completed jobs are cleaned up after `finished_job_state_clean_up_interval_seconds`, and everything is gone when the scheduler restarts. #1923 asks for a Spark History Server equivalent, where a standalone server replays durable event logs and serves the same `/api/*` responses so the existing TUI can browse completed jobs with no scheduler running.

The design is write once, replay verbatim. When a job finishes, the scheduler builds the REST responses once against the live execution graph and writes them into the event log's terminal record; the history server deserializes those values and re-serializes them unchanged. Byte-identical output is therefore a structural property, not something two implementations have to keep agreeing on.

Two things have to be true before any of that can be built, and both are refactors rather than features:

1. The response types have to live somewhere every party to the `/api/*` contract can reach. Today they are private types inside `ballista-scheduler`.
2. Response construction has to be callable outside an axum handler, since the event-log writer runs at job completion rather than on an HTTP request. Today it is inlined in the handler bodies.

This PR does both, and nothing else. No new feature, no new config, no behavior change. Landing it separately keeps the reviewable question down to one thing: is this behavior preserving?

## Why `ballista-api-types` and not `ballista-history`

The crate started out named for the history server, but the `/api/*` contract has three parties, not one. The scheduler serves it, the history server will serve replayed copies of it, and **the web TUI already consumes it** — with its own hand-maintained duplicate of these structs in `ballista-cli/src/tui/domain/jobs*.rs`.

That duplicate has already drifted: `2ed3464df` changed `TaskSummary::partition_id` from `u32` to `Vec<u32>` and the TUI copy was not updated, so it still declares `u32`. See #2257. Naming the crate after the contract rather than after one of its consumers is what makes it natural for the TUI to drop its copy, which is the follow-up in #2258.

The crate stays serde-only, which is what lets the TUI depend on it from a `wasm32` build.

**New `ballista-api-types` crate.** A leaf crate whose only dependency is `serde`. It is the single definition of what `/api/*` puts on the wire: `JobResponse`, `TaskSummary`, `TaskStatus`, `Percentiles`, `QueryStageSummary`, `QueryStagesResponse`, and `PlanFormat`. The crate doc names the three consumers and the rule for what is admitted.

Types describing live scheduler internals (`SchedulerStateResponse`, `SchedulerVersionResponse`, `CancelJobResponse`) stay in `handlers.rs`; no other party needs them. `ExecutorResponse` stays too, for a different reason worth being explicit about: it embeds `ballista-core` types, and keeping this crate serde-only is what makes it cheap to depend on from a `wasm32` build.

**New `api::dto_build` module.** The graph-to-DTO translation moves out of the handler bodies into pure functions: `job_overview_to_response`, `graph_to_job_response`, and `graph_to_query_stages`, plus the formatting and percentile helpers they use. State in, DTO out, no I/O and no wall-clock reads. `handlers.rs` drops from 1427 lines to 768 and is now mostly HTTP concerns.

These builders stay in `ballista-scheduler` on purpose. They need the live execution graph, and nothing on the replay path calls them, so there is no reason to push them into the shared crate.

**Some consolidation the extraction made obvious.** The three `ExecutionStage` arms in `get_query_stages` were near-identical; they collapse into one destructuring match, which also removes the mutable placeholder-zero summary. `render_stage_plan`, `task_summaries`, `percentiles_of`, `percent_complete`, `min_start_time`, and `format_millis` each replace two or three copies of the same code.

**Two things pulled out of the handler layer.** `PlanFormat` is part of the wire contract, so it moved to `ballista-api-types`, and the builders take it by value rather than taking the axum `JobQueryParams` extractor. And `graph_to_query_stages` takes `now` from its caller instead of reading the clock, so replaying a stored log renders stable elapsed times rather than ones that grow on every request.

**One type change.** `JobResponse::job_id` is a `String` rather than `ballista_core::JobId`, so the new crate can stay serde-only. `JobId` is `#[serde(transparent)]` over `String`, so the serialized JSON is identical.

**`impl From<&task_status::Status> for TaskStatus` becomes the free function `task_status_to_dto`.** Both types are now foreign to `ballista-scheduler`, so the impl would violate the orphan rule.

**Release tooling.** `ballista-api-types` is registered in `dev/update_ballista_versions.py` (both the crate list and the inter-crate dependency list), in the publish order in `dev/release/README.md`, and in `dev/release/crate-deps.dot`. Without this the next version bump would leave the crate behind while `ballista-scheduler` pinned the old version.

The 12 helper unit tests move to `dto_build.rs` alongside the functions they cover. The 8 `get_webtui` tests stay in `handlers.rs`.

No. REST responses are byte-identical, no public API of `ballista-scheduler` changes, and no configuration is added. `ballista-api-types` is new but nothing outside the scheduler depends on it yet.

Verified locally: `cargo test -p ballista-scheduler --lib` passes (324 tests), clippy is clean for `ballista-scheduler` and `ballista-api-types` with `--all-features -D warnings`, `cargo doc` is clean, `cargo fmt --all --check` and taplo are clean, and the `--no-default-features` check CI runs still passes.

# Follow-ups

The remaining three slices of #1925, in order:

2. Event schema, writer, and reader in a new `ballista-history` crate depending on this one (no scheduler changes).
3. Scheduler event-log wiring behind a new `event_log_dir` config, off by default.
4. The history server itself, its binary, docs, and the byte-identical-JSON end-to-end test.

One design question that belongs to step 3 rather than here, but is worth flagging early: because responses are rendered once at write time, the stored record carries a single plan format. A history server cannot honour `?plan_format=tree` unless step 3 stores every rendering, stores the plan and renders at read time, or the server simply ignores the parameter. The crate doc notes the constraint.

# Open question for reviewers

`ballista-api-types` does not set `publish = false`, so it would be published on the next release. Given the TUI will depend on it that is probably right, but it is a permanent commitment and worth deciding deliberately rather than by omission. The release tooling above assumes it is published; if we would rather it were not, that is a one-line change here plus dropping it back out of the publish order.


andygrove:
cc @phillipleblanc

phillipleblanc (approved):
Makes sense to me. I verified that 'ballista-api-types' isn't taken on crates.io

milenkovicm (approved):
Thanks @andygrove make sense to me

And thank you for reminding me of good old times of JEE, 3 layered architecture, Gang of Four book ... with Data Travel Objects naming 😀😂

PR #2258, opened 2026-08-08
fix(tui): use the shared API wire types instead of local copies

### LINKED ISSUE #2257: TUI's duplicated API response types have drifted from the scheduler, breaking the stages popup
## Describe the bug

The web TUI keeps its own hand-written copies of the scheduler's `/api/*` response types, and they have drifted from what the scheduler actually serves.

`ballista-cli/src/tui/domain/jobs/stages.rs:59` declares:

```rust
pub struct StageTaskResponse {
    pub id: usize,
    pub status: StageTaskStatus,
    pub partition_id: u32,
    ...
}
```

but the scheduler has served an array there since 2ed3464df ("Multi-partition tasks: partition_slice plumbing + SortShuffleWriter refactor", #2038, 2026-07-19), which changed `TaskSummary::partition_id` from `u32` to `Vec<u32>`. That commit touched `ballista-cli/src/main.rs` but not the TUI's copy of the type.

The field has no `#[serde(default)]`, so deserializing `[0]` into `u32` should fail the whole `JobStageResponse`, taking the stage detail popup with it.

A second, more benign drift in the same file: `StageTaskStatus::Failed { reason }` vs the scheduler's `TaskStatus::Failed { reason, error }`. Serde ignores the extra field, so this one just means the TUI cannot show the error text.

## To Reproduce

Run a job against a current scheduler and open the stages popup for it in the TUI.

## Expected behavior

The stage detail popup renders, showing per-task partitions.

## Additional context

This is a structural problem rather than a one-off. There are currently three independent declarations of the same wire contract:

| TUI type (`ballista-cli/src/tui/domain/jobs*.rs`) | Scheduler type |
| --- | --- |
| `Job` | `JobResponse` |
| `JobStagesResponse` | `QueryStagesResponse` |
| `JobStageResponse` | `QueryStageSummary` |
| `StageTaskResponse` | `TaskSummary` |
| `StageTaskStatus` | `TaskStatus` |
| `TaskPercentiles` | `Percentiles` |

Nothing makes them move together, so a scheduler-side field change compiles and ships while the TUI silently stops parsing.

The fix is for the TUI to depend on the shared wire-type crate rather than redeclare it. #2256 extracts those types into `ballista-api-types`, a serde-only leaf crate that a `wasm32` build can depend on. A follow-up stacked on it deletes the TUI's copies and imports the shared types instead, so this class of drift becomes a compile error.

andygrove:

Closes #2257.

# Rationale for this change

The TUI declared its own copies of the scheduler's `/api/*` response types. Nothing kept the two sets in step, and they have drifted: `2ed3464df` (multi-partition tasks, #2038) changed `TaskSummary::partition_id` from `u32` to `Vec<u32>` and updated `ballista-cli/src/main.rs` but not the TUI's copy, which still declared `u32`. The field has no `#[serde(default)]`, so the whole `JobStageResponse` fails to deserialize and the stage detail popup breaks.

A second, quieter drift in the same file: `StageTaskStatus::Failed { reason }` against the scheduler's `TaskStatus::Failed { reason, error }`. Serde ignores the extra field, so the TUI simply could not show the error text.

The fix is not to re-sync the copies, it is to stop having copies. #2256 extracts the wire types into `ballista-api-types`, a serde-only leaf crate. This PR points the TUI at them, which turns this whole class of drift into a compile error.

`ballista-cli` takes a dependency on `ballista-api-types` under both the `tui` and `web` features, and the local declarations are replaced with re-exports of the shared types under the names the TUI already used:

| Was (local) | Now (shared) |
| --- | --- |
| `Job` | `JobResponse` |
| `JobStagesResponse` | `QueryStagesResponse` |
| `JobStageResponse` | `QueryStageSummary` |
| `StageTaskResponse` | `TaskSummary` |
| `StageTaskStatus` | `TaskStatus` |
| `TaskPercentiles` | `Percentiles` |

`Job`'s four status predicates (`is_queued`, `is_running`, `is_completed`, `is_failed`) become a `JobStatusExt` extension trait, since the type is now defined in another crate and `ballista-api-types` should not carry TUI concerns.

The renames the shared names imply (`stage.id` to `stage.stage_id`, and so on) account for most of the line count. Three changes are more than mechanical:

- **`partition_id` renders as a list.** Single-partition tasks look the same as before; multi-partition tasks now show every partition they own instead of failing to parse. This is the actual bug fix.
- **Failed tasks can show the error.** The status arm destructures `Failed { reason, .. }` against the richer shared enum.
- **`stage_plan` is `Option<String>`** rather than a `#[serde(default)]` `String`, so the plan popup renders an empty plan as empty rather than relying on the default.

`start_time` / `end_time` are `u64` on the shared type where the TUI had `i64`; the duration calculation uses `saturating_sub` accordingly, which also removes an underflow if a job ever reports `end_time < start_time`.

Two regression tests are added, both against payloads shaped like real scheduler output: one deserializes a task with `"partition_id": [0, 1, 2]`, and one deserializes `Failed` with both `reason` and `error`.

Yes, all fixes:

- The stage detail popup works again against a current scheduler.
- Multi-partition tasks display all their partitions instead of breaking the response.
- Failed tasks can surface the scheduler's error text.

No API changes. `ballista-cli` gains an internal dependency, and `ballista-api-types` is serde-only so the `wasm32` web build is unaffected.

Verified locally: `cargo test -p ballista-cli --no-default-features --features cli,tui` passes (247 tests, including the two new ones), the `web` feature builds, and clippy is clean for both feature sets with `-D warnings`.


ballista-cli/src/tui/ui/main/jobs/mod.rs
```diff
@@ -359,14 +360,14 @@ fn render_jobs_table(
 }
 
 fn render_job_start_time_cell<'a>(job: &'a Job, app: &App) -> Cell<'a> {
-    let start_time = app.format_datetime(job.start_time);
+    let start_time = app.format_datetime(job.start_time as i64);
```
martin-g:
```suggestion
    let start_time = i64::try_from(job.start_time)
        .map_or_else(|_| INVALID_DATE.to_string(), |time| app.format_datetime(time));
```

ballista-cli/src/tui/ui/main/jobs/stage_tasks_popup.rs
```diff
@@ -106,7 +106,7 @@ fn build_stage_task_row(i: usize, task: &StageTaskResponse, app: &App) -> Row<'s
         StageTaskStatus::Successful => {
             Text::from("Successful").style(app.theme.status_completed)
         }
-        StageTaskStatus::Failed { reason } => {
+        StageTaskStatus::Failed { reason, .. } => {
```
martin-g: Do we want to show the `error` too/instead ? Or it may contain sensitive information ?
The `reason` is just the category.
https://github.com/apache/datafusion-ballista/blob/f485adb91781be1d357bfae1d1796fa8c6842c14/ballista/api-types/src/dto.rs#L66-L71

ballista-cli/Cargo.toml
```diff
@@ -34,6 +34,7 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 # CLI-only deps (DataFusion/Ballista) — gated by `cli` feature
 ballista = { path = "../ballista/client", version = "54.0.0", features = ["standalone"], optional = true }
+ballista-api-types = { path = "../ballista/api-types", version = "54.0.0", optional = true }
```
martin-g:
```suggestion
```

ballista-cli/Cargo.toml
martin-g, line 42:
```suggestion
# TUI/web shared deps
ballista-api-types = { path = "../ballista/api-types", version = "54.0.0", optional = true }
```
The dependency is used only by TUI/Web, not by the CLI.

Maybe it should be declared in the workspace Cargo.toml, for easier management with the rest of the Ballista crates ?!

martin-g:
@andygrove Let me fix the build since I suggested the improvement that broke it!

martin-g:
@andygrove May I take over here ?

martin-g (approved):

milenkovicm:
thanks @andygrove & @martin-g

PR #2260, opened 2026-08-08
feat(history): add the event-log schema, writer, and reader

andygrove:

Part of #1923. Second of four PRs splitting up #1925.

# Rationale for this change

#1923 asks for a Spark-History-Server equivalent: the scheduler records a durable log of each job, and a standalone server replays those logs and serves the same `/api/*` responses so the existing TUI can browse completed jobs with no scheduler running.

This PR adds the format and the machinery to read and write it, and nothing else. Nothing in the scheduler calls it yet, so it can be reviewed purely on its own terms: is the on-disk schema right, and does the writer behave under load?

Those are the two questions worth isolating. The schema is durable and cross-machine, so it is the part that is expensive to change later. And the writer sits next to the scheduler's hot path, so its backpressure behaviour matters more than its line count suggests.

**New `ballista-history` crate**, depending on `ballista-api-types` plus serde, tokio and log.

`event.rs` — the versioned JSONL schema. Every line is a `LogRecord`: a self-describing envelope carrying `ev`, `version`, and an opaque `data` payload. `JobStart`, `StageStart`, `StageEnd` and `TaskEnd` form an incremental timeline; the terminal `JobEnd` carries the finished `/api/*` responses.

Storing the built responses is what makes replay trustworthy: the history server re-serves exactly what the scheduler produced rather than re-deriving it, so there is one definition and one place that populates it instead of two implementations that have to keep agreeing. The consequence, worth knowing up front, is that anything not captured at write time cannot be recovered later.

Identifiers are fixed-width `u32` rather than `usize`, since a log written by a 64-bit scheduler has to mean the same thing to any reader.

`writer.rs` — an async buffered `EventLogWriter`. All file I/O happens on a background task, so the scheduler's event loop never waits on disk. One append-mode handle per job is held for the life of the process.

The backpressure split is deliberate. `append` never blocks and drops the event if the queue is full, because losing a progress record is better than stalling scheduling. `append_final` instead waits for capacity, because a job whose `JobEnd` was dropped is invisible to the history server entirely. `finish_job` then flushes and closes the handle, ordered after the terminal event by the single-consumer channel.

`reader.rs` — folds a completed log back into the payload a server would serve. A file is "completed" once it contains a `JobEnd`. Unrecognised lines are skipped rather than treated as fatal, so a log truncated by a crash still yields its job if the terminal record survived, and a future scheduler can add record kinds without breaking today's reader.

## Compatibility

A log is written once and may be read years later by a much newer Ballista, so the guarantee runs one way: **a reader accepts any log whose version is not newer than its own**. There is no way to upgrade the writer of a file that already exists.

That is deliberately the opposite of `BALLISTA_PROTOCOL_VERSION`, the scheduler/executor handshake, which is strict equality and rejects on mismatch. Both ends of that handshake are live and upgraded together, so refusing to proceed is the safe move. Here there is nothing to negotiate with, and refusing means losing the job.

Four things make that guarantee real rather than aspirational:

**The envelope.** Routing on `ev` and `version` before parsing `data` is what lets a reader tell "written by a newer Ballista" apart from "corrupt". Parsing the whole record up front collapses both into the same indistinguishable failure. Previously only `JobStart` and `JobEnd` carried a version at all, and nothing ever read it.

**The stored responses are opaque.** `JobEnd` holds them as raw JSON rather than typed structs, alongside a small frozen `JobIndex` the history server uses to list and sort. This matters because those responses are `ballista-api-types` shapes, which evolve with the live REST contract: `TaskSummary::partition_id` went from `u32` to `Vec<u32>` and `TaskStatus::Failed` gained a field, both within a single release cycle. Stored typed, either change would have made every older log fail to deserialize, and the job would have vanished from the UI. Stored raw, nothing ever parses the inner shape and the log is immune to REST churn. It also makes replayed output byte-identical rather than merely equivalent, which a typed round trip cannot promise — `serde_json::Value` sorts object keys.

**Unreadable is no longer indistinguishable from absent.** `read_completed_job` used to return `Ok(None)` both when a log had no `JobEnd` and when it had one that could not be parsed, and `HistoryStore::load` treated that as "still running" and logged nothing. A version-incompatible log meant a job silently disappearing. It now returns a `ReadError` separating an unsupported version from a malformed record.

**A frozen golden log.** `testdata/schema-v1.eventlog` is checked in and replayed by CI on every build. It deliberately contains a record kind this build has never heard of, and a stages payload carrying the multi-partition `partition_id` array, so it exercises both forward-compatibility paths. It is never regenerated: the point is that it was written by an older Ballista and must stay readable.

I verified it actually fires rather than just existing. Renaming a field in a stored type — a change that compiles cleanly everywhere — fails all four compatibility tests with `Malformed("missing field \`name\`")`.

**`JobConfig` returns to `ballista-api-types`.** I removed it in #2256 as dead; the `JobEnd` record is its real consumer.

**`TaskEnd` names a task, not a partition.** Under the multi-partition task model a task owns a slice of partitions, and `TaskStatus` carries `task_id` rather than `partition_id`. #1925 predates that change and recorded a `partition` field that no longer has a source.

**Release tooling.** `ballista-history` is registered in `dev/update_ballista_versions.py`, the publish order, and the crate dependency graph.

No. The crate is new and nothing depends on it yet. No scheduler code changes, no configuration, no API changes.

Verified locally: `cargo test -p ballista-history` passes (14 tests, 4 of them the golden-log compatibility suite), `cargo check -p ballista-scheduler` is unaffected, clippy is clean for both crates with `--all-features -D warnings`, and fmt/taplo are clean.

# Follow-ups

3. Scheduler event-log wiring behind a new `--event-log-dir` flag, off by default.
4. The history server binary, its docs, and the byte-identical-JSON end-to-end test.

Both are already written and forward-ported; they are waiting on this landing rather than on being figured out.

Separately, a written cross-version compatibility policy covering all three of Ballista's version-sensitive surfaces (the executor handshake, the REST API, and this log) is worth its own issue. Two of the three now have a mechanism; the REST API has none, which is what #2257 was.


milenkovicm (approved):
Thanks @andygrove

andygrove:
Thanks for the review @milenkovicm

PR #2262, opened 2026-08-08
feat(core): RangeFilterExec — sorted-input per-partition range filter (purely additive)

avantgardnerio:
## Summary

Extracted from #2223 (working parallel-windows acceleration PR).

**Purely additive.** Adds `RangeFilterExec` alongside `PerPartitionFilterExec`. No in-tree callers, no mutations. Follow-up PRs wire it into the parallel-window path.

`RangeFilterExec` is the sorted-input specialisation of `PerPartitionFilterExec`:
- Applies a half-open `[lo, hi)` predicate per input partition, widened by a lower/upper halo (for `RANGE`-frame boundary context in bounded window ops).
- Fast path when the input advertises ordering on `routing_expr` ascending: batch min/max bound the range, so batches entirely inside/outside the window skip `filter_record_batch` — `RecordBatch::slice` (zero-copy) or `Arc`-clone pass-through.
- Falls back to `filter_record_batch` when the input isn't sorted or has nulls in the routing column.

## Design notes

- Bounds are late-bound via `resolve_bounds()`. Wire encode refuses before resolution — over-the-wire plans always carry resolved bounds. Two constructors: `try_new_pending` (rule path — bounds resolved later by the scheduler) and `try_new_resolved` (wire decode + task-restriction path).
- **Primitive-generic APIs**: halos and bounds use `ScalarValue` at the API + proto surface so future KLL migration to non-Float64 numeric primitives doesn't break callers. Internal impl is Float64-only today; errors loudly on other numeric types.
- Scope: RFE is a pure per-partition filter. Range-partitioning concerns (cuts → per-partition ranges, task-slice mapping) belong to the scheduler in follow-up work; not in this PR.

## Test plan
- [x] `cargo test -p ballista-core --lib -- range_filter` — 22 tests (fast/slow paths, halo widening, sorted-key detection, null handling, pending/resolved constructors, resolve validation)
- [x] Serde roundtrip + pending-refuses-to-serialize tests
- [x] `cargo test --workspace` — 1200 pass / 0 fail / 8 ignored
- [x] `cargo clippy --all-targets` clean
- [x] `cargo fmt --all --check` clean
- [x] `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps` clean

### DIFF
--- a/ballista/core/src/execution_plans/mod.rs
+++ b/ballista/core/src/execution_plans/mod.rs
@@ -48,7 +48,7 @@ pub use ordered_range_repartition::OrderedRangeRepartitionExec;
 pub use partitioned_bounded_window_agg::PartitionedBoundedWindowAggExec;
 pub use per_partition_filter::{PerPartitionFilterExec, range_partition_predicates};
 pub use plan_algebra::{preserves_distribution, preserves_partitioning};
-pub use range_filter::RangeFilterExec;
+pub use range_filter::{RangeBound, RangeFilterExec, WidenedBound};
 pub use range_shuffle_reader::RangeShuffleReaderExec;
 pub use runtime_stats::{
     MergedRuntimeStats, RuntimeStatsExec, TaskRuntimeStats,
--- a/ballista/core/src/execution_plans/mod.rs
+++ b/ballista/core/src/execution_plans/mod.rs
@@ -26,6 +26,7 @@ mod ordered_range_repartition;
 mod partitioned_bounded_window_agg;
 mod per_partition_filter;
 pub mod plan_algebra;
+mod range_filter;
 mod range_repartition_common;
 mod range_shuffle_reader;
 mod runtime_stats;
@@ -47,6 +48,7 @@ pub use ordered_range_repartition::OrderedRangeRepartitionExec;
 pub use partitioned_bounded_window_agg::PartitionedBoundedWindowAggExec;
 pub use per_partition_filter::{PerPartitionFilterExec, range_partition_predicates};
 pub use plan_algebra::{preserves_distribution, preserves_partitioning};
+pub use range_filter::RangeFilterExec;
 pub use range_shuffle_reader::RangeShuffleReaderExec;
 pub use runtime_stats::{
     MergedRuntimeStats, RuntimeStatsExec, TaskRuntimeStats,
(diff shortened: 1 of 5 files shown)


avantgardnerio:
@phillipleblanc FYI

ballista/core/src/execution_plans/mod.rs
```diff
@@ -47,6 +48,7 @@ pub use ordered_range_repartition::OrderedRangeRepartitionExec;
 pub use partitioned_bounded_window_agg::PartitionedBoundedWindowAggExec;
 pub use per_partition_filter::{PerPartitionFilterExec, range_partition_predicates};
 pub use plan_algebra::{preserves_distribution, preserves_partitioning};
+pub use range_filter::RangeFilterExec;
```
andygrove: Small composability nit. `RangeBound` is `pub` inside `range_filter`, but `range_filter` itself is a private module and only `RangeFilterExec` is re-exported here, so downstream crates can't name the type. It still compiles for them because a type alias is transparent and they can spell out `Vec<(Option<ScalarValue>, Option<ScalarValue>)>`, but rustdoc renders an unlinkable name in the `try_new_resolved` and `raw_bounds` signatures, which is not a great first impression for anyone building on the operator.

One line fixes it:

```rust
pub use range_filter::{RangeBound, RangeFilterExec};
```

ballista/core/src/execution_plans/range_filter.rs
```diff
@@ -0,0 +1,1095 @@
...
+    /// [`Self::widened_bounds`] or expand `raw_bounds[k] ± (halo_lo, halo_hi)`
+    /// themselves.
+    pub fn raw_bounds(&self) -> Option<Vec<RangeBound>> {
+        self.bounds.lock().as_ref().map(|s| s.raw.clone())
+    }
+
+    /// Snapshot the halo-widened bounds — what the runtime filter actually
+    /// applies. Convenience for callers that already have the halos and just
+    /// want the resolved form.
+    pub fn widened_bounds(&self) -> Option<Vec<WidenedBound>> {
```
andygrove, line 243: Same idea as the `RangeBound` note over in `mod.rs`. `widened_bounds` is public but `WidenedBound` on line 93 is a private alias, so the rendered signature reads as a type nobody outside the crate can refer to. Either make the alias `pub` and export it alongside `RangeBound`, or just spell the tuple out in the return type here.

Not a blocker either way. Easier to sort out now while the operator has no callers than after it ships.

andygrove (approved):
Thanks @avantgardnerio. LGTM with a couple of nits.

phillipleblanc (approved):

PR #2265, opened 2026-08-09
feat(scheduler): add the history server

### LINKED ISSUE #1923: Implement equivalent of Spark History Server

andygrove:

Closes #1923. Last of four PRs splitting up #1925, after #2256 (merged), #2260 and #2264.

> **Stacked on #2264.** GitHub shows the whole chain until the earlier ones merge, so review the last three commits here (`feat(scheduler): add the history server`, `refactor(history): index event logs instead of loading them into memory` and `fix(history): list completed jobs newest first`).

# Rationale for this change

#2264 makes the scheduler write durable per-job logs, but nothing reads them. This serves them.

The scheduler forgets a job shortly after it finishes: completed jobs are cleaned up after `finished_job_state_clean_up_interval_seconds`, and everything is gone on restart. With this, the TUI can point at a history server and browse completed jobs with no scheduler running at all.

`user-personas.md` lists "a history server / UI" among the things Persona 2 depends on, so this fills in a guarantee already written down.

**`ballista-history-server --event-log-dir <dir>`** indexes every completed log at startup and serves the same `/api/*` paths as the live scheduler: `jobs`, `job/{id}`, `job/{id}/stages`, `job/{id}/config`, `job/{id}/dot`.

Corrupt or unreadable logs are logged and skipped rather than failing startup, so one file truncated by a crash cannot hide every other job. `/api/executors` returns empty and `/api/state` a static payload, since there is no cluster behind a history server and the TUI calls both at startup.

**A user-guide page** covering both flags, the endpoints, and the operational caveats.

## What is kept in memory

Only each job's `JobIndex`, which is exactly what `GET /api/jobs` reports. Everything else is read back out of the job's log per request, inside `spawn_blocking`.

The alternative, holding the loaded payload for every job, means keeping both plan-bearing REST responses, the session config and the DOT graph resident. A job with many tasks stores megabytes of per-task detail, so a directory retained for any length of time would put the server's memory use at the mercy of the retention policy, for data almost none of which anyone will open. Request rate on the detail endpoints is whatever a person clicks through a UI, so a read per request is the cheaper side of that trade.

Startup still scans the directory once, since a job's metadata only exists inside its log, but it decodes only the `index` field out of the `JobEnd` record rather than materialising payloads it would immediately drop.

Two consequences, both covered by tests:

- Corruption confined to the payloads is not seen at index time. It surfaces as a 500 naming the failing log when that job is opened.
- A log that loses its terminal record after indexing (rotated, truncated) returns 404.

## Job list ordering

Newest first, by start time, with the job id as a tiebreaker so the order is total. A job id is a random 7-character string (`TaskManager::generate_job_id`), so ordering by it would have been arbitrary. This also matches what the TUI does with the list once it has it, and is the order a future `?limit=` would want to truncate.

There is no paging yet. `GET /api/jobs` returns every job in one response, which is fine for the job counts a single directory realistically holds but will not stay fine forever. Adding it here alone would leave the history server and the live scheduler with different `/api/jobs` contracts, and the TUI talks to both, so it belongs in a change that covers both endpoints. Tracked in #2270, and noted as a limitation on the user-guide page in the meantime.

## How the payloads are served

The stored `/api/job/{id}` and `/api/job/{id}/stages` payloads are relayed as raw JSON rather than deserialized and re-serialized. Clients get the exact bytes the live scheduler produced, and a later change to the REST types cannot make an existing log unservable, which is the compatibility property #2260 is built around.

The job list is the one endpoint that needs structure, since it omits the plan fields. It is rebuilt from the frozen `JobIndex` rather than by editing a stored payload, so that path never parses a response it would mostly discard, and it is also the reason the list endpoint never touches disk.

## The end-to-end guarantee

`history_store_serves_byte_identical_json_to_live_scheduler` was held out of #2264 because it needs `HistoryStore`. It lands here. It emits a real `JobEnd` through the real async writer, loads it back through `HistoryStore::load`, and compares the served bytes against what the live builders produce for the same graph.

Storing raw JSON made it a stronger test than it was in #1925: it is now a literal byte comparison rather than a structural one.

Yes, all additive:

- The TUI can point at it with `BALLISTA__SCHEDULER__URL` to browse completed jobs with no live scheduler. Its `--host` / `--port` flags set the gRPC address queries run against and do not move the TUI, so the user-guide page documents the environment variable instead.

No API changes, and nothing about a cluster that has not enabled event logging.

Verified locally: `cargo test -p ballista-scheduler --lib` passes (342 tests) and `cargo test -p ballista-history` passes (18 tests), clippy is clean with `--all-features -D warnings`, fmt, taplo and prettier are clean, and the `--no-default-features` check still passes.

I also ran the binary against the checked-in `schema-v1.eventlog` fixture and curled every endpoint. It indexes the job, serves the list, the job, the stages (relaying the multi-partition `partition_id` array verbatim) and the config, and 404s on an unknown job.

# Known limitations

Documented on the user-guide page rather than left to be discovered:

- **Local filesystem only**, and the directory is scanned once at startup. Restart to pick up newly finished jobs.
- **Disk is not reclaimed automatically.** Logs accumulate until pruned.
- **Plans are rendered once, at write time**, so `?plan_format=` has no effect against a history server.
- **Only the terminal record is served.** The per-task timeline is captured so a future UI can show a job progressing, but nothing reads it yet.
- **`GET /api/jobs` is unpaged.** Tracked in #2270.

Now that nothing is cached, dropping the first of those is mostly a matter of rescanning the directory, but that is out of scope here.

# Verified against a live cluster

The two halves have now been exercised together, which earlier revisions of this description listed as outstanding.

A scheduler started with `--event-log-dir` and one executor ran a TPC-H join and aggregate over Parquet, producing a real event log. A history server pointed at that directory indexed the job and served all five endpoints, and `/api/job/{unknown}` returned 404.

Diffing the history server's responses against the still-running scheduler's, `/api/jobs`, `/api/job/{id}`, `/api/job/{id}/stages` and `/api/job/{id}/dot` are byte-identical, which is the property the raw-JSON relay is meant to give.

`/api/job/{id}/config` is the one endpoint whose bytes differ, and the cause is on the live side: it serializes a `HashMap`, so two consecutive requests to the scheduler itself disagree on key order. The parsed maps compare equal, and the history server's copy is the stable one because it is a frozen snapshot.

Driving the TUI against the history server is how the `--host` / `--port` documentation error above surfaced. With `BALLISTA__SCHEDULER__URL` set, the header shows the history server's address and the job list renders from it.

Two things this did not cover:

- The per-job TUI screens. Selecting a row could not be driven from a scripted terminal, so those screens are unverified, though every endpoint behind them is checked above.
- The static `/api/state` payload sets `started` to 0, which the TUI header renders as `1969-12-31`. Cosmetic, and left as is here.


ballista/scheduler/src/history/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::Arc;
+
+/// In-memory store of completed jobs, loaded once at startup from a directory
+/// of `<job_id>.eventlog` files.
+#[derive(Default)]
+pub struct HistoryStore {
+    /// Completed jobs keyed by job id.
+    pub jobs: HashMap<String, ReplayedJob>,
```
milenkovicm: why do we want to keep them all in the memory? cant we just read file on request

ballista/scheduler/src/history/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+
+impl HistoryStore {
+    /// Load every completed job found under `dir`. Missing directories yield
+    /// an empty store rather than an error.
+    ///
+    /// A single unreadable/corrupt `.eventlog` file (e.g. truncated by a
+    /// crash mid-write) is logged and skipped rather than failing the whole
+    /// load — one bad log must not hide every other completed job. Only a
+    /// failure to read the directory itself is propagated.
+    pub fn load(dir: &Path) -> std::io::Result<HistoryStore> {
```
milenkovicm, line 101: same like previous, why do we need to load it all instead of reading it on request ?

milenkovicm (commented):
do we need to load all jobs to memory instead of reading them from FS when they are needed ?

ballista/scheduler/src/history/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+        end_time: index.end_time,
+        logical_plan: None,
+        physical_plan: None,
+        stage_plan: None,
+    }
+}
+
+/// The one endpoint that touches every job, and the reason the index is held
+/// in memory at all: it is served without going near the disk.
+async fn get_jobs(State(store): State<Arc<HistoryStore>>) -> Json<Vec<JobResponse>> {
```
milenkovicm, line 227: We would need to add paging if we plan to keep more than few jobs in history store 

Making jobs id sorted would help i believe

ballista/scheduler/src/history/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    let mut jobs: Vec<JobResponse> =
+        store.jobs.values().map(|j| list_entry(&j.index)).collect();
+    jobs.sort_by(|a, b| a.job_id.cmp(&b.job_id));
```
milenkovicm: Should jobs be sorted on time rather than name (if job ids are not monotonically increasing)

milenkovicm (commented):
A few more comments

ballista/scheduler/src/history/mod.rs
andygrove: Thanks, this wasn't intended ... fixed now

ballista/scheduler/src/history/mod.rs
andygrove: Good catch. Fixed.

ballista/scheduler/src/history/mod.rs
andygrove: I'd like to do this as a separate PR since it will involve API changes. Filed https://github.com/apache/datafusion-ballista/issues/2270 to track

milenkovicm: Makes sense

ballista/scheduler/src/history/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// A single unreadable/corrupt `.eventlog` file (e.g. truncated by a
+    /// crash mid-write) is logged and skipped rather than failing the whole
+    /// load — one bad log must not hide every other completed job. Only a
+    /// failure to read the directory itself is propagated.
+    ///
+    /// Each log is read once here, but only its summary is decoded, so this
+    /// costs a pass over the directory rather than a copy of it in memory.
+    /// Corruption confined to the payloads therefore surfaces when the job is
+    /// requested rather than at startup.
+    pub fn load(dir: &Path) -> std::io::Result<HistoryStore> {
```
milenkovicm, line 101: will this be called ONLY when history server starts? 

can we handle situation where we have one history server running and we have one or more scheduler servers producing events, will new events be visible to history server (after history server startup)

milenkovicm (commented):
one question regarding visibility of new events after history server starts, not sure if i'm missing something but to me it looks job entries are read at (history server) startup

can we handle situation where we have one history server running and we have one or more scheduler servers producing events (to a directory which history server reads), will new events be visible to history server (after history server startup)

milenkovicm:
simple test scenario 
- i have a running server which writes events to predefined directory,
- i have a history server which reads same predefined directory 
- run a job 
- new event file gets written to predefined directory 

number of jobs returned by history server does not change, new job not detected

milenkovicm:
#2269 landed, it generates monotonically increasing ids, perhaps it could be used to avoid loading and caching data from a fs

andygrove:
Sorry for the quality of this PR @milenkovicm and thanks for reviews so far. Moving to draft until this is more complete.

milenkovicm:
Not a bother at all, this is good functionality to have, happy to help

milenkovicm:
I'll use this comments for follow up ideas, we can filter them later 

- Https support
- authentication 
- web TUI integration

docs/source/user-guide/history-server.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+The TUI reads its URL from configuration rather than from the `--host` and
+`--port` flags, which set the gRPC scheduler address used for running queries.
+Pointing the TUI somewhere else means setting `BALLISTA__SCHEDULER__URL`, or
+`scheduler.url` in the TUI's config file. Passing `--port 50060` alone leaves
+the TUI on its default of `http://localhost:50050`, where it either finds your
+live scheduler or reports that the scheduler is down.
+
+The directory is scanned once at startup, so restart the history server to pick
```
milenkovicm, line 86: Do you want history sever to serve only static list of jobs discovered at startup @andygrove ?

This was my main comment last time, I think it would make sense to discover new events automatically during execution rather than on startup. 

Anyway your call, please let me know what you think

milenkovicm:
This fall of the radar, it would be good to get it in for 55 release 


@andygrove do you want to merge it as it is, even with static job list, we can iterate on it. Or you prefer to address comments

docs/source/user-guide/history-server.md
andygrove: I have now added auto refresh based on an interval

andygrove:
> This fall of the radar, it would be good to get it in for 55 release
> 
> @andygrove do you want to merge it as it is, even with static job list, we can iterate on it. Or you prefer to address comments

I have now addressed feedback. Thanks for the reviews, I'll go ahead and merge and we can keep iterating on this.

milenkovicm:
Thanks @andygrove

PR #2269, opened 2026-08-09
feat: pluggable job id generator

milenkovicm:

Closes #.

 # Rationale for this change

Previously job ids were created randomly and as such they were not sortable, also, there was a chance for collision in case we save and keep them for a long time on top of that different users have different requirements, hence we nee to be flexible how job ids are generated 

Replaced random job ID generation with monotonically increasing, sortable Snowflake-style IDs. 

- Added [ferroid](https://lib.rs/crates/ferroid) dependency. Ferroid is high-performance ULID and Snowflake-style IDs. Unique, monotonic, and lexicographically sortable IDs optimised for low-latency services and async workloads.
- `generate_job_id()` has been removed from `TaskManager`, there is no real reason to have it there.
- new trait `JobIdGenerator` has been introduced to support pluggable generation strategies
- `DefaultJobIdGenerator ` using  `Arc<AtomicSnowflakeGenerator<SnowflakeMastodonId, MonotonicClock>>` has been added. Currently `Mastodon` style snowflake id layout has been used which **does not hold machine id**. [more details about layouts](https://lib.rs/crates/ferroid#readme-supported-layouts). The reason for this is simplification as machine id generation might need some infrastructure to run.
   - Ids are encoded used base32 encoding, they look something like `037Z8BV46C000`
- Scheduler config has new property to configure external job id generator 
 
Reason for not using UUID v7 or ULID is size of generated id, which might not be too user friendly. 

Net effect: default job IDs are now time-ordered/sortable rather than random, which is useful for chronological ordering, log correlation, and roughly indicates submission order.

No, apart from fact that job id increased in length

### DIFF
--- a/ballista/scheduler/src/scheduler_server/mod.rs
+++ b/ballista/scheduler/src/scheduler_server/mod.rs
@@ -100,7 +100,7 @@ struct DefaultJobGenerator {
 impl Default for DefaultJobGenerator {
     fn default() -> Self {
         let generator = AtomicSnowflakeGenerator::new(
-            0, // machine id is hard codded to 0
+            0, // machine id is hard-coded to 0
             MonotonicClock::<1>::with_epoch(ferroid::time::UNIX_EPOCH),
         );
 
--- a/ballista/scheduler/src/scheduler_server/mod.rs
+++ b/ballista/scheduler/src/scheduler_server/mod.rs
@@ -269,7 +269,7 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerServer<T
         subscriber: Option<JobStatusSubscriber>,
     ) -> Result<JobId> {
         log::debug!("Received submit request for job {job_name}");
-        let job_id = self.state.task_manager.generate_job_id();
+        let job_id = self.state.task_manager.generate_job_id().await;
         self.query_stage_event_loop
             .get_sender()?
             .post_event(QueryStageSchedulerEvent::JobQueued {
--- a/ballista/scheduler/src/scheduler_server/mod.rs
+++ b/ballista/scheduler/src/scheduler_server/mod.rs
@@ -99,9 +99,11 @@ struct DefaultJobGenerator {
 
 impl Default for DefaultJobGenerator {
     fn default() -> Self {
+        // default implementation does not support multi machine
+        // setups
         let generator = AtomicSnowflakeGenerator::new(
             0, // machine id is hard-coded to 0
-            MonotonicClock::<1>::with_epoch(ferroid::time::UNIX_EPOCH),
+            MonotonicClock::<1>::default(),
         );
 
         Self { generator }
--- a/ballista/scheduler/src/scheduler_server/mod.rs
+++ b/ballista/scheduler/src/scheduler_server/mod.rs
@@ -30,6 +30,11 @@ use datafusion::physical_plan::ExecutionPlan;
 use datafusion::prelude::{SessionConfig, SessionContext};
 use datafusion_proto::logical_plan::AsLogicalPlan;
 use datafusion_proto::physical_plan::AsExecutionPlan;
+use ferroid::base32::Base32SnowExt;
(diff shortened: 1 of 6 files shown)


milenkovicm:
cc @phillipleblanc id like to get your review is possible, as you use multi scheduler setup and we could make snowflake ids multi scheduler if needed

avantgardnerio (commented):
LGTM, except I'd really love to see HA return - I think it will be essential for rolling upgrades if nothing else. 

Would you be opposed to a format that includes some bits for machine ID, so we don't have to change it later and risk collision?

phillipleblanc:
> cc @phillipleblanc id like to get your review is possible, as you use multi scheduler setup and we could make snowflake ids multi scheduler if needed

This change won't affect Spice, because we generate our own query IDs (using UUIDv7) and pass that into a fork method we have `submit_job_with_id` that accepts our own IDs instead of the ballista generated ones. That might be a potential upstream candidate that would address the multi-scheduler concern? Basically you have the option to use the nice ballista generated jobIds from the snowflake method, or you fully opt into your own ID generation.

phillipleblanc (approved):
Looks good to me

milenkovicm:
So deffinetly there is need for machine specific ids, but also there is need for bring your own id generator.

I think BYO ID generator makes sense, it would tick both boxes. 
will change the proposal to address requirements, thanks @avantgardnerio & @phillipleblanc

milenkovicm:
please have a look @phillipleblanc  & @avantgardnerio if it make sense i will deal with python related failures before merge

ballista/scheduler/src/scheduler_server/mod.rs
```diff
@@ -69,6 +67,48 @@ pub(crate) mod query_stage_scheduler;
...
+/// A monotonically increasing sortable snowflake job id generator
+/// which does not capture machine id as part result.
+struct DefaultJobGenerator {
+    generator: AtomicSnowflakeGenerator<SnowflakeMastodonId, MonotonicClock>,
+}
+
+impl Default for DefaultJobGenerator {
+    fn default() -> Self {
+        let generator = AtomicSnowflakeGenerator::new(
+            0, // machine id is hard codded to 0
```
martin-g:
```suggestion
            0, // machine id is hard-coded to 0
```

martin-g: So, for HA it is left to the user to provide its own generator, right ?

martin-g (approved):

ballista/scheduler/src/scheduler_server/mod.rs
milenkovicm: Yes

avantgardnerio (approved):
Thanks @milenkovicm , it's great to support the (future/spice) HA use-case explicitly!

milenkovicm:
we need to merge #2277  first to make CI green

ballista/scheduler/src/scheduler_server/mod.rs
villebro: If possible, I think it would be nice if we could automatically incorporate a machine ID in the HA case, while still allowing users to BYO generator when they have more specific requirements. My preference would be for the default configuration to remain safe when moving from single-scheduler to HA, rather than introducing a potential collision risk unless the user explicitly configures a different generator.

milenkovicm: for that specific case we would need to use `ULID`s or `UUIDv7`, if we keep snowflake style IDs we need to provide machine id somehow, which probably means users need to configure it manually or add some kind of infrastructure to provide it. 

i suggest if we decide to go `ULID`s we can do it as a follow up, my primary concern with this PR was sortable JIDs. 

wdyt @villebro

villebro: Sure, if we can continue tweaking the default implementation in a follow up then no problem. My main goal is just to make running HA as painless as possible, and the fewer knobs an admin has to tune, the better.

milenkovicm: perfect, thanks

PR #2276, opened 2026-08-10
refactor(core): use DataFusionError::External for structured task errors

villebro:
Follow-up to #2202 (review discussion [r3737946864](https://github.com/apache/datafusion-ballista/pull/2202#discussion_r3737946864)).

# Rationale for this change

#2202 kept shuffle-fetch failures structural until task-failure classification, but carried the structured `BallistaError` across DataFusion as `ArrowError::ExternalError(Box<BallistaError>)`. Because DataFusion's `RecordBatchStream` yields `Result<RecordBatch, DataFusionError>`, that Arrow error auto-converts (via `From<ArrowError> for DataFusionError`) into `DataFusionError::ArrowError(ArrowError::ExternalError(..))` — a double wrap that recovery had to unwind through both the DataFusion and Arrow layers.

`DataFusionError::External` is the purpose-built variant for errors originating outside DataFusion, and `find_root()` already sees through it (traversing any `Shared` / `Context` / `Diagnostic` layers). Using it removes the Arrow intermediate and collapses recovery to a single downcast:

```
before:  DataFusionError::ArrowError(ArrowError::ExternalError(Box<BallistaError>))
after:   DataFusionError::External(Box<BallistaError>)
```

This is an internal refactor: errors are classified into the `FailedTask` proto inside the executor process before crossing any wire, so the wrapping convention can change without affecting the protocol.

- `BallistaError::into_datafusion` is the single helper for the carry step, used at the shuffle-reader fetch-stream boundary and the shuffle-writer drain handoff.
- `find_fetch_failed` / `fetch_failed_in_arrow` / `fetch_failed_in_datafusion` collapse into one `find_root()` + `External` downcast, and `is_retryable_io` sees through the same wrapper so IO errors stay retryable.
- Sites that already held a `DataFusionError` (`summaries_to_batch` in both writers, the `execute_query_pull` / `execute_query_push` client result streams) no longer wrap it in Arrow just to have it flattened back.
- The `distributed_query` fetch path now routes through the same `into_datafusion` helper instead of hand-building `DataFusionError::External`.
- The error-classification tests are consolidated to exercise the bare, `External`, `Shared`, and `Context` shapes in one place instead of a separate test per shape.

No API or documentation changes.

Free-text task error messages lose the redundant wrapper layer: the `Arrow error: External error:` prefix collapses to `External error:`, and the two IO-error messages (`Task failed due to Ballista IO error` / `Task failed due to DataFusion IO error`) are unified to `Task failed due to IO error`. Structured failure classification — the fetch-partition, retryable-IO, task-killed, and execution-error reasons the scheduler acts on — is unchanged, including under `Shared` / `Context` layers.

### DIFF
--- a/ballista/core/src/error.rs
+++ b/ballista/core/src/error.rs
@@ -77,14 +77,8 @@ impl<T> Into<Result<T>> for BallistaError {
 }
 
 impl BallistaError {
-    /// Carries a Ballista error across DataFusion's stream boundary as a single
-    /// [`DataFusionError::External`].
-    ///
-    /// This is the one wrapping convention for Ballista-origin errors. It keeps
-    /// the error structural (rather than stringifying it) so that task-failure
-    /// classification can recover it: `find_root()` traverses any `Shared` /
-    /// `Context` / `Diagnostic` layers DataFusion adds on top and lands on this
-    /// `External` node, whose payload is downcast back to a `BallistaError`.
+    /// Keeps a Ballista error structural across DataFusion's stream boundary, so
+    /// task-failure classification can recover it instead of parsing a string.
     pub fn into_datafusion(self) -> DataFusionError {
         DataFusionError::External(Box::new(self))
     }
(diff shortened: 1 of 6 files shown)


villebro:
@milenkovicm when you have a moment please review.

avantgardnerio (approved):
Nice cleanup! I could not find any faults with it.

milenkovicm:
thanks @villebro and @avantgardnerio, will merge this

PR #2278, opened 2026-08-10
fix(scheduler): key AQE shuffle output by job_id

### LINKED ISSUE #1859: `AdaptivePlanner` uses `job_name` to identify jobs rather than `JobId`
**Describe the bug**
The `AdaptivePlanner` struct holds a `job_name: String`, while instead it should use a `JobId` (or both, not sure if `job_name` is used for display purposes). This makes some calls awkward which expect a `job_id`. 

This was surfaced while experimenting with the newtype pattern for `JobId` in #1852 

**To Reproduce**
See [planner.rs](https://github.com/apache/datafusion-ballista/blob/d707465dd84efc82134ec7c2aa4a4ddf3d13893e/ballista/scheduler/src/state/aqe/planner.rs#L167-L168) for an example of awkward usage. 

**Expected behavior**
I think the `AdaptivePlanner` should hold both a `job_name` and `job_id`.

**Additional context**
N/A

villebro:
Closes #1859.

# Rationale for this change

On the adaptive (AQE) path, `AdaptivePlanner` carried the job *name* where it needed the job *id*. At the two spots that require a `JobId` — `handle_explain_plan` and `BallistaAdapter::adapt_to_ballista` (which builds shuffle-writer output paths) — it recovered one with `job_name.clone().into()`, flagged in the code as a probable bug:

```rust
// Note: the signature requires a JobId, but we are passing a JobName. The below is a
// dirty fix but this seems like a bug or a design flaw.
let job_id: JobId = job_name.clone().into();
```

Since `job_name` is frequently empty (e.g. `"job_name": ""` from the REST API), a job's shuffle output was keyed by an empty string instead of its `job_id`. Because the output path is `work_dir / job_id / stage_id / …`, that key is the per-job directory that isolates one job's shuffle files from another's — so every job with an empty name shared the same `work_dir/ /stage_id/…` tree and concurrent **jobs could collide on the same stage/partition paths**. The static planner was already correct.

`AdaptivePlanner` now holds `job_id: JobId` instead of `job_name: String`, and `AdaptiveExecutionGraph::try_new` passes the `job_id` it already has. The two `job_name.clone().into()` placeholders are gone.

Tests: existing planner tests just pass a `JobId` (`.to_owned()` → `.into()`); `should_split_plan_into_stages` now asserts the resolved shuffle writer is keyed by the job id; and a new `newly_built_adaptive_graph_has_expected_state` — the first test to build an `AdaptiveExecutionGraph` — checks a fresh graph reflects its constructor arguments (using a distinct job id and job name) and starts as a running job with runnable tasks once revived.

Under adaptive query planning, a job's shuffle output is now keyed by its `job_id` rather than its (frequently empty) `job_name`. No API or documentation changes.

### DIFF
--- a/ballista/scheduler/src/state/aqe/mod.rs
+++ b/ballista/scheduler/src/state/aqe/mod.rs
@@ -142,7 +142,7 @@ impl AdaptiveExecutionGraph {
         let session_id = ctx.session_id();
 
         let mut planner =
-            AdaptivePlanner::try_new(ctx, logical_plan, job_name.to_owned()).await?;
+            AdaptivePlanner::try_new(ctx, logical_plan, job_id.clone()).await?;
 
         let logical_plan = Some(logical_plan.display_indent().to_string());
 
--- a/ballista/scheduler/src/state/aqe/planner.rs
+++ b/ballista/scheduler/src/state/aqe/planner.rs
@@ -70,8 +70,8 @@ pub struct AdaptivePlanner {
     pub(crate) plan: Arc<dyn ExecutionPlan>,
     /// caches current runnable stages
     runnable_stage_cache: HashMap<usize, Arc<dyn ExecutionPlan>>,
-    /// job name
-    job_name: String,
+    /// Unique identifier for the job, used to key this job's shuffle output.
+    job_id: JobId,
 
     runnable_stage_output: HashMap<usize, StageOutput>,
 }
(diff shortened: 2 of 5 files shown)


milenkovicm (approved):
thanks for fixing this @villebro 
i take the blame for this issue

PR #2279, opened 2026-08-10
fix(scheduler): default no-executors grace to executor timeout (>= heartbeat)

akshaychitneni:
Which issue does this PR close?
  Follow-up to #2029 / #2212, addressing the timeout interaction surfaced by #2240 (https://github.com/apache/datafusion-ballista/pull/2240#discussion_r3732320008)

  Rationale for this change
  The empty-cluster grace timer defaulted to 30s, but the executor heartbeat interval defaults to 60s. Since #2240 routes task-launch-failure removals through the same ExecutorLost path (aggressive removal that also drops the heartbeat), a healthy executor removed by a transient blip re-registers only on its next heartbeat (~≤60s) — which the 30s timer could beat,
  spuriously failing the job.

  What changes are included in this PR?
  - no_executors_grace_period_seconds now defaults to executor_timeout_seconds (guaranteed longer than the heartbeat interval; the scheduler can't read the executor's heartbeat setting directly).
  - CLI arg is Option<u64>, resolving to executor_timeout_seconds when unset; SchedulerConfig::Default matches.
  - Removal stays immediate (fast rescheduling); only the fail decision waits, and the timer's re-check still cancels the failure if the executor re-registers. Explicit values (incl. 0 = fail-fast) honored.
  - Adds config tests (default resolution + grace >= executor_timeout invariant).
  
  Are there user-facing changes?
  Yes — the default no-executors grace period changes from 30s to executor_timeout_seconds (180s by default); --no-executors-grace-period-seconds is now optional and defaults to that.

### DIFF
--- a/ballista/scheduler/src/config.rs
+++ b/ballista/scheduler/src/config.rs
@@ -31,6 +31,7 @@ use ballista_core::extension::EndpointOverrideFn;
 use ballista_core::{ConfigProducer, JobId, config::TaskSchedulingPolicy};
 use datafusion_proto::logical_plan::LogicalExtensionCodec;
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
+use log::warn;
 use std::fmt::Display;
 use std::sync::Arc;
 
@@ -231,10 +232,9 @@ pub struct Config {
     /// cluster has lost its last executor before failing the running jobs.
     #[arg(
         long,
-        default_value_t = 30,
-        help = "Grace period, in seconds, to wait for an executor to (re)register after the last executor is lost before failing running jobs. Prevents jobs from hanging forever when every executor dies, while still tolerating a transient total loss (e.g. a rolling restart). Set to 0 to fail as soon as the loss is observed."
+        help = "Grace period, in seconds, to wait for an executor to (re)register after the last executor is lost before failing running jobs. Prevents jobs from hanging forever when every executor dies, while still tolerating a transient total loss (e.g. a rolling restart). Must be >= the executor heartbeat interval, or an executor removed by a transient task-launch failure may not re-register before this elapses; defaults to --executor-timeout-seconds (which is always longer than the heartbeat interval). Set to 0 to fail as soon as the loss is observed."
     )]
-    pub no_executors_grace_period_seconds: u64,
+    pub no_executors_grace_period_seconds: Option<u64>,
     /// Minimum number of registered executors before /readyz returns 200
     #[arg(
         long,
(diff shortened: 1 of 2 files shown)


ballista/scheduler/src/config.rs
```diff
@@ -399,7 +399,9 @@ impl Default for SchedulerConfig {
             grpc_client_max_message_size: 16777216,
             executor_timeout_seconds: 180,
             expire_dead_executor_interval_seconds: 15,
-            no_executors_grace_period_seconds: 30,
+            // Defaults to `executor_timeout_seconds` so the grace is always >=
+            // the executor heartbeat interval (see the field doc).
+            no_executors_grace_period_seconds: 180,
```
sandugood, line 405: Should this be 180 by default, if the `executor_heartbeat_interval_seconds` is 60 secs by default?

ballista/scheduler/src/config.rs
```diff
@@ -667,7 +662,12 @@ impl TryFrom<Config> for SchedulerConfig {
             executor_timeout_seconds: opt.executor_timeout_seconds,
             expire_dead_executor_interval_seconds: opt
                 .expire_dead_executor_interval_seconds,
-            no_executors_grace_period_seconds: opt.no_executors_grace_period_seconds,
+            // Default to the executor-liveness timeout when unset, so the grace
+            // is always >= the executor heartbeat interval and a transiently
+            // removed but healthy executor can re-register before it elapses.
+            no_executors_grace_period_seconds: opt
+                .no_executors_grace_period_seconds
+                .unwrap_or(opt.executor_timeout_seconds),
```
milenkovicm, line 699: Should it be multiple of value, just to avoid race ?

milenkovicm (commented):
Thanks @akshaychitneni this makes sense, one minor comment, and a question if we can add validation method now when we have dependencies between configuration values

akshaychitneni:
> Thanks @akshaychitneni this makes sense, one minor comment, and a question if we can add validation method now when we have dependencies between configuration values

Thanks. Added SchedulerConfig::validate(), called from start_server. It warns when no_executors_grace_period_seconds (nonzero) is below executor_timeout_seconds, and when the timeout isn't greater than the expire interval

ballista/scheduler/src/config.rs
akshaychitneni: Yes. 180 is the default executor_timeout_seconds (the liveness window) giving 3× margin for a transiently-removed executor to re-register before its jobs fail

ballista/scheduler/src/config.rs
akshaychitneni: Kept the default equal to executor_timeout_seconds rather than a multiple: 180s is already ~3× the 60s heartbeat. Increasing further might push total-loss failures to 6+ min with multiples

milenkovicm: Sorry I've misread property used, apologies

milenkovicm (approved):
Thanks @akshaychitneni

PR #2288, opened 2026-08-12
feat: embedded flight proxy explicit config entry

### LINKED ISSUE #2281: Replace the advertise_flight_sql_endpoint empty-string sentinel with an explicit flag
The scheduler decides whether to start its embedded Arrow Flight proxy by matching on an empty string:

  // ballista/scheduler/src/scheduler_process.rs:104
```
  match &config.advertise_flight_sql_endpoint {
      Some(proxy) if proxy.is_empty() => { /* start embedded flight proxy */ }
      // Some("host:port") => advertise an external endpoint
      // None            => no proxy
  }
```

  So one Option<String> overloads three meanings, and Some("") = "enable the embedded proxy" is non-obvious — a config/env value that happens to be empty silently flips a mode.

  Proposal: add an explicit enable_embedded_flight_proxy: bool (or an enum) separate from the "advertise this external endpoint" string, and treat an empty advertise_flight_sql_endpoint as
  unset rather than as a switch.

 Follow-up from [#2244 review](https://github.com/apache/datafusion-ballista/pull/2244#issuecomment-5226731477)

sandugood:

Closes #2281.

 # Rationale for this change

Before this change an empty string in the `advertise_flight_sql_endpoint` config entry would force the start of an embedded Flight proxy. 

Adding an explicit new config entry named `enable_embedded_flight_proxy` and refactored `SchedulerServer`'s method (as well as the REST API handlers, but they are not breaking)

For code consistency: there is backward compatibility, so user's `Scheduler` won't break if there is an emptry string in the `--advertise-flight-sql-endpoint`

Yes, adding a new `enable_embedded_flight_proxy` to `Scheduler`'s config.

Bare `--advertise-flight-sql-endpoint` (empty value) still enables the embedded proxy for backward compatibility, but logs a deprecation warning. Programmatically setting `advertise_flight_sql_endpoint` to an empty string no longer enables the proxy - should use `with_enable_embedded_flight_proxy(true)`


akshaychitneni:
LGTM. Thanks @sandugood

milenkovicm (commented):
this is a tricky one, what if user specifies

```
--enable-embedded-flight-proxy --advertise-flight-sql-endpoint=localhost..
```

at the moment it will respect `advertise-flight-sql-endpoint` rather than `enable-embedded-flight-proxy`, do we want to print warning or error, should we cover this case in the `verify` method ?

milenkovicm:
btw thanks for contribution @sandugood and @akshaychitneni for review

andygrove:
I thought that the Flight SQL functionality had been removed. I am reviewing this now.

milenkovicm:
@andygrove flightsql has been removed, but scheduler can provide a flight proxy for environments where clients cant connect to executors

milenkovicm:
it is disabled by default, and should be used carefully

andygrove:
> @andygrove flightsql has been removed, but scheduler can provide a flight proxy for environments where clients cant connect to executors

Got it. Looks like the naming is confusing. This is Flight not Flight SQL.

andygrove (commented):
> **Disclaimer:** this review was generated by an LLM (Claude Code) at my request, against `edf0d9a`. File/line references and the git history claims were checked against the code, but please treat the conclusions as suggestions and push back where it's wrong.

Thanks for picking this up — it's exactly what #2281 asked for, and the implementation is clean.

### Overview

Replaces the `Some("")` sentinel on `advertise_flight_sql_endpoint` (which overloaded one `Option<String>` with three meanings) with an explicit `enable_embedded_flight_proxy: bool`:

- New CLI flag `--enable-embedded-flight-proxy` + `SchedulerConfig` field + `with_enable_embedded_flight_proxy()` builder.
- `scheduler_process.rs` drops the `match ... if proxy.is_empty()` for a plain `if config.enable_embedded_flight_proxy`, removing a `_ => {}` arm and a level of nesting.
- `flight_proxy_config()` in `grpc.rs` rewritten: non-empty endpoint → `FlightProxy::External`, else embedded flag → `FlightProxy::Local(true)`, else `None`.
- Backward compat: bare/empty `--advertise-flight-sql-endpoint` maps to `enable_embedded_flight_proxy = true` + endpoint `None`, with a deprecation `warn!`.
- REST `/state` gains `embedded_flight_proxy: bool`.
- Tests for the CLI mapping and the `flight_proxy_config()` matrix.

`log::warn` is already imported (`config.rs:34`), `Config` and the two new CLI tests are correctly `#[cfg(feature = "build-binary")]`-gated, and the struct uses no clap `env =`, so the empty-env-var footgun from the issue is closed at the CLI layer.

### Correctness

The behaviour matrix is preserved and the client side stays sound:

| Input | Proxy started | Advertised |
|---|---|---|
| nothing | no | `None` |
| `--enable-embedded-flight-proxy` | yes | `Local(true)` |
| `--advertise-flight-sql-endpoint HOST:PORT` | no | `External(...)` |
| bare `--advertise-flight-sql-endpoint` | yes (+warn) | `Local(true)` |
| both | yes | `External(...)` |

`get_client_host_port` (`distributed_query.rs:806`) handles `External(addr)` by parsing `http://{addr}`, so `External("")` would fail to parse — the double filtering (builder *and* `flight_proxy_config`) correctly makes that unreachable. "Both set" isn't contradictory either: running the embedded proxy while advertising an LB in front of the scheduler is a real deployment, so external-wins-for-advertisement is right, and the help text ("Mutually independent from...") describes it well.

### Issues / suggestions

**1. The struct-literal construction path silently loses the proxy.**
`with_advertise_flight_sql_endpoint` filters empties, but the fields are `pub` and `Default` is implemented, so an embedder doing:

```rust
SchedulerConfig { advertise_flight_sql_endpoint: Some(String::new()), ..Default::default() }
```

previously got the embedded proxy and now silently gets nothing, with no warning on either path. `config.validate()` already runs at startup (`scheduler_process.rs:185`, before `start_grpc_service`) and is the established home for this guidance:

```rust
if self.advertise_flight_sql_endpoint.as_deref() == Some("") {
    warn!(
        "advertise_flight_sql_endpoint is set to an empty string, which is treated as \
         unset; use with_enable_embedded_flight_proxy(true) to start the embedded proxy."
    );
}
```

That covers every construction path uniformly and matches the existing warn-don't-reject style.

**2. While we're here: `advertise_flight_sql_endpoint` is a misnomer, and this PR is the natural place to fix it.**
Ballista removed Flight SQL in 559bcf29 (`feat: remove flight-sql`, #1228, released in 46.0.0) — there is no `FlightSqlService` in the tree. This config predates that removal (#472, 2022, when the scheduler *did* host a Flight SQL server); it survived and was then repurposed by ac49c18a (`feat: Add arrow flight proxy to scheduler`, #1351) for something else entirely — proxying **result partitions** over plain Arrow Flight. `flight_proxy_service.rs:117-145` decodes the ticket as a Ballista `Action::FetchPartition` and forwards it to the owning executor; `handshake`, `get_flight_info`, and `do_action` are all `Status::unimplemented`, so no Flight SQL or JDBC client can use this endpoint.

The new field is named well (`enable_embedded_flight_proxy`, no "sql"), but the PR leaves the old name in place and its new doc comment — "The route endpoint for proxying flight sql results via scheduler" — restates the misleading claim. Since the whole point of this PR is de-confusing this setting, consider going one step further: rename to `advertise_flight_endpoint` / `--advertise-flight-endpoint`, keeping the old flag as a deprecated clap alias next to the empty-value deprecation you're already adding. Two deprecations in one release is cheaper for operators than a second breaking pass later. If that's out of scope here, please at least drop "sql" from the doc comment and `--help` text and say "result partitions", and file a follow-up for the rename.

**3. Doc comment on the new field is thin.** `/// Whether to enable embedded proxy` under-sells the semantics compared to the good comment added to `flight_proxy_config()`. Worth saying it's independent of the advertise endpoint, and that a non-empty endpoint takes precedence for what clients are told.

**4. REST field naming.** `embedded_flight_proxy` diverges from the config field and from its sibling booleans in `SchedulerStateResponse`, which all use `*_support` (`substrait_support`, `keda_support`, …). `enable_embedded_flight_proxy` would map 1:1 to the config. Purely additive, and there's no `api-types` DTO or bundled UI mirroring this struct, so nothing else needs updating.

**5. Adjacent pre-existing bug on lines this PR moves.** `BallistaFlightProxyService::new` takes `(max_decoding_message_size, max_encoding_message_size, ...)` (`flight_proxy_service.rs:53-54`), but the call site passes encoding first, then decoding — the PR carries the swap over verbatim. These feed the scheduler→executor Flight client (`flight_proxy_service.rs:213-214`); the outer `FlightServiceServer` builder calls are correct. Latent today since both default to 16 MiB (`config.rs:406-407`), but wrong for anyone setting asymmetric limits. Since you're already rewriting this block, fixing it here (noted in the description) or filing a follow-up would be worthwhile.

### Tests

Good coverage: CLI → `SchedulerConfig` mapping for both flags, and the full `flight_proxy_config()` precedence matrix. Gaps worth one-liners:

- `with_advertise_flight_sql_endpoint(Some("".into()))` → `None` — the documented behaviour *change*, currently untested.
- The combined `--advertise-flight-sql-endpoint HOST:PORT --enable-embedded-flight-proxy` case at the CLI layer (covered only at the `SchedulerConfig` layer).

Minor style nit: the new `grpc.rs` test fully-qualifies `crate::test_utils::test_cluster_context` and `ballista_core::serde::BallistaCodec`, both already imported in `mod test`; neighbouring tests use the short forms.

### Docs, performance, security

- **Docs:** no update required. `configs.md` is generated from `ConfigEntry` definitions in `ballista/core/src/config.rs` (session settings), scheduler CLI flags aren't tabled anywhere, and no doc page mentions the embedded proxy. The `--help` text is the only user-facing documentation — see point 2.
- **Performance:** none; startup-time config only.
- **Security:** mildly positive. The old sentinel meant an accidentally-empty config value could silently expose an extra Flight service on the scheduler port; that now requires an explicit flag. The `/state` addition leaks nothing sensitive.

### Summary

Looks good to me once **(1)** is addressed, with **(2)** needing a scope call (rename now vs. follow-up — though the "sql" wording in the help text and doc comment should go either way) and **(3)–(5)** as cheap polish. Reviewed by reading; I did not compile or run the suite, but the new tests' helper signatures and feature gates check out against current `main`.

andygrove:
> > @andygrove flightsql has been removed, but scheduler can provide a flight proxy for environments where clients cant connect to executors
> 
> Got it. Looks like the naming is confusing. This is Flight not Flight SQL.

btw @villebro we still have the scheduler proxy for fetching results that we were discussing last night at the meetup.

milenkovicm:
we can add flight-proxy which can reside between client end executors, `--advertise-flight-endpoint` would tell clients where to go to get their shuffle data from, instead going to executors directly

villebro:
> btw @villebro we still have the scheduler proxy for fetching results that we were discussing last night at the meetup.

@andygrove thank you for the ping, I had indeed missed this. This does resolve the gap I was curious about.

> it is disabled by default, and should be used carefully

@milenkovicm have you observed issues with proxying the result requests via this endpoint? While fetching data directly from executors does have the benefit of removing the scheduler from the path, some form of proxying will be necessary in highly isolated environments where clients don't have direct network access to executors.

milenkovicm:
I haven't used this functionality but I would definitely recommend either making standalone proxy process or different thread pool for proxy

villebro:
> I haven't used this functionality but I would definitely recommend either making standalone proxy process or different thread pool for proxy

@milenkovicm this was precisely what I was going to propose. Based on experience from other popular distributed query engines, they typically suffer from scheduler congestion (especially the ones that don't have HA scheduler support), due both to complex planing and result proxying. If Ballista can decouple scheduler from result proxy and both can scale horizontally, we would be able to start with minimal footprint and adjust automatically where needed as traffic increases.

milenkovicm:
that would be great! 
the whole idea behind `--advertise-flight-endpoint` was to decouple control flow from data flow. 
if we do it smart we could have 

- current implementation, same process same port (not sure if we can have per tonic service thread pool ) 
- same process different port, different thread pool 
- different process 

pick and choose, whatever suits  :D

milenkovicm:
maybe @sandugood would it be possible to put something about this functionality in documentation (this PR or different PR whatever suits )

akshaychitneni:
> While fetching data directly from executors does have the benefit of removing the scheduler from the path, some form of proxying will be necessary in highly isolated environments where clients don't have direct network access to executors.

This makes sense. That's precisely why I used the proxy in k8s e2e where accessing executors ips can be challenging

sandugood:
Addressed all of the coments (naming, validation) + added doc entry/comments (with the help of my fellow LLM)

avantgardnerio:
Regarding "same process, different port": there's a special NAT pinhole justification for having it be the same socket, as far as the client is concerned. I'm not sure how important that is to various deployment scenarios, but I always thought it helped general acceptance in situations without a networking team.

milenkovicm (approved):
LGTM thanks @sandugood

andygrove (approved):
LGTM. Thanks @sandugood

PR #2293, opened 2026-08-13
docs: document DataFusion dependency cycle in contributor guide

andygrove:

Part of #1920.

## Rationale for this change

The process for tracking DataFusion (`main` pinned to a DataFusion git rev between releases, moving to the RC and then the final release before cutting a Ballista release branch, then back to a git rev) was only described in the issue and partly in `dev/release/README.md`. This documents it briefly in the contributor guide so contributors know why `Cargo.toml` points at a git rev and what happens at release time.

A short "DataFusion dependency" section in `docs/source/contributors-guide/development.md`.

## Are these changes tested?

Docs only; prettier formatting check passes.

Documentation only.

### DIFF
--- a/docs/source/contributors-guide/development.md
+++ b/docs/source/contributors-guide/development.md
@@ -182,6 +182,29 @@ This includes instructions for:
 - Comparing performance with Apache Spark
 - Running load tests
 
+## DataFusion dependency
+
+Between releases, `main` depends on a pinned git revision of
+[`apache/datafusion`](https://github.com/apache/datafusion) `main` rather than a crates.io release, so
+that breaking changes arrive in small increments instead of piling up at release time.
+
+The `datafusion*` entries in the workspace `Cargo.toml` move through one cycle per release:
+
+| Phase                        | `datafusion*` dependency                    |
+| ---------------------------- | ------------------------------------------- |
+| Between releases             | pinned git `rev` of DataFusion `main`       |
+| DataFusion release candidate | the RC tag, to validate early               |
+| DataFusion released          | the crates.io version                       |
+| Ballista release branch cut  | (branch keeps the crates.io version)        |
+| Back on `main`               | pinned git `rev` of DataFusion `main` again |
+
+A release branch therefore only ever depends on published crates. Each `rev` bump on `main` is a PR
+that ports any DataFusion API changes, refreshes affected golden files, and keeps `cargo test` and
+`cargo clippy --all-targets --workspace -- -D warnings` green.
+
+The remaining release steps are described in the
+[release process](https://github.com/apache/datafusion-ballista/blob/main/dev/release/README.md).
+
 ## Upgrade notes
 
 When a change breaks a public API, renames or removes a configuration key or CLI


andygrove:
cc @phillipleblanc

milenkovicm:
@andygrove, have they banned you from `claude.ai` when you failed over to `copilot`? :D

docs/source/contributors-guide/development.md
```diff
@@ -182,6 +182,29 @@ This includes instructions for:
...
+| Ballista release branch cut  | (branch keeps the crates.io version)        |
+| Back on `main`               | pinned git `rev` of DataFusion `main` again |
+
+A release branch therefore only ever depends on published crates. Each `rev` bump on `main` is a PR
+that ports any DataFusion API changes, refreshes affected golden files, and keeps `cargo test` and
+`cargo clippy --all-targets --workspace -- -D warnings` green.
+
+The remaining release steps are described in the
+[release process](https://github.com/apache/datafusion-ballista/blob/main/dev/release/README.md).
+
```
milenkovicm, line 207: how often we move ref ?

milenkovicm (approved):
LGTM, thanks @andygrove 

should we observe build times after we switch to git ref, i hope it build time wont increase

docs/source/contributors-guide/development.md
andygrove: It would be ideal if we could use dependabot or similar to do weekly upgrade PRs. I will look into this

andygrove:
> @andygrove, have they banned you from `claude.ai` when you failed over to `copilot`? :D  

It was an accidental click. I hope Claude doesn't get jealous.

andygrove:
> LGTM, thanks @andygrove 
> 
> should we observe build times after we switch to git ref, i hope it build time wont increase 

That's a good point. Yes we should check that before committing to this approach.

docs/source/contributors-guide/development.md
andygrove: Actually, we can add a skill to the repo for doing the upgrade pr. We can run manually to start with and then try GitHub agent feature to automate it perhaps

andygrove:
Good check — I just measured this on #2241.

For the Rust CI matrix, git-ref pinning is roughly neutral. Comparing latest main (DF 54 from crates.io) against `datafusion-55` (git tag 55.0.0-rc2):

| Job | main | datafusion-55 | Δ |
|---|---:|---:|---:|
| Ballista Rust | 32m 51s | 26m 35s | −6m |
| h2o | 29m 39s | 27m 40s | −2m |
| TPC-H SF10 | 35m 07s | 27m 07s | −8m |
| TPC-DS SF1 | 30m 23s | 29m 58s | −25s |
| Docker | **20m 27s** | **26m 27s** | **+6m** |
| Ballista Web TUI | 3m 00s | 1m 08s | −2m |
| Dev | 53s | 37s | −16s |

Two caveats: this is DF 54 → DF 55, not just source-type — most of the wins on the Rust jobs are probably upstream code-size differences, not git-vs-registry. And there's real runner noise; Docker on sibling DF 55 branches came in at 26m, 30m, and 33m on the same rev.

So my read is:
- **Rust matrix jobs**: no measurable git-ref penalty. `rust-cache` warms after the first fetch, and the compile graph drives the rest.
- **Docker job**: ~+5–10 min consistently. Worth a mention in the doc so contributors aren't surprised. Once DF 55 GA lands on crates.io and we can switch back to a version pin, that should recover.

(FWIW I also tested `[patch.crates-io]` vs `{ git = ... }` in `workspace.dependencies` — `Cargo.lock` comes out byte-identical either way, so the pin *syntax* doesn't matter; only the underlying source type does.)

avantgardnerio:
BTW, I think once you refer to a dependency with a relative path or git in any way (vs crates.io), clippy considers it fair game to go analyze that too. My local build time went up for this reason, and I had to tell Claude to pass `--no-deps`, though Claude denied this strongly I've seen it help.

PR #2294, opened 2026-08-13
feat(core): quantile sketching for any fixed-width ORDER BY key

avantgardnerio:
I tried making a PR to add KLL to RuntimeStatsExec, along with TDigest, but what accidentally came out is another KLL performance improvement PR:

1. RowConverter was slow and needed for NULLs because it does heap allocations, ~~has variable width~~ (I found FixedLengthEncoding, but it's private), and because it stores NULLs as another bit, it doesn't fit in a u64 register
2. Instead, Claude stole the RowConverter trick and turned f64s into sortable u64s which improved speed a good bit
3. We came up with this SortKeySketch concept that tracks NULLs out-of-band, so it can handle NULLs and it goes faster when they are present.

**This PR is purely additive.** Nothing in the tree constructs any of it, `RuntimeStatsExec` is untouched, and T-Digest is neither modified nor removed. `main` behaviour is byte-for-byte identical. It is the foundation for widening `RuntimeStatsExec` past the single non-nullable `Float64` sketch it accepts today, split out so the widening itself can be reviewed on its own.

## What it adds

`ballista/core/src/sort_key.rs`:

- **`SortKeyCodec`** is the complete ordering spec for one fixed-width `ORDER BY` key: its arrow type, its direction, and where its NULLs sort. It encodes values to an order-preserving `u64` and back. Covers signed and unsigned integers, `Float32`/`Float64`, and the temporal types that are `i32` or `i64` underneath. Returns `None` for anything else, which is the signal to fall back to arrow-row.
- **`SortKeySketch`** pairs that codec with a `KllSketch<u64>` over the encoded values and a count of the NULLs, and owns `ingest` / `merge` / `quantile` / `cuts` / `min` / `max`.

`ballista/core/src/kll.rs` gains `min()`, `max()`, `count()`, `at_rank()`, `at_ranks()`, `Clone` and `Debug`.

`benchmarks/benches/quantile_sketch.rs` gains arms that measure the shipped path rather than an open-coded stand-in, price a `u32` sketch key against the `u64` one, and measure cut extraction against ingest.

## Why

Two things the current sketch cannot express.

### NULLs have no representation

T-Digest has no NULL slot, and no sentinel `f64` can serve as one: `±inf` collides with real infinities, and there is nothing else outside the value range. `RuntimeStatsExec::try_new` therefore rejects nullable routing expressions outright, and `split_batch_by_range` sends NULL keys to partition 0 regardless of what `nulls_first` says (there is a TODO to that effect at `range_repartition_common.rs:162`).

`SortKeySketch` counts NULLs beside the sketch instead of encoding them. A NULL has no position among the values, only a side, and `nulls_first` fixes that side at plan time. Keeping it out of the key also leaves the key 8 bytes wide instead of 16, which the benchmark prices at 1.19x against 1.58x.

The cost is that a rank over the population is no longer a rank over the values, so the NULL run has to be stepped over. That remap lives in `SortKeySketch::value_rank` and nowhere else - `quantile` and `cuts` both go through it - which is the main reason the type exists rather than a loose `(sketch, null_count)` pair.

### A sketch can lose its extremes

`tdigest.rs:210-215` reads a batch's extremes from `first()` and `last()` only, then folds them with `f64::min` / `f64::max`, which are the NaN-ignoring variants:

```rust
let maybe_min = *sorted_values.first().unwrap();
let maybe_max = *sorted_values.last().unwrap();
if self.count() > 0.0 {
    result.min = self.min.min(maybe_min);
    result.max = self.max.max(maybe_max);
```

Sorted by `total_cmp`, any batch containing a NaN has NaN at one or both ends. That element is then discarded by the fold, and because nothing between the ends is examined, the batch's real extremes go with it:

```
input        : [1.0, 5.0, NaN, -NaN, +inf, -inf]
total_cmp    : min=NaN  max=NaN   <- the true extremes
one batch    : min=NaN  max=NaN
split batches: min=1.0  max=5.0
split, no NaN: min=-inf  max=inf
```

The last line is the control: same two-batch split without the NaN, and both infinities survive. Reproduction is four lines against `datafusion-functions-aggregate-common` alone.

This matters because `cut_partitions` routes shuffle files by exactly `[sketch.min(), sketch.max()]`. A file reporting `[1.0, 5.0]` while spanning `[-inf, +inf]` is routed only to buckets overlapping 1 to 5, and the rest of its rows are not routed anywhere. There is no error, and since the outcome depends on which batch arrived first it will not reproduce consistently.

`KllSketch` has no equivalent: `absorb_slice` scans every element using `Ord` on the key, and `-NaN` / `+NaN` are simply the smallest and largest `u64`. No shortcut, and no comparison that silently ignores a value.

## Cost

Ryzen 9 9950X, 64 MiB L3. `SortKeySketch` column is the shipped path.

| Rows | Column | T-Digest | SortKeySketch | Ratio |
|---|---|---|---|---|
| 100K | `Float64` | 1.38 ms (72 M/s) | 1.23 ms (81 M/s) | 0.89x |
| 100K | `Float64`, 10% NULL | unsupported | 1.27 ms (78 M/s) | 0.92x |
| 100K | `Int64` | unsupported | 1.22 ms (82 M/s) | 0.88x |
| 100K | `Timestamp(ns, UTC)` | unsupported | 1.21 ms (83 M/s) | 0.87x |
| 1M | `Float64` | 13.8 ms (73 M/s) | 16.4 ms (61 M/s) | 1.19x |
| 1M | `Float64`, 10% NULL | unsupported | 15.5 ms (65 M/s) | 1.13x |
| 1M | `Int64` | unsupported | 16.5 ms (61 M/s) | 1.20x |
| 1M | `Timestamp(ns, UTC)` | unsupported | 16.4 ms (61 M/s) | 1.20x |

T-Digest is `Float64`-only, so the integer and temporal rows have no baseline of their own; those ratios are against T-Digest sketching a float.

Three things to read off it:

- **Column type does not move the cost.** 1.19x, 1.20x and 1.20x for float, integer and timestamp. Compaction dominates ingest and the sketch only ever sees `u64`, so the encode difference is noise. One number covers every fixed-width column.
- **NULLs are a discount.** 1.13x against 1.19x, because a tenth of the rows are counted rather than sketched.
- **The sign flips with scale.** 11% to 13% faster than T-Digest at 100K, 13% to 20% behind at 1M, as KLL's per-row cost grows with its compaction stack.

`KLL_K` is chosen for worst-case rank-error parity with T-Digest at `max_size=100` (0.0016 against 0.0021 on a uniform 1M stream), so this is a matched race rather than a cheaper sketch looking fast. Rerun the parity check with `KLL_PARITY_CHECK=1 cargo bench --bench quantile_sketch`.

## Why not widen T-Digest instead

Worth stating plainly, because for the numeric and temporal types it is a real option. You can wrap the routing expression in a `CastExpr` to `Float64` and keep T-Digest. The precision loss is genuine but small enough not to matter: an `f64` at 2020s epoch nanoseconds quantizes onto a 256 ns grid, while T-Digest's own rank error over a day-wide partition is around three minutes. Nine orders of magnitude apart. Any argument resting on quantile precision is noise, and the sort_key docs say so explicitly so nobody over-claims it later.

What is left after discounting that:

- NULLs still have nowhere to go, and that is not a precision question.
- The extremes behaviour above is independent of column type.
- Casting is not free either. It needs a `CastExpr` planted at plan time in the AQE rule, and then `RangeFilterExec`'s halos have to be expressed in cast units rather than the column's own. That is plan surgery to avoid a code path that would otherwise already exist.
- Float precision is relative, so the cast does degrade for a narrow spread at a large magnitude. A partition covering a day is unaffected; one covering 100 µs at epoch-nanosecond magnitudes is past the point where the cast costs more than the sketch does. A corner, but a real one.

Exactness does pay in one place that is not statistical: `min` and `max` are exact by construction, tracked outside the compactor so no coin flip can move them, and they are the two values `cut_partitions` routes files on. There the error bars are zero, so anything a cast rounds away is error introduced where none existed.

## Correctness of the arrangement

Arrow's row format is treated as the definition of `ORDER BY` order throughout, since it is the only encoding that folds column type, `nulls_first` and `descending` into one memcmp order, and arrow's own sort agrees with it. The float transform here is the same one `arrow_row::fixed` applies, and both reduce to `total_cmp`, which is what `ArrowNativeTypeOp::compare` uses.

`integer_keys_order_identically_to_arrow_row` asserts the two induce the same permutation over a fixture containing ±NaN, ±0.0 and both infinities, so a divergence fails a test rather than surfacing as misrouted rows.

Following that order is also why there is no NaN handling anywhere in this PR. NaN has a defined place in `total_cmp`, beyond the infinity of its own sign, so it encodes to an ordinary key at one end of the `u64` range and comparisons against it behave normally.

## Testing

29 unit tests in `sort_key`, 27 in `kll`, full `ballista-core` suite green.

The NULL remap is validated differentially against `kll_norm_u128`, the design this PR prices and rejects. Putting NULLs inside the key needs no remap at all, which makes it an independent oracle for the arithmetic. The sweep covers NULL fractions from 0 to 100 of 100 rows, both placements, ASC and DESC, 21 quantiles each.

That oracle earned its keep immediately: it found an off-by-one that eleven hand-written tests had missed. Rescaling a rank into a fraction of the value run and letting the sketch multiply it back loses it, since with 99 values rank 59 becomes `59/99` and `59/99 * 99` is `58.999...`, truncating to 58. Fixed by keeping the subtraction in integers, which is what `KllSketch::at_rank` exists for.

The worked bit-pattern table in `impl_sortable_float`'s docs is asserted by `float_key_table_in_docs_is_accurate`, so the hex cannot rot.

## What this does not do

- Does not touch `RuntimeStatsExec`, T-Digest, or any routing code.
- `SortKeySketch` has no wire format yet. That lands with its first consumer, so the encoding is designed alongside the reader rather than guessed at.
- The arrow-row tier for multi-column and variable-width keys is measured here but not built.

## Follow-ups

1. `RuntimeStatsExec` holds a `SortKeySketch` per partition, plus the proto message.
2. Cuts become sort keys rather than floats through the router, which is what lifts the `Float64` gate in `ParallelWindowRule` and lets a `Timestamp` `ORDER BY` plan at all.
3. Multi-column and `Utf8` keys via the arrow-row tier.

### DIFF
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -202,6 +202,19 @@ fn level_capacity(k: usize, num_levels: usize, height: usize) -> usize {
     raw.max(MIN_LEVEL_WIDTH)
 }
 
+/// Summarizes rather than dumping the compactor stack, which holds on the
+/// order of `3k` items and would bury whatever else a caller was printing.
+impl<T: Ord + Clone> std::fmt::Debug for KllSketch<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("KllSketch")
+            .field("k", &self.k)
+            .field("count", &self.count())
+            .field("levels", &self.levels.len())
+            .field("retained", &self.levels.iter().map(Vec::len).sum::<usize>())
+            .finish()
+    }
+}
+
 /// Clone copies the compactor stack and the tracked extremes, and gives
 /// the copy a fresh PRNG rather than duplicating the original's position.
 ///
@@ -530,6 +543,33 @@ impl<T: Ord + Clone> KllSketch<T> {
         if total_weight == 0 {
             return None;
         }
+        self.at_rank((q * total_weight as f64) as u64)
+    }
+
+    /// Return the item at `rank`, counting cumulative weight from the
+    /// smallest item up. `None` if the sketch is empty.
+    ///
+    /// Semantics: the smallest retained item whose cumulative weight is at
+    /// least `rank`. Ranks at or beyond the ends give the tracked extremes,
+    /// which bypass the compactor so coin-flip history can't move them.
+    ///
+    /// This is the primitive [`Self::quantile`] is expressed in, and the
+    /// one to prefer whenever the caller already knows the rank it wants.
+    /// Converting a known rank into a fraction and back loses it: for 99
+    /// items, rank 59 becomes `59/99`, and multiplying that back by 99
+    /// yields `58.999…`, which truncates to 58. Callers that adjust a rank
+    /// — stepping over a run of NULLs, say — must stay in integers.
+    pub fn at_rank(&self, rank: u64) -> Option<&T> {
+        let total_weight = self.count();
+        if total_weight == 0 {
+            return None;
+        }
+        if rank == 0 {
+            return self.min.as_ref();
+        }
+        if rank >= total_weight {
+            return self.max.as_ref();
+        }
         let mut pairs: Vec<(&T, u64)> = self
             .levels
             .iter()
@@ -541,17 +581,16 @@ impl<T: Ord + Clone> KllSketch<T> {
             .collect();
         pairs.sort_by_key(|(item, _)| *item);
 
-        let target = (q * total_weight as f64) as u64;
         let mut cumulative = 0u64;
         for (item, weight) in &pairs {
             cumulative += weight;
-            if cumulative >= target {
+            if cumulative >= rank {
                 return Some(*item);
             }
         }
-        // total_weight > 0 and q ≤ 1 ⇒ cumulative reaches total_weight ≥ target
-        // on the final iteration, so the loop always returns.
-        unreachable!("quantile: threshold not reached despite non-empty sketch")
+        // rank < total_weight ⇒ cumulative reaches total_weight > rank on
+        // the final iteration, so the loop always returns.
+        unreachable!("at_rank: threshold not reached despite non-empty sketch")
     }
 
     /// Compact until every level fits within its (dynamically shrinking)
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -202,6 +202,26 @@ fn level_capacity(k: usize, num_levels: usize, height: usize) -> usize {
     raw.max(MIN_LEVEL_WIDTH)
 }
 
+/// Clone copies the compactor stack and the tracked extremes, and gives
+/// the copy a fresh PRNG rather than duplicating the original's position.
+///
+/// Written by hand because `StdRng` is not `Clone`. Reseeding is the
+/// correct behaviour anyway: two sketches sharing a coin-flip sequence
+/// would correlate their compaction decisions, and every quantile the
+/// clone can answer is already determined by the state that was copied.
+impl<T: Ord + Clone> Clone for KllSketch<T> {
+    fn clone(&self) -> Self {
+        Self {
+            levels: self.levels.clone(),
+            sorted: self.sorted.clone(),
+            k: self.k,
+            rng: StdRng::seed_from_u64(rand::random::<u64>()),
+            min: self.min.clone(),
+            max: self.max.clone(),
+        }
+    }
+}
+
 impl<T: Ord + Clone> KllSketch<T> {
     /// Construct an empty sketch with top-level compactor capacity `k`,
     /// seeded from OS entropy.
@@ -443,6 +463,35 @@ impl<T: Ord + Clone> KllSketch<T> {
         }
     }
 
+    /// The stream's true minimum, or `None` if nothing has been ingested.
+    ///
+    /// Tracked outside the compactor stack, so this is exact rather than
+    /// estimated — a coin flip can evict the smallest retained item but
+    /// cannot move this.
+    pub fn min(&self) -> Option<&T> {
+        self.min.as_ref()
+    }
+
+    /// The stream's true maximum, or `None` if nothing has been ingested.
+    /// Exact for the same reason as [`Self::min`].
+    pub fn max(&self) -> Option<&T> {
+        self.max.as_ref()
+    }
+
+    /// Number of items ingested so far.
+    ///
+    /// Exact, not estimated. Compaction halves an even buffer while
+    /// doubling the survivors' weight, and leaves the odd item behind at
+    /// its original weight, so the weighted sum over levels is invariant
+    /// under compaction and equals the ingested count.
+    pub fn count(&self) -> u64 {
+        self.levels
+            .iter()
+            .enumerate()
+            .map(|(h, level)| (1u64 << h) * level.len() as u64)
+            .sum()
+    }
+
     /// Return the estimated number of items strictly less than `x`.
     ///
     /// Sum over levels `h` of `2^h · | { y ∈ levels[h] : y < x } |`.
@@ -477,12 +526,7 @@ impl<T: Ord + Clone> KllSketch<T> {
         if q == 1.0 {
             return self.max.as_ref();
         }
-        let total_weight: u64 = self
-            .levels
-            .iter()
-            .enumerate()
-            .map(|(h, level)| (1u64 << h) * level.len() as u64)
-            .sum();
+        let total_weight = self.count();
         if total_weight == 0 {
             return None;
         }
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -244,6 +244,11 @@ impl<T: Ord + Clone> KllSketch<T> {
 
     /// Construct an empty sketch with top-level compactor capacity `k`
     /// and a fixed PRNG seed. Intended for deterministic tests.
+    ///
+    /// The seed does not survive [`Clone`], which reseeds from OS entropy
+    /// to keep two sketches from correlating their compaction decisions. A
+    /// test that needs two reproducible sketches builds both with
+    /// `with_seed` rather than cloning one.
     pub fn with_seed(k: usize, seed: u64) -> Self {
         Self {
             levels: vec![Vec::new()],
--- a/ballista/core/src/kll.rs
+++ b/ballista/core/src/kll.rs
@@ -565,16 +565,29 @@ impl<T: Ord + Clone> KllSketch<T> {
     /// yields `58.999…`, which truncates to 58. Callers that adjust a rank
     /// — stepping over a run of NULLs, say — must stay in integers.
     pub fn at_rank(&self, rank: u64) -> Option<&T> {
+        self.at_ranks(&[rank]).into_iter().next().flatten()
+    }
+
+    /// Answer several ranks against one pass over the retained items,
+    /// returning one answer per entry of `ranks`, positionally.
+    ///
+    /// Semantics per rank are [`Self::at_rank`]'s exactly. The difference is
+    /// cost: resolving a rank means materializing every retained item with
+    /// its level weight and sorting them, and that work does not depend on
+    /// the rank. Asking one at a time repeats it per rank, which makes
+    /// `cuts` over P partitions P-1 sorts of the whole retained set. Here it
+    /// happens once.
+    ///
+    /// `ranks` may be in any order; the answers come back in the order
+    /// asked. Internally they are walked smallest-first so a single
+    /// cumulative sweep serves all of them.
+    pub fn at_ranks(&self, ranks: &[u64]) -> Vec<Option<&T>> {
+        let mut answers: Vec<Option<&T>> = vec![None; ranks.len()];
         let total_weight = self.count();
         if total_weight == 0 {
-            return None;
+            return answers;
         }
-        if rank == 0 {
-            return self.min.as_ref();
-        }
-        if rank >= total_weight {
-            return self.max.as_ref();
-        }
+
         let mut pairs: Vec<(&T, u64)> = self
             .levels
             .iter()
@@ -586,16 +599,38 @@ impl<T: Ord + Clone> KllSketch<T> {
             .collect();
         pairs.sort_by_key(|(item, _)| *item);
 
+        let mut slots: Vec<usize> = (0..ranks.len()).collect();
+        slots.sort_by_key(|&slot| ranks[slot]);
+
+        // Resumed across ranks rather than restarted: `cumulative` is
+        // monotone and so are the ranks in `slots` order, so a rank already
+        // covered by the current item answers with that same item.
+        let mut consumed = 0usize;
         let mut cumulative = 0u64;
-        for (item, weight) in &pairs {
-            cumulative += weight;
-            if cumulative >= rank {
-                return Some(*item);
+        let mut current: Option<&T> = None;
+        for slot in slots {
+            let rank = ranks[slot];
+            // The extremes are tracked outside the compactor, so no
+            // coin-flip history can move them.
+            if rank == 0 {
+                answers[slot] = self.min.as_ref();
+                continue;
             }
+            if rank >= total_weight {
+                answers[slot] = self.max.as_ref();
+                continue;
+            }
+            while cumulative < rank {
+                let (item, weight) = pairs[consumed];
+                cumulative += weight;
+                current = Some(item);
+                consumed += 1;
+            }
+            // `rank < total_weight`, which is the sum of every pair's
+            // weight, so the sweep above always consumed at least one item.
+            answers[slot] = current;
         }
-        // rank < total_weight ⇒ cumulative reaches total_weight > rank on
-        // the final iteration, so the loop always returns.
-        unreachable!("at_rank: threshold not reached despite non-empty sketch")
+        answers
     }
 
     /// Compact until every level fits within its (dynamically shrinking)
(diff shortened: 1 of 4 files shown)


avantgardnerio:
@phillipleblanc purely additive PR, TDigest untouched. One thing I found while writing it though: TDigest can lose a batch's min/max entirely when the batch contains a NaN, and cut_partitions routes files on exactly those. Repro in the description.

Does that change your view on keeping both vs migrating off it later? Happy either way, just want to know before I build more on top.

andygrove:
> **Disclaimer:** This review was produced by an LLM (Claude Code) at my request. I've skimmed it, but treat the reasoning as unverified assistance rather than a considered human review — check the claims before acting on them. The test/clippy/fmt results quoted below were actually run against the PR head (`52b5fac`) locally.

## Overview

Purely additive, and the claim holds up — nothing in the tree constructs `SortKeyCodec` or `SortKeySketch`, and `RuntimeStatsExec` and T-Digest are untouched. Three parts:

- **`ballista/core/src/sort_key.rs`** (new): `SortKeyCodec` maps one fixed-width `ORDER BY` key to an order-preserving `u64` and back (24 arrow types: ints, `Float32/64`, and the `i32`/`i64`-backed temporals), folding `descending` into the key as a bitwise NOT. `SortKeySketch` pairs it with `KllSketch<u64>` plus an out-of-band NULL count.
- **`kll.rs`**: adds `min()`, `max()`, `count()`, `at_rank()`, hand-written `Clone` and `Debug`. `quantile` is refactored onto `at_rank`.
- **`quantile_sketch.rs`**: new arms measuring the shipped path plus the priced-and-dropped alternatives.

## Verification

| Check | Result |
|---|---|
| `cargo test -p ballista-core --lib sort_key` | 28 passed |
| `cargo test -p ballista-core --lib` | 303 passed, 2 failed |
| `cargo clippy -p ballista-core --all-targets` | clean |
| `cargo fmt --all --check` | clean |
| `cargo check --benches` | clean |

The two failures are `utils::tests::test_create_grpc_server_incoming_binds_eagerly` and `..._port_in_use` — socket-binding tests failing under a sandbox, in a module this PR doesn't touch. Not attributable to this change.

The NULL rank remap was checked independently against nulls-in-key semantics for both placements and all four boundary cases (`rank == 0`, inside the run, exactly on the run's end, `rank >= total`); it agrees. The `at_rank`-in-integers argument is correct, and the differential oracle is the right way to hold it. f32 encoding traced by hand in both directions — correct.

## Findings

**1. All-NULL sketch reports `min()`/`max()` as `None` (`sort_key.rs:588`)**

`extreme` only produces a typed NULL for the end the NULL run occupies. For an all-NULL column the *other* end falls through to the empty value sketch and returns `None` — which both methods document as "nothing was observed". Measured on 8 NULL `Int64` rows with `nulls_first: true`:

```
count()==8   min()==Some(Int64(NULL))   max()==None
```

and it inverts under `nulls_first: false`. `quantile()` gets this input right via its `values == 0` branch (`an_all_null_column_answers_null_everywhere` covers it), so the two APIs disagree. Since the module docs note that `cut_partitions` routes shuffle files on `[sketch.min(), sketch.max()]`, the announced follow-up consumer would see a half-absent range for an all-NULL partition.

Suggested fix in `extreme`: when `value_extreme` is `None` and `null_count > 0`, return the typed NULL regardless of which end was asked for.

**2. 20 of 24 dispatched types have no round-trip test (`sort_key.rs:293`)**

`dispatch_sortable!` admits 24 types; the tests exercise `Int64`, `UInt64`, `Float64` and `TimestampNanosecond`. That leaves the entire `impl_sortable_float!(f32, u32, 32)` instantiation unpinned — a distinct path from the f64 one, since `to_key` zero-extends a 32-bit key and `from_key` truncates, while under DESC the `orient` NOT applies over all 64 bits. Same for the narrow ints and every temporal but one. A loop asserting encode-monotonicity plus decode round-trip over one representative array per dispatched type would also make `try_new`'s allowlist and the decode arms self-checking.

## Smaller notes

- **`cuts()` (`sort_key.rs:673`) uses `.flatten()`**, silently dropping `None` quantiles and returning fewer than `partitions - 1` entries. Safe today (`quantile` returns `None` only when `total == 0`, all-or-nothing), but a consumer indexing cuts by partition has no signal if that invariant ever changes. `collect::<Option<Vec<_>>>()` or an explicit early return would make it structural.
- **`KLL_K = 800` is duplicated** — private in `sort_key.rs`, redeclared in the bench. The bench's claim that `kll_norm_u64` and `sort_key_sketch_f64` are comparable depends on them staying equal, with nothing enforcing it.
- **`encode` requires exact `DataType` equality**, so a codec built for `Timestamp(ns, Some("UTC"))` errors on an array tagged `Some("+00:00")` even though the timezone is irrelevant to the key. Strictness is defensible; worth knowing when the consumer lands.
- **`Clone` for `KllSketch` reseeds from `rand::random()`**, so cloning is non-deterministic. The decorrelation argument is sound and `new()` already seeds from OS entropy, so no regression — just a caveat for any future test that clones a sketch expecting reproducibility.
- **`quantile` computes `count()` and then `at_rank` recomputes it**, and `at_rank` allocates and sorts the ~3k retained pairs per call, so `cuts(P)` is O(P · m log m). Negligible against ingest, and pre-existing in shape.

## Quality

Encoding, direction handling, and NULL placement are each isolated where they belong, and the arrow-row agreement test is the right invariant to assert rather than argue. The float bit-table pinned by `float_key_table_in_docs_is_accurate` is a nice touch — that comment can't rot. Comment density is high relative to the rest of `ballista/core`, but it's load-bearing (the rank-remap off-by-one, the NaN/`total_cmp` reasoning, the priced-and-dropped alternatives) rather than restating code.

Both findings are contained to the two spots above; neither undermines the additive-only claim.

ballista/core/src/sort_key.rs
```diff
@@ -0,0 +1,1465 @@
...
+        nulls_are_on_this_end: bool,
+        value_extreme: Option<&u64>,
+    ) -> Result<Option<ScalarValue>> {
+        // With no values at all the run is unbounded on both sides, so the
+        // end it does not nominally occupy is a NULL too. Answering `None`
+        // there would claim nothing was observed while `count` says
+        // otherwise, and would hand `cut_partitions` half a range.
+        if self.null_count > 0 && (nulls_are_on_this_end || value_extreme.is_none()) {
+            return Ok(Some(self.codec.null_value()?));
+        }
```
avantgardnerio, line 601: > All-NULL sketch reports `min()`/`max()` as `None`

Confirmed and fixed. Reproduced as a failing test first, `an_all_null_column_is_null_at_both_extremes`, which failed on exactly the case you measured:

```
assertion `left == right` failed: nulls_first=true
  left: None
 right: Some(Int64(NULL))
```

> when `value_extreme` is `None` and `null_count > 0`, return the typed NULL regardless of which end was asked for

That is the fix, with the reason inline: with no values the NULL run is unbounded on both sides, so the end it does not nominally occupy is a NULL too. `extremes_account_for_where_nulls_sort` still pins the partial case, where the value end keeps its value.

ballista/core/src/sort_key.rs
```diff
@@ -0,0 +1,1569 @@
...
+    /// The narrow widths are their own code rather than a special case of
+    /// the 64-bit ones. `impl_sortable_float!(f32, u32, 32)` zero-extends a
+    /// 32-bit key that DESC then inverts across all 64 bits, and the narrow
+    /// signed arms sign-extend on the way out and truncate on the way back.
+    #[test]
+    fn every_dispatched_type_encodes_in_order_and_decodes_back() {
+        check_dispatched::<Int8Type>(vec![i8::MIN, -1, 0, 1, i8::MAX]);
+        check_dispatched::<Int16Type>(vec![i16::MIN, -1, 0, 1, i16::MAX]);
+        check_dispatched::<Int32Type>(vec![i32::MIN, -1, 0, 1, i32::MAX]);
+        check_dispatched::<Int64Type>(vec![i64::MIN, -1, 0, 1, i64::MAX]);
```
avantgardnerio, line 1497: > 20 of 24 dispatched types have no round-trip test

Fixed. `every_dispatched_type_encodes_in_order_and_decodes_back` walks one fixture per dispatched type: `try_new` accepts it, keys are strictly monotone in the direction asked for, and every key decodes back to the value it came from. Float fixtures carry both NaNs, both infinities and both zeros.

> That leaves the entire `impl_sortable_float!(f32, u32, 32)` instantiation unpinned

Mutation-tested rather than trusted green. Changing it to `impl_sortable_float!(f32, u32, 31)` fails the new test at `Float32 descending=false: 0x00000000ff800000 then 0x00000000c0000000`, and fails nothing else in the file. The arm was correct; it had no coverage.

ballista/core/src/sort_key.rs
```diff
@@ -0,0 +1,1581 @@
...
+    ///
+    /// Empty when `partitions < 2` or nothing was observed, and otherwise
+    /// exactly `partitions - 1` long so a caller can index it by output
+    /// partition. Entries may repeat where one value dominates, and may be
+    /// typed NULLs where the NULL run spans a boundary; both are faithful
+    /// answers about a skewed distribution rather than errors.
+    pub fn cuts(&self, partitions: usize) -> Result<Vec<ScalarValue>> {
+        if partitions < 2 || self.count() == 0 {
+            return Ok(Vec::new());
+        }
```
avantgardnerio, line 686: > uses `.flatten()`, silently dropping `None` quantiles and returning fewer than `partitions - 1` entries

Done, both halves. `count() == 0` early-returns empty, and a missing quantile past that point is now an error rather than a dropped entry, so the result is empty or exactly `partitions - 1`.

> Safe today (`quantile` returns `None` only when `total == 0`, all-or-nothing)

Confirmed while wiring it. `KllSketch::at_rank` clamps to `min`/`max` at both ends and returns `None` only on an empty sketch, which `quantile`'s `values == 0` branch already answers first. So the new error arm is unreachable today, and no test can tell the two versions apart. The change is the signal, not a behaviour fix.

benchmarks/benches/quantile_sketch.rs
```diff
@@ -38,9 +144,26 @@
...
+/// claim. Picked for worst-case rank-error parity with TDigest at
+/// `TDIGEST_MAX_SIZE=100` on the uniform 1M stream this bench feeds:
+///
+/// | sketch                       | worst rank err (9 deciles) |
+/// |------------------------------|---------------------------:|
+/// | TDigest max_size=100         |                     0.0021 |
+/// | KLL k=800                    |                     0.0016 |
+///
+/// Rerun via `KLL_PARITY_CHECK=1 cargo bench --bench quantile_sketch`.
+use ballista_core::sort_key::KLL_K;
```
avantgardnerio, line 175: > **`KLL_K = 800` is duplicated** — private in `sort_key.rs`, redeclared in the bench

Fixed. It is `pub` in `sort_key` now and the bench imports it, so the arms cannot be sized differently. The parity table moved onto the import.

ballista/core/src/sort_key.rs
```diff
@@ -0,0 +1,1588 @@
...
+    /// Errors if `array`'s type doesn't match the one this codec was built
+    /// for, which would mean the routing expression changed type between
+    /// planning and execution.
+    ///
+    /// The match is exact `DataType` equality, so a codec built for
+    /// `Timestamp(ns, Some("UTC"))` rejects an array tagged `Some("+00:00")`
+    /// even though the two encode to identical keys. Strict because the
+    /// codec's type is what [`Self::decode`] rebuilds a `ScalarValue` from,
+    /// so accepting an array of a type the codec doesn't carry would
+    /// relabel every cut it later produces.
```
avantgardnerio, line 410: > **`encode` requires exact `DataType` equality**

Documented rather than changed. The doc previously said the types must "match", which reads as a semantic type change, and the `data_type` field doc right above says the timezone is deliberately *not* in the key, so the rejection looked like a bug. It now states that the check is `DataType` equality and gives the reason: the codec's type is what `decode` rebuilds a `ScalarValue` from, so accepting an array of a type the codec doesn't carry would relabel every cut it produces.

> Strictness is defensible; worth knowing when the consumer lands.

Agreed, and left open. Codec and array come from the same routing expression today, so they cannot diverge. The case that would force this is the consumer building the codec in the scheduler from the table schema while the array comes from a scan whose file metadata tags the zone differently. If that turns up, the principled relaxation is comparing the dispatch arm rather than the full type, since the key only depends on the primitive. Deciding it there, where there is something to test against.

ballista/core/src/kll.rs
```diff
@@ -202,6 +202,39 @@ fn level_capacity(k: usize, num_levels: usize, height: usize) -> usize {
...
+}
+
+/// Clone copies the compactor stack and the tracked extremes, and gives
+/// the copy a fresh PRNG rather than duplicating the original's position.
+///
+/// Written by hand because `StdRng` is not `Clone`. Reseeding is the
+/// correct behaviour anyway: two sketches sharing a coin-flip sequence
+/// would correlate their compaction decisions, and every quantile the
+/// clone can answer is already determined by the state that was copied.
+impl<T: Ord + Clone> Clone for KllSketch<T> {
```
avantgardnerio, line 225: > **`Clone` for `KllSketch` reseeds from `rand::random()`**, so cloning is non-deterministic

I agree this is a tough choice. I've updated the docs to make the behavior explicit. I think it's safest if `clone()` re-seeds. I can add a `clone_with_seed()` test-only method if you think it's merited.

ballista/core/src/kll.rs
```diff
@@ -477,15 +544,50 @@ impl<T: Ord + Clone> KllSketch<T> {
...
+    /// cost: resolving a rank means materializing every retained item with
+    /// its level weight and sorting them, and that work does not depend on
+    /// the rank. Asking one at a time repeats it per rank, which makes
+    /// `cuts` over P partitions P-1 sorts of the whole retained set. Here it
+    /// happens once.
+    ///
+    /// `ranks` may be in any order; the answers come back in the order
+    /// asked. Internally they are walked smallest-first so a single
+    /// cumulative sweep serves all of them.
+    pub fn at_ranks(&self, ranks: &[u64]) -> Vec<Option<&T>> {
```
avantgardnerio, line 584: > `at_rank` allocates and sorts the ~3k retained pairs per call, so `cuts(P)` is O(P · m log m)

> Negligible against ingest, and pre-existing in shape.

Benched it rather than taking that on faith, and it was not negligible. `runtime_stats_cuts` is new and measures cut extraction against the ingest that built the sketch:

| n | `cuts(64)` | `cuts(256)` |
|---|---|---|
| 100K | 30% of ingest | 123% of ingest |
| 1M | 1.7% of ingest | 6.8% of ingest |

It scales the wrong way too: a smaller stage has less ingest to hide the sorts behind, so `cuts(256)` at 100K cost more than building the sketch did.

`KllSketch::at_ranks` now resolves a whole batch against one sorted pass, and `SortKeySketch::cuts` makes a single call. At n=1M that takes `cuts(256)` from 1.13 ms to 16.6 µs and `cuts(64)` from 279 µs to 7.56 µs, leaving cut extraction at 0.045% of ingest. The P-dependence is close to flat now, since only the scatter over the answers grows with P.

avantgardnerio:
Thanks so much for that review @andygrove ! As usual, you caught some real stuff. I believe I have addressed all of your concerns in threads with links to code. Please re-review when you get a chance.

phillipleblanc:
> @phillipleblanc purely additive PR, TDigest untouched. One thing I found while writing it though: TDigest can lose a batch's min/max entirely when the batch contains a NaN, and cut_partitions routes files on exactly those. Repro in the description.
> 
> Does that change your view on keeping both vs migrating off it later? Happy either way, just want to know before I build more on top.

Ok, then yeah I don't think it makes sense to preserve compatibility for TDigest once KLL is ready.

phillipleblanc (approved):
Looks good to me

PR #2297, opened 2026-08-14
test: centralize physical plan snapshot assertions

### LINKED ISSUE #2296: use `insta::assert_snapshot!` for all plan assertions

scheduler uses macro defined in 

https://github.com/apache/datafusion-ballista/blob/b6b6a21b78c7deed9a65b2ed7e535ff6aeea55df/ballista/scheduler/src/state/aqe/mod.rs#L1531

for all plan assertions in AQE submodule,

https://github.com/apache/datafusion-ballista/blob/b6b6a21b78c7deed9a65b2ed7e535ff6aeea55df/ballista/scheduler/src/state/aqe/test/alter_stages.rs#L61

use  of `insta::assert_snapshot!` crates a standardised way of plan assertions ad simplifies changes when/if plan changes. 

i would like to propose to move the macro to `core` module and use it as a standard way of plan assertion, in all tests 

- move macro to `core` module and use it for all plan assertions
- consider creating `assert_batch` macro which would use `insta::assert_snapshot!` for batch assertions 

**Additional context**

floze-the-genius:

Closes #2296.

# Rationale for this change

Physical-plan snapshots are useful outside the scheduler AQE module. Keeping the helper in that module forces unrelated tests to depend on a scheduler-local test macro and leaves some literal plan checks using ad hoc string equality.

- Move `assert_plan!` to `ballista-core` so Ballista crates can share one physical-plan snapshot helper.
- Update existing scheduler tests to import the macro from `ballista_core`.
- Convert the remaining literal physical-plan equality checks in `planner.rs` to inline snapshots.
- Leave semantic substring checks and dynamic before/after plan comparisons unchanged.

No. This only standardizes test assertions.

# Testing

- `git diff --check`
- The targeted scheduler test profile compiled successfully locally. Test execution was not completed because the local APFS volume ran out of space during post-link stripping; CI is the full test authority for this run.

AI assistance: OpenAI Codex was used to help implement and test this change.

### DIFF
--- a/ballista/core/src/lib.rs
+++ b/ballista/core/src/lib.rs
@@ -46,6 +46,22 @@ pub fn print_version() {
     println!("Ballista version: {BALLISTA_VERSION}")
 }
 
+/// Asserts an indented physical plan against an inline snapshot.
+///
+/// Keeping plan assertions behind one macro makes plan changes easier to
+/// review and update consistently across Ballista crates.
+#[macro_export]
+macro_rules! assert_plan {
+    ($plan:expr, @ $expected_lines:literal $(,)?) => {{
+        let plan = datafusion::physical_plan::displayable($plan)
+            .indent(true)
+            .to_string();
+        let actual_lines = plan.trim();
+
+        insta::assert_snapshot!(actual_lines, @ $expected_lines);
+    }};
+}
+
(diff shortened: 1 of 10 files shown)


andygrove (approved):
Makes sense to me. Thanks @floze-the-genius

PR #2300, opened 2026-08-14
chore: prune stale NOTICE.txt entries, add mini-redis attribution

andygrove:
> *Disclaimer: this PR was prepared by an LLM (Claude Code) at my request, against `ad129930`. The claim that each removed entry has no counterpart in the tree was checked by grepping the repo, and the mini-redis copyright line was fetched from upstream's LICENSE rather than recalled — but please treat the licensing judgement calls as suggestions and push back where they're wrong.*

N/A — housekeeping, no issue filed.

# Rationale for this change

`NOTICE.txt` was inherited when Ballista was split out of the Arrow repo, and still lists bundled works that Ballista does not contain. Everything removed below is an Arrow C++ or build-system artifact with no counterpart in this pure-Rust repo:

| Removed entry | Why it doesn't apply |
| --- | --- |
| LLVM | No C++; the only `llvm` match in the tree is a passing mention in a doc comment (`ballista/core/src/kll.rs:337`) |
| google-lint (cpplint) | No C++ to lint |
| CMake (two entries) | No CMake build |
| Jekyll | Docs are Sphinx; the only `jekyll` match is `touch .nojekyll` in `.github/workflows/docs.yaml` |
| Apache Kudu | No Kudu code |
| Apache ORC | No ORC code or ORC support |

Carrying entries for software we don't ship makes the file misleading about what's actually bundled, which is the one thing `NOTICE.txt` exists to communicate.

**`NOTICE.txt`** — reduced to the required ASF attribution, and the copyright end year refreshed to 2026:

```
Apache DataFusion Ballista
Copyright 2016-2026 The Apache Software Foundation

This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).
```

**`LICENSE.txt`** — while auditing what *is* bundled, `ballista/executor/src/shutdown.rs` turned out to be based on Tokio's [mini-redis](https://github.com/tokio-rs/mini-redis) example (MIT) and was not attributed anywhere. Added an entry after the existing Apache Aurora one, with the full MIT text inline, since MIT requires the notice and permission text travel with the code — a URL alone isn't sufficient the way it is for the Apache-2.0 Aurora entry.

The split follows [ASF policy](https://www.apache.org/legal/src-headers.html#notice): `NOTICE.txt` carries only required ASF attribution, and third-party bundled works belong in `LICENSE.txt`. Two other copied-from sites were checked and need nothing — `ballista-cli/src/tui/app.rs` and `.../aqe/execution_plan/dynamic_join.rs` both copy from DataFusion, which is ASF Apache-2.0 with a plain NOTICE.

No code or API changes. The contents of `NOTICE.txt` and `LICENSE.txt` ship in release artifacts, so this is user-visible in that narrow sense.


milenkovicm (approved):
Thanks @andygrove

PR #2301, opened 2026-08-15
chore(deps): bump the DataFusion pin to the 55.0.0-rc3 tag

andygrove:
> Drafted with the help of an LLM (Claude Code), reviewed by me before posting.

Follow-up to #2241, which pinned DataFusion to the `55.0.0-rc2` tag.

# Rationale for this change

`55.0.0-rc3` has been cut, so advance the pin to the latest release candidate ahead of the final 55.0.0 release.

- `[patch.crates-io]` in the root `Cargo.toml` now points at tag `55.0.0-rc3` (`d5552342`)
- `Cargo.lock` refreshed; no non-DataFusion dependency moves
- The `python/` workspace is untouched — it still consumes released 54.0.0 crates

# Are these changes tested?

Locally on macOS (aarch64):

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets` — clean, no new warnings
- `cargo test --workspace` — **the `ballista` client tests fail**: all 18 tests in `ballista/client/tests/context_basic.rs` error out at context creation with `Configuration("IO error: Operation not permitted (os error 1)")`. The same tests pass on `main` (rc2) in the same environment, so this looks like a genuine behavior change between rc2 and rc3, not flakiness. I have not yet root-caused it — leaving CI to confirm the failure independently before digging in.

No.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2222,7 +2222,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "55.0.0"
-source = "git+https://github.com/apache/datafusion.git?tag=55.0.0-rc2#209fd9406890d97d48c0fc396ce37a5e363270b4"
+source = "git+https://github.com/apache/datafusion.git?tag=55.0.0-rc3#d5552342012888b7d1a3ab88d92e3d292fc0cde0"
 dependencies = [
  "arrow",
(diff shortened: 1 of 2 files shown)


avantgardnerio (approved):
LGTM!

PR #2305, opened 2026-08-15
fix(scheduler): label every operator in the DOT graph

### LINKED ISSUE #2303: DOT graph renders SortMergeJoinExec and SortPreservingMergeExec as "Unknown Operator"
## Describe the bug

`ExecutionGraphDot` labels a physical operator by walking a chain of `downcast_ref` arms in `ballista/scheduler/src/state/execution_graph_dot.rs`, falling through to a literal `"Unknown Operator"` at line 353. Two operators that Ballista plans routinely are missing from that chain:

- **`SortMergeJoinExec`** — the chain handles `HashJoinExec` (line 292) and `CrossJoinExec` (line 311), but no sort-merge join.
- **`SortPreservingMergeExec`** — the chain handles `SortExec` (line 244) and `CoalescePartitionsExec` (line 282), but not the order-preserving merge.

This is not an edge case. `new_with_ballista()` sets `datafusion.optimizer.prefer_hash_join = false` (`ballista/core/src/extension.rs:840`), deliberately, because DataFusion's hash join cannot spill (#1648). Sort-merge is therefore the *default* join for Ballista, so an ordinary joined query renders an unlabeled box where the join should be. `SortPreservingMergeExec` typically sits at the top of a sorted plan, so sorted queries lose their final operator too.

The graph still has the right shape and edges. Only the label is lost, which makes the node useless for the thing the graph exists for: seeing what a stage actually does.

## To Reproduce

Against a scheduler built with `rest-api`, with `nation` and `region` registered as Parquet tables:

```sql
select r.r_name, count(*) as num_nations
from nation n join region r on n.n_regionkey = r.r_regionkey
group by r.r_name
order by r.r_name;
```

Then fetch the graph:

```shell
curl -s localhost:50050/api/job/<job_id>/dot
```

Two of the nodes come back unlabeled — one per unhandled operator:

```
	subgraph cluster0 {
		label = "Stage 1 [Successful]";
		stage_1_0 [shape=box, label="SortShuffleWriter [1 partitions]"]
		stage_1_0_0 [shape=box, label="Aggregate
groupBy=[r_name@0]
aggr=[count(Int64(1))]"]
		stage_1_0_0_0 [shape=box, label="Projection: ..."]
		stage_1_0_0_0_0 [shape=box, label="Unknown Operator"]
		stage_1_0_0_0_0_0 [shape=box, label="Sort: n_regionkey@0 NULLS FIRST"]
		stage_1_0_0_0_0_0_0 [shape=box, label="DataSourceExec: (...) [1 partitions]"]
...
	subgraph cluster2 {
		label = "Stage 3 [Successful]";
		stage_3_0 [shape=box, label="ShuffleWriter [1 partitions]"]
		stage_3_0_0 [shape=box, label="Unknown Operator"]
		stage_3_0_0_0 [shape=box, label="ShuffleReader [16 partitions]"]
```

The fallback arm logs what it could not name, so running the scheduler with

```shell
RUST_LOG=info,ballista_scheduler::state::execution_graph_dot=debug
```

identifies them directly:

```
DEBUG ballista_scheduler::state::execution_graph_dot: Unknown physical operator when producing DOT graph: SortMergeJoinExec
DEBUG ballista_scheduler::state::execution_graph_dot: Unknown physical operator when producing DOT graph: SortPreservingMergeExec
```

The same two nodes appear unlabeled in the TUI's job graph popup (`g` on a completed job), which is where this is likely to be noticed.

## Expected behavior

Both operators are labeled, in the style of the surrounding arms: the join naming its join keys the way the `HashJoinExec` arm does with `join_expr`, and the merge naming its sort expressions the way the `SortExec` arm does.

## Additional context

Worth considering alongside the specific fix: the fallback is silent at `info` level, so an operator dropping out of the chain shows up only as a blank box in a UI, and only if someone happens to look. DataFusion adds and renames physical operators every release, so this chain drifts by default rather than by accident. A test asserting that no node in a representative plan's graph is labeled `"Unknown Operator"` would catch the next one at CI time instead of in a screenshot.

Found while verifying the history server in #2265. Not caused by it — the history server relays the scheduler's stored bytes, and the two graphs are byte-identical. It reproduces against a live scheduler on `main`.

andygrove:

Closes #2303.

# Rationale for this change

`ExecutionGraphDot` named each physical operator by walking a chain of `downcast_ref` arms and fell through to a literal `"Unknown Operator"`. `SortMergeJoinExec` and `SortPreservingMergeExec` were both missing from that chain, so an ordinary joined or sorted query rendered an unlabeled box where the join should be, and lost the operator at the top of a sorted plan. Sort-merge is Ballista's default join because `new_with_ballista()` sets `prefer_hash_join = false`, so this was not an edge case.

Adding two more arms would fix the two reported operators and leave the underlying problem in place. DataFusion adds and renames physical operators every release, so a hand-maintained chain drifts by default, and it drifts silently: the fallback logs at `debug`, so the only visible symptom is a blank box in a UI.

`DisplayAs` is a supertrait of `ExecutionPlan`, so every operator already knows how to describe itself, and that description is exactly what `EXPLAIN` prints. Labeling from it removes the fall-through case entirely, so no operator can render blank again, and it keeps the graph consistent with the plan text users already read.

- Label a node as the operator's `name()` on the first line and its `DisplayFormatType::Default` rendering on the second, with the repeated name stripped from the detail. Two lines keeps the boxes narrow.
- Drop the fifteen-arm `downcast_ref` chain and the `"Unknown Operator"` fallback, along with the now-unused `format_partitioning` and `format_expr_list` helpers.
- Keep two special cases, only where Ballista knows better than the default: `DataSourceExec`, whose default rendering lists every file group and is far too long for a graph node, so the existing `get_file_scan` summary is retained; and `ShuffleWriterExec`, whose rendering is the constant `partitioning: None` because it never repartitions, so the input partition count is shown instead. Both are enhancements rather than requirements, so an operator dropping out of them degrades to the DataFusion default rather than to nothing.
- Add `dot_labels_every_operator`, which plans a join and an `ORDER BY` under `new_with_ballista()` and asserts both `SortMergeJoinExec` and `SortPreservingMergeExec` appear in the graph and that no node carries an empty label.
- Update the four existing golden strings for the new labels.

The node labels in `/api/job/<id>/dot`, `/api/job/<id>/stage/<n>/dot`, and the TUI job graph popup change. Nothing was lost. Names are now the DataFusion type names (`HashJoinExec` rather than `HashJoin`), and the labels carry more detail than before: join type and mode on joins, output partitioning on the sort shuffle writer and on unresolved shuffles. No public API changes.

### DIFF
--- a/ballista/scheduler/src/state/execution_graph_dot.rs
+++ b/ballista/scheduler/src/state/execution_graph_dot.rs
@@ -19,30 +19,16 @@
 
 use crate::state::execution_graph::ExecutionGraph;
 use ballista_core::execution_plans::{
-    ShuffleReaderExec, ShuffleWriterExec, SortShuffleWriterExec, UnresolvedShuffleExec,
+    ShuffleReaderExec, ShuffleWriterExec, UnresolvedShuffleExec,
 };
 use datafusion::datasource::listing::PartitionedFile;
 use datafusion::datasource::memory::MemorySourceConfig;
 use datafusion::datasource::physical_plan::FileScanConfig;
 use datafusion::datasource::source::DataSourceExec;
-use datafusion::physical_plan::aggregates::AggregateExec;
-#[allow(deprecated)]
-use datafusion::physical_plan::coalesce_batches::CoalesceBatchesExec;
-use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
-use datafusion::physical_plan::filter::FilterExec;
-use datafusion::physical_plan::joins::CrossJoinExec;
-use datafusion::physical_plan::joins::HashJoinExec;
-use datafusion::physical_plan::limit::{GlobalLimitExec, LocalLimitExec};
-use datafusion::physical_plan::projection::ProjectionExec;
-use datafusion::physical_plan::repartition::RepartitionExec;
-use datafusion::physical_plan::sorts::sort::SortExec;
-use datafusion::physical_plan::union::UnionExec;
-use datafusion::physical_plan::{ExecutionPlan, Partitioning, PhysicalExpr};
-use log::debug;
+use datafusion::physical_plan::{DisplayFormatType, ExecutionPlan};
 use object_store::path::Path;
 use std::collections::HashMap;
 use std::fmt::{self, Write};
-use std::sync::Arc;
 
 /// Utility for producing dot diagrams from execution graphs
 pub struct ExecutionGraphDot<'a> {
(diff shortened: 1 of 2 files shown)


ballista/scheduler/src/state/execution_graph_dot.rs
```diff
@@ -510,37 +424,45 @@ filter_expr="]
         let expected = r#"digraph G {
 	subgraph cluster0 {
 		label = "Stage 1 [Resolved]";
-		stage_1_0 [shape=box, label="SortShuffleWriter [2 partitions]"]
-		stage_1_0_0 [shape=box, label="DataSourceExec: (Memory) [2 partitions]"]
+		stage_1_0 [shape=box, label="SortShuffleWriterExec
+partitioning=Hash([a@0], 48)"]
```
avantgardnerio, line 428: Claude had an interesting comment about how this is _output_ partitioning, and it isn't always easy to determine _input_ partitioning. Not a blocker.

avantgardnerio (approved):

andygrove:
Thanks for the review @avantgardnerio!

PR #2307, opened 2026-08-15
fix(core): report the real partitioning from ShuffleWriterExec

### LINKED ISSUE #2306: ShuffleWriterExec renders a hardcoded "partitioning: None" instead of its real partitioning
## Describe the bug

`ShuffleWriterExec`'s `DisplayAs` writes a string literal rather than a value (`ballista/core/src/execution_plans/shuffle_writer.rs:624`):

```rust
// "None" is retained for plan-shape stability: this writer never
// repartitions, so the value can only ever be None.
DisplayFormatType::Default | DisplayFormatType::Verbose => {
    write!(f, "ShuffleWriterExec: partitioning: None")
}
```

It used to print a value. The writer carried a `shuffle_output_partitioning: Option<Partitioning>` field, and #2106 ("Remove the hash-partitioning shuffle writer") deleted it, leaving `shuffle_output_partitioning()` as a method that returns `None` unconditionally (line 803). Rather than drop the token, that PR froze it as a literal so the 22 approved TPC-H plan-stability goldens would not churn in the same diff. That was a reasonable expedient at the time, but the comment reads as a permanent justification rather than a follow-up, and it has stayed frozen since.

Three problems with the output as it stands:

1. **It carries no information, and is kept because it carries none.** A constant cannot drift, so it cannot break a golden file. That is the golden files driving the display rather than the other way round.
2. **It is misleading.** "partitioning: None" reads as "no output partitioning", which is not true. The writer preserves its input partitioning, and `try_new` sets its own `PlanProperties` to exactly the child's output partitioning (line 442). The `None` refers to the absence of a *repartitioning scheme*, which is an internal distinction a plan reader has no way to infer.
3. **The useful value is available and not shown.** `self.properties().output_partitioning()` is the real partitioning, including the partition count, and it renders exactly the way `SortShuffleWriterExec` already renders its own (`partitioning=Hash([c_custkey@0], 16)`).

Point 3 has a concrete downstream cost. In #2305 the DOT graph was changed to label nodes from each operator's `DisplayAs`, which works for every operator except this one: because the text is a constant, that PR has to keep a hand-written special case for `ShuffleWriterExec` just to recover the partition count. A display impl that forces consumers to special-case around it is not doing its job.

## To Reproduce

Every approved golden shows it, two or three times per query:

```shell
grep -c "ShuffleWriterExec: partitioning: None" ballista/scheduler/tests/tpch_plan_stability/approved/q1.txt
```

```
ShuffleWriterExec: partitioning: None
```

while the sibling writer in the same file reports something real:

```
SortShuffleWriterExec: partitioning=Hash([c_custkey@0], 16)
```

## Expected behavior

`ShuffleWriterExec` reports its actual output partitioning, the same as `SortShuffleWriterExec` does, so `EXPLAIN` output and the plan-stability goldens say something true and the partition count is visible.

## Additional context

The goldens regenerate with `BALLISTA_GENERATE_GOLDEN=1 cargo test -p ballista-scheduler --test tpch_plan_stability`, and the harness plans against a synthetic stats table, so no TPC-H data is needed. The resulting diff is large but mechanical.

Two related observations, not proposed as part of the fix:

- The two writers use different separators, `partitioning: ` versus `partitioning=`, so they read inconsistently side by side in the same plan.
- `ShuffleWriterExec` also has a plain `impl std::fmt::Display` (line 384) printing job id, stage id, work dir and the full child plan. Two display paths with very different content, and the informative one is not the one `EXPLAIN` uses.

andygrove:

Closes #2306.

# Rationale for this change

`ShuffleWriterExec`'s `DisplayAs` wrote a string literal rather than a value:

```rust
// "None" is retained for plan-shape stability: this writer never
// repartitions, so the value can only ever be None.
write!(f, "ShuffleWriterExec: partitioning: None")
```

It used to print a value. #2106 removed the `shuffle_output_partitioning` field along with the hash-partitioning writer, and the token was frozen as a literal so the 22 approved TPC-H goldens would not churn in the same diff. That was a fine expedient, but it left the output carrying no information and kept *because* it carries none: a constant cannot drift, so it cannot break a golden file.

It was also misleading. The `None` refers to the absence of a repartitioning scheme, an internal distinction a plan reader cannot infer. The writer preserves its input's partitioning, and `try_new` already sets its `PlanProperties` to exactly that, so the real value was available the whole time.

There was a downstream cost too. #2305 labels DOT graph nodes from each operator's `DisplayAs`, which works everywhere except here: because the text was a constant, that PR had to keep a hand-written special case for `ShuffleWriterExec` to recover the partition count.

- Render `self.properties().output_partitioning()` in both `Default`/`Verbose` and `TreeRender`, matching how `SortShuffleWriterExec` already reports its own.
- Add `display_as_reports_real_partitioning`, which renders two writers over inputs with different partitioning and asserts each reflects its own. A constant cannot pass it.
- Regenerate the 22 approved goldens with `BALLISTA_GENERATE_GOLDEN=1`.

The golden diff is 45 lines and entirely mechanical. Each new value is checkable against the child plan on the line below it, for example in `q1.txt`:

```
=== Stage 2 ===
ShuffleWriterExec: partitioning: Hash([l_returnflag@0, l_linestatus@1], 16)
  ...
        UnresolvedShuffleExec: stage=1, partitioning: Hash([l_returnflag@0, l_linestatus@1], 16)

=== Stage 3 ===
ShuffleWriterExec: partitioning: UnknownPartitioning(1)
  SortPreservingMergeExec: [l_returnflag@0 ASC NULLS LAST, l_linestatus@1 ASC NULLS LAST]
```

Stage 2 inherits the hash partitioning it writes; stage 3 collapsed to one partition through the merge. Both were previously reported as `None`.

`EXPLAIN` output and the scheduler's rendered stage plans now show the writer's actual output partitioning instead of a constant `None`. No public API changes.

Two related warts noted in the issue are deliberately left alone here, since neither is needed for the fix: the two writers use different separators (`partitioning: ` versus `partitioning=`), and `ShuffleWriterExec` also carries a plain `impl Display` whose content differs from its `DisplayAs`.

Once this lands, the `ShuffleWriterExec` special case in #2305 can be dropped and that operator labeled generically like every other one.


avantgardnerio (approved):
LGTM

Issue #2003, opened 2026-07-12
[DISCUSSION] Should Ballista add support for streaming eager execution?


andygrove:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Ballista currently implements a batch execution model directly modeled after Apache Spark. Query stages are executed to completion, with output written to shuffle files. The plan is optionally re-optimized based on statistics from completed query stages (AQE). This is a well proven design that scales massively for ETL and analytical workloads.

This issue is to discuss whether we should consider extending Ballista to also support a model that is closer to DataFusion's in-process execution model, where the whole plan is executed eagerly and results are streamed between stages, bypassing the shuffle files. This can lead to faster time to first results in some cases, and lower latency queries. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

andygrove:
@milenkovicm @avantgardnerio Curious to hear your thoughts on this. This isn't a priority for me right now, just thinking about the future scope of the project.

milenkovicm:
Definitely worth exploring, IMHO. 

I guess we'd need to sit down and think this through, perhaps we could copy spark streaming shuffle ?

- https://www.databricks.com/blog/introducing-real-time-mode-apache-sparktm-structured-streaming
- https://issues.apache.org/jira/plugins/servlet/mobile#issue/SPARK-52330

Issue #408, opened 2022-10-20
Design: Implement bubble execution mode.


liurenjie1024:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
(This section helps Arrow developers understand the context and *why* for this feature, in addition to  the *what*)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

r4ntix:
Very good proposal. Can you add some designs of popular systems such as Presto and Clickhouse for comparison?

liurenjie1024:
> Very good proposal. Can you add some designs of popular systems such as Presto and Clickhouse for comparison?

Good suggestion, I'll add later.

PR #2308, opened 2026-08-15
docs: explain why Ballista uses a blocking shuffle

andygrove:

N/A. This is a documentation-only change and does not have a tracking issue.

# Rationale for this change

Ballista blocks between query stages: a stage runs to completion and materializes its output to local disk before any downstream stage starts. Other DataFusion-based distributed engines, notably [DataFusion Distributed](https://datafusion-contrib.github.io/datafusion-distributed/) and [Sail](https://github.com/lakehq/sail), stream data between stages instead, and they are faster on interactive workloads as a result.

We had nothing written down explaining why Ballista chose the blocking model, so the question keeps coming up and the answer lives in people's heads. The existing docs cover the mechanics of the shuffle writer in the tuning guide, but not the design rationale.

That gap cuts both ways. Users evaluating Ballista against a pipelined engine have no way to tell whether the blocking shuffle is a deliberate trade-off or just something nobody has gotten around to fixing. Contributors proposing changes to the exchange layer have no shared statement of which properties the current design is protecting.

Worth noting that DataFusion Distributed's own documentation already points at Ballista for large batch and ETL workloads that benefit from materializing intermediate results, so this completes a comparison they started rather than opening a new one.

A new contributors-guide page, `docs/source/contributors-guide/shuffle.md`, covering:

- **The barrier, concretely.** The three places it is implemented: shuffle-write summaries not existing until files close, `UnresolvedStage::resolvable()` requiring every input stage to be complete, and `ShuffleReaderExec` fetching concrete `PartitionLocation`s.
- **Why the barrier is there.** Five properties that follow from materializing stage output: producer and consumer tasks never compete for slots, so a stage can be far wider than the cluster; a lost executor costs re-run tasks rather than the whole query; the filesystem absorbs producer/consumer skew so shuffles larger than cluster memory stay safe; completed stages give the adaptive planner exact statistics; and executors can join or leave between stages.
- **What it costs.** Straggler stall, write amplification, interactive latency, the local-disk requirement, and the absence of any path to streaming.
- **How pipelined engines differ**, including a note that pipelining does not rule out adaptive query execution. DataFusion Distributed does AQE without a barrier by sampling in-flight batches, so the real distinction is exact versus sampled statistics. Empty-stage elimination is the one rule that genuinely needs a completed stage.
- **A decision table** for when each model wins.
- **Directions that do not require abandoning the model**, framed as open questions rather than commitments: partial-input early start, hybrid exchange, and a remote shuffle service (#1539).

Supporting changes:

- A user-facing FAQ entry, "Why does Ballista write shuffle data to disk instead of streaming it between stages?", with the short version and links onward.
- A pointer from the Shuffle section of the architecture page.
- A toctree entry in `docs/source/index.rst`.

Documentation only. No code or API changes.

### DIFF
--- a/docs/source/contributors-guide/architecture.md
+++ b/docs/source/contributors-guide/architecture.md
@@ -194,6 +194,10 @@ Each executor will re-partition the output of the stage it is running so that it
 stage. This mechanism is known as an Exchange or a Shuffle. The logic for this can be found in the [ShuffleWriterExec]
 and [ShuffleReaderExec] operators.
 
+The shuffle is _blocking_: a stage runs to completion and materializes its output to local disk before any
+downstream stage starts. See [Shuffle Design](shuffle.md) for why Ballista made that choice, what it costs, and
+how it compares to the pipelined shuffle used by engines such as DataFusion Distributed and Sail.
+
 [shufflewriterexec]: https://github.com/apache/datafusion-ballista/blob/main/ballista/core/src/execution_plans/shuffle_writer.rs
 [shufflereaderexec]: https://github.com/apache/datafusion-ballista/blob/main/ballista/core/src/execution_plans/shuffle_reader.rs
 
--- a/docs/source/contributors-guide/architecture.md
+++ b/docs/source/contributors-guide/architecture.md
@@ -194,9 +194,12 @@ Each executor will re-partition the output of the stage it is running so that it
 stage. This mechanism is known as an Exchange or a Shuffle. The logic for this can be found in the [ShuffleWriterExec]
 and [ShuffleReaderExec] operators.
 
-The shuffle is _blocking_: a stage runs to completion and materializes its output to local disk before any
-downstream stage starts. See [Shuffle Design](shuffle.md) for why Ballista made that choice, what it costs, and
-how it compares to the pipelined shuffle used by engines such as DataFusion Distributed and Sail.
+The shuffle is _blocking_: a stage materializes its output to local storage, and today a downstream stage waits
+for the whole upstream stage to complete before it starts. Waiting on the whole stage is a property of the current
+scheduler rather than of the design — the output of a finished task is readable as soon as its files are closed,
+so a downstream stage could in principle start on partial input when the cluster has capacity to spare. See
+[Shuffle Design](shuffle.md) for why Ballista materializes shuffle output at all, what that costs, and how it
+compares to the pipelined shuffle used by engines such as DataFusion Distributed and Sail.
 
 [shufflewriterexec]: https://github.com/apache/datafusion-ballista/blob/main/ballista/core/src/execution_plans/shuffle_writer.rs
 [shufflereaderexec]: https://github.com/apache/datafusion-ballista/blob/main/ballista/core/src/execution_plans/shuffle_reader.rs
--- a/docs/source/contributors-guide/shuffle.md
+++ b/docs/source/contributors-guide/shuffle.md
@@ -20,7 +20,7 @@
 # Shuffle Design
 
 Ballista uses a **blocking shuffle**: a query stage runs to completion and
-materializes its output to local disk before any downstream stage starts. This
+materializes its output to local storage before any downstream stage starts. This
 is the same model Apache Spark uses, and it is deliberately different from the
 **pipelined shuffle** used by engines such as Apache Flink,
 [DataFusion Distributed], and [Sail], where a downstream stage streams data from
@@ -101,10 +101,15 @@ machinery handles executors disappearing outright
 Repeated failures are capped: once a stage exceeds its retry budget the job
 fails rather than looping.
 
-A pipelined engine generally cannot do this. There is no durable copy of the
-producer's output, so a consumer failure forces the producer to run again, and
-in the common case the whole query restarts. DataFusion Distributed is explicit
-about this in its own documentation:
+A pipelined engine has to work harder for the same guarantee. Without a durable
+copy of the producer's output, recovery means either re-running the producer or
+restoring from a checkpoint, and checkpointing a distributed dataflow is a hard
+problem in its own right — Flink got there with
+[asynchronous barrier snapshotting], but it took years of work, and it recovers
+by rewinding a whole pipeline region to the last snapshot rather than by
+re-running the handful of tasks that were actually lost. Engines that have not
+taken that on restart the query. DataFusion Distributed is explicit about this
+in its own documentation:
 
 > If any node fails mid-query, the whole query fails; there are no retries.
 > There's no persistence of intermediate results, so queries can't checkpoint or
@@ -114,6 +119,8 @@ For a query that runs for seconds, restarting is cheap and this hardly matters.
 For an ETL job that runs for an hour on a cluster where a spot instance
 disappears every few minutes, it is the difference between a job that finishes
 and one that never does.
+
+[asynchronous barrier snapshotting]: https://nightlies.apache.org/flink/flink-docs-stable/docs/concepts/stateful-stream-processing/
 
 ### The filesystem absorbs producer/consumer skew
 
@@ -157,6 +164,14 @@ downstream exchange with an empty node and propagates emptiness up the plan,
 requires knowing a stage produced zero rows, and no sample can establish that
 before the stage ends.
 
+Exact per-partition counts are also what makes skew mitigation tractable. The
+AQE `CoalescePartitionsRule` already uses them to merge undersized shuffle
+partitions before the next stage is scheduled, and the same information is what
+a future rule would need to go the other way and split an oversized partition,
+or to choose range boundaries that spread a hot key across several downstream
+tasks. Skew is listed below as a cost of the barrier, but the barrier is also
+where the statistics to fix it come from.
+
 ### Executors can come and go between stages
 
 Because a stage's inputs are files rather than live connections, cluster
@@ -166,6 +181,12 @@ autoscaling (including the KEDA scaler) straightforward. Shuffle files are
 cleaned up after the job finishes, on the interval set by
 `finished-job-data-clean-up-interval-seconds`.
 
+For deployments that scale continuously with demand and run on spot capacity,
+this and partial re-execution above are the same property seen twice: a fleet
+whose size changes under the query only works if a departing executor costs a
+few re-run tasks and an arriving one can be handed work with no setup. Both
+follow from stage output being a file rather than a connection.
+
 ## What the barrier costs
 
 The trade-off is real, and it is worth stating plainly.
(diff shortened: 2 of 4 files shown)


andygrove:
cc @phillipleblanc @villebro @akshaychitneni

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+executor, rolls back that downstream stage, and resubmits only the map tasks
+needed to regenerate the dropped partitions. Map output that lives on surviving
+executors is not recomputed, and unrelated stages are not touched. The same
+machinery handles executors disappearing outright
+(`reset_stages_on_lost_executor`, `rollback_running_stage`,
+`rerun_successful_stage` in `ballista/scheduler/src/state/execution_graph.rs`).
+Repeated failures are capped: once a stage exceeds its retry budget the job
+fails rather than looping.
+
+A pipelined engine generally cannot do this. There is no durable copy of the
```
avantgardnerio: Perhaps it would be more charitable to say "Piplined engines must implement check pointing to do so, a famously hard problem that Flink eventually solved after years of work" (or whatever). Maybe just a link to how Flink solved it as a footnote.

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+buffered batches then flow on into the consumer.
+
+So the real distinction is **exact versus sampled**. Sampling is enough to size
+a stage or pick a join side. It is not enough for decisions that depend on a
+final answer — Ballista's empty-stage elimination rule, which replaces a
+downstream exchange with an empty node and propagates emptiness up the plan,
+requires knowing a stage produced zero rows, and no sample can establish that
+before the stage ends.
+
+### Executors can come and go between stages
```
avantgardnerio, line 182: And this is key for users like Coralogix: we scale up and down constantly based on demand and spot instance termination.

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+cleaned up after the job finishes, on the interval set by
+`finished-job-data-clean-up-interval-seconds`.
+
+## What the barrier costs
+
+The trade-off is real, and it is worth stating plainly.
+
+- **Straggler stall.** A stage finishes when its slowest task finishes. Until
+  then no downstream work can start, even if 99% of the input is ready. On a
+  skewed join key or a slow node, most of the cluster sits idle waiting. This is
```
avantgardnerio, line 203: But it also allows dynamic range repartitioning, which can prevent the skewed key :)

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+  the single largest source of avoidable latency in Ballista today.
+- **Write amplification.** Every intermediate byte is written, then read, then
+  often served over the network — even when the intermediate result is a few
+  kilobytes. For small queries this dominates the runtime.
+- **Poor interactive latency.** The cost above is roughly fixed per stage
```
avantgardnerio: I don't think it's worth changing the docs, but is worth highlighting: we use it for ~1 second queries (I don't know exactly, but I suspect we even have subsecond ones)

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+  boundary, so a query with many small stages pays it repeatedly. Pipelined
+  engines report substantially better latency on interactive benchmarks, and the
+  gap is structural, not an implementation detail.
+- **Local storage is required.** Executors need writable local disk sized for
```
avantgardnerio: Our instance sizes come with 96GB of ram or something. ~10TB for a cluster. We never write anything to disk - only lazily to s3 to prevent having to re-run a failed task.

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,228 @@
...
+  the largest shuffle they will handle, plus cleanup.
+- **No path to streaming.** A blocking barrier cannot express an unbounded
```
avantgardnerio, line 223: I'm still interested in streaming shuffles, but mostly academically. It could cause many of our queries to move to sub-second latencies. Not worth getting hung up on though.

avantgardnerio (approved):
Everything you said is generally true. I don't think anything I said is a blocker.

docs/source/contributors-guide/architecture.md
```diff
@@ -194,6 +194,10 @@ Each executor will re-partition the output of the stage it is running so that it
 stage. This mechanism is known as an Exchange or a Shuffle. The logic for this can be found in the [ShuffleWriterExec]
 and [ShuffleReaderExec] operators.
 
+The shuffle is _blocking_: a stage runs to completion and materializes its output to local disk before any
```
Dandandan: In principle some next stages can start when partitions are written (without waiting on the full stage) - currently it waits on the full stage but in some cases it can beneficial (when enough cluster capacity) to already start the next stage...

Dandandan:
Looks good in general.

I think there is quite some room to push the shuffle model further to make it equal / favour the shuffle model in some cases (also for performance reasons).

Some smaller/bigger ideas:

* In memory shuffle. Currently it writes it to disk always and reads it back. (Streaming avoids this cost but at some other cost).
* Reduce the number of files by coalescing partitions, cross partition batch coalescing
* Optimize shuffle format (e.g. hash partitioning might spill data multiple times to an indexed file instead of creating N files)
* Avoid metadata write / read, optimize shuffle format
* Shuffle affinity (route tasks to where shuffle data lives)
* Better parallelization during write/read
* Overlap multiple stages, instead of always waiting on full stage
* Bubble execution
* Make better use of output statistics

etc.

andygrove:
Thanks for the feedback so far @avantgardnerio and @Dandandan. I had Claude incorporate it.

villebro:
Love the context @andygrove, this is really valuable! It would be great to add similar breadcrumbs for other design decisions so we can avoid repeating earlier trial-and-error iterations.

Having said that, I agree with @Dandandan that there's probably room for supporting alternative models here, maybe via config flags or similar.

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,275 @@
...
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Shuffle Design
+
+Ballista uses a **blocking shuffle**: a query stage runs to completion and
+materializes its output to local storage before any downstream stage starts. This
+is the same model Apache Spark uses, and it is deliberately different from the
+**pipelined shuffle** used by engines such as Apache Flink,
```
milenkovicm: Nit, add link to flink, the only one without link

docs/source/contributors-guide/shuffle.md
```diff
@@ -0,0 +1,275 @@
...
+  the filesystem regardless of size. Keeping them in memory, with a spill path
+  to files when they grow, would remove most of the write amplification for
+  short queries while leaving the durable path in place for the ones that need
+  it.
+- **A cheaper shuffle format.** The write side currently produces a file per
+  output partition per task and a metadata round-trip per partition, which is a
+  lot of small files and small reads at high partition counts. Writing one
+  indexed file per task, coalescing batches across partitions, and trimming the
+  per-partition metadata would cut the fixed cost of a stage boundary without
+  changing where the data lives.
```
milenkovicm: We have pluggable policies in place, maybe it's worth prototyping this.

milenkovicm (approved):
Thanks @andygrove good overview

Dandandan:
> Other DataFusion-based distributed engines, notably [DataFusion Distributed](https://datafusion-contrib.github.io/datafusion-distributed/) and [Sail](https://github.com/lakehq/sail), stream data between stages instead, and they are faster on interactive workloads as a result.

Based on some experiments I did today, I was being able to recover (most) of the performance by keeping shuffle data in memory (RAM disk), this yields > 40% improvement on TPC-H, suggesting that most of the difference doesn't come from the blocking shuffle but from (avoiding) IO.

Dandandan:
In my experiments today, I was being able to recover most of the performance difference between ballista and datafusion-distributed mostly (after enabling AQE + some other configs, so it can use hash join / broadcast join which in itself gives 2-3x improvement) by keeping shuffle data in memory (RAM disk), yielding roughly -40% with a 4GiB RAM disk (e.g. keeping almost everything in memory).

This suggests we should have an option to keep shuffle data in memory, and only spill once it exceeds a configured buffer, instead of always writing to disk.

andygrove:
Pushed another round of changes based on the latest feedback.

Doc changes:

- **Flink link** (@milenkovicm). Linked in the intro, it was the only engine named without one.
- **Shuffle affinity is prototypable today** (@milenkovicm). The bullet now says so: task distribution is already pluggable, so a `TaskDistributionPolicy::Custom` implementation can bind consumer tasks to the executors holding their `PartitionLocation`s and be measured against bias and round-robin without touching the core.
- **Exact vs sampled statistics is not a clean split.** While checking the above I noticed the page overstated the case. Ballista's range repartition operators already sample in flight: `RuntimeStatsExec` keeps a T-Digest over the routing expression and the repartition operator snapshots it on the first batch to pick quantile cuts. Added a paragraph saying so, and noted that the same mechanism is what would spread a hot key across downstream tasks, which is @avantgardnerio's point about dynamic range repartitioning preventing skew.
- **In-memory shuffle** (@Dandandan). Rewrote the bullet around the bounded-buffer-with-spill shape and your RAM disk result, and called it out as the largest single win available, since it says the cost is the filesystem round trip rather than the barrier.
- **Config gating** (@villebro). The directions section now states that where one of these changes behavior it would be config-gated, with today's model as the default.
- Each open direction now links to its tracking issue.

Filed issues for the ideas that were not already tracked, so they do not get lost:

- #2318 — buffer shuffle output in memory and spill to disk only when a configured budget is exceeded
- #2319 — shuffle affinity: prototype a task distribution policy that places consumer tasks where their input already lives
- #2320 — allow a stage to start on partially complete inputs instead of waiting for the whole upstream stage

Already tracked, so no new issue: cheaper shuffle format and fewer small files (#660, epic #1319), streaming/push-based exchange and all-at-once scheduling (#1151, discussion in #2003), bubble execution (#408), remote shuffle service (#1539).

andygrove:
Merged. Thanks @avantgardnerio @Dandandan @milenkovicm @villebro

Dandandan:
> In my experiments today, I was being able to recover most of the performance difference between ballista and datafusion-distributed mostly (after enabling AQE + some other configs, so it can use hash join / broadcast join which in itself gives 2-3x improvement) by keeping shuffle data in memory (RAM disk), yielding roughly -40% with a 4GiB RAM disk (e.g. keeping almost everything in memory).
> 
> This suggests we should have an option to keep shuffle data in memory, and only spill once it exceeds a configured buffer, instead of always writing to disk.

My bad, turned to be false info (queries were failing after first iteration after filling the disk).

Although I think there is some ways to avoid / improve IO still.

avantgardnerio:
>  most of the performance difference between ballista 

I'm all for performance of any kind, but I'd like to highlight the scales they benchmark at are below the threshold at which Ballista would generally be the correct solution:

image

PR #2310, opened 2026-08-15
fix(scheduler): drop an executor's cached gRPC client when it is removed

Dandandan:
`ExecutorManager` caches an `ExecutorGrpcClient` per executor in `self.clients`, but nothing ever removes an entry: the map is only inserted into by `get_client` and read back. `remove_executor` drops the executor from the cluster state and leaves the client behind.

Closes #.

 # Rationale for this change

Removes some state when lots of executors exit the scheduler over time.


milenkovicm (approved):
Good catch, thanks @Dandandan

Issue #2321, opened 2026-08-16
Q10 SF1000 regression after #2315: SortPreservingMergeExec exhausts fair memory pool under AQE default-on


andygrove:
**Describe the bug**

TPC-H Q10 at SF1000 fails under the AQE plan produced after #2315, hitting a memory-pool OOM in `SortPreservingMergeExec`. On the pre-#2315 baseline Q10 completed in ~53s on the same cluster and config; after #2315 it errors at runtime.

Full error:
```
DataFusion error: Arrow error: External error: Execution error:
Job 0380C6NW0A000 failed: Job failed due to stage 6 failed:
Task failed due to runtime execution error:
DataFusionError(Shared(External(DataFusionError(ResourcesExhausted(
  "Failed to allocate additional 111.2 MB for SortPreservingMergeExec[0]
   with 5.5 GB already allocated for this reservation
   - 19.4 MB remain available for the total memory pool:
   fair(pool_size: 5.6 GB)"
)))))
```

**To Reproduce**

- Ballista built from `f21c958` (merge commit of #2315)
- TPC-H SF1000, parquet/zstd, on Kubernetes: 32 executors × 8 concurrent tasks
- Memory pool: `fair` pool sized at 70% of a 64 GiB executor container ≈ 44.8 GB total, divided across 8 concurrent tasks → ~5.6 GB per-task budget (matches the error trace)
- AQE default-on (this PR flips it)

**Expected behavior**

Q10 should complete without OOM at this scale. It passed at ~53s on the same infrastructure before #2315.

**Additional context**

- Full 22-query run: 19 passed, Q10 regressed as above, Q21/Q22 hit unrelated cluster networking issues.
- Q11 was fixed by this PR (previously erroring, now 15.2s).
- Overall wall-clock improved ~12% (632s vs 721s baseline) despite the Q10 regression.
- The per-task pool sizing didn't change on our side — the new AQE plan for Q10 evidently needs more sort/merge memory than the previous plan. Possible causes: the removed redundant shuffle changing partition sizes, or an AQE-driven join/coalesce decision concentrating data.
- Original benchmark report: https://github.com/apache/datafusion-ballista/pull/2315#issuecomment-5308644923

Dandandan:
Perhaps we could try some distributed sort here 
@avantgardnerio 

;)

Dandandan:
> 44.8 GB total, divided across 8 concurrent tasks → ~5.6 GB per-task budget (matches the error trace)

[...]

PR #2334, opened 2026-08-17
perf(core): push a consumer's row limit into RangeShuffleReaderExec

Dandandan:
Related: https://github.com/apache/datafusion-ballista/issues/2321

## Summary

`RangeShuffleReaderExec` merges all of its input streams to completion even
when the `SortPreservingMergeExec` directly above it only wants the first few
rows. Its `StreamingMerge` was built without a fetch:

```rust
StreamingMergeBuilder::new()
    .with_streams(sub_streams)
    .with_batch_size(config.batch_size())
    .with_reservation(reservation)
    .build()?                       // no .with_fetch(...)
```


milenkovicm (approved):
LGTM thanks @Dandandan

PR #2337, opened 2026-08-17
feat(core): Replace TDigest with KLL

avantgardnerio:
Range repartitioning routed on a T-Digest, which is `Float64`-only, single-column, and has no representation for a NULL. Those three limits were enforced as construction gates on four operators and one planner rule, so a query ordering by an `Int64`, a `Timestamp`, or any nullable column could not be range-repartitioned at all.

This replaces it with the `SortKeySketch` from #2294 and removes the gates. **Not purely additive**: the T-Digest is deleted, `QuantileSketchState` leaves the proto, and `BALLISTA_PROTOCOL_VERSION` goes to 2.

## What can be routed now

| | before | after |
|---|---|---|
| key type | `Float64` | any fixed-width type (24 arrow types) |
| nullability | non-null only | nullable, run placed by `nulls_first` |
| direction | ASC (`RangeFilterExec` fast path still ASC-only) | either, folded into the key |
| halo | `f64`, cast from any frame bound | the frame bound's own type |

One definition of routable, `SortKeyCodec::try_new`, asked in four places instead of four different type checks that could drift apart.

## NULLs get a partition the same way a value does

A NULL has no position among the values, only a side, and NULLs are indistinguishable from one another. So a NULL run cannot be split across partitions and have the split be reproducible on the read side. That is not a special case: **a repeated value cannot be split either**, and cuts have always been allowed to repeat and produce a dead partition rather than break a run apart.

Sizing therefore targets the population, NULLs included, and only repairs when the run is genuinely too big to fit one partition:

```text
n <= N/K, no repair                     n > N/K, repaired
┌──────┬──────────────────────────┐     ┌────────────────────┬─────────┐
│NULLs │ values                   │     │ NULLs              │ values  │
└──────┴──────────────────────────┘     └────────────────────┴─────────┘
0      n                          N     0                    n         N

cut[c] = population rank, minus n       run takes one partition whole,
run shares a partition with values      values above it split evenly
```

| NULLs | values | K | cuts | partition sizes |
|------:|-------:|--:|---|---|
| 10 | 1..=90 | 4 | `[15, 40, 65]` | 24 / 25 / 25 / 26 |
| 60 | 1..=40 | 4 | `[1, 14, 27]` | 60 / 13 / 13 / 14 |
| 90 | 1..=10 | 4 | `[1, 4, 7]` | 90 / 3 / 3 / 4 |

A run shorter than one partition's share costs **no** balance at all: it shares the lowest partition with the values below the first cut. A longer one fixes that partition's size, which nothing can improve, and the remaining partitions come out within one row of each other.

One rank per boundary, one of two rules picked by whether the run outgrows a partition, then a floor:

```rust
let rank = if !nulls_outgrow_a_partition {
    population_rank.saturating_sub(null_count)    // step over the run
} else {
    cut * sketch_count / not_null_parts + 1       // split the values evenly among the rest
};
rank.max(cut + 1)                                 // keep boundaries distinct
    .min(sketch_count)
```

Without the even split the top partition keeps everything the run displaced: `[60, 1, 13, 26]` where `[60, 13, 13, 14]` is available. Without the floor the first two boundaries both name the minimum, because a KLL rank is a cumulative-weight threshold so ranks 0 and 1 coincide. `nulls_last` is the mirror: `min` with ceilings instead of `max` with floors.

## Every cut is a real value, structurally

`cuts()` ranks among the values, so there is no rank a NULL could answer. This is not a filter applied afterward, it is unrepresentable.

That matters because a NULL boundary fails silently in every consumer. `RangeFilterExec` and `PerPartitionFilterExec` turn a cut into `BinaryExpr(key, Lt, Literal(cut))`, and arrow's comparison kernels propagate nulls into the mask while `filter_record_batch` reads a null mask entry as exclude. The rows do not error, they disappear. `cut_partitions` would separately compute `NULL + halo`.

Worth being precise about the mechanism, because "SQL semantics" is the wrong name for it: **arrow has two comparison families and the predicate path uses the null-propagating one.** Arrow's *ordering* comparisons (the row format, `ArrowNativeTypeOp::compare`) give NULL a defined position. The comparison kernels a `BinaryExpr` reaches for do not. So `RangeFilterExec` never asks a NULL to fail a comparison. The partition that claims the run tests for the run directly:

```rust
predicate = BinaryExpr(predicate, Or, IsNullExpr(filter_expr))
```

## Which partition claims the run, without a partition number

`RangeFilterExec` sees a task-local slice of the bounds, so it cannot know its global partition index. It does not need one: only a globally-outermost partition has an unbounded end, and that survives the scheduler slicing bounds down to a task.

```rust
takes_nulls = nulls_first ? lo.is_none() : hi.is_none()
```

The scatter derives the same answer from the same bit (`nulls_first ? 0 : K-1`), so write side and read side agree without either being told a partition number, and there is no third place for them to disagree.

Construction refuses a nullable key whose plan states no order for it. Without a stated order there is no fact about where the run belongs, and defaulting would hand it to whichever end the default named, by accident.

## Halo arithmetic: generalized, and skipped when it is a no-op

Halos are now typed, using `ScalarValue::add`/`sub`, shared between `RangeFilterExec` and `cut_partitions` so the two cannot drift on what a halo means. **No coercion**: a `Float64` halo against a `Timestamp` key is a planner bug, and silently coercing a duration is how a nanosecond bound acquires a 256 ns grid.

Two things make that livable rather than restrictive:

**A zero halo widens by nothing, and is short-circuited before any arithmetic.** The scheduler passes `Float64(0.0)` for every consumer with no halo at all, so without this a zero halo of one type would refuse a key of every other type. This is what makes non-window range repartitioning type-agnostic: no halo means no arithmetic means the key type never has to pair with anything.

**A halo is a delta, not a value of the key's type.** For a numeric key those coincide; for a temporal one the delta is an interval. Rather than enumerate which pairs work, `halo_from_bound` attempts the exact operation the filter will perform:

```rust
let probe = ScalarValue::new_zero(key_type).ok()?;
probe.sub(&halo).ok()?;   // Timestamp - Interval -> Timestamp
probe.add(&halo).ok()?;
```

A pair that does not typecheck declines the rewrite, so the query runs unparallelized rather than failing at execute time. `RANGE INTERVAL '1' DAY PRECEDING` on a `Timestamp` key is now expressible end to end. Previously every bound was cast through `f64`.

## Wire format

`SortKeySketchState` carries `k`, the NULL count, the exact extremes as `repeated ScalarValue`, and the compactor stack as one **Arrow IPC stream**: one row per level, a single `List<Struct<...>>` column, items ascending.

IPC rather than a packed blob because the arrow schema states what a key is. A multi-column key becomes struct siblings rather than a second payload shape needing a discriminant field, and the key's in-memory representation stays out of the wire format. Bulk items go to IPC, individual scalars go to `ScalarValue`: 816 retained items would pay ~10 bytes each as `ScalarValue`, where two extremes pay ~20 bytes total and stay readable in a log line.

Levels are sorted before serialization, so a decoder takes ascending order as given rather than being told per level. Direction and NULL placement are deliberately absent: they live once per report in the `order_by` tag every consumer already reads.

### Payload size

Per sketch is larger. Per **task** is smaller, because the size no longer scales with partition count:

```text
                            per entry    per task at K=8
T-Digest, per partition      2,072 B      16.6 KB
KLL, per partition           7,460 B      59.7 KB   (never shipped)
KLL, merged + triples        7,460 B       7.7 KB   <- what ships
```

No consumer reads a per-partition *distribution*. `merge_reports` folds them all together for global cuts, and `cut_partitions` needs only each file's extremes to route it. So one merged sketch per report is lossless for every consumer that exists, and the executor is where the merge is cheapest since the task already holds every partition's sketch when it builds the message.

Measured: 7,460 B encoded against 6,528 B raw at 816 retained items, so **1.14x for IPC framing**.

## Measurements

h2o window Q8 at 1e7, release, 2 executors, K=8, `max_partitions_per_task=4`.

**Correctness.** 10,000,000 rows on every run. Cuts moved by less than the two sketches' combined rank error, which is what a different algorithm approximating the same distribution should look like:

| cut | T-Digest | KLL | delta |
|---|---|---|---|
| 1 | 13.354 | 13.388 | 0.034 |
| 2 | 25.726 | 25.674 | 0.052 |
| 3 | 38.111 | 38.199 | 0.088 |
| 4 | 50.480 | 50.544 | 0.064 |
| 5 | 62.879 | 62.852 | 0.027 |
| 6 | 75.238 | 75.403 | 0.165 |

**Sketch ingest** (`cargo bench --bench quantile_sketch`, n=1M):

| arm | time | vs T-Digest |
|---|---:|---:|
| `tdigest` | 13.737 ms | 1.00x |
| `sort_key_sketch_f64` | 15.933 ms | 1.16x |
| `sort_key_sketch_f64_nulls` | 15.236 ms | 1.11x |
| `sort_key_sketch_i64` | 15.980 ms | 1.16x |
| `sort_key_sketch_timestamp_ns` | 15.944 ms | 1.16x |

# Future directions

**Delta encoding is worth ~3.7x on integer keys.** Sorted levels make it available. Measured over 816 items:

| encoding | Float64 | Int64 dense | Timestamp ns |
|---|---:|---:|---:|
| raw LE u64 | 6,528 B | 6,528 B | 6,528 B |
| delta + varint | 0.85x | **0.28x** | 0.72x |
| delta + FOR bitpack | 0.80x | **0.27x** | 0.63x |
| byte-split + zlib | 0.79x | 0.31x | 0.75x |

Floats are near their information-theoretic floor already: 816 sorted samples from a continuum carry about `816 x (log2(range/816) + 1.44)` bits, which is ~4.7 KB against a best measured 4,966 B. There is no clever encoding waiting to be found there. Dense integers are a different story, and that is exactly the key family this PR adds. Adopting it bumps `BALLISTA_PROTOCOL_VERSION` rather than needing a version field in the message.

**Range repartitioning on strings.** Outside a window there is no halo, so there is no arithmetic, so nothing about the routing needs a numeric key. Two things block a `Utf8` key today: `SortKeyCodec` has no fixed-width encoding for it (the arrow-row tier is the documented path, priced at 3.75x in the bench), and `RangeFilterExec` still gates on `numeric || temporal`, which is stricter than its zero-halo path requires. Neither is the halo machinery.

**One of the two `RuntimeStatsExec` taps is redundant.** RSE#1 (below the Sort) and RSE#2 (above the ORRE) observe the same multiset, since the ORRE redistributes rows without dropping any. RSE#1 exists to feed local cut discovery in-process and never crosses the wire; RSE#2 ships. What RSE#2 uniquely provides is per-output-partition extremes for file routing, and those are a min/max fold rather than a sketch. Removing one full sketch ingest from the hot path is available.

**`cut_partitions` learned a case nothing can reach yet.** A file with a null count and no value range is all NULLs, and belongs to the partition the run occupies rather than to whatever an overlap check says about a range it does not have.

**The fast path could handle NULLs.** Sorted input puts the run at one end, so the values occupy a contiguous span and the run stays contiguous with the selection, keeping the slice zero-copy where it currently copies. It would make plan-declared null placement load-bearing for slicing, which today's bail deliberately avoids, so it wants the existing path counters as its evidence.

**`sorted_on_key` still requires ascending.** A descending input silently takes the predicate path. Correct but slower, and visible in `slow_batches`.

## What this does not do

No query in either benchmark suite exercises the widened keys. Both suites contain exactly two finite-RANGE window queries, both h2o `ORDER BY v2`, a non-null `Float64`. Every unpartitioned h2o query on the `Int64` `id3` column uses a ROWS frame or an unbounded RANGE, so they are excluded by frame shape rather than key type. TPC-DS has no finite-RANGE frame at all.

So Q8 proves no regression, not new capability. The `Int64`, nullable, and `Timestamp`-with-interval paths are covered by unit tests. A bounded RANGE frame over a timestamp is the shape most likely to matter in production and needs a dataset we do not have.


avantgardnerio:
@phillipleblanc FYI

ballista/core/src/sort_key.rs
```diff
@@ -672,48 +718,321 @@ impl SortKeySketch {
...
+    /// | true        |    60 | 1..=40 | 4 | `[1, 14, 27]`  | 60 / 13 / 13 / 14 |
+    /// | false       |    60 | 1..=40 | 4 | `[14, 27, 40]` | 13 / 13 / 13 / 61 |
+    /// | false       |    90 | 1..=10 | 4 | `[4, 7, 10]`   | 3 / 3 / 3 / 91    |
+    ///
+    /// Empty when `partitions < 2` or no value was observed.
+    ///
+    /// Should only error if:
+    /// 1. invalid sketch: min/max is empty but levels are not - guarded against in proto decode
+    /// 2. codec.decode() failure - guarded against in try_new
+    pub fn cuts(&self, partition_cnt: usize) -> Result<Vec<ScalarValue>> {
```
phillipleblanc, line 751: I believe this will conflate "no rows were observed" with "rows were observed, but every key was NULL." For an all-NULL key, the report has a valid sketch and nonzero row count, but this method would return empty, leading to the query failing.

phillipleblanc (approved):
Looks good! One minor edge case that we might want to fix first.

ballista/core/src/sort_key.rs
avantgardnerio: great catch as usual, thank you!

PR #2348, opened 2026-08-19
feat: remove serialisation of built-in file formats from ballista 

milenkovicm:

Closes #.

# Rationale for this change

datafusion `DefaultLogicalExtensionCodec` support serde of build in file formats since https://github.com/apache/datafusion/issues/16944 there is no need to have duplicated code in ballista.

remove handling of build in file formats from ballista logical code, and delegate it to datafusion 

No

### DIFF
--- a/ballista/core/src/serde/mod.rs
+++ b/ballista/core/src/serde/mod.rs
@@ -27,10 +27,6 @@ use datafusion::common::{DataFusionError, Result};
 use datafusion::execution::TaskContext;
 use datafusion::logical_expr::Extension;
 use datafusion::physical_plan::{ExecutionPlan, Partitioning};
-use datafusion_proto::logical_plan::file_formats::{
-    ArrowLogicalExtensionCodec, AvroLogicalExtensionCodec, CsvLogicalExtensionCodec,
-    JsonLogicalExtensionCodec, ParquetLogicalExtensionCodec,
-};
 use datafusion_proto::physical_plan::from_proto::parse_physical_sort_exprs;
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_hash_partitioning;
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_partitioning;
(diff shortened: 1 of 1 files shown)


ballista/core/src/serde/mod.rs
```diff
@@ -190,49 +185,12 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> BallistaCodec<T,
...
-    /// returns a position of codec in the list and result.
-    ///
-    /// position is important with encoding process
-    /// as position of used codecs is needed
-    /// so the same codec can be used for decoding
-    fn try_any<R>(
-        &self,
-        mut f: impl FnMut(&dyn LogicalExtensionCodec) -> Result<R>,
-    ) -> Result<(u32, R)> {
-        let mut last_err = None;
```
avantgardnerio, line 207: I think this changes the protobuf encoding wire numbers? If so, we should bump BALLISTA_PROTOCOL_VERSION

avantgardnerio (approved):

ballista/core/src/serde/mod.rs
milenkovicm: https://github.com/apache/datafusion-ballista/pull/2337 ( you 🫵🏻 ) update it two days ago 😀. i believe, we as we haven't released it yet, no need to update it. do i miss something @avantgardnerio ?

avantgardnerio: I was not sure what convention we wanted to support here. I'm not sure if folks are tracking ballista/main in their own projects. It seemed like monotonic bumps couldn't hurt. @phillipleblanc ?

milenkovicm: I have no problem to increment it, id argue that we should have API checkpoint on release

avantgardnerio: Makes me wonder if it should just be the hash of the protobuf, or a git SHA. Not advocating for either, just thinking out loud.

milenkovicm: we could do it, it would be easy to automate, it would be hard to capture backward compatible changes, though. 

i believe this change should produce same binary output (as the logic has been moved from here to datafusion), i would not put my money on that claim 😀

martin-g (approved):

ballista/core/src/serde/mod.rs
phillipleblanc: Yeah, I don't think we need to bump between versions.

phillipleblanc (approved):

andygrove:
Heads up that this turned out to break the wire format for `CopyTo` plans across versions. I filed #2376 with the details.

The short version is that the old private `FileFormatProto` and DataFusion's `FileFormatProto` are structurally identical on the wire, a varint followed by a bytes field, so an old message decodes cleanly and the format tag just gets reinterpreted. The old `encoder_position` indexed into `[Parquet, Csv, Json, Arrow, Avro]` while `FileFormatKind` is `UNSPECIFIED=0, CSV=1, JSON=2, PARQUET=3, ARROW=4, AVRO=5`.

Feeding each old position to the new codec gives:

```
pos=0 (Parquet) -> Err(This feature is not implemented: Unspecified file format kind)
pos=1 (Csv)     -> Ok(csv)       lines up by coincidence
pos=2 (Json)    -> Ok(json)      lines up by coincidence
pos=3 (Arrow)   -> Ok(parquet)   silently wrong format
pos=4 (Avro)    -> Ok(arrow)     silently wrong format
```

Parquet at least fails loudly, and that is what CI caught in #2374 where a released 54.0.0 Python client talks to a cluster built from the branch. The Arrow and Avro cases are the ones that worry me. `ArrowLogicalExtensionCodec::try_encode_file_format` writes an empty payload and the Parquet decoder treats empty bytes as all defaults, so `COPY ... STORED AS ARROW` from a 54 client runs as Parquet on a 55 scheduler with no error at all.

This is not a hypothetical skew for us because `pyballista` re-exports datafusion-python types, so the Python bindings cannot move to 55 until there is a matching `datafusion-python` release. Every Python user runs a 54 client against a 55 cluster in the meantime.

No objection to the cleanup itself, the old positional scheme was clearly worth removing. It is more that we probably want an old client to get a clear version mismatch error instead of a mis-decode. Bumping `BALLISTA_PROTOCOL_VERSION` and validating it for clients, which is #2370, would cover this and the next change of this shape. Happy to hear what you think is cleanest.

PR #2354, opened 2026-08-22
fix(scheduler): fix test job-completion polling 

NoahKusaba:

Closes #.

 # Rationale for this change

The job-completion poll loop in `SchedulerTest::run_with_subscriber` used `_ => continue`,
which skips the `sleep(100)` below the match. Once a job reports a non-terminal status the
loop polls with no delay and spins a core until the job finishes.

`await_completion` and `await_completion_timeout` had the same bug and no callers.

- `run_with_subscriber`: collapsed the inner `match` into the outer `if let` so
  non-terminal statuses fall through to the sleep.
- Deleted `await_completion` and `await_completion_timeout` (no callers).

We can also not delete the unused functions, but this is a very clear bug regardless.

No

### DIFF
--- a/ballista/scheduler/src/test_utils.rs
+++ b/ballista/scheduler/src/test_utils.rs
@@ -601,76 +601,6 @@ impl SchedulerTest {
             .await
     }
 
-    /// Waits for job completion with a timeout in milliseconds.
-    pub async fn await_completion_timeout(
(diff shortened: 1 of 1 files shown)


milenkovicm (approved):
thanks @NoahKusaba good catch

PR #2358, opened 2026-08-23
perf(scheduler): stop re-encoding the session config for every task

NoahKusaba:

Closes #.

 # Rationale for this change

Release build, 32-task batch, 200 iterations, measured on
`prepare_multi_task_definition` end to end:

| | per batch | per task |
|---|---|---|
| before | 3.53 ms | 110.5 µs |
| after  | 1.77 ms | **55.5 µs** |

Encode the session config once per task batch instead of once per task, cloning it into each task's definition.


Dandandan (approved):

phillipleblanc (approved):
🚀

andygrove (approved):
Thanks @NoahKusaba

PR #2359, opened 2026-08-24
perf(core): reuse shuffle partition-index buffers across batches

NoahKusaba:

Closes #.

 # Rationale for this change

`compute_partition_indices` allocated a fresh `Vec<Vec<u32>>` per input batch
`num_partitions` vectors grown from zero, then dropped once `push_batch` copied
the indices out. The churn scales with the output partition count.

Copying the hash_buffer implementation that already exists in compute_partition_indices,
it now writes into a caller-owned buffer whose inner `Vec`s are cleared rather
than dropped, so capacity survives into the next batch. Partition assignment is
unchanged.

AI Benchmark:
100 batches x 8192 rows, hash on `l_orderkey`, release, best of 5 — timing
hash + index build + `push_batch`:

| partitions | before | after | speedup |
|---|---|---|---|
| 16 | 74.8 µs/batch | 61.5 µs/batch | 1.22x |
| 64 | 92.2 µs/batch | 59.4 µs/batch | 1.55x |
| 200 | 125.8 µs/batch | 62.0 µs/batch | 2.03x |
| 400 | 162.4 µs/batch | 71.0 µs/batch | 2.29x |
| 800 | 207.4 µs/batch | 76.1 µs/batch | 2.73x |

No

### DIFF
--- a/ballista/core/src/execution_plans/sort_shuffle/writer.rs
+++ b/ballista/core/src/execution_plans/sort_shuffle/writer.rs
@@ -607,6 +607,7 @@ impl SortShuffleWriterExec {
                     .register(&context.runtime_env().memory_pool);
 
             let mut hash_buffer: Vec<u64> = Vec::new();
+            let mut per_partition_rows: Vec<Vec<u32>> = Vec::new();
             let mut spill_events: u64 = 0;
             let mut spill_triggers = SpillTriggerCounts::default();
             // Absolute buffered-bytes counter, independent of the runtime
@@ -625,11 +626,12 @@ impl SortShuffleWriterExec {
 
                 // Compute partition assignment for every row.
                 let timer = metrics.repart_time.timer();
-                let per_partition_rows = compute_partition_indices(
+                compute_partition_indices(
                     &input_batch,
                     &exprs,
                     num_output_partitions,
                     &mut hash_buffer,
+                    &mut per_partition_rows,
                 )?;
                 timer.done();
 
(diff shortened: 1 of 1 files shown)


ballista/core/src/execution_plans/sort_shuffle/writer.rs
```diff
@@ -1271,11 +1276,14 @@ fn compute_partition_indices(
         hash_buffer,
     )?;
 
-    let mut out: Vec<Vec<u32>> = (0..num_partitions).map(|_| Vec::new()).collect();
+    out.resize_with(num_partitions, Vec::new);
+    for rows in out.iter_mut() {
+        rows.clear();
+    }
     for (row, &h) in hash_buffer.iter().enumerate() {
         out[(h % num_partitions as u64) as usize].push(row as u32);
```
Dandandan, line 1284: This is also a slow path (modulo is extremely slow on CPUs), in DataFusion I went so far to replace it with strength reduce https://github.com/apache/datafusion/pull/21900

NoahKusaba: Interesting, I'll try to incorporate that too!

NoahKusaba: Actually I'll make a follow up PR today / tomorrow for that change, to make it easier for maintainers to review/approve. Thanks for the suggestion and an implementation to reference :)

Dandandan: Amazing!

PR #2373, opened 2026-08-25
fix(executor): forward arrow-ipc-optimizations to ballista-core

### LINKED ISSUE #2368: arrow-ipc-optimizations in ballista-executor does not forward to ballista-core
Both `ballista-core` and `ballista-executor` declare a feature named `arrow-ipc-optimizations`, and both have it on by default:

- `ballista/core/Cargo.toml:38`, defaulted at line 40
- `ballista/executor/Cargo.toml:36`, defaulted at line 46

They are unrelated features that happen to share a name. The executor's gates exactly one call site, `flight_service.rs:159`. Core's gates the rest of the shuffle read path. The executor's does not forward to core's, unlike `spark-compat` right next to it at `ballista/executor/Cargo.toml:47`.

This makes `ballista/client/Cargo.toml:33-35` misleading. It pulls the executor with `default-features = false, features = ["arrow-ipc-optimizations"]`, which reads as "keep the IPC optimization on" but only enables the executor's half. Core's happens to be on anyway through the client's direct dependency on `ballista-core` at line 32, so nothing is broken today, but it works by accident.

Anyone building the executor standalone with `--no-default-features --features arrow-ipc-optimizations` gets one of the four decode sites optimized and the other three validating, with nothing to indicate it.

Suggested fix, in `ballista/executor/Cargo.toml`:

```toml
arrow-ipc-optimizations = ["ballista-core/arrow-ipc-optimizations"]
```

Default behavior is unchanged since both are already on by default. Worth noting in the fix that this still does not give a full off switch, because Cargo features are additive and anything else in the graph enabling core's default turns it back on.

Related: `examples/Cargo.toml:61` sets `default-features = false` on `ballista-core` to get `build-binary` and silently loses `arrow-ipc-optimizations` as a side effect. Same root cause, may be worth fixing in the same PR.

stag7824:

Closes #2368.

# Rationale for this change

`ballista-core` and `ballista-executor` each declare a feature named `arrow-ipc-optimizations`, both on by default, but they are unrelated features that happen to share a name. The executor's feature did not forward to core's — unlike `spark-compat` immediately below it in the same `[features]` table.

The two halves gate different code. Core's side is now five sites (a fourth landed in `range_shuffle/ipc_file.rs` after this PR was opened):

| Feature | Gated site |
|---|---|
| `ballista-executor/arrow-ipc-optimizations` | `ballista/executor/src/flight_service.rs` |
| `ballista-core/arrow-ipc-optimizations` | `ballista/core/src/client.rs` |
| | `ballista/core/src/execution_plans/shuffle_reader.rs` |
| | `ballista/core/src/execution_plans/sort_shuffle/multi_stream_reader.rs` |
| | `ballista/core/src/execution_plans/range_shuffle/ipc_file.rs` |

So building the executor standalone with `--no-default-features --features arrow-ipc-optimizations` skips IPC validation at **one** of core's five sites and validates at the rest, with nothing to indicate the difference.

`ballista/client/Cargo.toml` pulls the executor with `default-features = false, features = ["arrow-ipc-optimizations"]`, which reads as "keep the IPC optimisation on" but only enabled the executor's half. Nothing is broken today, because the client also depends on `ballista-core` directly and core's feature arrives through core's own defaults — but it works by accident.

**`ballista/executor/Cargo.toml`** — the fix the issue was filed for:

```toml
-arrow-ipc-optimizations = []
+arrow-ipc-optimizations = ["ballista-core/arrow-ipc-optimizations"]
```

**`examples/Cargo.toml`** — the issue's related note, same root cause. This crate sets `default-features = false` on `ballista-core` to opt into `build-binary`, which silently drops `arrow-ipc-optimizations` as a side effect:

```toml
 ballista-core = { path = "../ballista/core", version = "54.0.0", default-features = false, features = [
     "build-binary",
+    "arrow-ipc-optimizations",
 ] }
```

`examples/examples/mtls-cluster.rs` constructs `ballista_executor::flight_service::BallistaFlightService` directly, so this crate genuinely exercises the gated code, not just links it in unused.

# Are these changes tested?

Yes, both with `cargo tree`'s feature resolver.

**Executor fix** — `-p ballista-executor --no-default-features --features arrow-ipc-optimizations -e features -i ballista-core`:

Before, core's feature is reachable only through core's own defaults:

```
ballista-core v54.0.0
├── ballista-core feature "arrow-ipc-optimizations"
│   └── ballista-core feature "default"
│       └── ballista-executor v54.0.0
│           └── ballista-executor feature "arrow-ipc-optimizations" (command-line)
└── ballista-core feature "default" (*)
```

After, the executor's feature enables core's directly:

```
ballista-core v54.0.0
├── ballista-core feature "arrow-ipc-optimizations"
│   ├── ballista-executor feature "arrow-ipc-optimizations" (command-line)
│   └── ballista-core feature "default"
│       └── ballista-executor v54.0.0
│           └── ballista-executor feature "arrow-ipc-optimizations" (command-line)
└── ballista-core feature "default" (*)
```

**Examples fix** — `-p ballista-examples -e features,dev -i ballista-core`:

Before, `ballista-core feature "arrow-ipc-optimizations"` does not appear in the graph under `ballista-examples` at all. After:

```
ballista-core v54.0.0
├── ballista-core feature "arrow-ipc-optimizations"
│   [dev-dependencies]
│   └── ballista-examples v54.0.0
│       ├── ballista-examples feature "default" (command-line)
│       └── ballista-examples feature "standalone"
│           └── ballista-examples feature "default" (command-line)
├── ballista-core feature "aws-config"
│   └── ballista-core feature "build-binary"
...
```

`ci/scripts/rust_toml_fmt.sh` passes on both `Cargo.toml` files.

Default behaviour is unchanged, since both features are already on by default in every case this touches.

One config is worth calling out explicitly, per review: someone depending on the executor today with

```toml
ballista-core = { version = "54", default-features = false }
ballista-executor = { version = "54", default-features = false, features = ["arrow-ipc-optimizations"] }
```

is deliberately keeping IPC validation on in core while opting the executor's half in. After this change they get core's validation-skip too, since the executor's feature now forwards to it. `with_skip_validation` is an unsafe operation, so that's a real behavioural shift for that specific configuration, not just a naming cleanup — even though it makes the feature name mean what it already claimed to mean, and the alternative (leaving the name misleading) is worse.

Worth noting for anyone reading the issue later: this still does not give a full off switch. Cargo features are additive, so anything else in the graph that enables `ballista-core`'s default features will turn core's half back on regardless.

https://claude.ai/code/session_01AycxQUU3D4C9D45UQfDEhb


ballista/executor/Cargo.toml
```diff
@@ -33,7 +33,7 @@ path = "src/bin/main.rs"
 required-features = ["build-binary"]
 
 [features]
-arrow-ipc-optimizations = []
+arrow-ipc-optimizations = ["ballista-core/arrow-ipc-optimizations"]
```
andygrove, line 36: Worth a sentence in the description I think. This does widen what the executor's feature turns on, and cargo feature declarations are part of what a library user pins against.

Someone doing this today:

```toml
ballista-core = { version = "54", default-features = false }
ballista-executor = { version = "54", default-features = false, features = ["arrow-ipc-optimizations"] }
```

is deliberately keeping validation on in core, and after this change they quietly lose it. Given `with_skip_validation` is unsafe, that flip is safety relevant rather than just a perf tweak.

I still think the change is right. The pre-existing half-on state is worse, and making the name mean what it says is the whole point. It just makes "Are there any user-facing changes? No" a bit too strong for that one config, and it costs a line to say so.

andygrove (commented):
Thanks for this, and thanks for putting the cargo tree output in the description. I reproduced it on main with `cargo tree -p ballista-executor --no-default-features --features arrow-ipc-optimizations -e features -i ballista-core` and got your "before" graph exactly, so the diagnosis holds up and this looks like the minimal fix for it.

No API breaks and no default behaviour change, so no `api change` label or upgrade guide entry needed. Default builds already had core's half on, so there is no plan shape or execution change to benchmark either.

Couple of small things:

- #2368 also flags `examples/Cargo.toml:61`, which sets `default-features = false` on `ballista-core` to pick up `build-binary` and silently drops `arrow-ipc-optimizations` as a side effect. The issue suggests fixing that in the same PR and it is still there on this branch. `cargo build -p ballista-examples` on its own loses the feature, a full workspace build hides it through feature unification. Fine either way, I mostly want to make sure it does not get lost when the issue closes.
- The site table in the description is one entry stale. `ballista/core/src/execution_plans/range_shuffle/ipc_file.rs:194` picked up a fourth core side `cfg!(feature = "arrow-ipc-optimizations")` in 7e939fe, so it is five sites in total rather than four. Nothing about the fix changes.

One optional thought for later. If you want the duplicate name gone rather than just wired up, core could export something like `pub const SKIP_IPC_VALIDATION: bool = cfg!(feature = "arrow-ipc-optimizations")` and `flight_service.rs` could read that, which would let the executor feature be deleted entirely and leave one knob instead of two. It does not fix the missing off switch, since additivity defeats that either way, so definitely not a reason to hold this up.

stag7824:
Thanks for reproducing it and for the careful read. All three addressed:

1. **`examples/Cargo.toml:61`** — fixed in dedb6f3, added `arrow-ipc-optimizations` alongside `build-binary` on the `ballista-core` dev-dependency. Verified with `cargo tree -p ballista-examples -e features,dev -i ballista-core`: before, core's feature doesn't appear in the graph under `ballista-examples` at all; after, there's a direct edge from it. Worth noting `mtls-cluster.rs` constructs `BallistaFlightService` directly, so this crate isn't just linking the gated code in unused.
2. **Site count** — corrected to five in the description, and added the `range_shuffle/ipc_file.rs` row.
3. **The safety-relevant config** — added a paragraph under "Are there any user-facing changes?" spelling out exactly the case you flagged: `ballista-executor` opted in with `default-features = false` on `ballista-core` now loses that isolation, since the executor's feature forwards to core's. Said plainly that it's a real behavioural shift for that one configuration, not just a rename.

Left the `SKIP_IPC_VALIDATION` const idea for a follow-up rather than folding it in here, since as you said it doesn't change whether this needs merging first.

stag7824:
Hi @andygrove — a gentle nudge on this one, no rush if it's simply queued behind other work.

I suspect it's stuck rather than waiting on me: no CI workflow has ever run against the head commit (`5c310359`). `actions/runs?head_sha=5c310359` comes back empty, while a recently merged PR from another external contributor shows 25 checks on its head. That looks like the `action_required` approval gate this repo applies to fork PRs, which needs a committer to approve the run — not something I can trigger from my side.

For what it's worth, the three points from your Aug 30 review are addressed in `dedb6f3`, and I merged `upstream/main` on Sep 3 so the branch is current with no conflicts.

Happy to rebase, split it up, or add anything else that would make it easier to review.

andygrove (approved):
Thanks @stag7824

stag7824:
Thanks for the approval @andygrove.

`test linux crates` went red on the run that followed, but it isn't this change — all five failures are in `examples/tests/object_store.rs` and they are image pulls, not assertions:

```
called `Result::unwrap()` on an `Err` value: Client(PullImage {
  descriptor: "minio/minio:RELEASE.2025-02-28T09-55-16Z",
  err: DockerResponseServerError { status_code: 404,
    message: "pull access denied for minio/minio, repository does not exist ..." } })
```

It's the stale-branch version of the problem #2447 just fixed: MinIO withdrew `minio/minio` from Docker Hub on Sep 11, my run hit that the same evening, and `ab98c3f` has since pointed the tests at `quay.io/minio/minio`. My branch was last synced with main on Sep 3, so it was still pulling the withdrawn image. Re-running wouldn't have helped — the image is genuinely gone from Docker Hub.

I've merged `upstream/main` (`1d44275`), which brings in that registry override. No conflicts, and the diff against main is still just the two `Cargo.toml` files.

Re-checked the forwarding after the merge:

```
$ cargo tree -e features -i ballista-core -p ballista-executor --features arrow-ipc-optimizations
ballista-core v54.0.0
├── ballista-core feature "arrow-ipc-optimizations"
│   ├── ballista-executor feature "arrow-ipc-optimizations" (command-line)
```

CI on fork PRs needs a maintainer to approve the run, so it will need one more click when you have a moment.

PR #2378, opened 2026-08-25
feat: scheduler REST & GRPC returns ballista version headers 

### LINKED ISSUE #2370: Validate BALLISTA_PROTOCOL_VERSION for clients, not just executors

The scheduler validates `BALLISTA_PROTOCOL_VERSION` when an executor registers, but there is no equivalent check for clients. `check_protocol_version` in `ballista/scheduler/src/scheduler_server/grpc.rs` is only wired into the executor registration RPCs, and nothing in `ballista/core/src/execution_plans/distributed_query.rs` or the client crate sends or verifies a protocol version.

The result is that a client built against a different Ballista version connects happily, plans and runs a job successfully, and then fails much later with a confusing error when it tries to read the results.

This came up while investigating #2367, where the client asked the executor for `.../{job}/3/0/data.arrow` while the executor had written `.../{job}/3/0/data-0.arrow`. The shuffle file layout for the final stage changed after 54.0.0: the passthrough `ShuffleWriterExec` used to report `file_id: None` and write `data.arrow`, and now reports `file_id: Some(task_id)` and writes `data-{task_id}.arrow`. A client and cluster that disagree about that surface as:

```
Failed to open partition file at ".../data.arrow": Os { code: 2, kind: NotFound, ... }
```

which reads like data loss or a cleanup race rather than a version mismatch. The job even reports as `Completed`, so the first thing a user suspects is the shuffle machinery.

Extend the existing protocol version handshake to the client:

- Have the client send its compiled-in `BALLISTA_PROTOCOL_VERSION` on `ExecuteQuery` (or on a dedicated handshake RPC).
- Have the scheduler reject a mismatch with a clear `failed_precondition` naming both versions, the same way it already does for executors.
- Bump `BALLISTA_PROTOCOL_VERSION` whenever the on-disk shuffle layout or the client-visible wire contract changes, and document that rule next to the constant in `ballista/core/src/lib.rs`.

Only bumping the version on wire-format changes and leaving layout changes alone. That does not help here, since the proto messages (`PartitionLocation`, `ShuffleWritePartition`, `FetchPartition`) are byte-identical between 54.0.0 and main. What actually diverged is the meaning of `file_id`, so the version rule needs to cover the shuffle layout too.

Alternatively, make the failure message itself smarter (e.g. when a partition file is missing, check whether the sibling name under the other layout exists and say so). Useful as a diagnostic, but it papers over the mismatch instead of refusing it up front.

**Additional context**

Related: #2367.

milenkovicm:

Closes #2370.

# Rationale for this change

scheduler returns version headers which client can compare with expected version.

schedule returns two additional header 

```
server: ballista/VERSION
x-powered-by: datafusion/VERSION
```

client which check `server/VERSION` and logs warning in case of mismatch 

### DIFF
--- a/ballista/scheduler/src/scheduler_process.rs
+++ b/ballista/scheduler/src/scheduler_process.rs
@@ -40,10 +40,11 @@ use datafusion::DATAFUSION_VERSION;
 use datafusion_proto::logical_plan::AsLogicalPlan;
 use datafusion_proto::physical_plan::AsExecutionPlan;
 use datafusion_proto::protobuf::{LogicalPlanNode, PhysicalPlanNode};
-use http::StatusCode;
+use http::{HeaderName, HeaderValue, StatusCode};
 use log::info;
 use std::{net::SocketAddr, sync::Arc};
 use tonic::service::RoutesBuilder;
+use tower_http::set_header::SetResponseHeaderLayer;
 /// Creates as initialized scheduler service
 /// without exposing it as a grpc service
 pub async fn create_scheduler<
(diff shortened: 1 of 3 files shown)


milenkovicm:
not sure if this could cover #2370 requirements.

avantgardnerio (commented):
I strongly support this, but the `warn!` could be an issue if we intend to keep the `54 -> 55` version compatibility @andygrove suggested. If it causes users to see a warning all the time, even in supported use-cases, they'll just ignore it. I'll wait for Andy to weigh in.

ballista/scheduler/src/scheduler_process.rs
```diff
@@ -85,6 +86,28 @@ pub async fn create_scheduler<
     Ok(scheduler_server)
 }
 
+/// Wraps a router so every response carries `Server` and `X-App-Version`
```
martin-g:
```suggestion
/// Wraps a router so every response carries `Server` and `x-powered-by`
```

martin-g (approved):

avantgardnerio (approved):

milenkovicm:
thanks @martin-g & @avantgardnerio

PR #2379, opened 2026-08-25
Revert "remove redundant logic from ballista logical codec (#2348)"

### LINKED ISSUE #2376: CopyTo plans from a 54 client silently decode as the wrong file format on a 55 scheduler
**Describe the bug**

#2348 changed `BallistaLogicalExtensionCodec` to delegate `try_encode_file_format` / `try_decode_file_format` to DataFusion's `DefaultLogicalExtensionCodec`, removing Ballista's private `FileFormatProto`. This is a breaking wire-format change for `CopyTo` plans, and the two encodings are structurally identical, so the mismatch is not detected as a decode error.

| | field 1 | field 2 |
|---|---|---|
| Ballista <= 54 | `encoder_position: u32`, an index into `[Parquet, Csv, Json, Arrow, Avro]` | `blob` |
| Ballista post-#2348 | `kind: FileFormatKind` (`UNSPECIFIED=0, CSV=1, JSON=2, PARQUET=3, ARROW=4, AVRO=5`) | `encoded_file_format` |

Both are a varint followed by a bytes field, so prost decodes the old message happily and the format tag is simply reinterpreted with the wrong meaning.

Decoding every old position with the new codec gives:

```
pos=0 (Parquet) -> Err(This feature is not implemented: Unspecified file format kind)
pos=1 (Csv)     -> Ok(csv)       correct only by coincidence
pos=2 (Json)    -> Ok(json)      correct only by coincidence
pos=3 (Arrow)   -> Ok(parquet)   silently wrong format
pos=4 (Avro)    -> Ok(arrow)     silently wrong format
```

Parquet fails loudly. CSV and JSON happen to line up numerically. **Arrow and Avro silently decode as the wrong format**: `ArrowLogicalExtensionCodec::try_encode_file_format` writes an empty payload, and the Parquet decoder accepts empty bytes as "all defaults", so `COPY ... STORED AS ARROW` from a 54 client is executed as Parquet by a 55 scheduler with no error at all.

**To Reproduce**

The failure is already reproduced in CI by the new job added in #2374, where a released 54.0.0 Python client runs against a cluster built from the branch:

```
FAILED python/tests/test_context.py::test_write_parquet - Exception: DataFusion error: Arrow error:
External error: Execution error: Status { code: InvalidArgument, message: "Could not parse plan:
DataFusion error: This feature is not implemented: Unspecified file format kind" }
```

The full matrix above can be reproduced with a unit test in `ballista-core` that feeds the old wire bytes (field 1 = varint position, empty blob, i.e. `[0x08, pos]`) to `BallistaLogicalExtensionCodec::default().try_decode_file_format`.

**Expected behavior**

Either the 54 encoding keeps working against a 55 scheduler, or the client is rejected with a clear version-mismatch error. Silently writing Arrow output as Parquet is the outcome to avoid.

**Additional context**

This matters for the 55.0.0 release (#2369) because client/cluster skew is the steady state rather than an edge case: `pyballista` re-exports datafusion-python types, so the Python bindings cannot move to 55 until there is a matching `datafusion-python` release. Until then every Python user runs a 54 client against a 55 cluster.

Possible directions:

- Keep the new delegation but make the old positional encoding detectable, so an old client gets a clear error rather than a mis-decode.
- Bump `BALLISTA_PROTOCOL_VERSION` and reject 54 clients outright. This is the same gap described in #2370, where the protocol version is validated for executors but not for clients.

At minimum the Arrow and Avro cases should not fail silently.

Related: #2348, #2374, #2369, #2370

andygrove:

Closes #2376.

# Rationale for this change

#2348 delegated file format serde to DataFusion's `DefaultLogicalExtensionCodec`. That is a good cleanup in isolation, but it changed the `CopyTo` wire format in a way that older clients cannot detect, and the two encodings are structurally identical so nothing errors at the transport layer.

| | field 1 | field 2 |
|---|---|---|
| Ballista <= 54 | `encoder_position: u32`, an index into `[Parquet, Csv, Json, Arrow, Avro]` | `blob` |
| Ballista post-#2348 | `kind: FileFormatKind` (`UNSPECIFIED=0, CSV=1, JSON=2, PARQUET=3, ARROW=4, AVRO=5`) | `encoded_file_format` |

Both are a varint followed by a bytes field, so prost decodes an old message happily and the format tag is simply reinterpreted:

```
pos=0 (Parquet) -> Err(This feature is not implemented: Unspecified file format kind)
pos=1 (Csv)     -> Ok(csv)       lines up by coincidence
pos=2 (Json)    -> Ok(json)      lines up by coincidence
pos=3 (Arrow)   -> Ok(parquet)   silently wrong format
pos=4 (Avro)    -> Ok(arrow)     silently wrong format
```

Parquet at least fails loudly, which is what the new job in #2374 caught when a released 54.0.0 Python client ran against a cluster built from the branch. Arrow is the one that worries me: `ArrowLogicalExtensionCodec::try_encode_file_format` writes an empty payload and the Parquet decoder treats empty bytes as all defaults, so `COPY ... STORED AS ARROW` from a 54 client executes as Parquet with no error at all.

This is not transient skew. `pyballista` re-exports datafusion-python types, so the Python bindings cannot move to 55 until there is a matching `datafusion-python` release, and until then every Python user runs a 54 client against a 55 cluster.

Worth noting that DataFusion deliberately kept its own logical proto backward compatible across 54 to 55: removed fields became `reserved 8; // was bool collect_stat`, `string location = 2` was kept and marked deprecated alongside a new `repeated string locations = 16`, and everything else is additive. So a 54 client's plan is meant to be readable by a 55 scheduler, and 73 of the 75 Python tests in #2374 confirm it is. This encoding change was the one thing that broke it.

- Reverts 6fcc7360b, restoring the `encoder_position` based `FileFormatProto`.
- Adds `decodes_file_format_bytes_from_a_54_client`, which pins the wire format by re-declaring the legacy layout locally and asserting that those bytes still decode to the right format.

The existing `file_format_serialization_roundtrip` test could not catch this because it encodes and decodes with the same codec, so it passes for any self-consistent encoding. The new test fails on the pre-revert code with exactly the error CI hit, and passes after the revert.

Avro is deliberately left out of the new test. DataFusion's `AvroLogicalExtensionCodec::try_decode_file_format` returns an `ArrowFormatFactory` in both 54 and 55, which looks like an upstream bug and is unrelated to this change.

# Are these changes tested?

- New regression test above, verified to fail before the revert and pass after.
- `cargo test -p ballista-core --lib serde::` passes, 32 tests.
- `cargo clippy --all-targets --workspace --all-features -- -D warnings` clean.
- `cargo fmt --all -- --check` clean.
- The real check is #2374, whose `Python client vs branch-built cluster` job should go green once this lands.

Yes, in the sense that it restores compatibility. Released 54 clients can once again run `CopyTo` against a scheduler built from main.

I am not proposing we carry the duplicated codec forever. The better end state is client side validation of `BALLISTA_PROTOCOL_VERSION` (#2370), after which #2348 can land again on a bumped protocol version and old clients get a clear error instead of a mis-decode. That is a larger change than I would want to rush before 55.0.0 (#2369), and on its own it would not help the Python client, since it would turn silent breakage into a hard block that users cannot act on until datafusion-python 55 exists.

@milenkovicm sorry to revert your cleanup, happy to help re-land it behind a version check once the client handshake is in place.

### DIFF
--- a/ballista/core/src/serde/mod.rs
+++ b/ballista/core/src/serde/mod.rs
@@ -27,6 +27,10 @@ use datafusion::common::{DataFusionError, Result};
 use datafusion::execution::TaskContext;
 use datafusion::logical_expr::Extension;
 use datafusion::physical_plan::{ExecutionPlan, Partitioning};
+use datafusion_proto::logical_plan::file_formats::{
+    ArrowLogicalExtensionCodec, AvroLogicalExtensionCodec, CsvLogicalExtensionCodec,
+    JsonLogicalExtensionCodec, ParquetLogicalExtensionCodec,
+};
 use datafusion_proto::physical_plan::from_proto::parse_physical_sort_exprs;
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_hash_partitioning;
 use datafusion_proto::physical_plan::from_proto::parse_protobuf_partitioning;
@@ -188,12 +192,49 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> BallistaCodec<T, 
 #[derive(Debug)]
 pub struct BallistaLogicalExtensionCodec {
     default_codec: Arc<dyn LogicalExtensionCodec>,
+    file_format_codecs: Vec<Arc<dyn LogicalExtensionCodec>>,
+}
+
+impl BallistaLogicalExtensionCodec {
+    /// looks for a codec which can operate on this node
+    /// returns a position of codec in the list and result.
+    ///
+    /// position is important with encoding process
+    /// as position of used codecs is needed
+    /// so the same codec can be used for decoding
+    fn try_any<R>(
+        &self,
+        mut f: impl FnMut(&dyn LogicalExtensionCodec) -> Result<R>,
+    ) -> Result<(u32, R)> {
+        let mut last_err = None;
+        for (position, codec) in self.file_format_codecs.iter().enumerate() {
+            match f(codec.as_ref()) {
+                Ok(result) => return Ok((position as u32, result)),
+                Err(err) => last_err = Some(err),
+            }
+        }
+
+        Err(last_err.unwrap_or_else(|| {
+            DataFusionError::Internal(
+                "List of provided extended logical codecs is empty".to_owned(),
+            )
+        }))
+    }
 }
 
 impl Default for BallistaLogicalExtensionCodec {
     fn default() -> Self {
         Self {
             default_codec: Arc::new(DefaultLogicalExtensionCodec {}),
+            // Position in this list is important as it will be used for decoding.
+            // If new codec is added it should go to last position.
+            file_format_codecs: vec![
+                Arc::new(ParquetLogicalExtensionCodec {}),
+                Arc::new(CsvLogicalExtensionCodec {}),
+                Arc::new(JsonLogicalExtensionCodec {}),
+                Arc::new(ArrowLogicalExtensionCodec {}),
+                Arc::new(AvroLogicalExtensionCodec {}),
+            ],
         }
     }
 }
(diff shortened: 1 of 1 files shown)


milenkovicm (approved):
no problem @andygrove 
I did not really expect this is going to break this way, but good time to catch it

PR #2387, opened 2026-08-27
feat: cleanup & simplify ballista config 

milenkovicm:

Closes #.

# Rationale for this change

We had quite a lot of boiler plate code to expose ballista config option, also exposed names might differ. 
Also we did not have uniform naming strategy for exposed configuration properties 

- macro which generates ballista config getters and setters
- change to some method names 
- deprecations to methods not following naming strategy

Yes, there are method name and type changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1133,7 +1133,7 @@ dependencies = [
  "md-5 0.11.0",
  "object_store",
  "parking_lot",
- "paste",
+ "pastey",
  "prost",
  "prost-types",
  "rand 0.10.2",
@@ -5100,10 +5100,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "paste"
-version = "1.0.15"
+name = "pastey"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
+checksum = "2ee67f1008b1ba2321834326597b8e186293b049a023cdef258527550b9935b4"
 
 [[package]]
 name = "pathdiff"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1133,6 +1133,7 @@ dependencies = [
  "md-5 0.11.0",
  "object_store",
  "parking_lot",
+ "paste",
  "prost",
  "prost-types",
  "rand 0.10.2",
@@ -5099,6 +5100,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
+
+[[package]]
 name = "pathdiff"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6977,7 +6984,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.3.4",
+ "getrandom 0.4.3",
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6984,7 +6984,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.3",
+ "getrandom 0.3.4",
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ itertools = "0.15"
 mimalloc = { version = "0.1" }
 object_store = "0.13.2"
 ordered-float = "5"
+paste = "1.0"
 prost = "0.14"
 prost-types = "0.14"
 rstest = { version = "0.26" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ itertools = "0.15"
 mimalloc = { version = "0.1" }
 object_store = "0.13.2"
 ordered-float = "5"
-paste = "1.0"
+pastey = "0.2.3"
 prost = "0.14"
 prost-types = "0.14"
 rstest = { version = "0.26" }
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -809,7 +809,7 @@ impl BallistaConfig {
     }
 
     /// Deprecated alias for [`Self::use_tls`].
-    #[deprecated(note = "renamed to `use_tls`")]
+    #[deprecated(since = "55.0.0", note = "renamed to `use_tls`")]
     pub fn client_use_tls(&self) -> bool {
         self.use_tls()
     }
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -581,8 +581,14 @@ impl BallistaConfig {
     }
 
     /// Returns the standalone processing parallelism level.
-    pub fn default_standalone_parallelism(&self) -> usize {
+    pub fn standalone_parallelism(&self) -> usize {
         self.get_usize_setting(BALLISTA_STANDALONE_PARALLELISM)
+    }
+
+    /// Deprecated alias for [`Self::standalone_parallelism`].
+    #[deprecated(note = "renamed to `standalone_parallelism`")]
+    pub fn default_standalone_parallelism(&self) -> usize {
+        self.standalone_parallelism()
     }
 
     /// Returns the maximum number of concurrent shuffle reader requests.
@@ -798,8 +804,14 @@ impl BallistaConfig {
     }
 
     /// should client use TLS to communicate with ballista cluster
-    pub fn client_use_tls(&self) -> bool {
+    pub fn use_tls(&self) -> bool {
         self.get_bool_setting(BALLISTA_CLIENT_USE_TLS)
+    }
+
+    /// Deprecated alias for [`Self::use_tls`].
+    #[deprecated(note = "renamed to `use_tls`")]
+    pub fn client_use_tls(&self) -> bool {
+        self.use_tls()
     }
 
     /// Returns the number of retries for IO operations in the Ballista client.
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -752,8 +752,8 @@ impl BallistaConfig {
 
     /// Returns the target post-coalesce partition byte size in bytes
     /// (Spark's `advisoryPartitionSizeInBytes`).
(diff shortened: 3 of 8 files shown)


ballista/core/src/config.rs
```diff
@@ -581,10 +581,16 @@ impl BallistaConfig {
     }
 
     /// Returns the standalone processing parallelism level.
-    pub fn default_standalone_parallelism(&self) -> usize {
+    pub fn standalone_parallelism(&self) -> usize {
         self.get_usize_setting(BALLISTA_STANDALONE_PARALLELISM)
     }
 
+    /// Deprecated alias for [`Self::standalone_parallelism`].
+    #[deprecated(note = "renamed to `standalone_parallelism`")]
```
```suggestion
    #[deprecated(since = "55.0.0", note = "renamed to `standalone_parallelism`")]
```
this makes it easier to decide when it is time to remove the deprecated method.

ballista/core/src/config.rs
```diff
@@ -798,10 +804,16 @@ impl BallistaConfig {
     }
 
     /// should client use TLS to communicate with ballista cluster
-    pub fn client_use_tls(&self) -> bool {
+    pub fn use_tls(&self) -> bool {
         self.get_bool_setting(BALLISTA_CLIENT_USE_TLS)
     }
 
+    /// Deprecated alias for [`Self::use_tls`].
+    #[deprecated(note = "renamed to `use_tls`")]
```
```suggestion
    #[deprecated(since = "55.0.0", note = "renamed to `use_tls`")]
```

Cargo.toml
```diff
@@ -59,6 +59,7 @@ itertools = "0.15"
 mimalloc = { version = "0.1" }
 object_store = "0.13.2"
 ordered-float = "5"
+paste = "1.0"
```
martin-g: https://github.com/dtolnay/paste is not maintained since few years.
Some tooling (`cargo deny` ?!) disallows it.
I believe https://github.com/as1100k/pastey is the most used fork of it.

ballista/core/src/extension.rs
```diff
@@ -1208,6 +1090,41 @@ mod test {
         )
     }
 
+    #[test]
+    fn should_round_trip_all_macro_generated_options() {
+        let plain = SessionConfig::new_with_ballista()
+            .with_ballista_standalone_parallelism(123)
```
martin-g:
```suggestion
        let plain = SessionConfig::new_with_ballista()
            .with_ballista_adaptive_query_planner(true)
            .with_ballista_standalone_parallelism(123)
```

ballista/core/src/extension.rs
```diff
@@ -1208,6 +1090,41 @@ mod test {
...
+            .with_ballista_shuffle_reader_maximum_concurrent_requests(7)
+            .with_ballista_coalesce_target_partition_bytes(2048)
+            .with_ballista_shuffle_reader_force_remote_read(true)
+            .with_ballista_shuffle_reader_remote_prefer_flight(true)
+            .with_ballista_use_tls(true)
+            .with_ballista_coalesce_enabled(true)
+            .with_ballista_coalesce_small_partition_factor(1.5)
+            .with_ballista_coalesce_merged_partition_factor(2.5);
+
+        assert!(plain.ballista_shuffle_reader_force_remote_read());
```
martin-g:
```suggestion
        assert!(plain.ballista_adaptive_query_planner_enabled());
        assert!(plain.ballista_shuffle_reader_force_remote_read());
```

ballista/core/src/extension.rs
```diff
@@ -1208,6 +1090,41 @@ mod test {
...
+        assert_eq!(plain.ballista_broadcast_join_threshold_bytes(), 789);
+        assert_eq!(plain.ballista_broadcast_join_threshold_rows(), 42);
+        assert_eq!(plain.ballista_hash_join_max_build_partition_bytes(), 999);
+        assert_eq!(plain.ballista_coalesce_target_partition_bytes(), 2048);
+        assert_eq!(plain.ballista_coalesce_small_partition_factor(), 1.5);
+        assert_eq!(plain.ballista_coalesce_merged_partition_factor(), 2.5);
+        assert_eq!(
+            plain.ballista_shuffle_reader_maximum_concurrent_requests(),
+            7
+        );
```
martin-g, line 1133: Should the deprecated getters be tested too ?

milenkovicm: this one is not deprecated,

ballista/core/src/extension.rs
```diff
@@ -277,9 +277,9 @@ pub trait SessionConfigExt {
     fn with_ballista_coalesce_enabled(self, enabled: bool) -> Self;
 
     /// Returns the target post-coalesce partition byte size in bytes.
-    fn ballista_coalesce_target_partition_bytes(&self) -> u64;
+    fn ballista_coalesce_target_partition_bytes(&self) -> usize;
     /// Sets the target post-coalesce partition byte size in bytes.
-    fn with_ballista_coalesce_target_partition_bytes(self, bytes: u64) -> Self;
+    fn with_ballista_coalesce_target_partition_bytes(self, bytes: usize) -> Self;
```
andygrove, line 282: `SessionConfigExt` is a public trait, so re-signing a method breaks external implementors and not just callers. I think this one is unavoidable, you cannot have two trait methods with the same name and different types, and inventing a new name just to dodge it would be worse. So I am fine taking the break here.

It does need an entry in `docs/source/upgrading/55.0.0.md` though, under the existing "API changes" heading. Something short saying these two moved from `u64` to `usize` and callers passing an explicit `u64` will need a cast.

ballista/core/src/config.rs
```diff
@@ -746,8 +752,8 @@ impl BallistaConfig {
 
     /// Returns the target post-coalesce partition byte size in bytes
     /// (Spark's `advisoryPartitionSizeInBytes`).
-    pub fn coalesce_target_partition_bytes(&self) -> u64 {
-        self.get_usize_setting(BALLISTA_COALESCE_TARGET_PARTITION_BYTES) as u64
+    pub fn coalesce_target_partition_bytes(&self) -> usize {
```
andygrove, line 755: This one is a hard break with no deprecation path, which makes the PR a bit inconsistent with itself. `default_standalone_parallelism` and `client_use_tls` both got a `#[deprecated]` alias a few hundred lines up, but this changes a public return type in place.

It could have been additive. Either keep a deprecated `u64`-returning shim under the old name and add the new one, or just leave this returning `u64` and let the macro-generated trait method do the cast. Happy either way as long as it is a deliberate call rather than a side effect of lining the types up for the macro, but if it stays as is then it belongs in the upgrade guide alongside the trait change.

ballista/core/src/extension.rs
```diff
@@ -373,6 +373,80 @@ impl SessionStateExt for SessionState {
...
+        $self.set_usize($const, $val)
+    };
+    (u64, $self:expr, $const:expr, $val:expr) => {
+        $self.set_u64($const, $val)
+    };
+    (f64, $self:expr, $const:expr, $val:expr) => {
+        $self.set_str($const, &$val.to_string())
+    };
+}
+
```
andygrove, line 393: Worth calling out in this doc comment that the macro now couples two public naming schemes. Because the trait method name is derived from the `BallistaConfig` method name, `BallistaConfig`'s public getters and `SessionConfigExt`'s methods can no longer be renamed independently. That is what forced the `default_standalone_parallelism` and `client_use_tls` renames, and it is a reasonable trade, but the next person renaming a config getter should know they are also renaming a public trait method.

The `as <setter_name>` escape hatch already covers the one case where the two diverge, which is a nice touch.

ballista/core/src/extension.rs
```diff
@@ -1208,6 +1090,44 @@ mod test {
         )
     }
 
+    #[test]
+    fn should_round_trip_all_macro_generated_options() {
+        let plain = SessionConfig::new_with_ballista()
```
andygrove: `new_with_ballista()` installs the `BallistaConfig` extension, so this only exercises the `if` branch of each generated setter. The `else` branch, where the setter is called on a `SessionConfig` that has no ballista extension yet and has to install one first, never runs. For a refactor whose whole claim is that the generated code does what the hand-written code did, that is the half worth covering. A second pass over a plain `SessionConfig::new()` would do it, and the variable name here suggests that was the intent.

Small thing too, `ballista_shuffle_reader_force_remote_read()` is asserted twice below, so one of those can go.

Cargo.lock
```diff
@@ -6977,7 +6984,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.3.4",
+ "getrandom 0.4.3",
```
andygrove: This `getrandom` bump under `tempfile` looks like incidental lockfile regen churn rather than anything `pastey` pulled in. Probably cleaner to drop it from the diff so the lockfile change is just the new dependency.

andygrove (commented):
Nice cleanup. The macro removes a lot of genuinely repetitive code, and the two renames that kept deprecated aliases are exactly the right pattern.

I reviewed this against the user personas in `docs/source/contributors-guide/user-personas.md`. Persona 3 (the library user building on Ballista) trips one red flag, "Breaking a public API signature without an upgrade-guide entry". Three signatures change type here and none of them are covered in `docs/source/upgrading/55.0.0.md`. Details inline, but the short version is that the trait break is probably unavoidable and fine, the `BallistaConfig` one could have followed the same deprecation pattern as the other two renames, and either way this needs a note in the upgrade guide under the existing "API changes" heading. The `api-change` label is already on, so that half is done.

One thing about the PR description. It says "some config names have changed" under user-facing changes, but as far as I can tell no `BALLISTA_*` key strings moved, only Rust method names, so nobody's `SET ballista.x = y` breaks. Worth correcting so the release notes do not scare people.

Everything else came back clean. Persona 1 is unaffected, no change to results or semantics and the `datafusion.*` override path is untouched. Persona 2 is unaffected, the AQE coalesce rule change is a pure cast with no behavior change and the stage/task model, AQE gating, and observability are all intact. No extension point narrowed, no codec argument dropped, `RuntimeProducer` and `ConfigProducer` untouched, no wire-format change. And since the only change under the AQE optimizer is that cast, I do not think this needs TPC-H numbers.

ballista/core/src/extension.rs
milenkovicm: `SessionConfigExt` implements extension pattern there should be no other implementers. 

`docs/source/upgrading/55.0.0.md` should definitely be updated

milenkovicm: done

ballista/core/src/extension.rs
milenkovicm: remove duplicated assert, added new test method to cover case when configuration is created with `new()` rather than `new_with_ballista()`

ballista/core/src/config.rs
milenkovicm: yes, this was deliberate, as all other public methods exposed in `SessionConfigExt` return `usize` instead of `u64`

milenkovicm: if we revert it to previous we would have `usize as u64` then in the other method we'd have `u64 to usize`.  maybe we should avoid using usize and use u64 everywhere, we can change it now as one big break?

Cargo.lock
milenkovicm: changed, cargo looks happy with the change

ballista/core/src/config.rs
andygrove: I think keeping `usize` is more consistent with upstream DataFusion

andygrove (approved):
Thanks @milenkovicm

PR #2389, opened 2026-08-27
feat: add build-gated Hudi read integration

wirybeaver:
## Dependencies

- Stacked on #2388; review the second commit (`feat: add build-gated Hudi read integration`) for this PR's own diff.
- Uses the DataFusion 55 compatibility work in apache/hudi-rs#698, pinned to commit `c3e4009136443057be564b717c41c99420ae821c` until that change is released.

Proves the build-gated provider boundary proposed for #1241 with a real open-table-format read.

## What changes are included?

- Adds an independently built `integrations/hudi` crate excluded from Ballista's workspace.
- Implements `HudiProviderCodec` for hudi-rs `HudiDataSource`.
- Transports only the table base URI and caller-supplied options, then reconstructs the provider on the scheduler.
- Uses hudi-rs's standard Parquet `DataSourceExec` path for copy-on-write reads.
- Adds a focused workflow for the excluded integration crate.

The hudi-rs dependency remains outside Ballista's root `Cargo.lock`, default dependency tree, published package, and release graph. Scope remains read-only Parquet copy-on-write; MOR custom execution, writes, and dynamic loading are intentionally excluded.

- `cargo check -p ballista --no-default-features --locked`
- `cargo check -p ballista --no-default-features --features contrib-hudi --locked`
- `cargo test --manifest-path integrations/hudi/Cargo.toml --test distributed_read`
- `cargo clippy --manifest-path integrations/hudi/Cargo.toml --all-targets --no-deps -- -D warnings`
- `cargo package -p ballista --allow-dirty --no-verify`
- Verified with `cargo tree` that neither the default Ballista build nor the build-gate feature contains a Hudi crate.
- `git diff --check`

The focused test runs a projected and filtered Hudi COW query through a two-worker standalone Ballista cluster and verifies the returned rows.

### DIFF
--- a/integrations/hudi/Cargo.toml
+++ b/integrations/hudi/Cargo.toml
@@ -1,0 +1,32 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "hudi-ballista"
+version = "54.0.0"
+edition = "2024"
+rust-version = "1.94.0"
+license = "Apache-2.0"
+publish = false
+
+[dependencies]
+ballista = { path = "../../ballista/client", version = "54.0.0", features = ["contrib-hudi", "standalone"] }
+hudi-datafusion = { git = "https://github.com/wirybeaver/hudi-rs.git", rev = "c3e4009136443057be564b717c41c99420ae821c", default-features = false }
+hudi-test = { git = "https://github.com/wirybeaver/hudi-rs.git", rev = "c3e4009136443057be564b717c41c99420ae821c" }
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+
+[workspace]
--- a/.github/workflows/hudi.yml
+++ b/.github/workflows/hudi.yml
@@ -1,0 +1,45 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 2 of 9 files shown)


ballista/core/src/serde/mod.rs
```diff
@@ -188,14 +188,117 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> BallistaCodec<T,
...
+    fn decode_plan(
+        &self,
+        payload: &[u8],
+        inputs: &[Arc<dyn ExecutionPlan>],
+        ctx: &TaskContext,
+        proto_converter: &dyn PhysicalProtoConverterExtension,
+    ) -> Result<Arc<dyn ExecutionPlan>>;
+}
+
+const DATA_SOURCE_CODEC_PREFIX: &[u8] = b"ballista-data-source-v1\0";
```
NoahKusaba: ?

milenkovicm (commented):
just a heads up @wirybeaver, i would be reluctant to merge this to main at the moment for multiple reasons 

1. hudi does not really look maintained 
2. this would increase workload on maintainers
3. this would add to our dependencies and increase our release complexity

i would suggest for now to keep this as a standalone project out of ballista, if change are needed to ballista to accommodate that please let me know  

hope you understand

integrations/hudi/Cargo.toml
```diff
@@ -17,10 +15,18 @@
...
+name = "hudi-ballista"
+version = "54.0.0"
+edition = "2024"
+rust-version = "1.94.0"
+license = "Apache-2.0"
+publish = false
+
+[dependencies]
+ballista = { path = "../../ballista/client", version = "54.0.0", features = ["contrib-hudi", "standalone"] }
+hudi-datafusion = { git = "https://github.com/wirybeaver/hudi-rs.git", rev = "c3e4009136443057be564b717c41c99420ae821c", default-features = false }
```
milenkovicm, line 28: this depends on unreleased version,

milenkovicm (commented):
I'm still failing to understand why we need to bring this integration to ballista ?

PR depends on unmaintained, private fork of hudi, hence my comments are same like last time, i would be reluctant to merge this to ballista main

integrations/hudi/Cargo.toml
wirybeaver: I understand that part. I need Hudi to Drive Hudi community to absorb my contributions. That's the reason I make this PR as Draft status

wirybeaver:
Thank you for the concern. I agree that an unreleased fork should not become part of Ballista's release dependencies. **#2389 is intentionally still a Draft and I am not proposing that it be merged in its current state.** It should only be reconsidered after hudi-rs has an official compatible release.

```text
#2389: validation only (Draft)

Ballista workspace / root Cargo.lock / published package
                         |
                         | no hudi-rs edge
                         v
                 +----------------+
                 | build gate/SPI |
                 +-------+--------+
                         |
              explicit standalone build
                         |
                         v
              +-----------------------+
              | integrations/hudi     |
              | separate crate        |
              | excluded from workspace|
              | hudi-rs pin lives here|
              +-----------------------+
```

The purpose of this Draft is to prove that the #2388 boundary can execute a realistic projected and filtered distributed read while keeping hudi-rs outside Ballista's main crate, root lockfile, published package, and release graph. The private-fork pin is temporary test scaffolding confined to that excluded crate.

The intended long-term convention is one independent crate per open table format under the Apache project umbrella. Iceberg, Delta Lake, Lance, and Hudi integrations would follow the same build-gated boundary and own their dependencies separately. Hudi was selected because Iceberg and Delta work already exists, but not yet under this proposed isolation model. If useful, I can add Lance as a second example so the proposal can be evaluated as a general convention rather than as a request to merge Hudi specifically.

PR #2390, opened 2026-08-27
fix(ci): resolve datafusion.proto from datafusion-proto-models

### LINKED ISSUE #2377: Vendored datafusion.proto sync check silently passes without checking datafusion.proto
**Describe the bug**

`ci/scripts/rust_proto_check.sh` currently passes without checking `datafusion.proto` at all, and the vendored copy has drifted 471 diff lines from the DataFusion version we build against.

In DataFusion 55 the logical plan proto moved out of the `datafusion-proto` crate into a new `datafusion-proto-models` crate:

- 54: `datafusion/proto/proto/datafusion.proto`, shipped in the `datafusion-proto` crate
- 55: `datafusion/proto-models/proto/datafusion.proto`, shipped in the `datafusion-proto-models` crate

`dev/update_datafusion_proto.py` still looks for it under `datafusion-proto`:

```python
FILES = {
    "datafusion.proto": ("datafusion-proto", "datafusion.proto"),
    "datafusion_common.proto": ("datafusion-proto-common", "datafusion_common.proto"),
}
```

The lookup misses, and the "some releases don't publish the .proto" escape hatch that was added for pre-54 `datafusion-proto-common` swallows it:

```
note: datafusion-proto@55.0.0 does not ship proto/datafusion.proto; leaving vendored datafusion.proto unchanged
```

The script then exits 0, so the CI job "Check vendored DataFusion proto is in sync" is green while checking only `datafusion_common.proto`.

**To Reproduce**

```
$ python3 dev/update_datafusion_proto.py --check
note: datafusion-proto@55.0.0 does not ship proto/datafusion.proto; leaving vendored datafusion.proto unchanged
$ echo $?
0
```

The real source is present and resolvable. `datafusion-proto-models` is already in `Cargo.lock`, and the crate does ship the file, so `cargo metadata` can find it. Comparing the vendored copy against it (after the usual import rewrite) reports `STALE` with 471 diff lines, including things like `reserved 8; // was bool collect_stat`, the new `RangeRepartition` variant on `Partitioning`, `repeated string locations = 16` on the scan node, and the `MERGE_INTO` DML additions.

**Expected behavior**

The check should resolve `datafusion.proto` from `datafusion-proto-models` and fail on drift, and a missing source file for a crate that is expected to ship one should be an error rather than a skip.

**Additional context**

There is no build breakage today. The vendored files are build-time stubs only, with no Rust generated from them, since `ballista/core/build.rs` maps their packages to the real crates via `extern_path`. `ballista.proto` refers to `datafusion.Partitioning`, `datafusion.PhysicalPlanNode` and friends only by type name, so protoc still resolves everything and the generated Rust comes from `datafusion-proto` itself.

The problem is that the guard described in CLAUDE.md is not currently guarding, so a future change that does matter would land unnoticed.

Suggested fix:

- Point the `datafusion.proto` entry at `datafusion-proto-models` (with a fallback to `datafusion-proto` if we want the script to keep working on 54).
- Tighten the skip path so it only applies to a known allowlist, or drop it now that both crates ship their protos, so a missing file fails loudly.
- Re-vendor and commit the refreshed `datafusion.proto`.

Found while investigating #2376.

AnuragRaut08:

Closes #2377.

# Rationale for this change

`dev/update_datafusion_proto.py` was silently passing CI without actually checking `datafusion.proto`. In DataFusion 55 the logical-plan proto moved from the `datafusion-proto` crate into a new `datafusion-proto-models` crate. The script still targeted `datafusion-proto`, hit a broad "crate doesn't ship this file" escape hatch, printed a `note:` to stderr, and **exited 0** — so the CI job could remain green while validating nothing for `datafusion.proto`.

Reproducer (before this fix):

```text
$ python3 dev/update_datafusion_proto.py --check
note: datafusion-proto@55.0.0 does not ship proto/datafusion.proto; leaving vendored datafusion.proto unchanged
$ echo $?
0
```

**`dev/update_datafusion_proto.py`**

* `FILES` values are now `([ordered crate list], filename)`. For `datafusion.proto` the list is `["datafusion-proto-models", "datafusion-proto"]` — the script tries each in order and uses the first one that actually ships the file. This keeps the script working on both v55+ (`datafusion-proto-models`) and v54 (`datafusion-proto`) without a separate code path.
* `_ALL_CRATES` is derived automatically from `FILES`, so `crate_source_dirs()` always looks up every candidate without a separate maintenance step.
* The silent-skip path (`if not src.exists(): continue`) is **removed**. If none of the candidate crates ships the expected file, the script now exits with a diagnostic listing which crates were tried, so a future rename fails loudly instead of silently passing CI.
* The crate lookup no longer relies on a separate hard-coded list of required crates; validation is derived from the entries in `FILES`.

**`ballista/core/proto/datafusion.proto`**

Re-vendored from `datafusion-proto-models@55.0.0` (211 insertions, 39 deletions vs the previously stale copy). Key additions:

* `reserved 8/9` annotations on `ListingTableScanNode` (was `collect_stat` / `target_partitions`)
* `RangeRepartition` variant on `RepartitionNode` + new `RangeSplitPoint` message
* `repeated string locations = 16` on `CreateExternalTableNode`
* `AnalyzeNode` gains `analyze_level`, `analyze_categories`, `format` fields
* `ExplainNode` gains `show_statistics` field
* `MERGE_INTO` added to `DmlNode.Type` enum

No. The vendored `.proto` files are build-time stubs only — `ballista/core/build.rs` maps their packages to the real DataFusion crates via `extern_path`, so no Rust code is generated from them in Ballista. This change only makes the CI guard actually guard.

### DIFF
--- a/ballista/core/proto/datafusion.proto
+++ b/ballista/core/proto/datafusion.proto
@@ -92,8 +92,8 @@ message ListingTableScanNode {
   datafusion_common.Schema schema = 5;
   repeated LogicalExprNode filters = 6;
   repeated PartitionColumn table_partition_cols = 7;
-  bool collect_stat = 8;
-  uint32 target_partitions = 9;
+  reserved 8; // was bool collect_stat
+  reserved 9; // was uint32 target_partitions
   oneof FileFormatType {
     datafusion_common.CsvFormat csv = 10;
     datafusion_common.ParquetFormat parquet = 11;
@@ -148,7 +148,17 @@ message RepartitionNode {
   oneof partition_method {
     uint64 round_robin = 2;
     HashRepartition hash = 3;
+    RangeRepartition range = 4;
   }
+}
+
+message RangeSplitPoint {
+  repeated datafusion_common.ScalarValue value = 1;
+}
+
+message RangeRepartition {
+  repeated SortExprNode sort_expr = 1;
+  repeated RangeSplitPoint split_point = 2;
 }
 
 message HashRepartition {
@@ -163,7 +173,8 @@ message EmptyRelationNode {
 message CreateExternalTableNode {
   reserved 1; // was string name
   TableReference name = 9;
-  string location = 2;
+  string location = 2; // deprecated; use repeated locations
+  repeated string locations = 16;
   string file_type = 3;
   datafusion_common.DfSchema schema = 4;
   repeated string table_partition_cols = 5;
@@ -224,12 +235,22 @@ message ValuesNode {
 message AnalyzeNode {
   LogicalPlanNode input = 1;
   bool verbose = 2;
+  // Statement-level override for `datafusion.explain.analyze_level`.
+  // Absent means "fall back to session config".
+  optional datafusion_common.MetricType analyze_level = 3;
+  // Statement-level override for `datafusion.explain.analyze_categories`.
+  // Absent means "fall back to session config".
+  optional datafusion_common.ExplainAnalyzeCategoriesNode analyze_categories = 4;
+  datafusion_common.ExplainFormat format = 5;
 }
 
 message ExplainNode {
   LogicalPlanNode input = 1;
   bool verbose = 2;
   datafusion_common.ExplainFormat format = 3;
+  // Statement-level override for `datafusion.explain.show_statistics`.
+  // Absent means "fall back to session config".
+  optional bool show_statistics = 4;
 }
 
 message AggregateNode {
@@ -293,7 +314,7 @@ message FileFormatProto {
 }
 
 message DmlNode{
-   enum Type {
+  enum Type {
     UPDATE = 0;
     DELETE = 1;
     CTAS = 2;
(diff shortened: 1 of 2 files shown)


andygrove:
Nice catch on the escape hatch, I reproduced the silent pass on the parent commit and confirmed the check actually fails on drift now.

One small thing in `crate_source_dirs()`. You derive `_ALL_CRATES` from `FILES` automatically, which is great, but then the required check right under it hard-codes `required = {"datafusion-proto-common"}` while the comment above says "at least one crate per FILES entry must be present". Not a bug today since the `src is None` path catches a missing crate with a good message, but it is the same shape of hidden hard-code that caused the original problem, so a third entry in `FILES` would quietly not be covered. Either derive it from `FILES` or drop the check and let the `src is None` path own it.

AnuragRaut08:
Good catch. I’ve removed the hard-coded `required` check and now rely on the existing `src is None` path to handle missing crates, keeping the validation fully derived from `FILES`.

AnuragRaut08:
@andygrove The hard-coded crate requirement has been removed as discussed, and the PR is still passing all 24 checks. This should now be ready for review/merge when you get a chance.

andygrove (approved):
Thanks @AnuragRaut08

comphead (approved):
Thanks @AnuragRaut08

comphead:
My question would be: will CI keep track on proto inconsistencies?

AnuragRaut08:
Yes. The existing CI job runs `dev/update_datafusion_proto.py --check`, and with this change the script resolves `datafusion.proto` from `datafusion-proto-models` and compares it against the vendored copy. Any drift now causes the check to fail instead of silently passing.

milenkovicm (approved):
This makes sense to me thanks @AnuragRaut08

AnuragRaut08:
@andygrove Thanks! All requested changes are addressed, with 3 approvals, all checks passing, and no conflicts. Feel free to merge when convenient.

PR #2393, opened 2026-08-27
fix(executor): wake graceful shutdown after tasks drain

### LINKED ISSUE #2356: Executor graceful shutdown hangs when tasks are active
**Describe the bug**

When an executor receives SIGTERM or Ctrl-C while running a task, it hangs indefinitely instead of exiting after the task completes.

**To Reproduce**

1. Start a scheduler and executor.
2. Submit a query that keeps an executor task active.
3. Send SIGTERM or Ctrl-C to the executor.
4. Observe that the executor does not exit after the task finishes.

**Expected behavior**

The executor should exit after its active tasks have drained.

**Additional context**

`TasksDrainedFuture` in `ballista/executor/src/executor.rs` returns `Poll::Pending` when tasks are active but does not register a waker. Removing the final task from `abort_handles` therefore does not wake the shutdown future.

This is distinct from #2105/#2107, which fixed shutdown notifications not reaching subscribers.

1fanwang:
Fixes https://github.com/apache/datafusion-ballista/issues/2356

An executor can hang after receiving SIGTERM or Ctrl-C while a task is active.
The task eventually finishes, but `TasksDrainedFuture` returned `Pending`
without registering a waker, so shutdown never checked the task map again.

The drain future now registers an `AtomicWaker` before checking active tasks.
Cancellation leaves the task registered until `execute_query_stage` finishes
its cleanup. Removing the last task then wakes the drain future and lets
shutdown continue.

### Testing

The regression drives the real task registration and cancellation path. On
`7e939fe26c8cfacc41ed16a656704d69c94ea380`, with only the regression applied:

```text
$ cargo test -p ballista-executor executor::test::test_task_cancellation --lib -- --nocapture
thread 'executor::test::test_task_cancellation' panicked:
assertion failed: tokio::time::timeout(Duration::from_secs(1), tasks_drained).await.is_ok()
test executor::test::test_task_cancellation ... FAILED
test result: FAILED. 0 passed; 1 failed; 52 filtered out
```

On this branch:

```text
$ cargo test -p ballista-executor executor::test::test_task_cancellation --lib -- --nocapture
test executor::test::test_task_cancellation ... ok
test result: ok. 1 passed; 0 failed; 52 filtered out

$ cargo test -p ballista-executor --lib
test result: ok. 53 passed; 0 failed; 0 ignored; 0 measured
```

### DIFF
--- a/ballista/executor/src/executor.rs
+++ b/ballista/executor/src/executor.rs
@@ -315,7 +315,9 @@ mod test {
         DefaultSessionRuntimeCache, MemoryPoolPolicy, SessionRuntimeCache,
     };
     use ballista_core::RuntimeProducer;
+    use ballista_core::error::BallistaError;
     use ballista_core::execution_plans::ShuffleWriterExec;
+    use ballista_core::serde::protobuf;
     use ballista_core::serde::protobuf::ExecutorRegistration;
     use ballista_core::serde::scheduler::TaskKey;
     use ballista_core::utils::default_config_producer;
@@ -449,21 +451,12 @@ mod test {
         }
     }
 
-    #[tokio::test]
-    async fn test_task_cancellation() {
-        let work_dir = TempDir::new().unwrap().path().to_str().unwrap().to_string();
+    /// The result `execute_query_stage` hands back once a spawned task unwinds.
+    type TaskOutcome = Result<Vec<protobuf::ShuffleWritePartition>, BallistaError>;
 
-        let shuffle_write = ShuffleWriterExec::try_new(
-            "job-id".into(),
-            1,
-            Arc::new(NeverendingOperator::new()),
-            work_dir.clone(),
-        )
-        .expect("creating shuffle writer");
-
-        let query_stage_exec =
-            DefaultQueryStageExec::new(ShuffleWriterVariant::Passthrough(shuffle_write));
-
+    /// Builds an executor over `work_dir`, along with the session context whose
+    /// runtime its tasks run on.
+    fn never_ending_executor(work_dir: &str) -> (Arc<Executor>, SessionContext) {
         let executor_registration = ExecutorRegistration {
             id: "executor".to_string(),
             ..Default::default()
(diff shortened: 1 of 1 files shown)


ballista/executor/src/executor.rs
```diff
@@ -269,7 +282,7 @@ impl Executor {
         stage_id: usize,
         task_id: usize,
     ) -> Result<bool, BallistaError> {
-        if let Some((_, handle)) = self.abort_handles.remove(&TaskKey {
+        if let Some(handle) = self.abort_handles.get(&TaskKey {
```
milenkovicm, line 287: why do we need to change from `remove` to `get` why cant we remove the handle if task is cancelled

milenkovicm (commented):
thanks @1fanwang i hope we finally fix this issue, i have a comment please have a look when you get chance

ballista/executor/src/executor.rs
1fanwang: thanks for the pointer! fixed in https://github.com/apache/datafusion-ballista/commit/468dd81137748edad1d66ac780786072b6d42956.

ballista/executor/src/executor.rs
```diff
@@ -269,7 +282,8 @@ impl Executor {
         stage_id: usize,
         task_id: usize,
     ) -> Result<bool, BallistaError> {
-        if let Some((_, handle)) = self.abort_handles.remove(&TaskKey {
+        // Keep the handle registered until execute_query_stage finishes cleanup.
+        if let Some(handle) = self.abort_handles.get(&TaskKey {
```
milenkovicm, line 287: i believe task ower there will return error, at the cancellation and cleanup / remove would be performed there ? Am I correct? if I am, please update this comment adding those details, also please add comment over there so that there is dependency to `cancel_task`

milenkovicm (commented):
@1fanwang thanks a lot, one more comment

ballista/executor/src/executor.rs
1fanwang: Done in 8ea0335a.

avantgardnerio (approved):
This looks like a real improvement to me.

I proposed an addition test, take it if you think its worthwhile https://github.com/1fanwang/datafusion-ballista/pull/1

milenkovicm (approved):
thanks @1fanwang please let us know if you would like to add proposed test

1fanwang:
I added the test in ebfb5f9e. Thanks @avantgardnerio: it covers the multi-task drain case my original test missed, and I confirmed it fails on `assertion failed: wake_counter.0.load(Ordering::SeqCst) > wakes_before_last` when the waker registration is removed.

milenkovicm:
Thanks @1fanwang & @avantgardnerio

PR #2395, opened 2026-08-29
fix(core): stop forwarding BallistaConfig defaults to the scheduler

### LINKED ISSUE #2371: Python client silently forces the static planner (pinned to ballista 54.0.0 defaults)
**Describe the bug**

#2315 flipped `ballista.planner.adaptive.enabled` to default `true`, so a client built from `main` gets the adaptive (AQE) planner. The published Python client does not, and it silently forces the static planner instead.

`python/Cargo.toml` pins the Rust side to the released crates:

```toml
ballista = { version = "=54.0.0" }
ballista-core = { version = "=54.0.0" }
```

In 54.0.0 the default for that key is `false`. `BallistaConfig` materialises every setting including defaults and ships it to the scheduler with the session config, so the scheduler sees an explicit `ballista.planner.adaptive.enabled=false` and honours it. The user never asked for that, and there is nothing in the logs at INFO level to tell them which planner ran.

Confirmed against a cluster built at `apache/main` with scheduler debug logging on:

```
DEBUG ballista_scheduler::state::task_manager: Using adaptive query planner (AQE) for job planning   <- Rust client from main
DEBUG ballista_scheduler::state::task_manager: Using static query planner for job planning           <- ballista 54.0.0 from PyPI
```

Same query, same cluster, different planner purely because of which client submitted it.

**To Reproduce**

1. Build and start a scheduler and executor from `main`.
2. Run any TPC-H query with the Rust client, and again with `pip install ballista` (54.0.0 or 54.1.0).
3. With `RUST_LOG=ballista_scheduler::state::task_manager=debug` on the scheduler, compare the "query planner" lines.

**Expected behavior**

Python users should get the same planner as everyone else, and if a client does pin a non-default value it should be because the user asked for it.

Two things worth separating here:

1. Any config key whose default changes on the Rust side is invisible to a Python client pinned to an older release, and the mismatch is silent. It would be better if a client only sent keys the user actually set, so the scheduler's own defaults apply to everything else.
2. The version pin itself. `python/Cargo.toml` has the path dependencies commented out and depends on crates.io, so the Python bindings on `main` are not built against the code on `main`. That is fine for release builds but means the bindings lag every default and behaviour change until the next release.

   Worth noting that this pin is not something we can simply drop. `pyballista` re-exports datafusion-python types (`datafusion_python::dataframe::PyDataFrame` and friends in `python/src/lib.rs`), so `datafusion-python`, `datafusion`, and `ballista-core` all have to link the same `datafusion`. The workspace is currently on the `55.0.0-rc3` git tag while crates.io `datafusion-python` is still at `54.0.0`, so the bindings cannot move until there is a matching datafusion-python release. In practice the Python client will normally lag the cluster by a DataFusion release cycle, which makes fix (1) above the one that actually helps.

**Additional context**

Found while investigating #2367. Related: #2315, and the CI gap in #2372.

AnuragRaut08:

Closes #2371.

# Rationale for this change

`BallistaConfig::entries()` currently emits every registered configuration key with its default value, even when the user has not explicitly configured that key.

These defaults come from the client's version of the configuration registry. When a client and scheduler run different Ballista versions, the client can therefore send stale defaults that override the scheduler's current defaults.

For example, a Python client built against an older Ballista version can send:

```text
ballista.planner.adaptive.enabled = false
```

during session creation even though the user never configured this setting. A newer scheduler whose default is `true` receives this as an explicitly provided value and applies `false`, silently changing the planner behaviour.

This means that a configuration default changing between releases can be unintentionally overridden by an older client, without any explicit user configuration.

### `ballista/core/src/config.rs`

`BallistaConfig::entries()` now only emits values that are present in `self.settings`, meaning values that were explicitly configured by the user.

Previously, the method used the registered default value when a key was absent from `self.settings`. It now returns `None` for unset keys.

The receiving side already skips `None`-valued configuration entries, so unset settings are left to the scheduler/executor to resolve using their own current defaults.

This also preserves the existing behaviour for local configuration reads. Methods such as `get_bool_setting` and `get_usize_setting` continue to fall back to `CONFIG_ENTRIES` when a setting has not been explicitly configured.

Two Rust unit tests were added:

* `default_ballista_config_emits_no_values_over_wire` — verifies that a default `BallistaConfig` does not emit concrete values for unset configuration keys.
* `explicitly_set_key_is_present_in_entries` — verifies that explicitly configured values are still included and forwarded correctly.

### `python/python/tests/test_context.py`

A Python integration regression test was added:

* `test_unset_ballista_config_does_not_override_scheduler_defaults` — starts a test cluster with no client-side `cluster_config`, queries `information_schema.df_settings` for `ballista.planner.adaptive.enabled`, and verifies that the scheduler's own default (`true`) is preserved.

This covers the cross-version behaviour described in #2371.

Yes. This is a behavioural fix to configuration propagation.

When a client has not explicitly configured a Ballista setting, the setting is no longer sent with the client's potentially stale default. The scheduler/executor can therefore apply its own current default.

In particular, this prevents an older Python client from unintentionally overriding a newer scheduler's `ballista.planner.adaptive.enabled` default.

Explicitly configured values continue to be forwarded unchanged, so user-specified configuration retains the same behaviour.

### DIFF
--- a/ballista/core/src/config.rs
+++ b/ballista/core/src/config.rs
@@ -889,15 +889,15 @@ impl datafusion::config::ExtensionOptions for BallistaConfig {
     }
 
     fn entries(&self) -> Vec<datafusion::config::ConfigEntry> {
+        // Emit only keys the user explicitly set; absent keys produce value: None.
+        // The receiving side (scheduler/executor) skips None entries and applies
+        // its own current defaults, so a client built against an older release
+        // cannot silently override newer scheduler defaults with stale values.
         Self::valid_entries()
             .iter()
             .map(|(key, value)| datafusion::config::ConfigEntry {
                 key: key.clone(),
-                value: self
-                    .settings
-                    .get(key)
-                    .cloned()
-                    .or(value.default_value.clone()),
+                value: self.settings.get(key).cloned(),
                 description: &value.description,
             })
             .collect()
@@ -1055,5 +1055,41 @@ mod tests {
             .get(BALLISTA_SHUFFLE_SORT_BASED_BATCH_SIZE)
             .expect("entry is registered");
         assert_eq!(batch_size.doc_default(), Some("8192"));
+    }
+
+    // A default BallistaConfig must not emit any key with a concrete value.
+    // Stale client defaults must not override the scheduler's current defaults.
+    #[test]
+    fn default_ballista_config_emits_no_values_over_wire() {
+        let config = BallistaConfig::default();
+        let entries = datafusion::config::ExtensionOptions::entries(&config);
+        for entry in &entries {
+            assert!(
+                entry.value.is_none(),
+                "BallistaConfig::default() should not emit a value for `{}` \
+                 (got {:?}); only explicitly-set keys should be sent to the scheduler",
+                entry.key,
+                entry.value,
+            );
+        }
+    }
+
+    // An explicitly configured key must still be forwarded to the scheduler.
+    #[test]
+    fn explicitly_set_key_is_present_in_entries() {
+        use datafusion::config::ExtensionOptions;
+        let mut config = BallistaConfig::default();
+        config.set("planner.adaptive.enabled", "false").unwrap();
+
+        let entries = datafusion::config::ExtensionOptions::entries(&config);
+        let adaptive = entries
+            .iter()
+            .find(|e| e.key == BALLISTA_ADAPTIVE_PLANNER_ENABLED)
+            .expect("adaptive planner key must be present after explicit set");
+        assert_eq!(
+            adaptive.value.as_deref(),
+            Some("false"),
+            "explicitly-set value must be forwarded to the scheduler"
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


milenkovicm (commented):
As we have two session contexts one in the client one in the server we took the client session context as authoritative, the config values client sees have precedence over scheduler values. 

proposed change would break this. From your example, client might see AQE which has default value `false` but scheduler would behave with `true`. 

there are ongoing discussions about client server version compatibility, until we get that sorted I propose to delay this PR merge

AnuragRaut08:
Thanks for the clarification. That makes sense.

I was treating the scheduler's defaults as authoritative for unset client values, but I see that this conflicts with the current model where the client session context takes precedence. In a mixed-version setup, my proposed change could therefore make the client and scheduler session contexts disagree.

I'll leave the PR unchanged for now and defer the implementation until the client/server version compatibility discussion determines the intended configuration semantics.

PR #2397, opened 2026-08-30
feat(scheduler): document REST API with OpenAPI and serve /api/openapi.json

### LINKED ISSUE #2351: Use OpenAPI to document public REST interfaces
> For REST, could we use:
> 
> 1. OpenAPI to document our format to clients?
> 2. [utoipa](https://github.com/juhaku/utoipa) to generate OpenAPI spec from our Rust types we serde_json
> 3. [oasdiff](https://github.com/Tufin/oasdiff) in CI to prevent breaking changes
> 
> This would enable strict enforcement and formally describe our API 

 _Originally posted by @avantgardnerio in [#2261](https://github.com/apache/datafusion-ballista/issues/2261#issuecomment-5232282999)_

Initially we can start from scheduler API

DebadityaHait:

Closes #2351

# Rationale for this change

Ballista's public REST interfaces currently lack a machine-readable OpenAPI specification. As proposed in #2351, this PR addresses Phase 1 by using `utoipa` to generate OpenAPI documentation directly from our Rust DTO types and serving the spec at `GET /api/openapi.json`.

- Added `utoipa` (5.5) as a workspace dependency with `macros` enabled.
- Added an optional `utoipa` feature to `ballista-api-types` and derived `utoipa::ToSchema` on shared DTO types (`JobResponse`, `TaskStatus`, `TaskSummary`, `Percentiles`, `QueryStageSummary`, `QueryStagesResponse`, `PlanFormat`).
- Added `#[utoipa::path(...)]` annotations to all public scheduler endpoints in `ballista-scheduler` (`/api/state`, `/api/version`, `/api/executors`, `/api/executor/{executor_id}`, `/api/jobs`, `/api/job/{job_id}`, `/api/job/{job_id}/config`, `/api/job/{job_id}/stages`, `/api/job/{job_id}/dot`, `/api/job/{job_id}/stage/{stage_id}/dot`, `/api/metrics`, and `/api/job/{job_id}/dot_svg` when `graphviz-support` is enabled).
- Added `ballista_scheduler::api::openapi` defining the `ApiDoc` structure and `GET /api/openapi.json` route handler.
- Added comprehensive unit tests in `ballista_scheduler` and `ballista_api_types` asserting schema registration, path coverage, serialization, and endpoint response.

The scheduler REST API now exposes `GET /api/openapi.json` returning the OpenAPI v3 JSON specification. No existing endpoints or serialization schemas were modified.

### DIFF
--- a/ballista/api-types/Cargo.toml
+++ b/ballista/api-types/Cargo.toml
@@ -35,4 +35,5 @@ serde = { workspace = true, features = ["derive"] }
 utoipa = { workspace = true, optional = true }
 
 [dev-dependencies]
+ballista-api-types = { path = ".", features = ["utoipa"] }
 serde_json = { workspace = true }
--- a/ballista/api-types/Cargo.toml
+++ b/ballista/api-types/Cargo.toml
@@ -33,3 +33,6 @@ utoipa = ["dep:utoipa"]
 [dependencies]
 serde = { workspace = true, features = ["derive"] }
 utoipa = { workspace = true, optional = true }
+
+[dev-dependencies]
+serde_json = { workspace = true }
--- a/ballista/api-types/Cargo.toml
+++ b/ballista/api-types/Cargo.toml
@@ -26,5 +26,10 @@ authors = { workspace = true }
 edition = { workspace = true }
 rust-version = { workspace = true }
 
+[features]
+default = []
+utoipa = ["dep:utoipa"]
+
 [dependencies]
 serde = { workspace = true, features = ["derive"] }
+utoipa = { workspace = true, optional = true }
--- a/ballista/api-types/src/dto.rs
+++ b/ballista/api-types/src/dto.rs
@@ -187,7 +187,26 @@ pub type JobConfig = BTreeMap<String, String>;
 #[cfg(all(test, feature = "utoipa"))]
 mod tests {
     use super::*;
+    use std::collections::BTreeSet;
     use utoipa::ToSchema;
+
+    fn get_schema_property_names<T: ToSchema>() -> BTreeSet<String> {
+        let schema_ref = T::schema();
+        match schema_ref {
+            utoipa::openapi::RefOr::T(utoipa::openapi::Schema::Object(obj)) => {
+                obj.properties.keys().cloned().collect()
+            }
+            _ => panic!("Expected Object schema"),
+        }
+    }
+
+    fn get_json_property_names<T: Serialize>(value: &T) -> BTreeSet<String> {
+        let json = serde_json::to_value(value).expect("serialize to value");
+        match json {
+            serde_json::Value::Object(map) => map.keys().cloned().collect(),
+            _ => panic!("Expected JSON Object"),
+        }
+    }
 
     #[test]
     fn test_dto_schemas() {
@@ -198,5 +217,41 @@ mod tests {
         assert_eq!(QueryStageSummary::name(), "QueryStageSummary");
         assert_eq!(QueryStagesResponse::name(), "QueryStagesResponse");
         assert_eq!(PlanFormat::name(), "PlanFormat");
+
+        let task_summary = TaskSummary {
+            id: 1,
+            status: TaskStatus::Running,
+            partition_id: vec![0],
+            scheduled_time: 100,
+            launch_time: 100,
+            start_exec_time: 100,
+            end_exec_time: 200,
+            exec_duration: 100,
+            finish_time: 200,
+            input_rows: 10,
+            output_rows: 10,
+        };
+        assert_eq!(
+            get_schema_property_names::<TaskSummary>(),
+            get_json_property_names(&task_summary)
+        );
+
+        let percentiles = Percentiles {
+            min: 1,
+            p25: 2,
+            median: 3,
+            p75: 4,
+            max: 5,
+        };
+        assert_eq!(
+            get_schema_property_names::<Percentiles>(),
+            get_json_property_names(&percentiles)
+        );
+
+        let query_stages = QueryStagesResponse { stages: vec![] };
+        assert_eq!(
+            get_schema_property_names::<QueryStagesResponse>(),
+            get_json_property_names(&query_stages)
+        );
     }
 }
--- a/ballista/api-types/src/dto.rs
+++ b/ballista/api-types/src/dto.rs
@@ -23,6 +23,7 @@ use std::collections::BTreeMap;
 
 /// Summary of one job, served by `GET /api/jobs` and `GET /api/job/{job_id}`.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub struct JobResponse {
     /// A `String` rather than `ballista_core::JobId` so this crate stays a
     /// serde-only leaf. `JobId` is `#[serde(transparent)]` over `String`, so
@@ -57,6 +58,7 @@ pub struct JobResponse {
 
 /// Terminal or in-flight state of a single task.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub enum TaskStatus {
     /// Task is currently executing.
     Running,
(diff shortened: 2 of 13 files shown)


milenkovicm:
would you mind having a look @avantgardnerio ?

ballista/scheduler/src/api/handlers.rs
```diff
@@ -42,40 +42,62 @@ use http::{HeaderMap, StatusCode, header::CONTENT_TYPE};
...
+#[schema(as = ExecutorSpecification)]
+pub struct ExecutorSpecificationSchema {
+    /// Virtual cores assigned to this executor at runtime
+    pub vcores: u32,
+}
+
+/// Operating system level specification of an executor.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, utoipa::ToSchema)]
+#[schema(as = ExecutorOperatingSystemSpecification)]
+pub struct ExecutorOperatingSystemSpecificationSchema {
```
andygrove: This mirror struct declares three fields, but the real `ExecutorOperatingSystemSpecification` in `ballista/core/src/serde/scheduler/mod.rs` has nine. It also carries `os_ver_long`, `physical_cores`, `num_disks`, `total_disk_space`, `total_available_disk_space` and `open_files_limit`, and all of them get serialized.

Since `#[schema(value_type = ...)]` swaps the schema out wholesale, the published spec will under-describe what `/api/executors` and `/api/executor/{executor_id}` actually return, right from the first release. `ExecutorSpecificationSchema` above happens to match today, but it is the same hand-copied pattern and will drift the same way as soon as someone edits the core type.

Would it work to derive `ToSchema` on the real types in `ballista-core` behind an optional feature, the same way you have done it for `ballista-api-types`? That way there is only one definition to keep right.

If you would rather keep this PR small, filling in the six missing fields would fix the immediate problem. In that case it might be worth adding a test that asserts the schema's property set matches the JSON keys of `ExecutorOperatingSystemSpecification::default()`, so the next drift shows up as a failing test rather than a quietly wrong spec.

ballista/scheduler/src/api/openapi.rs
```diff
@@ -0,0 +1,256 @@
...
+)]
+pub struct ApiDoc;
+
+#[cfg(feature = "graphviz-support")]
+#[derive(OpenApi)]
+#[openapi(paths(crate::api::handlers::get_job_svg_graph))]
+struct GraphvizApiDoc;
+
+/// Generate the OpenAPI specification for the scheduler REST API.
+pub fn openapi_spec() -> utoipa::openapi::OpenApi {
```
andygrove: This rebuilds the whole document on every request to `/api/openapi.json`. The spec is fixed at compile time, so wrapping it in a `std::sync::LazyLock` would turn the handler into a cheap clone of a cached value.

ballista/scheduler/src/api/openapi.rs
```diff
@@ -0,0 +1,256 @@
...
+        crate::api::handlers::get_executor_info,
+        crate::api::handlers::get_jobs,
+        crate::api::handlers::get_job,
+        crate::api::handlers::cancel_job,
+        crate::api::handlers::get_job_config,
+        crate::api::handlers::get_query_stages,
+        crate::api::handlers::get_job_dot_graph,
+        crate::api::handlers::get_query_stage_dot_graph,
+        crate::api::handlers::get_scheduler_metrics,
+        crate::api::openapi::get_openapi_spec,
```
andygrove, line 64: `/healthz` and `/readyz` are missing here. They are mounted by `health_routes` unconditionally, so they are live even when `rest-api` is off, and they are exactly the two endpoints a platform owner wires into Kubernetes probes.

Anyone generating a client or a monitoring config from `openapi.json` will not see them. Happy for this to be a follow-up if you would rather keep Phase 1 focused.

ballista/api-types/src/dto.rs
```diff
@@ -172,3 +179,20 @@ pub enum PlanFormat {
...
 /// identically whether it is served live or replayed from a stored log.
 pub type JobConfig = BTreeMap<String, String>;
+
+#[cfg(all(test, feature = "utoipa"))]
+mod tests {
+    use super::*;
+    use utoipa::ToSchema;
+
+    #[test]
+    fn test_dto_schemas() {
```
andygrove, line 208: This one is mostly asserting that the derive macro does what it says, so I am not sure it earns its keep. A check that the generated schema's properties line up with the actual serde output would catch a lot more, things like a `skip_serializing_if` that the schema does not know about.

Worth knowing too that this only runs when feature unification turns the `utoipa` feature on from the scheduler side. `cargo test -p ballista-api-types` on its own will skip it silently.

andygrove (commented):
Thanks for picking this up. A machine-readable spec for the REST API is a genuinely useful thing to have, and the `cfg_attr` gating on `ballista-api-types` is the right call so downstream users of that crate see no change by default.

I left a few inline comments. The one about `ExecutorOperatingSystemSpecificationSchema` is the one I would like sorted before this lands.

One thing that does not fit on a line in the diff: `docs/source/user-guide/scheduler.md` has the table of REST endpoints around line 78. Could you add `/api/openapi.json` to it?

DebadityaHait:
Thank you for the review and suggestions @andygrove! I've addressed all feedback in the latest commit:

1. **Direct `ToSchema` derivation in `ballista-core`**:
   - Added optional `utoipa` feature to `ballista-core` and derived `ToSchema` directly on `ExecutorSpecification` and `ExecutorOperatingSystemSpecification` (including all 9 fields).
   - Removed the proxy mirror structs in `handlers.rs`.
   - Added property alignment tests comparing `utoipa::ToSchema` property sets directly against `serde_json` serialization output to guard against future schema drift.

2. **Cached OpenAPI Document**:
   - Wrapped the spec assembly in `std::sync::LazyLock` so `GET /api/openapi.json` returns a cached value.

3. **Kubernetes Health Probes (`/healthz` and `/readyz`)**:
   - Added `#[utoipa::path]` annotations to `healthz` and `readyz` in `api::health` and included them in the OpenAPI specification under the `health` tag.

4. **Schema Property Alignment Tests**:
   - Enhanced tests in both `ballista-api-types` and `ballista-scheduler` to verify that generated schema properties match the serialized JSON keys of DTO and response instances.

5. **Documentation**:
   - Added `GET /api/openapi.json` to the REST endpoints table in `docs/source/user-guide/scheduler.md`.

ballista/api-types/Cargo.toml
```diff
@@ -26,5 +26,13 @@ authors = ["Apache DataFusion <dev@datafusion.apache.org>"]
...
+[features]
+default = []
+utoipa = ["dep:utoipa"]
+
 [dependencies]
 serde = { workspace = true, features = ["derive"] }
+utoipa = { workspace = true, optional = true }
+
+[dev-dependencies]
+serde_json = { workspace = true }
```
avantgardnerio, line 39: Following up on @andygrove's note that this only runs when feature unification turns `utoipa` on from the scheduler side: `cargo test -p ballista-api-types` still reports `0 tests`. A self dev-dependency enables the feature for the test build only:

```suggestion
[dev-dependencies]
ballista-api-types = { path = ".", features = ["utoipa"] }
serde_json = { workspace = true }
```

The `#[cfg(all(test, feature = "utoipa"))]` gate in `dto.rs` should stay as it is, since the derives are gated the same way.

I ran this locally: `cargo test -p ballista-api-types` goes from `0 tests` to running `dto::tests::test_dto_schemas`, and downstream users are unaffected. Dev-dependencies are not part of `cargo build`, and `cargo package` drops the path-only self-dependency from the published manifest.

avantgardnerio (approved):
I see you addressed @andygrove 's feedback, so ✔️ with one nice-to-have in the comments.

ballista/api-types/Cargo.toml
DebadityaHait: Applied! Added the self dev-dependency in `ballista/api-types/Cargo.toml` and verified `cargo test -p ballista-api-types` directly runs the DTO schema tests now. Thank you @avantgardnerio!

milenkovicm:
Can you please rebase pr @DebadityaHait

DebadityaHait:
Rebased onto latest `main`, updated `Cargo.lock`, and verified that all tests, `--locked` cargo check, and machete checks pass. Thank you @milenkovicm!

milenkovicm:
thanks @DebadityaHait, @avantgardnerio & @andygrove will merge this on green ci

PR #2405, opened 2026-09-01
feat(history): mark running event logs with a .running suffix until job completion

milenkovicm:

Relates to #2267
Closes #.

# Rationale for this change

Event logs written by `EventLogWriter` (`ballista/history/src/writer.rs`) now go to `<job_id>.eventlog.running` while a job runs, and are flushed, closed, and renamed to `<job_id>.eventlog` when the job reaches a terminal state (success, failure, or cancellation) via `finish_job`; this lets a directory scan tell running/crashed jobs apart from finished ones by filename alone, since the history server's existing `.eventlog` extension filter (`ballista/scheduler/src/history/mod.rs`) already excludes `.running` files without opening them, avoiding the previous per-tick re-parsing of in-flight logs.

Doc comments in `ballista/history/src/lib.rs`,
`writer.rs`, `history/mod.rs`, and `docs/source/user-guide/history-server.md` were updated to describe the new two-name lifecycle, and tests were adjusted accordingly: writer.rs's tests were updated for the new path and a new no-op-finish test was added, history/mod.rs's `refresh_indexes_a_log_that_gains_its_terminal_record` was rewritten to exercise a real write-then-rename, and two tests in `scheduler_server/event_log.rs` that previously only called flush_job (and would otherwise never see the renamed file) now also call finish_job.

This would enable custom event log handlers (history server) to make difference between finished and running jobs. we can use `inotifiy` listeners to add finished jobs to history server state, or implement external object store synchroniser without reading whole directory every time synchronisation is needed

### DIFF
--- a/ballista/history/src/lib.rs
+++ b/ballista/history/src/lib.rs
@@ -20,9 +20,13 @@
 //! Durable event logs for completed Ballista jobs.
 //!
 //! The scheduler appends one JSONL record per event to
-//! `<event_log_dir>/<job_id>.eventlog` while a job runs. The history server
-//! reads those files back and serves the same `/api/*` responses the live
-//! scheduler does, long after the scheduler has forgotten the job.
+//! `<event_log_dir>/<job_id>.eventlog.running` while a job runs, then renames
+//! the file to `<event_log_dir>/<job_id>.eventlog` once the job reaches a
+//! terminal state. The history server only ever looks at `.eventlog` files,
+//! so the `.running` suffix alone marks a job as not yet ready to read,
+//! without opening it. The history server reads finished files back and
+//! serves the same `/api/*` responses the live scheduler does, long after the
+//! scheduler has forgotten the job.
 //!
 //! # Write once, replay verbatim
 //!
@@ -52,6 +56,12 @@
 //! that losing a progress record is better than stalling scheduling. The
 //! terminal `JobEnd` is the exception: it waits for queue capacity, because a
 //! job missing its `JobEnd` is invisible to the history server.
+//!
+//! Completion is signaled twice, deliberately redundantly: by the file's
+//! name (`.eventlog` vs `.eventlog.running`), which a directory scan can act
+//! on without opening the file, and by the `JobEnd` record's presence in the
+//! content, which guards against a file that was renamed but is otherwise
+//! corrupt.
 
 pub mod event;
 pub mod reader;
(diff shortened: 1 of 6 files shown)


andygrove (approved):
Nice, this is a good change. Answering "is this log finished?" by reading the file to the end never scaled, and turning it into an extension check on the dirent means the rescan stops re-parsing every in-flight log on every tick. It also unblocks the inotify and object store sync ideas in #2267, since a rename is something a listener can act on and "a JobEnd line appeared somewhere inside this file" is not.

Timing is good too. The history crate is not in 54.1.0 yet, so this is an on disk naming change to something unreleased and there is no compatibility path to owe anyone. That stops being true after the next release.

I like that completion is still signaled twice. Keeping the JobEnd check in the content means a file that got renamed but is otherwise corrupt is still skipped rather than served, so we are not trading the old safety property away for the cheap one.

One question, and I do not think it blocks merging. After `finish_job` removes the handle, a straggler event for that job would hit `open_for` and recreate an empty `<job_id>.eventlog.running` that nothing will ever rename. The served `.eventlog` is untouched, which is better than the old behaviour of appending records after `JobEnd` into the finished file, so the worst case is a stray file that reads as "this job crashed" for a job that actually succeeded. Do you think late appends after a terminal event are possible in practice, say a task status update arriving after `JobCancel` or after a job has already failed? If they are, it might be worth having `open_for` refuse to reopen a job that has already finished, so the crash signal stays trustworthy.

milenkovicm:
> One question, and I do not think it blocks merging. After `finish_job` removes the handle, a straggler event for that job would hit `open_for` and recreate an empty `<job_id>.eventlog.running` that nothing will ever rename. The served `.eventlog` is untouched, which is better than the old behaviour of appending records after `JobEnd` into the finished file, so the worst case is a stray file that reads as "this job crashed" for a job that actually succeeded. Do you think late appends after a terminal event are possible in practice, say a task status update arriving after `JobCancel` or after a job has already failed? If they are, it might be worth having `open_for` refuse to reopen a job that has already finished, so the crash signal stays trustworthy.
> 

Will have to investigate. Can't come up with corner case for suggested "check before open" but let me sleep on it.

I just wanted to land protocol change before we release. 

Just a heads up, will change history server as a follow up 

- Read directory on startup once
- Use inotify to listens for new event logs and populate state 

Will try to make both functions pluggable,

milenkovicm:
added check for late events, we have one extra syscall, should not be the problem, lets see.
 wdyt @andygrove

PR #2406, opened 2026-09-01
feat(core): order shuffle reads round-robin across producers, staggered per consumer

### LINKED ISSUE #2403: Deterministically stagger shuffle read fetch order instead of randomizing it

ShuffleReaderExec::execute orders its fetch list in two steps with separate goals:

// Sort partitions for evenly send fetching partition requests to avoid hot executors within one task
... .sorted_by(|(p1_idx, _), (p2_idx, _)| Ord::cmp(p1_idx, p2_idx)) ...
// Shuffle partitions for evenly send fetching partition requests to avoid hot executors within multiple tasks
partition_locations.shuffle(&mut rng());

The sort builds a round-robin so one task spreads its in-flight window across producers. The shuffle then discards it: Fisher-Yates yields a uniform permutation regardless of input order, so nothing of the sort survives. The intra-task goal is not met today.

This was raised in review on #256:

▎ thinkharderdev: Why do we sort by index before shuffling here?
▎
▎ yahoNanJing: Sorting first may be helpful for reducing the bias of random chosen. Maybe it's not necessary.

Sorting cannot reduce bias, since a uniform shuffle erases input order whatever it was; however, it produced exactly the spread its own comment describes, which the shuffle is discarding.

The cross-task goal also falls short. Each task draws its own random order, so peers can still open on the same producer. 

Nothing better is available today. restrict_plan_to_partitions re-indexes each task's slice from zero, so every reader is partition 0 locally and carries nothing that says which peer it is.

Stamp each local output partition with its global index among the stage's peers, carry it to the executor, and use it to rotate a deterministic round-robin over producers.

Round-robin gives the intra-task spread the sort was meant to provide. The rotation staggers peers exactly rather than probabilistically.

- Drop the sort, keep the RNG. Matches what the code effectively does today and answers the #256 question, but leaves peers colliding.
- Seed the RNG deterministically. Nothing stable to seed with. After restriction every reader is partition 0, so peers seed identically.
- Assign order in the scheduler at dispatch. Needs runtime state for a result a peer index gives locally.

NoahKusaba:

Closes #2403 

 # Rationale for this change

This PR should optimally distribute consumers across producers deterministically. 
Previously we relied on a RNG distribution, which could land several consumers on the same producers. `req_sem` caps how many blocks are in flight but not how many of them share an address, so nothing today stops all 64 targeting one executor. Round-robin gives the exact spread the original sort was reaching for, and makes fetch order a pure function of the block list and the consumer's index. Remaining variance comes from block skew and executor load rather than from the draw, so it is attributable. 

The order is also assertable, which several tests here rely on and an RNG made impossible.

Primary benefits:
- **More predictable query times.** 

- **Less chance of a hot producer:** This in theory should improve the stability of Ballista at higher cluster sizes.

- **Potentially faster (unproven see below).** If every consumers is optimally distributed, this may lead to performance gains due to lack of contention.

Side benefits:
- **We can drop the itertools dependency**
- **A bound on open files and buffered data.** Every local result waiting to be
  consumed holds an open file descriptor and a 256 KB buffer. The old shared
  channel let a stalled consumer pin `max_requests` of them; the split lane caps
  it at two. This predates the PR and was found while investigating the stall.
- **Less per-block overhead.** The per-address semaphore map is now touched only
  by the driver instead of contended across M tasks, and the executor id is no
  longer cloned twice per block. 

# Performance improvements + AI Benchmarking:

**Fetch ordering is cheaper than what it replaces.** The old path sorted all M blocks and then ran an O(M) shuffle over them. `interleave_by_executor` sorts only the N producers. Per `execute()` call, on the synchronous setup path before any fetch is issued:

| M | N | old (sort + shuffle) | new (round-robin) |
|---|---|---|---|
| 1000 | 20 | 837 us | 285 us |
| 5000 | 50 | 2.59 ms | 1.45 ms |
| 20000 | 200 | 16.96 ms | 6.11 ms |

## Help wanted 
The ordering benefit itself is unmeasured. I don't have a multi-executor cluster to test this on.

The numbers I would find most useful, against `main` on a shuffle-heavy query such as TPC-H q4 or q18:

- `ShuffleReaderExec`'s `permit_wait_time` and `fetch_time` metrics, which is
  where converging on one producer shows up
- reduce-stage wall time, especially its spread across tasks: the change is
  aimed at tail latency more than the mean
- whether more executors widens or narrows the gap, since N is what the
  round-robin has to spread across

A confirmation that nothing regresses would be just as valuable as a win.

**Ordering.** `interleave_by_executor` replaces the sort-then-shuffle. Blocks
are round-robined across the executors that produced them, starting at the
producer the consumer's own index selects. A skewed producer keeps its turn
each round, then drains alone.

**Peer identity.** New `ShuffleReaderExecNode.output_partition_indices`
(proto field 8), stamped by `restrict_plan_to_partitions` alongside the slice
it selects. Absent on the wire decodes to the identity mapping, so a plan from
an older encoder is unaffected; a stamp that is not parallel to `partition` is
a decode error rather than a silent fallback.

**One fetch driver instead of a task per block.** Ordering only matters if it
survives to the wire. With a task per block, M tasks race to acquire `req_sem`
and the admitted subset is up to the runtime, so the round-robin is computed
and then discarded. A single driver over `buffer_unordered` initiates in list
order; completion stays unordered.

**Decode off the driver.** One driver polling every block would decode them all
on a single thread, since IPC decode and decompression happen in `poll_next` and
hold whatever thread polls them. So the driver issues the request and hands the
drain to its own task (`collect_off_task`). Measured on 24 blocks of
ZSTD-compressed Arrow IPC, 337 KiB per block:

| where the block is drained | wall time (4 runs) |
|---|---|
| inside the driver | 58.8, 61.0, 61.5, 63.2 ms |
| on its own task (this PR) | 11.5, 12.5, 14.4, 19.1 ms |

`main` spawns a task per block, so its decode is already parallel. This keeps
that property rather than improving on it. Fetch order is unaffected: the
request is still issued from the driver, in list order.

**Separate local and remote lanes.** A queued local result pins an open fd and
a 256 KB buffer, so locals get a depth-1 lane; remotes mirror `req_sem`. The
two are merged with `select`, so locals arrive interleaved rather than
serialized behind remotes.

`permit_wait_time` now times from when a block becomes eligible rather than
from when the buffer constructs its future, keeping the metric comparable with
the previous task-per-block version. 

# Tests

Ordering: `interleave_visits_every_executor_once_per_round`,
`interleave_drains_a_skewed_producer_last`, `interleave_preserves_the_fetch_set`,
`producer_order_is_independent_of_input_order`, `interleave_handles_no_locations`,
`stamped_peers_round_robin_across_producers`,
`rotation_comes_from_the_stamp_not_the_local_index`.

Reaching the wire: `remote_fetches_reach_the_wire_in_list_order` records fetch
order through a stub client pool, on both sides of the governor window.

Concurrency: `decode_does_not_run_on_the_callers_task` fails unless the drains
run on threads of their own; `local_reads_do_not_run_ahead_of_the_consumer`
counts open descriptors under the work dir.

Stamping: `restriction_stamps_the_global_output_partition_indices`,
`stamp_stays_parallel_when_an_index_is_out_of_range`,
`broadcast_reader_under_a_collected_join_is_still_stamped`, plus codec
roundtrips for the stamped, absent, mismatched, and broadcast cases.

Every test above was checked by mutation: reverting the behavior it covers
fails that test and no other.

# How Spark handles the same problem

**Spark randomizes too**, exactly as Ballista did before this PR:
`fetchRequests ++= Utils.randomize(remoteRequests)`.

**Its producer spread comes from request sizing, not ordering.** Spark's fetch
unit is a batch of blocks from one address, capped at `maxBytesInFlight / 5`:

> the reason to keep them smaller than `maxBytesInFlight` is to allow multiple,
> parallel fetches from up to 5 nodes, rather than blocking on reading output
> from one node.

The byte budget itself forces several distinct sources in flight, whatever the
queue order. Randomization is only tie-breaking.

**This PR does not copy that** because it needs a batched multi-block request.
Ballista fetches one partition per Flight call, and `req_sem` bounds the fetch
*count*, which says nothing about address diversity: all 64 can be one executor.
Ordering is the cheaper lever on the existing design, and the two compose.

**Spark has no per-consumer stagger.** Nothing keys fetch order to the reducer's
own index. `output_partition_indices` has no Spark analog, which cuts both ways:
stronger than the state of the art, and unproven.

Source: [ShuffleBlockFetcherIterator.scala](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/storage/ShuffleBlockFetcherIterator.scala)

No

### DIFF
--- a/ballista/core/src/execution_plans/shuffle_reader.rs
+++ b/ballista/core/src/execution_plans/shuffle_reader.rs
@@ -49,11 +49,9 @@ use datafusion::physical_plan::{
     PlanProperties, RecordBatchStream, SendableRecordBatchStream, Statistics,
 };
 use datafusion::prelude::SessionConfig;
+use futures::stream::{BoxStream, select};
 use futures::{Stream, StreamExt, TryStreamExt, ready};
-use itertools::Itertools;
 use log::{debug, error, trace};
-use rand::prelude::SliceRandom;
-use rand::rng;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fs::File;
@@ -64,7 +62,7 @@ use std::pin::Pin;
 use std::result;
 use std::sync::Arc;
 use std::task::{Context, Poll};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore, mpsc};
 use tokio_stream::wrappers::ReceiverStream;
 
@@ -125,6 +123,10 @@ pub struct ShuffleReaderExec {
     /// is responsible for pre-concatenating the M-shape upstream
     /// `Vec<Vec<PartitionLocation>>` into K-shape before invoking `try_new_coalesced`.
     pub coalesce: Option<CoalescePlan>,
+    /// Index of each local output partition in the unsliced stage, parallel to
+    /// `partition`. Restriction re-indexes `partition` from zero, so this is all
+    /// that tells peers apart. Picks which producer `execute` fetches first.
+    pub output_partition_indices: Vec<usize>,
     /// Execution metrics
     metrics: ExecutionPlanMetricsSet,
     properties: Arc<PlanProperties>,
@@ -142,6 +144,7 @@ impl ShuffleReaderExec {
         partitioning: Partitioning,
     ) -> Result<Self> {
         let upstream_partition_count = partition.len();
+        let output_partition_indices = (0..partition.len()).collect();
         let properties = Arc::new(PlanProperties::new(
             datafusion::physical_expr::EquivalenceProperties::new(schema.clone()),
             partitioning,
@@ -155,6 +158,7 @@ impl ShuffleReaderExec {
             broadcast: false,
             upstream_partition_count,
             coalesce: None,
+            output_partition_indices,
             metrics: ExecutionPlanMetricsSet::new(),
             properties,
             work_dir: None,    // to be updated at the executor side
@@ -185,6 +189,7 @@ impl ShuffleReaderExec {
             broadcast: true,
             upstream_partition_count,
             coalesce: None,
+            output_partition_indices: vec![0],
             metrics: ExecutionPlanMetricsSet::new(),
             properties,
             work_dir: None,    // to be updated at the executor side
@@ -223,6 +228,7 @@ impl ShuffleReaderExec {
             "partitioning.partition_count() must equal coalesce.groups.len() (= K)",
         );
         let upstream_partition_count = coalesce.upstream_partition_count as usize;
+        let output_partition_indices = (0..partition.len()).collect();
         let properties = Arc::new(PlanProperties::new(
             datafusion::physical_expr::EquivalenceProperties::new(schema.clone()),
             partitioning,
@@ -236,6 +242,7 @@ impl ShuffleReaderExec {
             broadcast: false,
             upstream_partition_count,
             coalesce: Some(coalesce),
+            output_partition_indices,
             metrics: ExecutionPlanMetricsSet::new(),
             properties,
             work_dir: None,    // to be updated at the executor side
@@ -252,6 +259,7 @@ impl ShuffleReaderExec {
             broadcast: self.broadcast,
             upstream_partition_count: self.upstream_partition_count,
             coalesce: self.coalesce.clone(),
+            output_partition_indices: self.output_partition_indices.clone(),
             metrics: self.metrics.clone(),
             properties: self.properties.clone(),
             work_dir: Some(work_dir),
@@ -267,11 +275,33 @@ impl ShuffleReaderExec {
             broadcast: self.broadcast,
             upstream_partition_count: self.upstream_partition_count,
             coalesce: self.coalesce.clone(),
+            output_partition_indices: self.output_partition_indices.clone(),
             metrics: self.metrics.clone(),
             properties: self.properties.clone(),
             work_dir: self.work_dir.clone(),
             client_pool: Some(client_pool),
         }
+    }
+
+    /// Stamps each local output partition with its global index. `indices` must
+    /// be parallel to `partition`.
+    pub fn with_output_partition_indices(mut self, indices: Vec<usize>) -> Self {
+        debug_assert_eq!(
+            indices.len(),
+            self.partition.len(),
+            "output_partition_indices must be parallel to partition",
+        );
+        self.output_partition_indices = indices;
+        self
+    }
+
+    /// Producer-ring offset for a local output partition: its stamped global
+    /// index, falling back to the local one if the stamp is somehow short.
+    fn rotation_for(&self, partition: usize) -> u64 {
+        self.output_partition_indices
+            .get(partition)
+            .copied()
+            .unwrap_or(partition) as u64
     }
 }
 
@@ -350,6 +380,7 @@ impl ExecutionPlan for ShuffleReaderExec {
                 broadcast: self.broadcast,
                 upstream_partition_count: self.upstream_partition_count,
                 coalesce: self.coalesce.clone(),
+                output_partition_indices: self.output_partition_indices.clone(),
                 metrics: ExecutionPlanMetricsSet::new(),
                 properties: self.properties.clone(),
                 work_dir: self.work_dir.clone(),
@@ -389,22 +420,9 @@ impl ExecutionPlan for ShuffleReaderExec {
             config.ballista_shuffle_reader_maximum_concurrent_requests(),
             config.ballista_grpc_client_max_message_size()
         );
-        let mut partition_locations = HashMap::new();
-        for p in &self.partition[partition] {
-            partition_locations
-                .entry(p.executor_meta.id.clone())
-                .or_insert_with(Vec::new)
-                .push(p.clone());
-        }
-        // Sort partitions for evenly send fetching partition requests to avoid hot executors within one task
-        let mut partition_locations: Vec<PartitionLocation> = partition_locations
-            .into_values()
-            .flat_map(|ps| ps.into_iter().enumerate())
-            .sorted_by(|(p1_idx, _), (p2_idx, _)| Ord::cmp(p1_idx, p2_idx))
-            .map(|(_, p)| p)
-            .collect();
-        // Shuffle partitions for evenly send fetching partition requests to avoid hot executors within multiple tasks
-        partition_locations.shuffle(&mut rng());
+        let locations = &self.partition[partition];
+        let partition_locations =
+            interleave_by_executor(locations, self.rotation_for(partition));
 
         let work_dir = self
             .work_dir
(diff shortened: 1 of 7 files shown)


ballista/core/src/execution_plans/shuffle_reader.rs
```diff
@@ -125,6 +123,10 @@ pub struct ShuffleReaderExec {
     /// is responsible for pre-concatenating the M-shape upstream
     /// `Vec<Vec<PartitionLocation>>` into K-shape before invoking `try_new_coalesced`.
     pub coalesce: Option<CoalescePlan>,
+    /// Index of each local output partition in the unsliced stage, parallel to
+    /// `partition`. Restriction re-indexes `partition` from zero, so this is all
+    /// that tells peers apart. Picks which producer `execute` fetches first.
+    pub output_partition_indices: Vec<usize>,
```
avantgardnerio: How does this differ from [global_output_partition_ids](https://github.com/apache/datafusion-ballista/blob/7e939fe26c8cfacc41ed16a656704d69c94ea380/ballista/executor/src/execution_engine.rs#L205) which is also in the [proto](https://github.com/apache/datafusion-ballista/blob/e81f2ae5dba8a58e5f54eb0c107863e80c0cbd15/ballista/core/proto/ballista.proto#L634) and makes it to the executor - could that be reused instead?

avantgardnerio: I see now that this means "output" of the writer, which is _input_ of the stage. To keep naming (which is already confusing) consistent, could this be `upstream_partition_indexes`? The term `global_output_partitions` is already in use [here](https://github.com/apache/datafusion-ballista/blob/7e939fe26c8cfacc41ed16a656704d69c94ea380/ballista/core/proto/ballista.proto#L625), and the struct already has `upstream_partition_count` so I think the name change would align it correctly.

avantgardnerio:
**Cluster numbers, as requested in the PR description.**

**Setup.** 4 × `r6i.24xlarge` in EKS, 32 executor pods at 8 vCPU / 64 GiB each (`--vcores=8`, `--memory-pool-size=48103633715`, 1000 GiB gp3 shuffle volume), TPC-H SF1000 Parquet+ZSTD on S3, `target_partitions=256`, `max_partitions_per_task=1`, AQE on — the shape in `docs/source/contributors-guide/benchmarking.md`. Baseline is `b0dfd259`, this PR's merge base; treatment is `7684d651`. Same client image both arms.

**Method.** Counterbalanced ABBA (base → pr → pr → base), because a first sequential base→pr pair showed a gap that partly disappeared once the order was balanced. 5 iterations per query per run, first dropped as cold, so n = 8 warm iterations per arm.

| query | metric | base | pr | Δ | Welch t |
|---|---|---:|---:|---:|---:|
| q4 | wall (s) | 10.32 ± 0.97 | 10.68 ± 0.99 | +3.5% | 0.70 |
| q4 | `fetch_time` (s) | 6,773 ± 578 | 8,172 ± 1,571 | +20.7% | 2.36 |
| q4 | `permit_wait_time` (s) | 43,219 ± 4,523 | 46,082 ± 12,478 | +6.6% | 0.61 |
| q18 | wall (s) | 38.09 ± 0.76 | 38.69 ± 1.09 | +1.6% | 1.19 |
| q18 | `fetch_time` (s) | 59,751 ± 3,207 | 63,844 ± 3,248 | +6.9% | 2.54 |
| q18 | `permit_wait_time` (s) | 527,109 ± 21,846 | 579,635 ± 29,380 | +10.0% | 4.06 |

Metric columns are stage sums over 256 partitions.

**Wall clock is indistinguishable in either direction** (t = 0.70, 1.19). `fetch_time` is higher on both queries at t > 2.3.

**`permit_wait_time` is not strictly comparable across the arms.** This PR starts the timer at a single `queued_at` taken before the loop, where main starts each block's timer when its spawned task first runs, so blocks now accumulate wait from an earlier instant. Some of the +10% is definitional. `fetch_time` and wall clock keep their definitions.

**Tail spread** — per-task `exec_duration` across the 256 tasks of the shuffle-reading stage, which is what the change targets. q4: `max/median` 3.53 → 3.74, `cv` 0.372 → 0.376. q18: `max/median` 1.13 → 1.16, `cv` 0.069 → 0.073. No tightening. q18's baseline is already `cv = 0.069` across 256 tasks, so there is little imbalance available to remove at this shape.

**Scope.** One shape only: 32 producers, 256 consumers. The PR asks whether more executors widens or narrows the gap; that is untested here.

---
🤖 Run and written by [Claude Code](https://claude.com/claude-code) on @avantgardnerio's cluster. Raw logs and per-stage metric dumps available on request.

NoahKusaba:
> **Cluster numbers, as requested in the PR description.**
> 
> **Setup.** 4 × `r6i.24xlarge` in EKS, 32 executor pods at 8 vCPU / 64 GiB each (`--vcores=8`, `--memory-pool-size=48103633715`, 1000 GiB gp3 shuffle volume), TPC-H SF1000 Parquet+ZSTD on S3, `target_partitions=256`, `max_partitions_per_task=1`, AQE on — the shape in `docs/source/contributors-guide/benchmarking.md`. Baseline is `b0dfd259`, this PR's merge base; treatment is `7684d651`. Same client image both arms.
> 
> **Method.** Counterbalanced ABBA (base → pr → pr → base), because a first sequential base→pr pair showed a gap that partly disappeared once the order was balanced. 5 iterations per query per run, first dropped as cold, so n = 8 warm iterations per arm.
> query 	metric 	base 	pr 	Δ 	Welch t
> q4 	wall (s) 	10.32 ± 0.97 	10.68 ± 0.99 	+3.5% 	0.70
> q4 	`fetch_time` (s) 	6,773 ± 578 	8,172 ± 1,571 	+20.7% 	2.36
> q4 	`permit_wait_time` (s) 	43,219 ± 4,523 	46,082 ± 12,478 	+6.6% 	0.61
> q18 	wall (s) 	38.09 ± 0.76 	38.69 ± 1.09 	+1.6% 	1.19
> q18 	`fetch_time` (s) 	59,751 ± 3,207 	63,844 ± 3,248 	+6.9% 	2.54
> q18 	`permit_wait_time` (s) 	527,109 ± 21,846 	579,635 ± 29,380 	+10.0% 	4.06
> 
> Metric columns are stage sums over 256 partitions.
> 
> **Wall clock is indistinguishable in either direction** (t = 0.70, 1.19). `fetch_time` is higher on both queries at t > 2.3.
> 
> **`permit_wait_time` is not strictly comparable across the arms.** This PR starts the timer at a single `queued_at` taken before the loop, where main starts each block's timer when its spawned task first runs, so blocks now accumulate wait from an earlier instant. Some of the +10% is definitional. `fetch_time` and wall clock keep their definitions.
> 
> **Tail spread** — per-task `exec_duration` across the 256 tasks of the shuffle-reading stage, which is what the change targets. q4: `max/median` 3.53 → 3.74, `cv` 0.372 → 0.376. q18: `max/median` 1.13 → 1.16, `cv` 0.069 → 0.073. No tightening. q18's baseline is already `cv = 0.069` across 256 tasks, so there is little imbalance available to remove at this shape.
> 
> **Scope.** One shape only: 32 producers, 256 consumers. The PR asks whether more executors widens or narrows the gap; that is untested here.
> 
> 🤖 Run and written by [Claude Code](https://claude.com/claude-code) on @avantgardnerio's cluster. Raw logs and per-stage metric dumps available on request.

Thank you for your time, I suspected this might not amount to much, but I got a bit caught up imagining what an optimal implementation would be when I saw the sort -> shuffle. 

I'll just make a new PR to delete the useless sort instead and leave it alone.

avantgardnerio:
Thanks for trying it out! I'm sorry this one didn't pan out, but the next one will. Happy to throw some resources at it.

PR #2408, opened 2026-09-01
ci: share one cluster harness across the tpch, tpcds and h2o jobs

NoahKusaba:

Closes #.

 # Rationale for this change

The three benchmark workflows each inlined the same ~60 lines of setup and
cluster lifecycle. This extracts the shared halves into
`.github/actions/setup-benchmark-builder` (deps, toolchain, cache, build) and
`ci/scripts/ballista_cluster.sh` (`start` / `stop`). Workflows drop from ~90
lines of job body to ~40.

`stop` is now its own step under `if: always()`, so teardown and the log dump
also happen on a timeout or cancellation. The old `trap cleanup EXIT` only
fired if the step's own shell exited.

Fixes drift the three copies had accumulated:
- tpch and h2o uploaded logs under `failure()`, discarding them on a hang,
  which is the case they exist for. Now `!success()`, matching tpcds.
- tpch and h2o had no timeouts (6-hour default). All three now 45min job /
  30min step; observed passing jobs are 8-23min.
- tpcds used `--concurrent-tasks`, a deprecated alias that warns. Now `--vcores`.

Also fixes two readiness bugs present in all three copies. A dead process was
waited on for the full 30s instead of failing immediately, and an open port was
treated as proof of health (the executor binds its gRPC port before the Flight
server and exits if that later bind fails).

No new third-party actions,  the composite reuses the rust-cache pin already
used across the repo.

### DIFF
--- a/.github/actions/setup-benchmark-builder/action.yaml
+++ b/.github/actions/setup-benchmark-builder/action.yaml
@@ -1,0 +1,47 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


andygrove (approved):
Thanks @NoahKusaba

NoahKusaba:
The pr history looks weird, as I wanted to trim some comments and re-ran a saved push with --force-with-lease, as I normally like to compact all commits before making a PR. 
That was my mistake. Nothing functionally changed after Andy's approval.

PR #2409, opened 2026-09-01
chore: consolidate shared manifest metadata into the workspace

NoahKusaba:

Closes #.

 # Rationale for this change

Manifest-only cleanup. No resolved dependency or feature set changes.

- `[workspace.package]` gains `version`, `authors`, `license`, `homepage`,
  `repository`; all ten members inherit. `readme` stays per-crate (inherited,
  it resolves against the workspace root, not the crate).
- Internal crates moved to `[workspace.dependencies]`, so a release bump
  touches one file. Five use sites keep the explicit `path` + `version` form
  because Cargo rejects `default-features = false` on an inherited dep.
- Third-party stragglers now use the workspace table: `itertools`, `chrono`,
  `axum`, `serde_json` (was declared 5x across 3 version specs), `tokio` in
  benchmarks, and `serde` / `futures` / `tracing*` in ballista-cli. `reqwest`
  is left alone: the CLI wants default TLS, chaos-testing deliberately drops it.
- benchmarks was the only member hardcoding `edition` and omitting
  `rust-version`; both now inherit. Dropped the executor's empty
  `[dev-dependencies]` and `[build-dependencies]`.
- `python/Cargo.toml` moves to edition 2024, which needs Rust 1.85+, so
  `rust-version` goes to 1.94.0. The old 1.82.0 floor was already unreachable
  via `ballista = "=54.0.0"`. It can't inherit; it's in the workspace's
  `exclude` list.

Verified: `cargo tree --all-features` identical across all 557 packages before
and after; no `Cargo.lock` drift; `cargo check --workspace --all-targets
--locked` and the `--no-default-features` check pass; pyballista compiles under
edition 2024; `taplo format --check` clean.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,14 +32,24 @@ members = [
 resolver = "3"
 
 [workspace.package]
+authors = ["Apache DataFusion <dev@datafusion.apache.org>"]
 edition = "2024"
+homepage = "https://datafusion.apache.org/ballista/"
+license = "Apache-2.0"
+repository = "https://github.com/apache/datafusion-ballista"
 # we should try to follow datafusion version 
 rust-version = "1.94.0"
+version = "54.0.0"
 
 [workspace.dependencies]
 arrow = { version = "59.2.0", features = ["ipc_compression"] }
 arrow-flight = { version = "59.2.0", features = ["flight-sql-experimental"] }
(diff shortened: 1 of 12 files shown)


andygrove (approved):
Thanks @NoahKusaba

milenkovicm (approved):
LGTM, thanks @NoahKusaba

PR #2411, opened 2026-09-02
chore(ci): run cargo machete to catch unused dependencies

### LINKED ISSUE #2410: Run cargo machete in CI to catch unused dependencies

We do not currently check for unused dependencies anywhere in CI. Our Rust lint surface is `cargo fmt`, `cargo clippy`, `cargo tomlfmt`, plus the docs/proto/config-docs checks, and the `Dependencies` workflow only covers the MSRV check and the `python/Cargo.lock` sync check.

As a result, dependencies that are no longer used stay in our manifests. They still get resolved, downloaded, and (in many cases) compiled, which slows down builds, adds noise to `Cargo.lock`, and grows the surface we have to audit for supply chain and license reasons.

Running `cargo machete` locally today reports quite a few:

```
ballista-core -- ./ballista/core/Cargo.toml:
	aws-config
	aws-credential-types
	md-5
	prost-types
ballista-scheduler -- ./ballista/scheduler/Cargo.toml:
	insta
	prost-types
	tonic-prost
ballista-executor -- ./ballista/executor/Cargo.toml:
	arrow
	tracing
pyballista -- ./python/Cargo.toml:
	async-trait
msrvcheck -- ./dev/msrvcheck/Cargo.toml:
	time
ballista-cli -- ./ballista-cli/Cargo.toml:
	critical-section
	tracing-appender
```

Some of these are likely false positives (for example deps that are only referenced through macros or through generated code), but several look like genuinely dead entries.

1. Triage the list above, remove the dependencies that really are unused, and add the rest to `[package.metadata.cargo-machete] ignored = [...]` with a short comment explaining why each one is needed.
2. Add a `ci/scripts/rust_machete.sh` that runs `cargo machete`, wire it into `dev/rust_lint.sh` alongside the other lint steps, and add a job or step in CI so regressions are caught on PRs.

- `cargo udeps`, which is more accurate because it uses real build output, but it requires a nightly toolchain and is considerably slower. `cargo machete` is a source scan, so it is fast enough to sit next to clippy, at the cost of needing an ignore list for the false positives.
- Doing nothing and cleaning up dependencies opportunistically. That is what we do today, and it is how the list above accumulated.

**Additional context**

Both DataFusion and several other Arrow subprojects already run `cargo machete` in CI, so this would bring Ballista in line with them.

7487:

Closes #2410.

# Rationale for this change

We do not currently check for unused dependencies in CI, so dead manifest entries accumulate — they still get resolved, downloaded, and often compiled. This adds `cargo machete` to the lint surface, following what DataFusion and other Arrow subprojects already do.

**Removed dependencies that are genuinely unused** (no references anywhere in the source tree):

- `ballista-core`: `aws-config`, `aws-credential-types` (also dropped from the `build-binary` feature; S3 support goes through `object_store`'s own builders), `md-5`, `prost-types`
- `ballista-scheduler`: `prost-types`
- `ballista-executor`: `tracing` (also dropped from the `build-binary` feature; `tracing-subscriber`/`tracing-appender` pull it in transitively)
- `ballista-cli`: `tracing-appender` (also dropped from the `tui` feature)
- `pyballista`: `async-trait`
- root workspace: `prost-types` (no longer referenced by any member)

**Moved**: scheduler's `insta` from `[dependencies]` to `[dev-dependencies]` — it is only used through `ballista_core::assert_plan!` in `cfg(test)` modules, so it no longer compiles into release scheduler builds.

**Kept, with `[package.metadata.cargo-machete]` ignored entries and comments**:

- `ballista-scheduler`: `insta` (macro expansion, see above) and `tonic-prost` (used by the `keda-scaler` gRPC code generated into `OUT_DIR`, which machete cannot see)
- `ballista-executor`: `arrow` (not imported directly, but pulls arrow's `ipc_compression` feature into the build so shuffle IPC compression — lz4 by default — works through datafusion's arrow re-export)
- `ballista-cli`: `critical-section` (provides the critical-section implementation `ratatui-core` needs at link time in the wasm32 `web` build)
- `msrvcheck`: `time` (never imported; pins the transitive `time` version so `cargo` keeps compiling)

**CI wiring**: new `ci/scripts/rust_machete.sh`, added to `dev/rust_lint.sh`, and a `cargo-machete` job in the `Ballista Rust` workflow (installs `cargo-machete@0.9.2` via the already-pinned `taiki-e/install-action`).

Verified locally: `cargo machete` is clean; `taplo format --check` (0.10.0) passes; the CI clippy commands for the four touched crates pass with `-D warnings` (including `--all-features`, which exercises `keda-scaler`); `cargo check -p ballista-scheduler -p ballista-executor -p ballista-core -p ballista --no-default-features --locked` passes; scheduler tests using `assert_plan!` still pass; `python/Cargo.lock` stays in sync (`cargo metadata --locked`).

The implicit `aws-config`/`aws-credential-types` cargo features of `ballista-core` no longer exist, and `ballista-core`/`ballista-executor` lose a few transitive dependencies. No API or behavior changes.


andygrove (approved):
Thanks @7487

NoahKusaba:
@7487 can you resolve conflicts and update branch, if you are still managing this change?
I think it's really important we get this merged in before the pending ballista version update.

7487:
@NoahKusaba Done — merged upstream main and resolved the conflicts. Both were against the workspace-dependency refactor: `ballista-core` now takes `chrono = { workspace = true }` (aws-config/aws-credential-types stay removed — still no references in the source tree on current main), and `ballista-cli` keeps `tracing-appender` removed (still only used by the scheduler/executor binaries). Re-verified locally: `cargo machete` clean, `taplo format --check` passes, and both lockfiles are in sync.

NoahKusaba (approved):

7487:
Merged `main` again (dd8d5cf) and pushed one small follow-up (555c1a2):

- #2421 added a `tinyvec` version pin to `dev/msrvcheck` that is never imported — same shape as the existing `time` pin — so `cargo machete` on `main` + this PR reported it. Added it to the crate's `[package.metadata.cargo-machete]` ignored list. Without this the new job would have gone red right after merge.
- Aligned the `cargo-machete` job's `taiki-e/install-action` pin with the v2.87.5 SHA the rest of `rust.yml` uses since #2429.

Verified locally on the merged tree: `ci/scripts/rust_machete.sh` clean, `taplo format --check` passes.

The `pull_request` workflows on this branch have been sitting at "action_required" since the 09-04 update (they all passed on the original commit), so this just needs a maintainer to approve the runs. Ready to go from my side.

milenkovicm:
Thanks @7487 and @andygrove

PR #2414, opened 2026-09-02
perf(core): remove the inert shuffle-read sort

### LINKED ISSUE #2403: Deterministically stagger shuffle read fetch order instead of randomizing it

NoahKusaba:
Closes #2403 

 # Rationale for this change
The shuffle is uniformly random (Durstenfeld's Fisher-Yates) , so sorting before the random shuffle does nothing.

## AI Verification

Confirmed by counting permutations over 4 blocks (24 permutations), 480,000
trials per pipeline. Uniform if chi-square is below 35.17 (5% level, 23 df):

| shape | pipeline | chi-square |
| --- | --- | --- |
| even | sort + shuffle | 24.66 |
| even | shuffle only | 13.38 |
| skewed | sort + shuffle | 27.17 |
| skewed | shuffle only | 19.58 |

All four pass. Every trial also asserted the output is a permutation of the
input, so the set of blocks fetched is unchanged.

- Delete inert sort for partition_locations in shuffle reader.
- Delete now unused itertools crate.

No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1128,7 +1128,6 @@ dependencies = [
  "datafusion-proto-common",
  "datafusion-spark",
  "futures",
- "itertools 0.15.0",
  "log",
  "object_store",
  "parking_lot",
--- a/ballista/core/Cargo.toml
+++ b/ballista/core/Cargo.toml
@@ -55,7 +55,6 @@ datafusion-proto = { workspace = true }
 datafusion-proto-common = { workspace = true }
 datafusion-spark = { workspace = true, optional = true, features = ["datafusion"] }
 futures = { workspace = true }
-itertools = { workspace = true }
 log = { workspace = true }
 object_store = { workspace = true, features = ["aws", "http"], optional = true }
 parking_lot = { workspace = true }
--- a/ballista/core/src/execution_plans/shuffle_reader.rs
+++ b/ballista/core/src/execution_plans/shuffle_reader.rs
@@ -50,7 +50,6 @@ use datafusion::physical_plan::{
 };
 use datafusion::prelude::SessionConfig;
 use futures::{Stream, StreamExt, TryStreamExt, ready};
-use itertools::Itertools;
 use log::{debug, error, trace};
 use rand::prelude::SliceRandom;
 use rand::rng;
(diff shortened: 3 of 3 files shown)


NoahKusaba:
There might have still been some merit to merging the deterministic ordering PR, but I honestly don't think it's worth the complexity. I'm going to split out some of the cleanups / improvements from it into new branches.

avantgardnerio (approved):
Makes sense to me

andygrove (approved):
Thanks @NoahKusaba

Issue #2298, opened 2026-08-14
[DISCUSSION] Restore Flight SQL support as a pluggable query frontend


andygrove:
> **Disclaimer:** this issue was drafted with an LLM (Claude Code) at my request. The code, git, and issue references were verified against `main` at 5e54503, but the design proposal is a starting point for discussion, not a settled plan.

## Is your feature request related to a problem or challenge? Please describe what you are trying to do.

Ballista has no way for a generic SQL client to connect to it. Every client must be a Rust program that builds a DataFusion `LogicalPlan`, serializes it, and speaks `SchedulerGrpc`. That rules out JDBC/ODBC tooling, BI tools, DBeaver, Python (`adbc`/`pyarrow`), and anything else that expects to send SQL text and get Arrow back.

Flight SQL used to fill that gap, and was removed in 46.0.0 (#1227, #1228). It's worth being precise about *why*, because it wasn't "Flight SQL is a bad fit for Ballista":

- Nobody was maintaining it, and it had accumulated unaddressed bugs (#1012, #941, #839, #756).
- It was substantially incomplete. Of ~40 `FlightSqlService` methods in `ballista/scheduler/src/flight_sql.rs`, well over half were `Status::unimplemented`, including `CommandGetSqlInfo` — which carried the comment `// TODO: implement for FlightSQL JDBC to work`. Catalog metadata (`GetCatalogs`, `GetDbSchemas`, `GetTables` as a real RPC, `GetXdbcTypeInfo`, keys) was stubbed, so driver-level introspection didn't work.
- It was structurally coupled to scheduler internals: it took a concrete `SchedulerServer<LogicalPlanNode, PhysicalPlanNode>`, reached into `server.state.config`, faked catalog responses by smuggling well-known strings (`"get_flight_info_tables"`) through the `job_id` field of a `FetchPartition` ticket, and carried its own inline Flight proxy.
- Authentication was `if user != "admin" || pass != "password"`, Basic-only.
- It had **zero tests** (`grep -c '#\[test\]' <old file>` → 0).

So the removal was right, and reviving that file as-is would be wrong. But the capability is still valuable, and @avantgardnerio made the point directly in #2249: *"using this to restore FlightSQL could add a significant audience as well."* This issue is to design a new implementation properly.

## Why now

Several things have landed since 46.0.0 that make this materially easier than it was in 2023–2025:

[...]

Issue #1227, opened 2025-04-04
Remove `flight-sql` from ballista in 46.0.0


milenkovicm:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

I would like to propose to remove `flight-sql` as the part of #1068, it does not quite fit with current ballista direction.
There is no effort put into it and there is effort needed to improve it, without anybody interested to contribute.

Few relevant issues: 

- #1012 
- #941
- #839

**Describe the solution you'd like**

- Remove `flight-sql` from codebase in 46.0.0

**Describe alternatives you've considered**

- move it to `datafusion-contrib` if someone is interested to maintain it 

**Additional context**

milenkovicm:
issues which may be relevant to `flight-sql`:

-  #941
- #839

Issue #1012, opened 2024-04-30
wrong location in FlightEndpoint causing not support Flight  SQL JDBC Driver


Curricane:
**Describe the bug**
use flight-sql-jdbc-driver can not execute "select 1" in ballista

**To Reproduce**
os: wsl2 ArchLinux
## start ballista
RUST_LOG=info ballista-scheduler
RUST_LOG=info ballista-executor -c 4  
## connect ballista with java code
os: windows11 
ide: Intellij
```java
package com.test;

import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.ResultSet;
import java.sql.Statement;
import java.util.Properties;
import org.apache.arrow.compression.CommonsCompressionFactory;
import org.apache.arrow.compression.Lz4CompressionCodec;

public class App {
    public static void main(String[] args) {
        try {
            // Load the JDBC driver
            Class.forName("org.apache.arrow.driver.jdbc.ArrowFlightJdbcDriver");

            String url = "jdbc:arrow-flight-sql://localhost:50050?useEncryption=0";

            Properties p = new Properties();
            p.setProperty("user", "admin");
            p.setProperty("password", "password");
            Connection connection = DriverManager.getConnection(url, p);

            // Create a statement
            Statement statement = connection.createStatement();

            // Execute a query
            String query = "SELECT 1"; // Replace "your_table" with your actual table name
            ResultSet resultSet = statement.executeQuery(query);

            // Process the result set
            while (resultSet.next()) {
                System.out.println(resultSet.getInt(1)); // Adjust the index or use column names depending on your
                                                            // schema
            }

            // Close the resources
            resultSet.close();
            statement.close();
            connection.close();

        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```                       
pom.xml
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.fusiongrid</groupId>
    <artifactId>arrow-flight-sql-example</artifactId>
    <version>1.0-SNAPSHOT</version>
    <properties>
        <maven.compiler.source>11</maven.compiler.source>
        <maven.compiler.target>11</maven.compiler.target>
[...]

PR #2416, opened 2026-09-03
feat: Arrow Flight SQL frontend for the scheduler + ADBC support

andygrove:

Part of #2298. Does not close it.

# Rationale for this change

Ballista has no way for a generic SQL client to connect. Flight SQL was removed
in 46.0.0 (#1227, #1228) because it was unmaintained, substantially incomplete,
structurally coupled to `SchedulerServer`, and had zero tests. This is a fresh
implementation aimed at those four objections, not a revival of the old
`flight_sql.rs`.

The concrete motivation was querying a Ballista cluster from Python via
`adbc_driver_flightsql`, bypassing the Ballista Python bindings.

**New crate `ballista/flight-sql`**, depending only on `ballista-core` — it
never references `SchedulerServer`.

- `backend.rs` — a `QueryBackend` trait (session open/close, submit-and-await,
  cancel), the transport-neutral seam #2298 asked for.
- `service.rs` — `BallistaFlightSqlService<B: QueryBackend>`: handshake,
  `CommandStatementQuery`, `DoGet`, prepared statements, `CommandStatementUpdate`,
  `CancelQuery`.
- `metadata.rs` — `SqlInfo`, `XdbcTypeInfo`, and catalog introspection sourced
  from the session's real DataFusion catalog.
- `auth.rs` — pluggable `Authenticator`, no built-in credentials.
- `session.rs` — TTL'd stores for sessions, prepared statements, and results.

**Scheduler wiring** — `ballista/scheduler/src/flight_sql.rs` (~100 lines)
implements `QueryBackend`, behind a non-default `flight-sql` feature and a
`--flight-sql` runtime flag.

**Supporting changes** — `flight_proxy_service.rs` moves from
`ballista/scheduler/src` to `ballista/core/src` unchanged so both crates can use
it; `start_grpc_service_with_listener` extracted so callers can bind port 0;
`docs/source/user-guide/flightsql.md` replaces the page deleted in #1228;
`examples/python/adbc_flight_sql.py`; `benchmarks/tpch_adbc.py`; a CI step
running the tests with the feature on.

**Fixes outside the new crate**, each found by driving a real cluster with the
generic ADBC driver and each covered by a test:

- `ballista-core`'s `LocalRun` visitor is exposed as `scans_only_local_tables`
  and reused by the frontend. `SHOW TABLES` and other `information_schema`
  reads were being distributed, where their physical form is a
  `StreamingTableExec` the codec cannot serialize, so `GetFlightInfo` hung for
  ever — which is what a BI tool does on connect. `BallistaQueryPlanner`
  already answered these locally, hence `ballista-cli` never hit it.
- `TaskManager::launch_multi_task` now reports the jobs whose task definitions
  could not be prepared, and `SchedulerState` fails them. It used to log and
  drop them, leaving the job Running for the life of the scheduler with every
  caller — `GetFlightInfo`, `DistributedQueryExec` — blocked with no timeout.
  The pull-based `poll_work` path already did this.
- `distributed_explain.rs` declares its two output columns not-null, matching
  both DataFusion's `ExplainExec` and `DistributedExplainAnalyzeExec`, so
  `EXPLAIN` describes itself the same way however it is run.

# Design decisions worth discussing

Detail and rationale for each is in #2298; summarising what this PR chose:

1. **Endpoints carry no location**, so clients redeem tickets on the connection
   they already have and no executor address leaks. Direct fix for #1012/#1349.
2. **The frontend subsumes the plain proxy** (only one can serve
   `FlightService`); `do_get_fallback` keeps Ballista's own tickets working.
3. **Catalog model: option (b)** — embedder-supplied via `SessionBuilder`, plus
   session-scoped DDL.
4. **A separate `--flight-sql` flag**, distinct from the Flight result-proxy
   configuration. #2281 landed on main while this was in progress and renamed
   that knob to `enable_embedded_flight_proxy` / `advertise_flight_endpoint`,
   which settles #2298 item 8; when the frontend is enabled it subsumes the
   embedded proxy on the same port.
5. **CTAS, `INSERT`/`UPDATE`/`DELETE`/`COPY` are refused** rather than silently
   executed on the scheduler.
6. **`CleanJobData` not wired** — the scheduler's existing delayed cleanup owns
   result retention.

7. **A running query cannot be cancelled.** Without `PollFlightInfo`,
   `GetFlightInfo` blocks, and `CancelQuery` takes an `ActionCancelQueryRequest`
   carrying a `FlightInfo` the client only holds once the query has already
   finished — so there is nothing left to cancel. A client that disconnects
   leaves its query running to completion.

   Cancelling on disconnect is the obvious fix and is deliberately **not** in
   this PR: #2418 is a pre-existing bug where cancelling a running job never
   returns its executor vcores, so under the default `PushStaged` policy a
   single cancellation can leave the scheduler unable to bind any further work.
   Until that is fixed, cancelling an abandoned query would turn a wasted query
   into a wedged scheduler. #2418 reproduces on `main` with `ballista-cli` and
   the scheduler's own `CancelJob`, with no Flight SQL involved.

Not implemented: `PollFlightInfo` (so `GetFlightInfo` blocks until the query
finishes), cancellation of a running query (see 7), bound parameters,
transactions, Substrait, key metadata.

# How are these changes tested?

Unit and integration tests for the new crate, plus end-to-end tests in
`ballista/scheduler/tests/flight_sql.rs` that drive a real in-process cluster
over the Flight SQL protocol: a distributed query, session-scoped catalog,
prepared statements, `CommandGetTables`/`CommandGetSqlInfo`, agreement between
the advertised schema and the delivered data, and catalog queries answered
without distributing them. CI runs these with the feature on.

Manually, against a scheduler and executor on TPC-H SF100 Parquet, driven only
by `adbc_driver_flightsql` with no Ballista client library: all 22 TPC-H
queries pass, and pass again on the same scheduler without a restart.

Only opt-in. `flight-sql` is a non-default compile-time feature and
`--flight-sql` is off at runtime, so no existing deployment changes behaviour.
Adds a user guide page and a Python example.


milenkovicm:
not sure why i see a empty line after each line 
image

villebro:
> not sure why i see a empty line after each line

@milenkovicm I've been seeing this from time to time. It's a GitHub rendering issue, and goes away with a browser refresh. Very annoying.

comphead:
Thanks @andygrove the direction makes a lot of sense to feed downstream BI/reporting/visualize layers using Arrow format, considering the layer already supports the Arrow capable endpoint. Apache Superset https://github.com/apache/superset is one of the good examples.

ballista/flight-sql/src/service.rs
```diff
@@ -0,0 +1,852 @@
...
+        &self,
+        query: ActionClosePreparedStatementRequest,
+        _request: Request<Action>,
+    ) -> Result<(), Status> {
+        let handle = prepared_handle(&query.prepared_statement_handle)?;
+        self.store.remove_prepared(&handle);
+        Ok(())
+    }
+
+    async fn do_action_cancel_query(
```
comphead, line 711: **Blocker.** This can never cancel a running query, and calling it deletes the results the client was just given.

- A client only holds the `FlightInfo` after `GetFlightInfo` returns, so by then the job has already succeeded (design note 7).
- `cancel_job` posts `JobCancel`, and that handler always calls `clean_up_failed_job`. That removes the job's data on the executors without the `finished_job_data_clean_up_interval_seconds` delay, so the endpoints in that `FlightInfo` stop working.
- `result: 1` is `CANCEL_RESULT_CANCELLED`. For a cancel that completes in the background, the proto asks for `CANCEL_RESULT_CANCELLING` (2).
- `ActionCancelQueryRequest` has been deprecated since Flight SQL 13.0.0 in favor of `CancelFlightInfo`, which arrow-flight 59.2.0 routes to `do_action_fallback`.

Suggest removing `do_action_cancel_query`, `job_id_from_flight_info`, `QueryBackend::cancel` and `cancelling_a_query_reaches_the_backend`, reporting `FlightSqlServerCancel` as `false` (`metadata.rs:65`), and dropping "query cancellation" from the crate README. Cancel could come back together with `PollFlightInfo`.

ballista/flight-sql/src/metadata.rs
```diff
@@ -0,0 +1,295 @@
...
+    );
+    builder.append(SqlInfo::FlightSqlServerCancel, true);
+    builder.append(SqlInfo::FlightSqlServerBulkIngestion, false);
+
+    builder.append(SqlInfo::SqlDdlCatalog, false);
+    builder.append(SqlInfo::SqlDdlSchema, true);
+    builder.append(SqlInfo::SqlDdlTable, true);
+    // DataFusion folds unquoted identifiers to lower case and preserves the
+    // case of quoted ones: SQL_CASE_SENSITIVITY_LOWERCASE / _CASE_INSENSITIVE.
+    builder.append(SqlInfo::SqlIdentifierCase, 2i32);
```
comphead, line 73: **Major.** These numbers don't match the comment. In `SqlSupportedCaseSensitivity`, 2 is `SQL_CASE_SENSITIVITY_UPPERCASE` and 3 is `SQL_CASE_SENSITIVITY_LOWERCASE`. So this reports unquoted identifiers as stored in upper case and quoted ones as lower case, and clients that read these values will get DataFusion's identifier rules backwards.

`arrow_flight::sql` re-exports the enum, which makes the intent explicit:

```rust
builder.append(
    SqlInfo::SqlIdentifierCase,
    SqlSupportedCaseSensitivity::SqlCaseSensitivityLowercase as i32,
);
```

Same for `SqlQuotedIdentifierCase` on line 75, with the value picked deliberately.

ballista/flight-sql/src/service.rs
```diff
@@ -0,0 +1,852 @@
...
+             the distributed write path is not implemented",
+        ),
+        LogicalPlan::Ddl(DdlStatement::CreateMemoryTable(_)) => Disposition::Unsupported(
+            "Ballista Flight SQL does not support CREATE TABLE AS SELECT, \
+             because it would execute on the scheduler rather than the cluster; \
+             use CREATE EXTERNAL TABLE over data the executors can read",
+        ),
+        // Other DDL only edits the session catalog, and `SET`-style statements
+        // only edit session config; neither has anything to distribute.
+        LogicalPlan::Ddl(_) | LogicalPlan::Statement(_) => Disposition::RunOnScheduler,
```
comphead, line 307: **Major.** `disposition` only looks at the root node, so a few statements still run on the scheduler or get past the write refusal:

- `EXECUTE q` is a `LogicalPlan::Statement`, so it lands here. In DataFusion 55, `execute_logical_plan` turns it into the prepared query (`execute_prepared`), and `execute_locally` then runs that `SELECT` on the scheduler (the default `SessionBuilder` uses DataFusion's local planner) and keeps the whole result in memory as a `LocalResult`. That is the same reason CTAS is refused.
- `EXPLAIN ANALYZE INSERT ...` plans as `Analyze` over `Dml` (`explain_to_plan` plans any statement), so it is classified `Distribute` instead of being refused.

Suggest refusing `Statement::Execute` and checking the whole plan, subqueries included, for `Dml`, `Copy` and `CreateMemoryTable`. Related: the comment on `scans_only_local_tables`.

ballista/core/src/planner.rs
```diff
@@ -165,6 +162,19 @@ impl<T: 'static + AsLogicalPlan> QueryPlanner for BallistaQueryPlanner<T> {
...
+/// Returns `true` if every table `plan` scans lives in `information_schema`.
+///
+/// Such a plan reads the catalog of the node it is planned on and has nothing
+/// to distribute; worse, the physical form of those scans (`StreamingTableExec`)
+/// cannot be serialized for an executor, so distributing one produces a job
+/// that can never run. Callers that hand plans to the cluster should answer
+/// these locally instead.
+pub fn scans_only_local_tables(plan: &LogicalPlan) -> bool {
+    let mut local_run = LocalRun::default();
+    let _ = plan.visit(&mut local_run);
```
comphead, line 174: **Major.** `visit` doesn't descend into subquery expressions, while `visit_with_subqueries` does (DataFusion 55, `logical_plan/tree_node.rs`). The Flight SQL frontend calls this on the unoptimized plan from `create_logical_plan`, so for

```sql
SELECT table_name FROM information_schema.tables
WHERE table_name IN (SELECT name FROM big)
```

the only scan it sees is `information_schema.tables`. The statement is classified `RunOnScheduler` and `big` is scanned on the scheduler. `BallistaQueryPlanner` receives the optimized plan, where most subqueries are already decorrelated into joins, which mostly hides this there. `plan.visit_with_subqueries(&mut local_run)` fixes both callers.

ballista/scheduler/src/state/mod.rs
```diff
@@ -305,23 +340,35 @@ impl<T: 'static + AsLogicalPlan, U: 'static + AsExecutionPlan> SchedulerState<T,
...
-                            // It's OK to remove executor aggressively,
-                            // since if the executor is in healthy state, it will be registered again.
-                            state
-                                .remove_executor(&executor_id, Some(err_msg), &sender)
-                                .await;
-
-                            false
-                        } else {
-                            true
+                            Ok(unpreparable) => {
```
comphead, line 348: **Major.** Failing the job is right, but the vcores bound to these tasks never come back. This branch reports success, so the spawned task returns `vec![]` and `unbind_tasks` never refunds what `bind_one` reserved. `abort_job` then sends `CancelTasks` for tasks the executor never received, so no status arrives to refund them either. Under `PushStaged`, each such job permanently lowers that executor's capacity, which is the wedge described in #2418. I found this by reading the code and haven't reproduced it.

Suggest returning the unpreparable tasks' `vcores_consumed` as `ExecutorSlot`s along with failing the job.

Two related points:
- The pull path already fails jobs for the same reason through `SchedulerServer::fail_job` (`scheduler_server/grpc.rs:174`). One helper shared by both paths would keep them consistent, including `queued_at`, which is looked up here and set to now in `fail_job`.
- This changes scheduling for every `PushStaged` user, not only Flight SQL, so it may be easier to review and backport as its own PR.

ballista/scheduler/src/state/task_manager.rs
```diff
@@ -1456,4 +1471,42 @@ mod tests {
         assert_eq!(manager.running_job_number(), 0);
         Ok(())
     }
+
+    /// A task whose definition cannot be prepared is never sent to an executor,
+    /// so no status will ever come back for it. `launch_multi_task` has to name
+    /// the job it dropped, or the job stays Running for the life of the
+    /// scheduler and everything awaiting its status blocks with no timeout.
+    #[tokio::test]
+    async fn tasks_that_cannot_be_prepared_are_reported_against_their_job() -> Result<()>
```
comphead, line 1480: **Test.** This checks what `launch_multi_task` returns, not the behavior the fix is about, so it still passes if the `fail_unlaunchable_job` call in `state/mod.rs` is reverted. A `SchedulerTest` case under `PushStaged` that asserts the job ends up `Failed`, and that the executor's vcores are back, would fail without the fix.

ballista/flight-sql/src/service.rs
```diff
@@ -0,0 +1,852 @@
...
+    async fn get_flight_info_xdbc_type_info(
+        &self,
+        query: CommandGetXdbcTypeInfo,
+        request: Request<FlightDescriptor>,
+    ) -> Result<Response<FlightInfo>, Status> {
+        let schema = query.into_builder(&self.xdbc_info).schema();
+        Self::metadata_info(query, schema, request.into_inner())
+    }
+
+    async fn do_get_statement(
```
comphead, line 564: **Major (security).** With an authenticator installed, the token is only checked where a session context is needed. `do_get_statement`, `get_flight_info_prepared_statement`, `do_action_close_prepared_statement`, `do_action_cancel_query` and `do_get_fallback` never call `session_id`.

Partition tickets describe themselves fully and aren't signed, and job IDs are listed at `GET /api/jobs` on the same port (with the default `rest-api` feature). So a client with no token can forge a ticket and read any job's result partitions.

Suggestions:
- Check the token in every handler that reads data or changes state.
- Bind prepared handles and local results to the session that created them.
- Sign partition tickets, or swap them for a server-side handle.

`do_get_fallback` has to stay open for native Ballista clients, which is one more reason the docs shouldn't present the authenticator as a security boundary (see the comment in `config.rs`). `Identity` is currently only logged. It's the natural thing to bind handles to, otherwise it can go.

ballista/scheduler/src/config.rs
```diff
@@ -397,6 +405,19 @@ pub struct SchedulerConfig {
...
     pub stage_max_failures: usize,
+    /// Whether to serve Arrow Flight SQL on the scheduler's gRPC port.
+    ///
+    /// The frontend subsumes the plain Flight result proxy when enabled, so
+    /// Ballista's own clients keep working on the same port.
+    #[cfg(feature = "flight-sql")]
+    pub flight_sql: bool,
+    /// Authenticates Arrow Flight SQL handshakes.
+    ///
+    /// Without one the frontend accepts every client, so this is the only way
```
comphead, line 416: **Major (docs).** This overstates what the authenticator gives. The same port also serves `SchedulerGrpc` and the REST API (including `PATCH /api/job/{id}`) with no authentication, and several Flight SQL paths skip the token check (see the comment on `do_get_statement`). Suggest describing it as per-client session isolation and keeping the "trusted network only" guidance from the user guide.

ballista/flight-sql/src/service.rs
```diff
@@ -0,0 +1,852 @@
...
+
+    /// Runs a planned statement and describes where to collect its results.
+    async fn flight_info_for(
+        &self,
+        ctx: Arc<SessionContext>,
+        plan: LogicalPlan,
+        descriptor: FlightDescriptor,
+        job_name: &str,
+    ) -> Result<FlightInfo, Status> {
+        if let Disposition::Unsupported(reason) = disposition(&plan) {
```
comphead, line 201: **Minor.** `disposition(&plan)` runs twice here, and each call walks the plan through `scans_only_local_tables`. One `match disposition(&plan) { ... }` over the three variants, as `do_put_statement_update` already does, avoids the second pass.

ballista/core/src/serde/scheduler/mod.rs
```diff
@@ -71,11 +71,27 @@ impl PartitionLocation {
...
-        if self.is_sort_shuffle {
-            ShuffleLayout::Sort
-        } else {
-            ShuffleLayout::Passthrough
-        }
+        shuffle_layout(self.is_sort_shuffle)
+    }
+}
+
+impl crate::serde::protobuf::PartitionLocation {
```
comphead, line 78: **Minor (reuse).** The only caller is `partition_handle` in the new crate, while `fetch_partition` in `execution_plans/distributed_query.rs` (line 899) still inlines the same `is_sort_shuffle` rule on the same protobuf type. Either use this there too, or add one constructor for `Action::FetchPartition` from a `PartitionLocation` that both `partition_handle` and `BallistaClient::fetch_partition_proxied` use.

ballista/flight-sql/src/session.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Takes a local result. Results are single-use: a ticket is redeemed once.
+    pub(crate) fn take_result(&self, handle: &str) -> Option<LocalResult> {
+        self.results.remove(handle).map(|(_, entry)| entry.value)
+    }
+
+    /// Evicts everything idle for longer than the TTL.
+    ///
+    /// Returns the session ids that no longer have any live token, so the
+    /// caller can release them in the backend.
+    pub(crate) fn sweep(&self) -> Vec<String> {
```
comphead, line 166: **Minor.** This iterates `sessions` to collect the expired tokens and then removes them one by one. A single `retain` pass can collect the released session ids as it goes, and their contexts can be dropped afterwards. That also leaves `remove_session` used only by a test.

ballista/flight-sql/src/session.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    #[test]
+    fn session_survives_while_touched_and_expires_when_idle() {
+        let store = SessionStore::new(Duration::from_millis(50));
+        store.insert_session("token".to_string(), "session".to_string());
+
+        assert_eq!(store.session("token"), Some("session".to_string()));
+        assert!(store.sweep().is_empty());
+
+        std::thread::sleep(Duration::from_millis(60));
```
comphead, line 231: **Minor (flaky test).** The first `sweep()` has to run within 50 ms of `insert_session`, which a loaded CI runner can miss. The idle-eviction tests in `executor/src/client_pool.rs` backdate the timestamp (`Instant::now() - age`) instead of sleeping. Setting `last_used` in the past here would make this deterministic.

comphead (commented):
Thanks for picking this up. The design notes make the trade-offs easy to follow.

I reviewed `250500ce` against DataFusion 55.0.0 and arrow-flight 59.2.0. Details are inline. The main points:

- **Blocker:** `CancelQuery` can only reach jobs that have already finished, and it deletes their results.
- **Major:** the `SqlIdentifierCase` values are wrong. `EXECUTE` and subqueries over `information_schema` still run on the scheduler. The fix for jobs that can't be launched still leaks their vcores. Auth isn't enforced on the data paths.

Points not tied to a single line:

- **Scope.** The `PushStaged` job-failure fix, the `EXPLAIN` nullability change, and the ADBC benchmark and example (about 430 lines of Python) could each land separately. That would make the new crate easier to review.
- **Missing tests.** A result with more than one endpoint (every test here seems to produce a single one), `PushStaged`, auth on DoGet and prepared statements, and refusal of `EXECUTE`.
- **Duplicate tests.** `the_default_service_allows_anonymous_access` overlaps `an_authenticator_makes_tokenless_requests_fail`. `GetTables` is tested in `metadata.rs`, in `tests/service.rs` and in the scheduler end-to-end test. `sql_info_is_buildable` and `xdbc_type_info_is_buildable` add nothing, because every `BallistaFlightSqlService::new` builds both. `local_results_are_single_use` and `plain_ddl_runs_on_the_scheduler` repeat `ddl_runs_on_the_scheduler_and_its_result_is_single_use`. The two bearer-token tests can be one.
- **Comments.** `LocalResult` still mentions DML (`session.rs:46`). `query_failed` says a failure shouldn't be "an opaque 500" but returns `Status::internal`. The `QueryBackend::session` doc says the context is created on first use, while the scheduler builds a new one on every call. The notes about the pre-46.0.0 implementation in the code and tests would fit better in the PR description.
- **Robustness.** `BallistaFlightSqlService::new` spawns the reaper task, so it panics outside a Tokio runtime. Sessions, prepared statements and local results have no cap, and every handshake builds a `SessionContext` up front.
- **Performance.** All result bytes go through the scheduler, with a new executor connection per `DoGet`. That's fine for TPC-H-sized answers, but large results will make the scheduler the bottleneck, and nothing measures that yet.

ballista/flight-sql/src/service.rs
```diff
@@ -0,0 +1,852 @@
...
+        Ok(response)
+    }
+
+    /// Redeems tickets that are not Flight SQL commands.
+    ///
+    /// Ballista's own Rust client fetches shuffle output with a
+    /// `ballista.protobuf.Action` ticket. When the Flight SQL frontend is
+    /// mounted it replaces the standalone proxy on the scheduler's port, so it
+    /// has to keep serving those tickets.
+    async fn do_get_fallback(
```
comphead, line 469: **Minor.** The ticket is decoded here only to return `InvalidArgument`, and then `BallistaFlightProxyService::do_get` decodes it again. Mapping the decode error to `Status::invalid_argument` inside the proxy removes the second decode and gives the plain proxy the same error code.

.github/workflows/rust.yml
```diff
@@ -99,6 +99,12 @@ jobs:
           export ARROW_TEST_DATA=$(pwd)/testing/data
           export PARQUET_TEST_DATA=$(pwd)/parquet-testing/data
           cargo test --profile ci --features=testcontainers
+      - name: Run Arrow Flight SQL tests
+        # The frontend is a non-default feature, so the workspace run above
+        # compiles its integration test away.
+        run: |
+          export PATH=$PATH:$HOME/d/protoc/bin
+          cargo test --profile ci -p ballista-scheduler --features flight-sql --test flight_sql
```
comphead, line 107: **Minor (CI).** `ci/scripts/rust_clippy.sh` lints an explicit list of packages and doesn't include `ballista-flight-sql`. `--package ballista-scheduler --all-features` only lints the scheduler's own code, so the new crate is never linted. Suggest adding it to that script.

Also, this step compiles the scheduler a second time with a different feature set. Adding `ballista-scheduler/flight-sql` to `--features` in the step above should cover it in one build. I haven't measured the difference.

ballista/scheduler/src/flight_sql.rs
```diff
@@ -0,0 +1,111 @@
...
+    ) -> Result<QueryResult> {
+        // The schema handed to the client has to be the one the shuffle files
+        // actually carry, which is the physical plan's, not the logical plan's:
+        // the two disagree about nullability often enough to matter (`version()`
+        // is nullable logically and not-null in the data; a list literal goes the
+        // other way), and a Flight SQL client that is promised one schema and
+        // handed another -- ADBC among them -- rejects the result outright.
+        // Planning twice costs a few milliseconds against a job that is about to
+        // run on the cluster.
+        let schema = ctx.state().create_physical_plan(&plan).await?.schema();
```
comphead, line 70: **Question.** This builds the physical plan once only to read its schema, and the job is planned again on submit. For listing tables on object storage, physical planning lists files (and may collect statistics), so that work happens twice per query. Could the schema come from the job's final stage plan once it succeeds, since that is the plan that wrote the files?

ballista/flight-sql/tests/service.rs
```diff
@@ -0,0 +1,593 @@
...
+    );
+    assert_ne!(
+        err.code(),
+        tonic::Code::InvalidArgument,
+        "the ticket itself must be accepted: {err}"
+    );
+}
+
+#[tokio::test]
+async fn sessions_are_isolated_per_token() -> std::result::Result<(), BallistaError> {
```
comphead, line 582: **Test.** This only exercises `StubBackend` and `SessionContext`, and `_service` is never used. Either drop it or make it go through the frontend: two handshakes, DDL in one session, and the table absent from the other.

ballista/flight-sql/tests/service.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    assert_eq!(err.code(), tonic::Code::Unauthenticated);
+}
+
+/// The frontend replaces the standalone Flight proxy when mounted, so
+/// Ballista's own client tickets must still be recognised. The fallback is
+/// only reachable if arrow-flight's dispatcher does not mistake a Ballista
+/// action for a Flight SQL command.
+#[tokio::test]
+async fn ballista_client_tickets_are_recognised_by_the_fallback() {
```
comphead, line 533: **Test.** The doc comment says this checks that arrow-flight's dispatcher routes a Ballista ticket to the fallback, but the test calls `do_get_fallback` directly, so the dispatcher never runs. `ballista_tickets_route_to_the_fallback` in `service.rs` covers half of it through `Any::decode`. One test that calls `FlightService::do_get(&service, ...)` with a Ballista ticket would cover both, and the other can go.

ballista/flight-sql/tests/service.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    let err = expect_err(
+        service.do_get_statement(ticket, ticket_for(&info, 0)).await,
+        "a ticket is redeemable once",
+    );
+    assert_eq!(err.code(), tonic::Code::NotFound);
+}
+
+#[tokio::test]
+async fn writes_are_refused_rather_than_run_on_the_scheduler() {
```
comphead, line 292: **Test.** This and `ctas_is_refused_rather_than_run_on_one_node` exercise the same path. One test that loops over `INSERT`, CTAS, `COPY` and `EXECUTE` (or an `rstest` case list) would cover more statements with less code. The `EXECUTE` case fails today (see the comment on `disposition`).

ballista/scheduler/tests/flight_sql.rs
```diff
@@ -0,0 +1,339 @@
...
+    );
+
+    Ok(())
+}
+
+/// A table registered through DDL must still be there for the next request.
+/// `QueryBackend::session` builds a fresh `SessionContext` every call, so this
+/// only holds because the frontend caches it per session.
+#[tokio::test]
+async fn a_session_keeps_its_catalog_across_requests() -> Result {
```
comphead, line 203: **Test.** This is already covered by `flight_sql_runs_a_distributed_query` and `flight_sql_supports_prepared_statements`, which both register the table in one request and query it in another. Each test boots its own cluster, so dropping it also saves CI time.

ballista/scheduler/tests/flight_sql.rs
```diff
@@ -0,0 +1,339 @@
...
+use tonic::transport::Channel;
+
+/// Test errors are only ever printed, and the client, transport, and Ballista
+/// error types all implement `Error` — so one boxed type spares every call site
+/// a `map_err`.
+type Result<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+/// Boots a scheduler serving Flight SQL plus one executor, and returns the
+/// scheduler's URL.
+async fn start_cluster() -> Result<String> {
```
comphead, line 59: **Test.** Every end-to-end test runs under `PullStaged`, while the hang fix in `state/mod.rs` is on the `PushStaged` path. Covering push would need a push-mode executor, since `new_standalone_executor` only runs `poll_loop`. If that's too much for this PR, the `SchedulerTest` case suggested on `task_manager.rs` covers the fix itself.

Smaller points: this repeats most of `standalone::new_standalone_scheduler_with_builder`, and it sets `config.flight_sql = true` directly instead of using `with_flight_sql(true)`.

PR #2425, opened 2026-09-05
test(chaos): run #2029 executor-loss scenarios on the k8s backend

akshaychitneni:
Add the two failure scenarios with kind cluster backend

Closes #. N/A - follow-up test coverage for #2029

 # Rationale for this change

The #2029 scheduler fix (#2212) was only covered by the process-based chaos harness. The two failure modes that matter can't be reproduced without a real cluster: total executor loss (a Deployment scaled to zero, not a transient process/pod kill) and pod rescheduling-with-new-id.

Test-only, within chaos-testing. Adds the two scenarios that genuinely need a cluster to the kind backend:

  - G (#2029) — every executor is lost mid-query and the job must fail (not hang), under both AQE settings. The loss is a hard kill: scale the Deployment to 0 (so no replacement is rescheduled) then force-delete the pods (--grace-period=0 --force). A graceful scale 0 alone sends SIGTERM, and the executors drain their in-flight tasks to completion, so the query would succeed — not the condition #2029 is about. New K8sCluster::kill_all_executors_hard() primitive.
  - F — a force-deleted executor pod is rescheduled with a new id; the cluster reabsorbs it and keeps serving queries. AQE-off only (planner-independent).

  Plus:
  - Dedupe the scheduler REST polling shared by both backends into src/rest.rs (registered_executors, running_job_id, await_any_stage_running, await_executor_count, executor_ids); TestCluster and K8sCluster delegate to it. 
  - Add a K8sRun test helper mirroring ha.rs's ChaosRun.
  
  All k8s scenarios stay gated behind the k8s feature and CHAOS_BACKEND=kind.

No — test/harness code only, no public API or runtime changes.

### DIFF
--- a/chaos-testing/src/k8s.rs
+++ b/chaos-testing/src/k8s.rs
@@ -259,29 +259,37 @@ impl K8sCluster {
 
     /// How many executors the scheduler currently considers registered.
     pub async fn registered_executors(&self) -> Result<usize, String> {
-        // A short timeout so a stalled port-forward surfaces as a retryable
-        // error in the polling loop rather than hanging the whole wait.
-        let client = reqwest::Client::builder()
-            .timeout(Duration::from_secs(5))
-            .build()
-            .map_err(|e| e.to_string())?;
-        let body: serde_json::Value = client
-            .get(format!("{}/api/executors", self.rest_url()))
-            .send()
-            .await
-            .map_err(|e| e.to_string())?
-            .json()
-            .await
-            .map_err(|e| e.to_string())?;
-        Ok(body.as_array().map(|a| a.len()).unwrap_or(0))
+        crate::rest::registered_executors(&self.rest_url()).await
     }
 
-    /// Scale the executor Deployment. `0` is a total loss that stays lost (the
-    /// controller does not recreate the pods); scaling back up recovers.
-    ///
-    /// Not yet exercised by a scenario — this is the k8s primitive the executor
-    /// kill/loss scenarios (the #2029 follow-ups) will drive; the baseline test
-    /// only needs a healthy cluster. Kept here so the backend is complete.
+    /// The ids of every executor the scheduler currently lists. A killed
+    /// executor's pod is rescheduled with a fresh id, so a scenario can compare
+    /// this before and after a kill to prove the replacement is genuinely new.
+    pub async fn executor_ids(&self) -> Result<Vec<String>, String> {
+        crate::rest::executor_ids(&self.rest_url()).await
+    }
+
+    /// The id of the single job the scheduler currently knows about.
+    pub async fn running_job_id(&self) -> Result<String, String> {
+        crate::rest::running_job_id(&self.rest_url()).await
+    }
+
+    /// Block until any task in any stage is Running, so a kill lands mid-flight.
+    pub async fn await_any_stage_running(&self, job_id: &str) -> Result<(), String> {
+        crate::rest::await_any_stage_running(&self.rest_url(), job_id).await
+    }
+
+    /// Block until the scheduler considers exactly `n` executors registered
+    /// (i.e. a killed executor has been reaped, or a rescheduled one re-joined).
+    pub async fn await_executor_count(&self, n: usize) -> Result<(), String> {
+        crate::rest::await_executor_count(&self.rest_url(), n).await
+    }
+
+    /// Scale the executor Deployment to `replicas`. Setting `0` stops the
+    /// controller from rescheduling, but it is a *graceful* scale-down: the pods
+    /// receive SIGTERM and the executors drain their in-flight tasks to
+    /// completion before exiting, so a query in flight will still succeed. To
+    /// lose work mid-query, use [`Self::kill_all_executors_hard`].
     pub async fn scale_executors(&self, replicas: usize) -> Result<(), String> {
         kubectl(&[
             "-n",
@@ -289,6 +297,28 @@ impl K8sCluster {
             "scale",
             &format!("deploy/{EXECUTOR_DEPLOYMENT}"),
             &format!("--replicas={replicas}"),
+        ])
+        .await
+        .map(|_| ())
+    }
+
+    /// Total, sustained executor loss *mid-query* (#2029): every executor dies at
+    /// once and none is rescheduled. Scales the Deployment to 0 first so the
+    /// controller will not recreate the pods, then force-deletes them
+    /// (`--grace-period=0 --force`) so they are SIGKILLed without the graceful
+    /// drain that [`Self::scale_executors`] alone allows — otherwise the
+    /// executors would finish the in-flight query and it would succeed.
+    pub async fn kill_all_executors_hard(&self) -> Result<(), String> {
+        self.scale_executors(0).await?;
+        kubectl(&[
+            "-n",
+            &self.namespace,
+            "delete",
+            "pod",
+            "-l",
+            "app=ballista-executor",
+            "--grace-period=0",
+            "--force",
         ])
         .await
         .map(|_| ())
--- a/chaos-testing/src/rest.rs
+++ b/chaos-testing/src/rest.rs
@@ -1,0 +1,159 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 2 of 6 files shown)


akshaychitneni:
cc @andygrove @milenkovicm

chaos-testing/src/k8s.rs
```diff
@@ -294,6 +302,28 @@ impl K8sCluster {
         .map(|_| ())
     }
 
+    /// Total, sustained executor loss *mid-query* (#2029): every executor dies at
+    /// once and none is rescheduled. Scales the Deployment to 0 first so the
+    /// controller will not recreate the pods, then force-deletes them
+    /// (`--grace-period=0 --force`) so they are SIGKILLed without the graceful
+    /// drain that [`Self::scale_executors`] alone allows — otherwise the
+    /// executors would finish the in-flight query and it would succeed.
+    pub async fn kill_all_executors_hard(&self) -> Result<(), String> {
```
andygrove, line 311: This races SIGTERM against SIGKILL, and I think it makes scenario G non-deterministic about which scheduler path it exercises.

`kubectl scale --replicas=0` returns as soon as the API server accepts it. The ReplicaSet controller then marks the pods for deletion and kubelet sends SIGTERM, and all of that happens concurrently with the second kubectl round trip that force-deletes them. So the executors may or may not see SIGTERM before the SIGKILL lands, and which one wins will vary run to run.

That matters because `start_executor_process` handles SIGTERM by setting `TERMINATING` and immediately sending a `Fenced` heartbeat to the scheduler (`ballista/executor/src/executor_process.rs`, the `tokio::select!` around L775 and the `notify_scheduler` block just after). The scheduler then reaps that executor through the `terminating && grace_period_expired` branch of `get_expired_executors` rather than the plain `expired` branch. Both end in a failed job, so the assertion passes either way, but some fraction of runs will be covering the graceful fence path instead of the abrupt-loss path #2029 is actually about. The `ha.rs` version uses a raw SIGKILL and has no such ambiguity, so k8s is the weaker of the two here, which is the opposite of the intent.

If you want a clean abrupt total loss with no controller left to reschedule, `kubectl delete deploy ballista-executor --cascade=orphan` followed by a force-delete of the now orphaned pods gets there with no SIGTERM in the picture at all.

chaos-testing/src/rest.rs
```diff
@@ -0,0 +1,159 @@
...
+/// and materialize stages differently, so rather than target a specific stage id
+/// we wait until the job is genuinely executing a task somewhere. Used where the
+/// scenario only needs a fault to land mid-flight.
+pub(crate) async fn await_any_stage_running(
+    rest_url: &str,
+    job_id: &str,
+) -> Result<(), String> {
+    let deadline = Instant::now() + Duration::from_secs(60);
+    loop {
+        let stages = stages(rest_url, job_id).await?;
```
andygrove, line 109: `REQUEST_TIMEOUT` is documented as making a stalled connection "surface as a retryable error inside a polling loop rather than hanging the whole wait", but only `await_executor_count` actually treats it that way with its `if let Ok(count)`. Both `running_job_id` (L79) and `await_any_stage_running` (L109) do `get_json(...).await?`, so a single timed out or refused request ends the wait instead of retrying to the deadline.

That was harmless while these two only ran over loopback, but scenario G is the first thing to drive them across the kubectl port-forward, and the harness has a whole supervisor that expects the forward to drop and come back (`PORT_FORWARD_RETRY_DELAY`). A restart landing inside either poll fails the test at `expect("job must appear")` with nothing useful to go on.

Retrying on error until the deadline, and carrying the last error into the timeout message, would match what the comment already promises.

chaos-testing/src/k8s.rs
```diff
@@ -259,29 +259,37 @@ impl K8sCluster {
...
+    }
+
+    /// Block until any task in any stage is Running, so a kill lands mid-flight.
+    pub async fn await_any_stage_running(&self, job_id: &str) -> Result<(), String> {
+        crate::rest::await_any_stage_running(&self.rest_url(), job_id).await
+    }
+
+    /// Block until the scheduler considers exactly `n` executors registered
+    /// (i.e. a killed executor has been reaped, or a rescheduled one re-joined).
+    pub async fn await_executor_count(&self, n: usize) -> Result<(), String> {
```
andygrove, line 284: I think this one is dead code, and a slightly worse copy of something already here.

On the process backend the pair is a real distinction, because `TestCluster::await_executors` waits for `count >= n` while `await_executor_count` waits for `count == n`. But `K8sCluster::await_executors` (L199) already waits for `count == n`, on the same 120s deadline, and it also calls `dump_diagnostics()` on timeout. So this method is that same wait minus the diagnostics, and nothing calls it. Scenario F rolls its own `await_replacement` instead.

Probably worth either dropping it or folding the two into one.

chaos-testing/tests/k8s.rs
```diff
@@ -27,87 +27,278 @@
...
-    let config = SessionConfig::new_with_ballista();
-    let state = SessionStateBuilder::new()
-        .with_config(config)
-        .with_default_features()
-        .build();
-    let ctx = SessionContext::remote_with_state(&cluster.scheduler_url(), state)
+    let result = tokio::time::timeout(Duration::from_secs(120), query)
         .await
-        .expect("client must connect to the scheduler");
+        .expect("job must terminate, not hang, after every executor is lost")
```
andygrove, line 222: Scenario G never dumps diagnostics when it fails. `await_replacement` in scenario F calls `cluster.dump_diagnostics()` before it panics, which is right, but G's three failure points (`running_job_id`, `await_any_stage_running`, and this 120s timeout) all just panic, and the namespace is deleted on drop immediately afterwards. A CI failure there leaves nothing behind to look at.

Worth a dump on at least this timeout, since it is the assertion the whole PR exists for.

chaos-testing/src/rest.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use serde_json::Value;
+use std::time::{Duration, Instant};
+
+/// Per-request timeout. Short so a stalled connection (e.g. a k8s port-forward
+/// that dropped) surfaces as a retryable error inside a polling loop rather than
+/// hanging the whole wait; harmless for the loopback (process) backend.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+fn client() -> Result<reqwest::Client, String> {
```
andygrove, line 35: Nit: this builds a fresh `Client`, and so a fresh connection pool, on every request. `await_any_stage_running` polls every 50ms for up to 60s, so that is roughly 1200 new connections through the port-forward over one wait.

Not a regression, since `reqwest::get` did the same thing before, but now that it is all in one place a `OnceLock<Client>` would fix it once and take a bit of pressure off the forward, which is also relevant to the retry comment above.

chaos-testing/tests/k8s.rs
```diff
@@ -27,87 +27,278 @@
...
+}
+
+/// Poll until the scheduler lists exactly two executors and at least one is not
+/// in `before` — i.e. the killed executor has been reaped and its rescheduled
+/// replacement (new id) has registered.
+async fn await_replacement(cluster: &K8sCluster, before: &[String]) -> Vec<String> {
+    let deadline = Instant::now() + Duration::from_secs(120);
+    loop {
+        if let Ok(ids) = cluster.executor_ids().await
+            && ids.len() == 2
```
andygrove, line 291: Nit: the expected count of 2 is hardcoded here while the caller already knows it from `before.len()`. Passing it in would let this helper serve a scenario with a different executor count later.

andygrove (commented):
Thanks for pushing these onto the cluster backend. The `rest.rs` dedupe is a good call, and these are the right two scenarios to run on kind rather than the process harness.

I checked the branch out locally. `cargo fmt --all -- --check` is clean, and `cargo clippy -p ballista-chaos --all-targets` is clean both with and without `--features k8s`. No public API surface in the four `missing_docs` crates is touched, no proto changes, and nothing in the planner, execution plans, physical optimizer, execution graph, or executor task loop, so no `api change` label and no benchmark numbers needed here.

Four things inline. The first one is the substantive one, since it affects whether scenario G is really covering the #2029 path. The rest are smaller. Nothing here is a blocker.

_Disclosure: I put this review together with LLM assistance (Claude Code). The branch was checked out and the lint runs above were actually executed._

PR #2430, opened 2026-09-07
feat: shuffle-affinity task distribution policy

### LINKED ISSUE #2319: Shuffle affinity: prototype a task distribution policy that places consumer tasks where their input already lives

Consumer tasks are assigned to executors without regard for where their input already lives. The scheduler's built-in distribution policies are `Bias` (fill each executor's slots eagerly) and `RoundRobin` (spread evenly), neither of which looks at the `PartitionLocation`s the task will read. A reduce task therefore frequently lands on an executor that holds none of its input and fetches every block over Arrow Flight, even when one executor already holds most of it on local disk.

`ShuffleReaderExec` already distinguishes the two paths, reading directly from the filesystem when a partition is local and going over Flight otherwise (unless `ballista.shuffle.force_remote_read` is set), so the win from placing the task well is immediate: local reads instead of network round trips, less pressure on the shuffle governor, and fewer connections to the producing executors.

An affinity-aware task distribution policy: when binding a task that reads shuffle input, prefer an executor that already holds the largest share of that task's input bytes, subject to available slots, and fall back to the current behavior when no candidate has a meaningful share or the preferred executor is saturated.

This does not need a core change to prototype. Task distribution is already pluggable via `TaskDistributionPolicy::Custom(Arc<dyn DistributionPolicy>)` (#1243), so an affinity policy can be implemented and measured against `Bias` and `RoundRobin` before deciding whether it belongs in the scheduler as a built-in option.

Open questions worth answering in a prototype:

- How much locality is actually available in practice? With N producer tasks spread across the cluster, a reduce partition's input is usually spread too, so the best candidate may only hold 1/N of the bytes. Affinity likely matters most at low executor counts, after AQE coalescing, and for pass-through (non-repartitioning) stage boundaries.
- What is the cost of the resulting slot idling? Waiting for the preferred executor's slot trades network for latency, so the policy needs a bound on how long it will hold out.
- Interaction with straggler behavior and with `ballista.shuffle.reader.max_bytes_in_flight`.

- Leave placement alone and make remote reads cheaper (`IO_BLOCK_TRANSPORT`, client pooling, larger blocks). Already partly done and complementary, but it cannot beat not making the request at all.
- Reduce the number of blocks per task instead, via a cheaper shuffle format (#660). Also complementary.

**Additional context**

- Discussed on #2308; suggested by @milenkovicm, who noted the pluggable policy hook makes this cheap to prototype.
- Design rationale for the current model: `docs/source/contributors-guide/shuffle.md`.

NoahKusaba:

Closes #2319

# Rationale for this change

A shuffle map task writes its output to the local disk of the executor that ran it. `bias` and `round-robin` place reduce tasks by free vcores alone, so a reduce task often fetches its input over Arrow Flight from an executor that already had it on disk. The scheduler already knows where the bytes are: each `PartitionLocation` names the executor that wrote a partition and how many bytes it wrote.

- `ShuffleAffinityPolicy` in the `examples` crate, installed through `TaskDistributionPolicy::Custom`. Each pending partition is offered to the three executors with free vcores that hold the most of its bytes, largest holdings first. The rest is bound bias-style, so no vcore idles. A single-task stage goes to its largest holder with room.
- Locality reporting through `ShuffleAffinityPolicy::stats`, an optional `LocalityObserver`, and a `debug!` line per scheduling round.
- A runnable example (`examples/examples/shuffle-affinity.rs`), usage docs in the user guide and examples README, and a design entry in the contributors guide.

The scheduler crate is unchanged; the policy copies the four scheduler rules it depends on.

# Evaluation

Share of shuffle bytes read on the executor running the task (`locality_benchmark_against_bias_and_round_robin`; free vcores are `executor_1` / `executor_2`):

| Scenario           | Free vcores | Bias | Round-robin | Affinity |
| ------------------ | ----------: | ---: | ----------: | -------: |
| Skewed partitions  |       8 / 8 |  50% |         50% |      90% |
| Skewed partitions  |      4 / 16 |  50% |         50% |      90% |
| Uniform partitions |       8 / 8 |  50% |         50% |      50% |
| Global aggregate   |      1 / 16 |  10% |         10% |      90% |

These are placement scores on synthetic layouts, not a speedup, and nothing has run on a cluster.

## Where it can help

Only where a partition's bytes are unevenly spread. An even hash shuffle leaves `1/E` of each partition on every executor, so every policy reads the same share. In the TPC-H plans in `tpch_plan_stability`, which assume evenly spread data, no query should change: single-task stages read already-reduced data, and every other stage reads an even hash shuffle.

Real data can be uneven. In the SF1000 plans from #2419 (date-partitioned Iceberg), a date filter leaves only a few file groups, so filtered scans are written by a few tasks:

- q8 s5 is a single task reading 80M rows, all written by one task.
- q5, q10, q14, q15 and q20 each have a stage where one executor holds at least 21–61% of a filtered scan's output.

Those stages take 1–7s of compute, while wall-clock time is dominated by stages reading Parquet, and the event log has no per-task executor ids or shuffle read metrics. A rerun under bias and affinity with those metrics, starting with q8, q10, q14, q15 and q20, would show whether it matters.

## Open questions from #2319

- **Available locality.** Holdings are summed per executor, so an even shuffle offers `1/E`, not `1/N`, and nothing to gain at any executor count. Non-repartitioning stage boundaries only exist under merges, so they feed single-task stages. AQE coalescing (off by default) averages holdings rather than concentrating them.
- **Slot idling.** None: the policy never waits, so a busy holder costs locality rather than latency.
- **Stragglers and `max_bytes_in_flight`.** There is no speculative execution, and a slow holder only receives tasks while it has free vcores. The fetch governor is per reader stream and covers only remote fetches. Local files are read one at a time while remote fetches run concurrently, so higher locality is not automatically faster.

# Follow-up: waiting for the best executor

I opted to have remaining pending tasks drain into any free executor, after the affinity task distribution, to reduce complexity. This makes it more in-line with being "bias, but better." 

A bounded wait, like Spark's `spark.locality.wait`, needs:
1. **A timer (core).** Binding runs only on events: job submission, task status, an executor registering or being lost, or a failed launch. A held partition would wait for the next event, not its deadline. `job_resubmit_interval_ms` is defined but never read.
2. **A hook (core).** `bind_tasks` cannot ask to run again. One option is a default trait method, `revive_after() -> Option<Duration>`, answered with a delayed `ReviveOffers`.
3. **Policy changes.** Track when each partition was first held, per stage attempt, and tell held partitions apart from unplaceable ones: any leftover partition currently ends the round for every remaining job.

# Help wanted: benchmarking on a multi-executor cluster

Placement scores can't show whether this lowers latency. If you have a cluster with several executors, ideally on separate hosts and with data partitioned like #2419, a comparison would settle it.

1. **Baseline:** run `ballista-scheduler --task-distribution bias`.
2. **Candidate:** run `cargo run --release --example shuffle-affinity -- --external-host <scheduler-host>`, and start executors with `--scheduler-host <scheduler-host>`. `--bind-host` defaults to `0.0.0.0`, as in `ballista-scheduler`. To capture plans in the same format as #2419, set `with_event_log_dir` on the example's `SchedulerConfig`.
3. **Workload:** for each scheduler, run `tpch benchmark ballista --host <scheduler> --port 50050 --path <data> --format parquet --iterations 3 --output <dir>`, then `tpch compare baseline.json candidate.json`. Start with q8, q10, q14, q15 and q20, with q1 and q6 as controls.
4. **Please report:** cluster shape (executors, vcores, network), per-query wall-clock time, the policy's local byte ratio from the scheduler log, and the executors' shuffle reader `local_partitions`, `remote_partitions` and fetch time.

Documentation only. There is no API change and no change to default behaviour: the policy is opt-in and lives outside the scheduler.


milenkovicm (commented):
thanks @NoahKusaba, question, can this be implemented as `TaskDistributionPolicy::Custom` policy rather than being part of ballista main ?

NoahKusaba:
Give me a few days to re-do everything.

NoahKusaba:
@milenkovicm you don't need to review it fully, but can you let me know if this is the direction you were looking at?

PR #2431, opened 2026-09-07
feat(history): replace the timer-based rescan with a filesystem watch

milenkovicm:

Closes #.

# Rationale for this change

The standalone history server kept its job index current by re-walking theevent-log directory on a fixed timer (`--update-interval-seconds`, default 10s). That has two problems:

- **Latency vs. cost trade-off is hard-coded.** A short interval means a  `stat` per file per tick over a directory that can hold thousands of  finished jobs; a long interval means a finished job is invisible for up to
  that long.
- **The source is assumed to be a flat local directory.** All listing and  reading was inline `std::fs`, so there was no seam for an alternative  backend (object store, archived snapshot, remote listing) or an  alternative "when to refresh" policy.

This change pulls "where logs come from" and "when to act on them" behind two traits, and wires in a `notify`-based directory watch so a completed log is indexed the moment it is renamed into place rather than on the next tick.

**New `history/source.rs` — `EventLogSource` trait**
- `scan_jobs() -> BoxStream<io::Result<PathBuf>>`, `read_job_index`,  `read_completed_job`.
- `LocalDirSource` is the only impl and carries over the previous flat-dir  `std::fs` logic unchanged (including the `.eventlog` vs `.eventlog.running`  extension filter).

**New `history/trigger.rs` — `ScanTrigger` / `EventLogTrigger` traits**
- `ScanTrigger::scan_tick` decides when a full directory pass runs;  `OnceTrigger` fires exactly once then parks, `NoopTrigger` never fires.
- `EventLogTrigger::next_eventlog` reports individual `*.eventlog` files as  they appear or get deleted; `NotifyTrigger` watches the directory with the `notify`  crate, treating both create and rename-into-place (`Modify(Name(..))`) as  "a new file appeared", filtering by the `eventlog` extension.
- Unit tests for `NotifyTrigger`: create, arrive-by-rename, and  ignore-running-logs-and-unrelated-files.

**`history/mod.rs` — `HistoryStore` rework**
- `HistoryStore::load` + `refresh` + `RefreshStats` + `FileStamp` + the  `seen` map are removed.
- `HistoryStore::new(dir)` sets up the `NotifyTrigger` watch and builds the  initial index synchronously; `dir` must already exist.
- `HistoryStore::new_static(dir)` builds a frozen snapshot (both triggers  noop) for tests and one-shot use over an archived directory.
- `spawn_refresh_task` becomes `spawn_service_tasks(store) -> ServiceTasks`,  which runs a watch loop (`index_one` per file from the trigger) and a  scan loop (full `load_index` pass per `scan_tick`). `ServiceTasks` aborts  both loops on drop.
- Indexing is now **upsert-only**: a log that disappears or is rewritten  under a new job id is no longer reconciled out of the index.
- `read_job` is now `async` and delegates I/O to the source;  `read_job_blocking` is gone, replaced by `impl From<JobReadError> for  SchedulerErrorResponse`.
- The timer/refresh/removal test suite is deleted; remaining tests move to  `new_static` / `.await`.

**`bin/history_server.rs`**
- Drops the `--update-interval-seconds` arg.
- `create_dir_all`s the event-log directory before `HistoryStore::new`  (the watch cannot be placed on a missing path).

**Dependencies**
- Adds `notify = "8"` as a workspace dependency, wired into the scheduler  crate's `rest-api` feature as optional `dep:notify`.
- `Cargo.lock` updated accordingly.

Yes, all behind the `rest-api` feature:

- **`--update-interval-seconds` is removed** from the `history_server`  binary. Refresh is now a single startup pass plus an event-driven  filesystem watch.
- The history server now **creates the event-log directory if it does not  exist** instead of treating a missing directory as empty.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,6 +1243,7 @@ dependencies = [
  "insta",
  "itertools 0.15.0",
  "log",
+ "notify",
  "object_store",
  "once_cell",
  "parking_lot",
@@ -3445,6 +3446,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "funty"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4114,6 +4124,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "inotify"
+version = "0.11.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4cc00ea907cab49550b7da656f80ebb97be1b997d931fbcd28d39734e17ce592"
+dependencies = [
+ "bitflags 2.13.1",
+ "inotify-sys",
+ "libc",
+]
+
+[[package]]
+name = "inotify-sys"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c033f80b2c113cdf91ab7a33faa9cbc014726dcad99880c8609af2a370edf37d"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "insta"
 version = "1.48.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


andygrove:
Nice cleanup of the internals, the source and trigger split reads well and the notify tests are thorough.

The one thing I would like to work through before this lands is dropping the periodic pass entirely. The old polling loop had a comment explaining why it polled rather than watched: history servers are usually pointed at a directory on a shared or network filesystem, written by schedulers on other hosts, and inotify does not see remote writes. With `OnceTrigger` as the only scan trigger, that deployment gets a job list frozen at startup with no error to explain it.

Could we keep both? Watch for the low latency path, plus a periodic trigger as a backstop, with `--update-interval-seconds` retained to tune it and `0` meaning watch-only. A longer default than 10s makes sense now that notify carries the common case. That also covers the coarse-backend case your own comment in `trigger.rs` mentions, where a missed removal currently has no next full scan to wait for.

Two smaller things. `HistoryStore::new` does not actually call `load_index`, so the rustdoc on it and the comment in `history_server.rs` about a synchronous initial walk are describing something that does not happen, and the store is empty until the scan loop runs. And `docs/source/user-guide/history-server.md` still documents the interval flag, the stat-per-file cost, and deletion reconciliation on the next pass, so that section needs updating with this.

Since `history` is a public module, `load`, `refresh`, `RefreshStats` and `spawn_refresh_task` going away is a breaking change for anyone embedding the scheduler crate. Could you add the `api change` label and a short entry in `docs/source/upgrading/55.0.0.md`? Several of those could stay as deprecated wrappers too, which would soften it.

Last one, the rationale sells the traits as a seam for object store and remote backends, but they are `pub(crate)` and there is no constructor taking them, so an embedder cannot plug one in yet. Worth either exposing them or trimming that from the description.

milenkovicm:
thanks @andygrove will address your comments 

> Last one, the rationale sells the traits as a seam for object store and remote backends, but they are pub(crate) and there is no constructor taking them, so an embedder cannot plug one in yet. Worth either exposing them or trimming that from the description.

I will need another pass to support object stores as PathBuf should be changed for object store equivalent

milenkovicm:
> Since history is a public module, load, refresh, RefreshStats and spawn_refresh_task going away is a breaking change for anyone embedding the scheduler crate. Could you add the api change label and a short entry in docs/source/upgrading/55.0.0.md? Several of those could stay as deprecated wrappers too, which would soften it.

I believe we haven't released this yet, hence no upgrade path needed

PR #2434, opened 2026-09-08
feat(scheduler): stage a join's build side before deciding the join

andygrove:

Closes #.

Follows up on the q8 analysis in #2419: https://github.com/apache/datafusion-ballista/pull/2419#issuecomment-5591677739 (items 2 and 3). No issue filed yet, happy to open one if this direction looks right.

# Rationale for this change

TPC-H q8 at SF1000 takes 164s, and 131s of that is a single stage: `SortShuffleWriterExec Hash([l_partkey], 256)` over the raw lineitem scan. All 6B rows of lineitem are shuffled across the network in order to join against 1,332,379 rows of filtered `part`, which is one `p_partkey` i64 column, roughly 10.7 MB, about 12x under the 128 MiB broadcast budget. That shuffle should not exist.

It exists because AQE decides the join before it can measure anything. The build side is `FilterExec(p_type = 'ECONOMY ANODIZED STEEL')` over `part`, and with no runtime statistics the planner falls back to `default_filter_selectivity` (20%). The resulting estimate sits over the broadcast budget, so `to_actual_join` picks `PartitionMode::Partitioned` and emits `JoinSelectionAction::Repartition`, which wraps **both** children in an `ExchangeExec`. The 131s is committed at that point.

Worse, it is unrecoverable. Stage 0 finishes in 1.4s and reveals the true size, but by then `selection_state == Repartitioned`, so the only arms `to_actual_join` can still reach are Hash and SortMerge. There is no path back to `CollectLeft`.

The same shape shows up in q9 s1 (58s), and the estimate/measurement gap is not specific to q8: any filtered dimension scan feeding a fact-table join hits it.

## Stage the build side before deciding (`JoinSelectionAction::StageBuildSide`)

A new action that shuffles only the prospective build side, on the join key, and leaves the probe side untouched. `selection_state` stays `Unknown`, so when `upstream_resolved()` goes true after that one cheap stage, `to_actual_join` runs again — this time against a resolved `ExchangeExec`, whose `partition_statistics` reports `Precision::Exact` row and byte counts read back from the shuffle's `PartitionStats`. For q8 that is 1.33M rows and a few MB, which promotes the join to `CollectLeft`. The probe side is then scanned once inside the join stage against a broadcast build side, and stages 1 and 2 (142s combined) collapse into roughly one 12s stage.

`ExchangeExec::to_broadcast` shares the `stage_id` and `shuffle_partitions` `Arc`s, so the staged shuffle is *read* as a broadcast rather than re-run.

`requires_build_staging` gates this on four conditions, each necessary:

- **the join type must be able to broadcast at all**, per `collect_left_broadcast_safe` on the post-swap join type. `partition_mode` is `CollectLeft` only when that same predicate holds, so for a `Full` join, or a left-sided semi/anti/outer join, no measurement can change the outcome — staging would buy a serialised stage boundary and no decision.
- **the build estimate must be `Inexact`.** An exact size over budget is a fact, not a guess, so measuring it again cannot flip the join and would only serialise two shuffles that could run concurrently.
- **the estimate must be at most `stage_build_side_max_estimate_multiple` over the byte budget.** Past that, no measurement brings it back under. q8's `part` scan estimates ~10x over and measures three orders of magnitude under.
- **the probe side must be at least `stage_build_side_min_probe_ratio` times the build side in bytes.** Staging serialises the two shuffles, so the deferral costs at most the *smaller* side's runtime; the ratio is what bounds the cost of being wrong.

It fires at most once per join, guarded on neither child already being an `ExchangeExec` — once it fires, one child is, so the guard can never pass again. When the measured size then turns out to be too large to broadcast, the join falls through to `Repartition`, which reuses an exchange already shuffling on the join key (`SelectJoinRule::ensure_exchange`) instead of nesting a second stage boundary inside the first.

The swap decision cannot rescue a non-broadcastable join type either: `supports_swap_join_order` is `left_bytes > right_bytes`, so a build side that is already the left input is the smaller one, and measuring it only shrinks it further.

## Configuration

Three keys, all consulted only when staging is enabled, and only for build sides whose size is an estimate rather than a measurement:

- `ballista.optimizer.stage_build_side` (Boolean, default `true`)
- `ballista.optimizer.stage_build_side_min_probe_ratio` (UInt64, default `10`)
- `ballista.optimizer.stage_build_side_max_estimate_multiple` (UInt64, default `32`)

`stage_build_side_ships_the_documented_defaults` pins the shipped values.

## Why one-sided shuffling is safe

Worth spelling out, since a half-shuffled join is obviously wrong if it ever executes. It never does, for two independent reasons.

*The intermediate plan is never executed.* `DynamicJoinSelectionExec::execute` is an error, and stages come only from `ExchangeExec` subtrees (`find_runnable_exchanges`). In the staged state the sole unresolved exchange is the build side's, whose subtree is the filtered dimension scan, so that is the only stage created. The probe side contains no exchange and is never scheduled; the join node sits *above* the exchange, not below it, so it is in no stage either.

*Every executable outcome satisfies the join's distribution requirement.* A `Partitioned` hash join or a `SortMergeJoin` is reachable only through the `JoinInputState::Repartitioned` arms of `to_actual_join`, and that state comes from exactly two places: `to_partitioned()`, which inserts both exchanges, or `inputs_already_partitioned()`, which verifies both children satisfy `KeyPartitioned` exactly. Staging sets neither. The two states it can reach are `CollectLeft`, which replicates the whole build side to every probe task and needs no co-partitioning, or a fallthrough to `Repartition`, which shuffles the probe side too. There is no arm that produces a co-partitioned join from a half-shuffled input.

The `CollectLeft` outcome leaves the build side with the staged exchange's partition count, which DataFusion's `CollectLeft` mode does not accept on its own. `SelectJoinRule` converts it with `to_broadcast` when the probe is partitioned, and `EnsureRequirements` gathers it with a `CoalescePartitionsExec` when it is not.

## Tests

Unit coverage in `dynamic_join` for `requires_build_staging` across every condition, `to_actual_join` returning `StageBuildSide` for the q8 shape, `Repartition` when the config is off, and `Repartition` rather than a second staging once a child is an exchange. `stages_exactly_the_join_types_that_can_broadcast` loops **all ten** join types and compares against `collect_left_broadcast_safe` as the oracle, so a join type added to DataFusion cannot slip through untested.

`state/aqe/test/stage_build_side.rs` drives `AdaptivePlanner` through **both passes** against declared statistics:

- `stages_the_build_side_then_broadcasts_the_measured_result` — pass one shows an exchange on the build side and a bare scan on the probe, the single stage is the build side alone, and after the shuffle reports its measured size pass two resolves to `HashJoinExec: mode=CollectLeft` over `broadcast=true`, with the fact table still never shuffled.
- `reuses_the_staged_exchange_when_the_build_side_measures_too_large` — the fallthrough, asserting exactly one exchange per side (the build's being the resolved staged one) rather than a nested pair.

`cargo test -p ballista-scheduler --lib`: **410 passed, 0 failed.** Clippy and rustfmt clean.

## Review rounds

Two review passes are folded in.

The first added the join-type guard above, which was a real gap: `Full` was affected unconditionally (`Full.swap() == Full`), and `Left`/`LeftSemi`/`LeftAnti`/`LeftMark` whenever the build side was already the left input.

The second was a reuse and simplification pass with no correctness findings, taken essentially in full for a net **−232 lines**: the `StageBuildSide` payload dropped (the rule already visits that node) and `JoinSide` replaced with a `build_is_right: bool` that makes the unreachable variant unrepresentable; the two `(Unknown, Partitioned)` arms collapsed into one; `with_selection_state` removed as a no-op; `BuildStagingLimits` replaced with plain params to match the sibling predicates; a dead clone dropped from `join_key_partitioning`; `exchange_on` renamed to `ensure_exchange` and made a method. On the test side, six single-assert staging tests folded into two, the two hand-partitioned join-type lists merged onto the `collect_left_broadcast_safe` oracle, `q8_shaped_action*` un-forked from `run_to_actual_join`, and the duplicated session/schema/register fixtures hoisted into `test/mod.rs`.

Only the four `stage_build_side.rs` snapshots changed across that pass, with identical plan shapes and `plan_id`s — which is the evidence it is behaviour-preserving.

## Split out of this PR

The `broadcast_join_threshold_rows` semantics change originally bundled here now lives in **#2444**. It is not what fixes q8 — at pre-execution q8's build side has an `Inexact` `total_byte_size`, so the byte branch short-circuits and the row path is never reached — so the two are easier to review and revert independently.

## Benchmarks

Measured on a cluster at TPC-H **SF1000**, 32 executors on `r6i.24xlarge`, Parquet on S3, AQE on: the 22-query suite went from 817.07s to 696.09s, with q8 2.36x, q9 2.08x and q21 1.46x.

**Those numbers are not being landed in the docs by this PR, and should not be read as final.** They were measured before the join-type guard, so they include stages that were staged and discarded on anti and outer joins. They are also one iteration per query: Q1 and Q6 are both join-free, and one improved 4% while the other regressed 7%, so the noise floor is around ±7% — wide enough to explain every small delta in the set, though not the three headline wins, which are an order of magnitude larger.

`benchmarking.md` is therefore untouched here. I will re-measure with the guard in place and land a docs commit separately after merge, stating the iteration count and the regression band rather than rounding them away.

## Known follow-ups, not addressed here

- The staged side is written as a hash shuffle into `target_partitions`, so the broadcast read pulls all of them per probe task. Correct, and cheap at 1.33M rows, but an unpartitioned write would be better.
- The measured `total_byte_size` is compressed shuffle-file bytes, which understates the in-memory build size. Pre-existing: `hash_join_max_build_partition_bytes` already reads the same figure.
- The exchange guard inspects only immediate children, so a build side shaped like `FilterExec(ExchangeExec(..))` can still gain a boundary above an existing one. The fire-at-most-once bound is unaffected, since staging always produces a direct exchange child.
- `to_actual_join` builds five independent `StatisticsContext`s per call, each with a fresh cache. Pre-existing; this PR adds no new statistics computation. Threading one through is worth its own PR.
- `CoalescePartitionsRule` has no way to *clear* a `CoalescePlan`, and both bail paths return without clearing. Not introduced here and not reachable in the staged shape, since `finalise_stage_internal` replans and re-decides the join before `actionable_stages` invokes the rule, but the invariant is order-dependent rather than structural.

No API changes.

Three new config keys, listed above. With `stage_build_side` on, a plan may gain one extra stage boundary and lose a much larger one, so stage counts and the plan shape for affected joins change. Generated config docs regenerated via `./dev/update_config_docs.sh`.


andygrove:
Looks promising:

```
      "queries": [
        {
          "query_id": "TPCH-8",
          "has_errors": false,
          "has_warnings": false,
          "mean_s": 37.76327075066666,
          "min_s": 36.831046083,
          "max_s": 39.35304816,
          "median_s": 37.105718009,
          "stdev_s": 1.3836203688704116,
          "iterations": 3
        }
      ]
```

comphead:
Reviewed this against the PR head (`1340cc97`). Baseline is green for me: 416/416 in `cargo test -p ballista-scheduler --lib`.

First, the part you asked people to push on — **the one-sided shuffle safety argument holds.** I traced it rather than taking it on faith: `find_runnable_exchanges` only pushes `ExchangeExec` nodes as stages and stops descending at a resolved one, so in the staged state the probe subtree creates no stage, and the join sits above the exchange rather than inside its subtree. `Repartitioned` is reachable only via `to_partitioned()` or `inputs_already_partitioned()`, and staging sets neither. `to_broadcast` sharing `stage_id`/`shuffle_partitions` checks out too.

Two things I'd want changed before merge.

### 1. Staging fires for join types that can never broadcast

The `StageBuildSide` guard in `to_actual_join` checks the config, `!null_aware`, the exchange guard and `should_stage_build_side` — but not `collect_left_broadcast_safe(build_side_join_type)`, which is already computed ~80 lines above.

`partition_mode` is `CollectLeft` only when `under_threshold && collect_left_broadcast_safe(..)`. That second term is a pure function of the join type, so no measurement can change it. For a non-broadcastable type the second pass can't reach `CollectLeft`, hits the (now failing) exchange guard, and falls to `Repartition` — one serialised stage boundary that never had a chance of changing the decision.

I confirmed it with a throwaway test on the q8 shape:

```
join_type=Full     collect_left_broadcast_safe=false  action=StageBuildSide
join_type=Left     collect_left_broadcast_safe=false  action=StageBuildSide
join_type=LeftSemi collect_left_broadcast_safe=false  action=StageBuildSide
join_type=LeftAnti collect_left_broadcast_safe=false  action=StageBuildSide
```

`Full` is affected unconditionally (`Full.swap() == Full`); `Left`/`LeftSemi`/`LeftAnti`/`LeftMark` whenever the build side is already the left input.

One line fixes it:

```rust
 (JoinInputState::Unknown, PartitionMode::Partitioned)
     if bc.stage_build_side_enabled()
         && !self.null_aware
+        // Measuring the build side can only change the outcome if
+        // CollectLeft is reachable at all, which is a property of the
+        // (post-swap) join type alone.
+        && collect_left_broadcast_safe(build_side_join_type)
         && !self.left.is::<ExchangeExec>()
```

Applied locally: **415/415 pass, no changes needed to any test in this PR** — they all use `Inner`.

**On the +3.1s band:** this is a candidate partial explanation, not a diagnosis. q16 (`NOT IN`, which `not_in_subquery_rewrite` turns into a plain anti join) and q22 (`NOT EXISTS`) are both anti joins and both sit in the band. It doesn't explain q5/q7/q12, which are plain inner joins — those look more like the serialisation cost you already acknowledge. Re-running the six with `stage_build_side=false` per query would separate the two.

### 2. Nothing tests the multi-pass path

All the new coverage is single-pass: `should_stage_build_side` in isolation, one `to_actual_join` call, `exchange_on`. The mechanism this PR is actually about — stage → resolve with measured stats → re-decide → `CollectLeft` — isn't exercised, and neither is the `EnsureRequirements`/`CoalescePartitionsExec` behaviour you say you verified by hand.

The harness for exactly this already exists in `aqe/test/join_selection.rs` (`AdaptivePlanner` + `mock_partitions_with_statistics` + `assert_plan!`). Two `assert_plan!` tests would cover it: the q8 shape resolving to a broadcast `CollectLeft`, and the fallthrough where the measured size is still too large and `exchange_on` reuse yields one exchange per side rather than a nested pair.

Worth noting `test_left_join_not_collected_left` (the #1055 guard) passes today only because its MemTable stats don't match the staging shape — nothing currently pins "a Left join must not stage."

### Smaller

- I'd split the `broadcast_join_threshold_rows` semantics change out. It's well argued and clearly flagged, but by your own account it isn't what fixes q8, and it replaces a test that asserted the old behaviour deliberately. Separate PRs means either half can be reverted alone if the SF1000 numbers move.
- The exchange guard only inspects immediate children, so a build side shaped like `FilterExec(ExchangeExec(..))` gets a second boundary stacked on the first. Edge case.
- `CoalescePartitionsRule`'s docs assume both join legs read upstream shuffle output, and attach the same `CoalescePlan` to every leaf to keep buckets aligned. In the staged state only one leg is an exchange. It's off by default so the risk is low, but does the leaf grouping handle a one-exchange join?

### On defaulting to `true`

Fix #1 first, then yes. As it stands, default-on ships a path that's provably unprofitable for `Full` and left-sided semi/anti joins, which is a bad trade for queries that get nothing back. After the guard I'd have no objection.

Benchmark methodology is good — real cluster, row counts cross-checked, regressions disclosed rather than smoothed. Two caveats worth stating inline: one iteration per query, and the current totals include wasted stages on any anti/outer join matching the shape, so the post-fix number should come in a little under 696.09s.

I did not check what happens to a resolved exchange's `shuffle_partitions` when the executor holding that shuffle is lost. Looks pre-existing for all exchanges rather than something this PR introduces, but staging makes one more stage's output load-bearing.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -212,6 +212,15 @@ impl DisplayAs for DynamicJoinSelectionExec {
...
 
 pub enum JoinSelectionAction {
+    /// Shuffle only the prospective build side, leaving the probe side
+    /// untouched, and re-decide the join once the build side's *measured* size
+    /// is known. See [`should_stage_build_side`].
+    StageBuildSide {
+        join: Arc<DynamicJoinSelectionExec>,
+        /// Which child to give an exchange: `true` for `left`, `false` for
+        /// `right`.
+        build_is_left: bool,
```
comphead: perhaps having enum BuildSide::Left, Right would be more explanatory and this is generic enum that can be reused it other parts

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -606,6 +664,82 @@ impl DynamicJoinSelectionExec {
...
+/// * the estimate must be **plausibly wrong enough to flip**. Past
+///   [`STAGE_BUILD_SIDE_MAX_ESTIMATE_MULTIPLE`] the side is large on any
+///   reading, and no measurement brings it under budget.
+/// * the probe side must be **much larger**, per
+///   [`STAGE_BUILD_SIDE_MIN_PROBE_RATIO`], which is what bounds the cost of
+///   being wrong.
+///
+/// Both sides are compared in bytes. A missing probe-side size declines: with
+/// nothing to compare against there is no evidence the round trip pays off.
+fn should_stage_build_side(
```
comphead: Would prefer something like 
`requires_build_staging`, `build_staging_is_neeed`, etc

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -606,6 +664,82 @@ impl DynamicJoinSelectionExec {
...
+
+/// How far over the broadcast budget an *estimated* build side may sit and
+/// still be worth measuring.
+///
+/// TPC-H q8's filtered `part` scan estimates roughly 10x over budget, because
+/// the planner falls back to `default_filter_selectivity` for
+/// `p_type = '...'`, and measures three orders of magnitude under it. A raw
+/// fact-table scan sits far beyond this multiple and is shuffled without the
+/// extra round trip.
+const STAGE_BUILD_SIDE_MAX_ESTIMATE_MULTIPLE: usize = 32;
```
comphead: any chances the user wants to override this without rebuilding from source?

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Agreed the bool was opaque. I went with DataFusion's existing `datafusion::common::JoinSide` rather than defining a new `BuildSide` — it is already the generic Left/Right/None enum for exactly this, and it is already imported elsewhere in this crate (`physical_optimizer/join_selection.rs`), so it is one less type to keep in sync.

The variant is now `build_side: JoinSide`, and `SelectJoinRule` matches on it instead of indexing children with a bool. `JoinSide::None` is rejected with an internal error, since the resolver only ever produces Left or Right.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Renamed to `requires_build_staging`.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Good call — both are config keys now, so they are tunable without a rebuild:

- `ballista.optimizer.stage_build_side_min_probe_ratio` (UInt64, default `10`)
- `ballista.optimizer.stage_build_side_max_estimate_multiple` (UInt64, default `32`)

They are read into a new `BuildStagingLimits` struct alongside `broadcast_join_threshold_bytes`, which is also where the reasoning for each figure now lives. Both are only consulted when `stage_build_side` is enabled.

Config docs regenerated via `./dev/update_config_docs.sh`, and `stage_build_side_ships_the_documented_defaults` pins the shipped values, since those are what the SF1000 numbers were measured under.

andygrove:
Thanks for tracing the safety argument rather than taking it on faith — that was the part I most wanted someone to actually walk, and your reading of `find_runnable_exchanges` and the two `Repartitioned` entry points matches mine.

Both blocking items are addressed, plus the three inline ones. Pushed.

### 1. Join-type guard — fixed

You're right, and it's a real waste rather than a cosmetic one. Applied exactly as you wrote it.

One thing worth adding, because it was the only way I could see the guard being *too* strict: the swap decision can't rescue those cases either. `supports_swap_join_order` is `left_bytes > right_bytes`, so `swap_inputs == false` means the build side is already the smaller input — and measuring it only shrinks it further, since the whole premise is that the estimate overshoots. So the swap can never flip `false → true` and turn a `Left` into a broadcastable `Right` after staging. The guard loses nothing.

Both directions are pinned now, across all ten join types: `does_not_stage_a_join_type_that_can_never_broadcast` and `stages_the_join_types_that_can_broadcast`.

### 2. Multi-pass coverage — added

New file `ballista/scheduler/src/state/aqe/test/stage_build_side.rs`, using the `AdaptivePlanner` + `StatsTable` + `assert_plan!` harness you pointed at. It needed one new fixture helper (`estimated_statistics`, for an `Inexact` `total_byte_size` — `sized_statistics` is `Exact`, which never stages) and a parameterised `mock_partitions_with_size`, with `mock_partitions_with_statistics` delegating to it.

Three tests:

- `stages_the_build_side_then_broadcasts_the_measured_result` — the mechanism end to end. Pass one asserts an `ExchangeExec` on the build side and a **bare scan** on the probe; `actionable_stages()` returns exactly one stage and that stage is the build side alone; then after the shuffle reports its measured size, pass two resolves to `HashJoinExec: mode=CollectLeft` over `broadcast=true`, with the fact table still unshuffled.
- `reuses_the_staged_exchange_when_the_build_side_measures_too_large` — the fallthrough. Asserts one exchange per side, the build's being the resolved staged one (`stage_id=0, stage_resolved=true`) and the probe's fresh, rather than a nested pair.
- `does_not_stage_a_left_join` — pins the thing you noted nothing currently pins. You were right that `test_left_join_not_collected_left` passes only incidentally; its MemTable stats don't match the staging shape. These ones do.

418 passed, 0 failed. Clippy and rustfmt clean.

### 3. Splitting the row-threshold change — done

Now #2444. Agreed on the reasoning: it isn't what fixes q8, and it deliberately replaces a test that asserted the old semantics, so the two should be revertable independently.

### Smaller

**Exchange guard depth.** Acknowledged, and left as is deliberately — I'd rather not widen it without a shape that actually motivates it. The fire-at-most-once bound is unaffected, because staging always produces a *direct* `ExchangeExec` child, which the guard does catch. And in the `FilterExec(ExchangeExec(..))` case the staged measurement is of the filter's output, not the inner shuffle's, so it isn't redundant work — just an extra boundary. A subtree-wide search would also decline legitimately useful cases, like a build side that is itself a join over an already-resolved broadcast exchange. Noted as a follow-up in the description; happy to tighten if you think the edge case is likelier than I'm reading it.

**`CoalescePartitionsRule` with a one-exchange join.** I traced it. Short answer: the leaf grouping handles it fine, and the risk you're pointing at isn't reachable — but it's safe by scheduling order rather than by construction, which I think is worth recording.

The rule is entirely join-agnostic: it collects leaf `ExchangeExec`s with no notion of which leg they feed, and the only cardinality guard is `leaves.is_empty()`. A one-leaf group is already routine (any single-input stage reading one shuffle), and `start_indices_to_partition_groups` covers `[0, M)` exactly, so coalescing one leg changes task count and never row count.

The window you're describing — build leaf carrying a plan from a one-leaf pass while the probe leaf carries none — can't occur, because the rule only runs *after* a full replan. `finalise_stage_internal` calls `replan_stages()`, which runs `SelectJoinRule` and re-decides the join; `CoalescePartitionsRule` runs later, inside `actionable_stages()`. So the instant the staged exchange resolves, the join is already re-decided, and by the time the rule looks the plan is either broadcast (it bails on `any leaf .broadcast`) or two leaves with the probe unresolved (it bails on the unresolved leaf). `m` is also read from `properties().partitioning`, which `set_coalesce` never rewrites, so re-running on an already-coalesced leaf re-derives the same `M` and the heterogeneous-`M` guard doesn't misfire.

The thing I'd raise separately, which this PR doesn't introduce but does sit close to: **nothing ever clears a `CoalescePlan`.** `set_coalesce` is the only writer, and both bail paths return without clearing — including the degenerate-`K` path that says so explicitly in its own debug line. If a leaf ever acquires a plan and the group's byte profile later changes such that the rule bails, the stale `K` persists on one leg while its partner stays at `M`. Only the replan-before-rule ordering keeps that from being reachable here. Clearing the slot on the two bail paths, or asserting the group agrees on `coalesce()` before returning, would make it structural. I'd rather do that in its own PR than widen this one — let me know if you'd prefer it bundled.

**Executor loss and a resolved exchange's `shuffle_partitions`.** Agreed it looks pre-existing for all exchanges rather than something staging introduces, and agreed that staging makes one more stage's output load-bearing. I haven't checked it either; not planning to in this PR.

### Defaults and benchmark numbers

Keeping `stage_build_side` at `true`, per your "fix #1 first, then yes".

Both caveats are now stated inline in the description rather than in a footnote: one iteration per query, and the totals were measured **before** the join-type guard, so they include stages that were staged and thrown away on anti and outer joins. The post-fix total should land slightly under 696.09s. I haven't re-measured on a cluster yet.

Your partial explanation for the +3.1s band holds up and I've written it into the description: q16 (`NOT IN` → anti join via `not_in_subquery_rewrite`) and q22 (`NOT EXISTS`) are both anti joins, and the guard stops both from staging at all now. It doesn't cover q5/q7/q12, which are plain inner joins — those look like the serialisation cost. Re-running the six with `stage_build_side=false` per query to separate the two is the next measurement I plan to take.

comphead:
some s3 tests are failing

comphead:
its the same minio issue observed in [ci: `datafusion-cli` MinIO docker login failed · Issue #25215 · apache/datafusion](https://github.com/apache/datafusion/issues/25215)
https://github.com/apache/datafusion/issues/25215

UPD: https://github.com/apache/datafusion-ballista/issues/2446

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -212,6 +212,16 @@ impl DisplayAs for DynamicJoinSelectionExec {
...
+    /// Shuffle only the prospective build side, leaving the probe side
+    /// untouched, and re-decide the join once the build side's *measured* size
+    /// is known. See [`requires_build_staging`].
+    StageBuildSide {
+        join: Arc<DynamicJoinSelectionExec>,
+        /// Which child to give an exchange. Only ever [`JoinSide::Left`] or
+        /// [`JoinSide::Right`] — the resolver picks the side the join would
+        /// build from, which is never absent.
+        build_side: JoinSide,
+    },
```
comphead, line 225: Two things collapse here.

`join` is a field-identical copy of the node `SelectJoinRule` is already visiting (see the `with_selection_state` call below), so the payload can go and the arm can rebuild from the visited `node`.

`JoinSide` is a three-variant type for a two-state fact, which is what forces the unreachable `JoinSide::None` arm in `join_selection.rs`. The `BuildSide { Left, Right }` I originally suggested would have made that unrepresentable. A plain `build_is_right: bool` does too, and the producer below already computes exactly that as `swap_inputs`.

~13 lines.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -385,6 +395,41 @@ impl DynamicJoinSelectionExec {
...
+                    // decision.
+                    && collect_left_broadcast_safe(build_side_join_type)
+                    // Firing at most once per join is what keeps this from
+                    // ping-ponging: once it fires, one child is an
+                    // `ExchangeExec` and this guard never passes again.
+                    && !self.left.is::<ExchangeExec>()
+                    && !self.right.is::<ExchangeExec>()
+                    && requires_build_staging(
+                        build_stats,
+                        if swap_inputs { &stats_left } else { &stats_right },
```
comphead: Three things in one guard.

- `!self.null_aware` (L400) is already implied. `partition_mode` is `if self.null_aware || (..) { CollectLeft } else { Partitioned }`, so matching `Partitioned` guarantees it.
- The 8-line comment at L401-408 sits inside a match *pattern*. The same rationale is already on `requires_build_staging`, on the join-type test, and in `stage_build_side.rs`. Keep one copy.
- `if swap_inputs { &stats_left } else { &stats_right }` (L417) is the third `swap_inputs` ternary in this function. L292 and L328 already do it for `build_side` and `build_stats`.

Structural: this arm and the fallthrough at L433 have the **identical** pattern, separated only by the guard, so a reader scans 25 lines past the pattern before learning what it matches. Hoisting `probe_stats` next to `build_stats` and folding the conditions into one `should_stage_build_side(..)` predicate collapses both into one arm, and retires the "the caller is responsible for the fourth condition" paragraph at L720.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -385,6 +395,41 @@ impl DynamicJoinSelectionExec {
...
+                        BuildStagingLimits::new(
+                            &bc,
+                            threshold_collect_left_join_bytes,
+                        ),
+                    ) =>
+            {
+                Ok(JoinSelectionAction::StageBuildSide {
+                    join: Arc::new(self.with_selection_state(JoinInputState::Unknown)),
+                    build_side: if swap_inputs {
```
comphead: `with_selection_state(Unknown)` is a no-op here. `JoinInputState` has two variants, and this arm is reachable only when the local `selection_state` (L366-372) is `Unknown`, which requires `self.selection_state == Unknown` already. So this copies all 11 fields and changes nothing.

Combined with dropping the payload, `with_selection_state` can go entirely and `to_partitioned` can return to its inline body. Caller counts today: `to_partitioned` 1, `with_selection_state` 2, one of which is `to_partitioned` itself.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -601,6 +657,107 @@ impl DynamicJoinSelectionExec {
...
+
+impl BuildStagingLimits {
+    fn new(bc: &BallistaConfig, broadcast_threshold_bytes: usize) -> Self {
+        Self {
+            broadcast_threshold_bytes,
+            min_probe_ratio: bc.stage_build_side_min_probe_ratio(),
+            max_estimate_multiple: bc.stage_build_side_max_estimate_multiple(),
+        }
+    }
+}
```
comphead: 37 lines to bundle three `usize`s for one call site, and it diverges from the convention in this same file. `supports_collect_by_thresholds(plan, threshold_byte_size, threshold_num_rows)` (L509), `hash_build_fits(..)` (L766) and `broadcast_build_side_pays_off(..)` (`join_selection.rs:176`) all take plain params. I grepped `ballista/scheduler` and `ballista/core`: this is the only limits bundle in the tree.

It is also internally inconsistent. Two limits come from the `&BallistaConfig` it receives, the third is passed by hand even though it is just `bc.broadcast_join_threshold_bytes()` from L274.

Plain params drop ~31 lines and take the `stage_limits` test helper with them. Worth keeping: the q8 `part`-scan anecdote at L680-684, relocated as one sentence onto `requires_build_staging`. That one is measured evidence rather than prose.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -601,6 +657,107 @@ impl DynamicJoinSelectionExec {
...
+///
+/// * the build estimate must be **inexact**. An exact size over budget is a
+///   fact rather than a guess, so measuring it again cannot change the outcome
+///   and would only serialise two shuffles that could run concurrently.
+/// * the estimate must be **plausibly wrong enough to flip**. Past
+///   [`BuildStagingLimits::max_estimate_multiple`] the side is large on any
+///   reading, and no measurement brings it under budget.
+/// * the probe side must be **much larger**, per
+///   [`BuildStagingLimits::min_probe_ratio`], which is what bounds the cost of
+///   being wrong.
```
comphead: These three bullets map 1:1 onto the three early returns immediately below.

Worth keeping: L698-706 (what `Repartition` actually costs, which is the real *why*) and L724-725 (why a missing probe size declines, not derivable from `else { return false }`).

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -1166,6 +1323,306 @@ mod tests {
...
+    fn sized_stats_exec(total_byte_size: Precision<usize>) -> Arc<dyn ExecutionPlan> {
+        Arc::new(StatisticsExec::new(
+            Statistics {
+                num_rows: Precision::Inexact(1_000),
+                total_byte_size,
+                column_statistics: vec![ColumnStatistics::new_unknown()],
+            },
+            Schema::new(vec![Field::new("k", DataType::Int32, false)]),
+        ))
+    }
```
comphead, line 1279: Fourth `StatisticsExec` builder in this module, alongside `sizeless_stats_exec` (L919), `stats_exec` (L1055) and `stats_exec_rows_only` (L1068). All four wrap the same one-column `Int32` schema and differ only in how `total_byte_size` is filled.

One parameterised pair, with the three existing ones as thin wrappers, would also let the six tests below drop their inline `Statistics { .. }` literals.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -1166,6 +1323,306 @@ mod tests {
...
+            &Statistics {
+                num_rows: Precision::Exact(1_000_000),
+                total_byte_size: Precision::Exact(FACT_PROBE_BYTES),
+                column_statistics: vec![],
+            },
+            stage_limits(0),
+        ));
+    }
+
+    // The shipped defaults are what the benchmark numbers were measured under,
```
comphead, line 1390: Six single-assert tests, ~115 lines, each differing in **one** `Precision` value buried in identical boilerplate.

This file already has the shape for it. `supports_collect(&plan)` (L934) and `is_collected(&action)` (L1162) are predicate helpers, and `known_total_byte_size_still_decides` (L996), `broadcast_threshold_bytes_drives_collect_decision` (L1223) and `to_actual_join_uses_post_swap_join_type` (L1207) each bundle positive and negative cases with explanatory messages.

Add `fn stages(build: Precision<usize>, probe: Precision<usize>, threshold: usize) -> bool` and fold to two or three tests. ~80 lines.

Not suggesting a named-case table harness. The file has none, and adding one would be the churn.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -1166,6 +1323,306 @@ mod tests {
...
+        )
+        .unwrap();
+        bc.set("optimizer.stage_build_side", &stage_build_side.to_string())
+            .unwrap();
+        config.extensions.insert(bc);
+        dj.to_actual_join(&config).unwrap()
+    }
+
+    // The whole point: a guessed-large build side against a fact-table probe
+    // side stages the build alone instead of committing to both shuffles.
```
comphead: This forks `run_to_actual_join` (L1081), documented in-file as *"Core driver: runs the join-strategy decision for the given children."* Same `on: JoinOn`, same 11-field struct literal, same `ConfigOptions` / `BallistaConfig` / `insert` / `to_actual_join` plumbing. Only the keys being `set` differ, so adding a field to `DynamicJoinSelectionExec` now breaks four test constructors instead of fewer.

Give `run_to_actual_join` a `stage_build_side` param and delete the fork, along with the middle link `q8_shaped_action_with_children` (L1486), which exists only to default `join_type` to `Inner` for two callers. ~42 lines.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -1166,6 +1323,306 @@ mod tests {
...
+            );
+
+            assert!(
+                matches!(action, JoinSelectionAction::StageBuildSide { .. }),
+                "{join_type:?} can reach CollectLeft, so measuring its build side \
+                 is worth a staging round trip"
+            );
+        }
+    }
+
```
comphead, line 1445: Two hand-partitioned five-element lists. `to_actual_join_collects_only_broadcast_safe_join_types` (L1176) already solves this exact problem 370 lines up: it loops **all ten** types and compares against `collect_left_broadcast_safe(join_type)` as the oracle.

```rust
for join_type in [/* all ten */] {
    let action = q8_shaped_action_with_join_type(..., join_type, true);
    assert_eq!(
        matches!(action, JoinSelectionAction::StageBuildSide { .. }),
        collect_left_broadcast_safe(join_type),
        "join_type {join_type:?}",
    );
}
```

The merge is sound here: with left `Inexact(100 MB)` and right `Exact(2000 MB)`, `supports_swap_join_order` is `100MB > 2000MB` = false, so the post-swap type equals the declared type in every case.

Add a join type to DataFusion and neither current list fails. The merged form does. It also subsumes `to_actual_join_stages_the_build_side_for_the_q8_shape` above. ~54 lines.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -134,6 +135,43 @@ impl SelectJoinRule {
...
+/// on its keys.
+fn join_key_partitioning(
+    join: &DynamicJoinSelectionExec,
+    partition_count: usize,
+) -> Vec<Partitioning> {
+    join._required_input_distribution()
+        .iter()
+        .map(|d| d.clone().create_partitioning(partition_count))
+        .collect()
+}
```
comphead: The clone here is dead. `_required_input_distribution()` returns an **owned** `Vec<Distribution>` (`dynamic_join.rs:236`), and `create_partitioning(self, ..)` **consumes** it (`datafusion-physical-expr-55.0.0/src/partitioning.rs:713`). `.into_iter()` removes two `Vec<Arc<dyn PhysicalExpr>>` allocations and 2N refcount round-trips.

While here: the `StageBuildSide` arm builds this two-element `Vec` and uses exactly one element, then `exchange_on` clones the `&Partitioning` again at L170. Returning `(Partitioning, Partitioning)` and taking it **by value** in `exchange_on` removes all three copies.

The clone predates the PR, but extracting the helper is the moment to drop it.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -134,6 +135,43 @@ impl SelectJoinRule {
...
+        && exchange.partitioning.as_ref() == Some(partitioning)
+    {
+        return child;
+    }
+    Arc::new(ExchangeExec::new(
+        child,
+        Some(partitioning.clone()),
+        plan_id(),
+    ))
+}
```
comphead: Not a defect first: the `impl FnOnce() -> usize` laziness **is** load-bearing, since `plan_id()` is `fetch_add` on an `AtomicUsize` (L104). Do not swap it for a plain `usize`.

But a method receiver gives the same laziness without the generic. `fn exchange_on(&self, child, partitioning)` calling `self.plan_id()` inline matches every other id consumer in this file (L220, L247, L308, L314, L360). `SelectJoinRule` derives `Default`, so the three tests cost one line each.

On the name: in DataFusion vocabulary "on" means join keys (`JoinOn`, and the `on=[(id@0, id@0)]` in this PR's own snapshots). `exchange_on(child, partitioning)` reads as "exchange on the join keys" but means "ensure an exchange with this partitioning". `ensure_exchange` is unambiguous and matches the `add_sort_above` convention.

One more: `DynamicJoinSelectionExec::inputs_already_partitioned` (`dynamic_join.rs:552`) answers the same "is this input already shuffled the way the join needs" question using the equivalence-aware `Partitioning::satisfaction`, where this uses struct equality. They agree today. If the plainer comparison is deliberate, one line in the doc saying so would stop someone unifying them later.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -294,31 +332,64 @@ impl PhysicalOptimizerRule for SelectJoinRule {
...
+                                let build_idx = match build_side {
+                                    JoinSide::Left => 0,
+                                    JoinSide::Right => 1,
+                                    JoinSide::None => {
+                                        return Err(DataFusionError::Internal(
+                                            "StageBuildSide requires a build side"
+                                                .to_owned(),
+                                        ));
+                                    }
+                                };
```
comphead: Unreachable by construction. The producer computes this from a `bool` two lines earlier, and the field's own doc says it is "never absent". Switching the variant to `build_is_right: bool`, or to a two-variant `BuildSide`, deletes this arm, the error string and the three-line doc, and drops the `JoinSide` import from both files.

If `JoinSide` stays: the repo precedent at `physical_optimizer/join_selection.rs:517` is `unreachable!()`, and the sibling arms in this same match use `DataFusionError::Execution`, not `Internal`.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -294,31 +332,64 @@ impl PhysicalOptimizerRule for SelectJoinRule {
...
 
-                                let right = Arc::new(ExchangeExec::new(
-                                    right,
-                                    Some(partitioning[1].clone()),
+                                let mut children =
+                                    vec![join.left.clone(), join.right.clone()];
+                                children[build_idx] = Arc::new(ExchangeExec::new(
+                                    children[build_idx].clone(),
+                                    Some(partitioning[build_idx].clone()),
```
comphead: Clones a `Vec` element it immediately overwrites. `std::mem::replace(&mut children[build_idx], staged)`, or building the exchange first and placing it, reads as intended. Cosmetic, but as written it looks like a slip.

comphead (commented):
Ran a reuse / simplification / efficiency / altitude pass over the diff, following up on my earlier review. **No correctness findings.** The one-sided shuffle argument still holds and I think the mechanism is at the right depth. Everything below is quality only, and none of it is blocking.

Rough total if all of it is taken: **~340 lines, out of ~917 lines of new Rust** (~90 production, ~250 test).

Best value-to-effort, all mechanical and behaviour-preserving:

- fold the six `requires_build_staging` tests (~80)
- un-fork `q8_shaped_action*` from `run_to_actual_join` (~42)
- merge the two join-type loops onto the existing `collect_left_broadcast_safe` oracle (~35)
- `BuildStagingLimits` to plain params, matching the three sibling predicates in the same file (~31)
- `mock_partitions_with_statistics_no_data` to `mock_partitions_with_size(0, 0)` (~21)

Two things I checked and am **not** asking for:

- Dropping `stage_build_side_max_estimate_multiple`. A *filtered* fact-table scan reports `Inexact` and can sit orders of magnitude over budget, so the `Inexact` gate does not already cover it. The multiple does real work.
- Turning the two knobs into constants. House style is config keys (all eight pre-existing numeric AQE knobs are), and I was the one who asked for them to be tunable.

Not introduced here, noted for later: `to_actual_join` builds five independent `StatisticsContext`s per call (L288, L306, L324, L326, plus one inside `supports_collect_by_thresholds`), each with a fresh empty `StatsCache`, so each child subtree is walked about 2.5 times with no memo sharing. To the PR's credit it adds zero new statistics computation and reuses what is already there. It just adds a fourth consumer, which makes this a natural moment to thread one context through.

Tiny, and invisible to CI: `dynamic_join.rs:901`'s `use datafusion::common::{ColumnStatistics, Statistics, stats::Precision};` is now fully shadowed by `use super::*`, since the parent import gained `Statistics` and `Precision`. rustc does not warn on an explicit `use` over a glob, so `clippy -D warnings` will not catch it.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -294,31 +332,64 @@ impl PhysicalOptimizerRule for SelectJoinRule {
...
+                            JoinSelectionAction::Repartition(dynamic_join) => {
+                                let partitioning = join_key_partitioning(
+                                    &dynamic_join,
+                                    config.execution.target_partitions,
+                                );
+
+                                // A build side already staged by
+                                // `StageBuildSide` carries an exchange on the
+                                // join key, so reuse it rather than shuffling
+                                // that side a second time.
```
comphead: This says what `exchange_on`'s own doc at L153-157 already says, where the behaviour actually lives. The same rationale appears a third time at L475-478.

ballista/scheduler/src/state/aqe/optimizer_rule/join_selection.rs
```diff
@@ -391,6 +462,69 @@ mod tests {
...
+    #[test]
+    fn exchange_on_wraps_a_non_exchange_child() {
+        let schema = Schema::new(vec![Field::new("k", DataType::Int32, false)]);
+        let child: Arc<dyn ExecutionPlan> =
+            Arc::new(EmptyExec::new(Arc::new(schema.clone())));
+
+        let wrapped = exchange_on(child, &key_partitioning("k", &schema), || 1);
+
+        assert!(wrapped.is::<ExchangeExec>());
+    }
```
comphead: This asserts that `Arc::new(ExchangeExec::new(..))` is an `ExchangeExec`. The unconditional-wrap behaviour predates the PR and is pinned end-to-end by four snapshots in `test/join_selection.rs` (L190, L217, L251, L278).

Keep the other two: `exchange_on_reuses_a_matching_exchange` guards the load-bearing new branch, and `exchange_on_wraps_an_exchange_on_a_different_key` guards a case no integration test can produce.

ballista/scheduler/src/state/aqe/test/mod.rs
```diff
@@ -46,6 +48,16 @@ use datafusion::prelude::{SessionConfig, SessionContext};
...
+    mock_partitions_with_size(42, 10)
+}
+
+/// Shuffle output reporting `num_rows` rows over `num_bytes` bytes in a single
+/// partition. Tests that turn on the *value* of a measured size, rather than
+/// just its presence, pick their own figures.
+pub(crate) fn mock_partitions_with_size(
+    num_rows: u64,
+    num_bytes: u64,
+) -> Vec<Vec<PartitionLocation>> {
```
comphead, line 97: Good extraction, but it stopped one function short. `mock_partitions_with_statistics_no_data` right below (L85-108) is still a full 24-line copy of the same `PartitionLocation` literal, differing only in `PartitionStats::new(Some(0), None, Some(0))`:

```rust
pub(crate) fn mock_partitions_with_statistics_no_data() -> Vec<Vec<PartitionLocation>> {
    mock_partitions_with_size(0, 0)
}
```

~21 lines. An extraction that covers two of three cases is the state most likely to rot.

ballista/scheduler/src/state/aqe/test/stage_build_side.rs
```diff
@@ -0,0 +1,229 @@
...
+    let state = SessionStateBuilder::new_with_default_features()
+        .with_config(config)
+        .build();
+    SessionContext::new_with_state(state)
+}
+
+fn register(ctx: &SessionContext, name: &str, stats: Statistics) {
+    ctx.register_table(name, Arc::new(StatsTable::new(join_schema(), stats, 4)))
+        .unwrap();
+}
```
comphead: I diffed these against `broadcast_thresholds.rs`:

- `ballista_ctx()` is **byte-identical** to `broadcast_thresholds.rs:80-88`
- `join_schema()` is **byte-identical** to its `narrow_schema()` (L71-76)
- `register()` is its L90-93 with the `schema` param dropped, so strictly less capable
- `const MB` is a third copy (L47 there, and `dynamic_join.rs:911`)

`test/mod.rs` is already the home for cross-file AQE fixtures (`mock_context`, `mock_schema`, `mock_partitions_*`). Hoisting these there and importing in both files nets ~25 lines, and more importantly stops "the shipped session config" from meaning two different things.

ballista/scheduler/src/state/aqe/test/stage_build_side.rs
```diff
@@ -0,0 +1,229 @@
...
+      ProjectionExec: expr=[val@1 as val]
+        DynamicJoinSelectionExec: plan_id=0, join_type=Left, on=[(id@0, id@0)] repartitioned=true
+          ExchangeExec: partitioning=Hash([id@0], 4), plan_id=1, stage_id=pending, stage_resolved=false
+            CooperativeExec
+              StatsExec: partitions=4, rows=Inexact(1000000), bytes=Inexact(209715200)
+          ExchangeExec: partitioning=Hash([id@0], 4), plan_id=2, stage_id=pending, stage_resolved=false
+            CooperativeExec
+              StatsExec: partitions=4, rows=Exact(1000000000), bytes=Exact(42949672960)
+    ");
+}
```
comphead: The justification in the doc does not quite hold up. `test_left_join_not_collected_left` (`test/join_selection.rs:190-198`) already snapshots the identical two-exchange `Left` shape, and `JoinType::Left` is already a case in `does_not_stage_a_join_type_that_can_never_broadcast`. The genuinely unique content is "with q8-shaped statistics too", which is one row of the merged loop.

Judgment call rather than a clear cut, since `broadcast_thresholds.rs` does pair integration positives and negatives. But it is a ~30-line plan snapshot that will need re-blessing on every unrelated `plan_id` shift.

ballista/scheduler/src/state/aqe/test/stats_table.rs
```diff
@@ -224,6 +224,26 @@ pub(crate) fn sizeless_statistics(schema: &Schema, num_rows: usize) -> Statistic
...
+) -> Statistics {
+    use datafusion::common::{ColumnStatistics, stats::Precision};
+
+    Statistics {
+        num_rows: Precision::Inexact(num_rows),
+        total_byte_size: Precision::Inexact(total_byte_size),
+        column_statistics: vec![ColumnStatistics::new_unknown(); schema.fields().len()],
+    }
+}
+
```
comphead, line 240: This is `sized_statistics` with `Exact` swapped for `Inexact`, and DataFusion ships the conversion:

```rust
pub(crate) fn estimated_statistics(schema: &Schema, num_rows: usize, total_byte_size: usize) -> Statistics {
    sized_statistics(schema, num_rows, total_byte_size).to_inexact()
}
```

(`Statistics::to_inexact`, `datafusion-common-55.0.0/src/stats.rs:485`.)

Counterpoint: three sibling constructors is this file's existing shape, so low value either way.

docs/source/contributors-guide/benchmarking.md
```diff
@@ -24,10 +24,22 @@ Current TPC-H **SF1000** results for Ballista, compared against a vanilla
...
+```{note}
+`da92ba8d` is the scheduler commit of
+[#2434](https://github.com/apache/datafusion-ballista/pull/2434) — `main`
+as of 2026-09-08 plus that PR's change, which stages a join's build side
+and reads its measured statistics before `DynamicJoinSelectionExec`
+commits to a strategy. It accounts for almost all of the improvement over
+the previous (`67b3a19b`) result set — Q8 96.02 → 40.68, Q9 107.89 →
+51.94, Q21 95.63 → 65.69 — so a run from `main` before #2434 lands will
+not reproduce these numbers. Re-pin this row to the merge commit once it
+does.
```
comphead: Checked-in docs pinned to an unmerged PR commit, carrying an unowned TODO ("Re-pin this row to the merge commit once it does"). This is also the only `{note}` admonition in the whole `docs/source` tree, so it is not an established pattern here.

Leaving the row on `67b3a19b` and landing a separate docs commit after merge avoids the self-invalidating state.

docs/source/contributors-guide/benchmarking.md
```diff
@@ -135,29 +147,29 @@ mean of 3 iterations (cold iteration dropped by the harness).
...
+|        14 |        15.70 |          7.00 |
+|        15 |        19.30 |         23.75 |
+|        16 |        17.40 |         23.41 |
+|        17 |        31.15 |         82.30 |
+|        18 |        51.41 |        129.40 |
+|        19 |        18.99 |         11.26 |
+|        20 |        33.49 |         19.22 |
+|        21 |        65.69 |        101.53 |
+|        22 |        12.40 |         12.71 |
+| **Total** |   **696.09** |    **845.21** |
```
comphead: Q6 is join-free (single-table scan, filter, aggregate on `lineitem`) and regressed 7% (14.25 to 15.28), while the equally join-free Q1 improved 4%. At one iteration per query that is a noise floor of roughly plus or minus 7% being published as measured signal, right next to a note saying this PR "accounts for almost all of the improvement".

The three headline wins are an order of magnitude above that band and survive it fine. The 13 small regressions do not, and the doc says nothing about them while the PR description does.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -601,6 +657,107 @@ impl DynamicJoinSelectionExec {
...
+    /// TPC-H q8's filtered `part` scan estimates roughly 10x over budget,
+    /// because the planner falls back to `default_filter_selectivity` for
+    /// `p_type = '...'`, and measures three orders of magnitude under it. A raw
+    /// fact-table scan sits far beyond this multiple and is shuffled without the
+    /// extra round trip.
+    max_estimate_multiple: usize,
+}
+
+impl BuildStagingLimits {
+    fn new(bc: &BallistaConfig, broadcast_threshold_bytes: usize) -> Self {
```
milenkovicm: nit: consider adding `max_probe_ratio` & `max_estimate_multiple`, instead of `BC`, "law of demeter" principle.  at the moment we have mix and match of paratemters as `broadcast_threshold_bytes` comes from BC as well

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -601,6 +657,107 @@ impl DynamicJoinSelectionExec {
...
+/// The configured limits that decide whether staging a join's build side on its
+/// own is worth an extra round trip. See [`requires_build_staging`].
+#[derive(Debug, Clone, Copy)]
+struct BuildStagingLimits {
+    /// `ballista.optimizer.broadcast_join_threshold_bytes`: the budget a build
+    /// side must come in under to be broadcast. `0` disables broadcast
+    /// promotion, and with it any reason to measure the build side again.
+    broadcast_threshold_bytes: usize,
+    /// `ballista.optimizer.stage_build_side_min_probe_ratio`: how many times
+    /// larger the probe side must be before staging pays off.
```
milenkovicm: is there a difference if statictics are exact or estimated ?

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
```diff
@@ -601,6 +657,107 @@ impl DynamicJoinSelectionExec {
...
+/// [`JoinSelectionAction::Repartition`] shuffles both inputs at once, which for
+/// a fact-table probe side commits to the single most expensive stage in the
+/// query *before* any measurement exists. When the build side's size is only a
+/// guess, that guess is the sole reason the join is not a broadcast, and the
+/// probe side dwarfs it, one cheap stage buys an exact number and often flips
+/// the join to `CollectLeft`, leaving the probe side never shuffled at all.
+///
+/// Each of the three conditions is necessary:
+///
+/// * the build estimate must be **inexact**. An exact size over budget is a
```
milenkovicm: this answers my question from, reasoning make sense to me.

ballista/core/src/config.rs
```diff
@@ -294,6 +319,31 @@ static CONFIG_ENTRIES: LazyLock<HashMap<String, ConfigEntry>> = LazyLock::new(||
                          which makes AQE use a hash join regardless of build size.".to_string(),
                          DataType::UInt64,
                          Some((64 * 1024 * 1024).to_string())),
+        ConfigEntry::new(BALLISTA_STAGE_BUILD_SIDE.to_string(),
+                         "Stages a join's prospective build side as its own shuffle before \
+                          choosing the join strategy, when that side's size is only an \
+                          estimate and the probe side is far larger. Lets AQE measure the \
```
milenkovicm: perhaps we should mention that we're working with estimated statistics only

milenkovicm (approved):
thanks @andygrove & @comphead , 

this makes sense to me, when initial DJ implementation was created i had a question if scheduling both sides at the same time makes sense. i believe this is a good balance between cluster utilisation and plan adaptility. 

maybe as a follow up we should consider work in #1916 and cardinality estimation, with hope it could bring better join decision

andygrove:
Thanks both. @milenkovicm for the approval, and @comphead for the second pass — I took essentially all of it. Net **−232 lines** (386 insertions, 618 deletions) across the branch, and **410 passed / 0 failed**, clippy and rustfmt clean.

The load-bearing check: the only snapshots insta accepted were the four in `stage_build_side.rs`, and their plan shapes and `plan_id`s are byte-identical to before the refactor. Nothing in `test/join_selection.rs` or the plan-stability files moved, which is the evidence that this round is behaviour-preserving.

I verified your three structural claims rather than taking them on faith, and all three hold: `create_partitioning(self, ..)` does consume (so the clone in `join_key_partitioning` was genuinely dead), `partition_mode` is `Partitioned` only when `!null_aware`, and the local `selection_state` can only be `Unknown` when `self.selection_state` already is — so `with_selection_state(Unknown)` really was copying eleven fields to change nothing.

### Taken

**Production.** Dropped the `join` payload (the arm rebuilds from the visited node); `build_side: JoinSide` → `build_is_right: bool`, which deletes the unreachable `JoinSide::None` arm, its error string and the `JoinSide` import from both files; collapsed the two `(Unknown, Partitioned)` arms into one with the join-type condition moved inside `requires_build_staging` (which retires the "caller is responsible for the fourth condition" paragraph); removed `with_selection_state` and restored `to_partitioned`'s inline body; `BuildStagingLimits` → plain params; `probe_stats` hoisted beside `build_stats`; `join_key_partitioning` returns a tuple by value over the owned `Vec`; `exchange_on` → `SelectJoinRule::ensure_exchange`.

On the naming: agreed `exchange_on` read as "on the join keys". The receiver keeps the `fetch_add` laziness without the generic, as you suggested. I also took the doc line about `inputs_already_partitioned` — the structural comparison is deliberate, since the only exchange worth reusing is one this rule just placed, and that is now written down so nobody unifies them later.

**Tests.** Six single-assert staging tests folded into two behind a `stages(build, probe, threshold)` predicate; the two hand-partitioned join-type lists merged into one loop over all ten with `collect_left_broadcast_safe` as the oracle (your point that a new DataFusion join type fails neither current list is exactly right, and the merged form does catch it); `q8_shaped_action*` un-forked from `run_to_actual_join`, which gained a `stage_build_side` param; the fourth `StatisticsExec` builder parameterised with the other three as wrappers; `mock_partitions_with_statistics_no_data` delegating to `mock_partitions_with_size`; `estimated_statistics` using `Statistics::to_inexact`.

You were right that the fixtures in `stage_build_side.rs` were a straight copy — `ballista_ctx`, `narrow_schema`, `register_stats_table` and `MB` now live in `test/mod.rs` and both files import them, so "the shipped session config" means one thing.

Dropped `does_not_stage_a_left_join` and `exchange_on_wraps_a_non_exchange_child`. On the first, your read was better than mine: `Left` is already a row of the merged loop with the same q8-shaped statistics, so the only unique content was a plan snapshot that needs re-blessing on every unrelated `plan_id` shift.

Also fixed the shadowed `use datafusion::common::{ColumnStatistics, Statistics, stats::Precision};` — good catch, since as you say `-D warnings` never would have.

### Not taken

**The exchange-guard depth**, deliberately. The fire-at-most-once bound is unaffected, because staging always produces a *direct* `ExchangeExec` child, which the guard does catch. In the `FilterExec(ExchangeExec(..))` shape the staged measurement is of the filter's output rather than the inner shuffle's, so it is an extra boundary but not redundant work, and a subtree-wide search would decline legitimately useful cases — a build side that is itself a join over an already-resolved broadcast exchange, for one. Noted as a follow-up; happy to tighten if you think the shape is likelier than I am reading it.

**Threading one `StatisticsContext` through `to_actual_join`.** Agreed it is the natural moment, and agreed this PR adds no new statistics computation. But it is a change to a path five callers share, with no test that would catch a memo-sharing mistake, so I would rather it not ride along behind a behaviour-preserving cleanup. Separate PR.

### @milenkovicm

Law of Demeter: taken, and it lines up with comphead's item — `requires_build_staging` now takes `min_probe_ratio` and `max_estimate_multiple` as plain params rather than reaching through `&BallistaConfig`, and `broadcast_threshold_bytes` is passed the same way, so the mix-and-match is gone.

Config wording: also taken. All three `stage_build_side*` entries now say explicitly that staging applies only where the build side's size is an estimate rather than a measurement — an exact size over the budget is a fact no further measurement can change. Docs regenerated.

And yes on #1916 and cardinality estimation as the follow-up. Better cardinalities would shrink the window this heuristic exists to cover, which is the right direction.

### Benchmarks and docs

I reverted the `benchmarking.md` change entirely, per your point. It pinned a row to an unmerged PR commit behind an unowned TODO, and it was the only `{note}` admonition in `docs/source`. The deeper problem is the one you found in Q1 vs Q6: both are join-free, one improved 4% and the other regressed 7%, so a ±7% noise floor was being published as measured signal next to a claim that this PR accounts for almost all of the improvement.

Those numbers also predate the join-type guard, so they are stale regardless. I will re-measure on the cluster and land the docs separately after merge, with the iteration count and the regression band stated rather than rounded away.

### CI

The S3 failures look like the MinIO issue you linked (#2446) rather than anything here.

ballista/scheduler/src/state/aqe/execution_plan/dynamic_join.rs
andygrove: Taken. `BuildStagingLimits` is gone: `requires_build_staging` now takes `min_probe_ratio` and `max_estimate_multiple` as plain `usize` params rather than reaching through `&BallistaConfig`, and `broadcast_threshold_bytes` is passed the same way — so the mix of "two read off BC, one passed by hand" is gone with it.

This also lines up with comphead's separate point that the three sibling predicates in this same file (`supports_collect_by_thresholds`, `hash_build_fits`, `broadcast_build_side_pays_off`) all take plain params, so the bundle was the odd one out.

ballista/core/src/config.rs
andygrove: Taken. All three `stage_build_side*` entries now say it explicitly — staging applies only where the build side's size is an estimate rather than a measurement, because an exact size over the broadcast budget is a fact that no further measurement can change.

The code says the same thing in one place now too: `requires_build_staging` returns early unless `total_byte_size` is `Precision::Inexact`.

Config docs regenerated via `./dev/update_config_docs.sh`.

comphead (approved):
Thanks @andygrove

andygrove:
Merged. Thanks @milenkovicm and @comphead

PR #2460, opened 2026-09-16
fix(ci): upload both cluster logs from the container benchmark jobs

### LINKED ISSUE #2147: CI: TPC-H SF10 failure artifact is missing executor.log
**Describe the bug**

When a `TPC-H SF10` job fails, the `Upload cluster logs on failure` step in `.github/workflows/tpch.yml` is supposed to attach both `scheduler.log` and `executor.log`. Only `scheduler.log` ever makes it into the artifact.

That is the more valuable of the two to lose. Most SF10 failures are task-level (spill, shuffle write, OOM, disk), and those only show up in the executor log. Without it you are left with the 200 line tail the cleanup trap echoes into the job output, which is usually not enough to see what led up to the failure.

**To Reproduce**

Two independent failed runs, different failure modes, same result:

- Run [29923718090](https://github.com/apache/datafusion-ballista/actions/runs/29923718090) (`No space left on device` during query 21), artifact `tpch-sf10-cluster-logs-aqe-on-mpt` contains only `__w/_temp/scheduler.log`.
- Run [29749959848](https://github.com/apache/datafusion-ballista/actions/runs/29749959848) (executor registration failure), artifact `tpch-sf10-cluster-logs` contains only `__w/_temp/scheduler.log`.

```
$ gh run download 29923718090 -n tpch-sf10-cluster-logs-aqe-on-mpt -D art
$ find art -type f
art/__w/_temp/scheduler.log
```

In both cases `executor.log` demonstrably existed at upload time, because the cleanup trap in the same step successfully ran `tail -n 200 "$EXECUTOR_LOG"` and its output is visible in the job log a fraction of a second earlier.

**Expected behavior**

The failure artifact contains both `scheduler.log` and `executor.log`.

**Additional context**

The step config is:

```yaml
      - name: Upload cluster logs on failure
        if: failure()
        uses: actions/upload-artifact@v7
        with:
          name: tpch-sf10-cluster-logs-${{ matrix.slug }}
          path: |
            ${{ runner.temp }}/scheduler.log
            ${{ runner.temp }}/executor.log
          if-no-files-found: ignore
```

I have not pinned down the mechanism, so treat the following as clues rather than a diagnosis. From the upload step's own output:

```
  path: /home/runner/work/_temp/scheduler.log
/home/runner/work/_temp/executor.log

Multiple search paths detected. Calculating the least common ancestor of all paths
The least common ancestor is /. This will be the root directory of the artifact
With the provided path, there will be 1 file uploaded
```

Two things stand out. The action only resolved one of the two search paths, and the least common ancestor came out as `/` rather than the shared `/home/runner/work/_temp` directory, which suggests the two paths did not resolve to the same root. Note also that this is a container job (`image: amd64/rust`), `${{ runner.temp }}` expands to the host path `/home/runner/work/_temp`, and the file lands in the artifact under the container path `/__w/_temp`. Some interaction between the container path mapping and multi-path glob resolution looks like the likely culprit, but that needs confirming.

Since `if-no-files-found: ignore` is set, the miss is silent.

Possible fixes, roughly in order of how much I like them:

- Point `path` at the directory or a glob (`${{ runner.temp }}/*.log`) so only a single search path is involved and the LCA logic is not exercised.
- Write both logs into a dedicated subdirectory (e.g. `${{ runner.temp }}/cluster-logs/`) and upload that one directory.
- Switch `if-no-files-found` to `warn` so a future silent miss is at least visible in the job log.

Found while investigating the SF10 disk exhaustion in #1996.

amitvijapur:

Closes #2147.

# Rationale for this change

The `Upload cluster logs on failure` step in the three container benchmark jobs (`tpch.yml`, `tpcds.yml`, `h2o.yml`) lists the two log files as a multi-line `path`. In a container job the runner translates host paths to container paths for action inputs with a single prefix replacement on the whole input string (`ContainerInfo.TranslateToContainerPath`, applied to every environment value in `StepHost.cs` before `docker exec`; node action inputs arrive there as `INPUT_*`). A two-line input starts with the host path once, so only its first line is rewritten. actions/runner#1174 is this defect, open since 2021.

upload-artifact then gets one path that exists inside the container and one that does not. From a failed PR leg on the current form of the step (run 35050467408):

```
The least common ancestor is /. This will be the root directory of the artifact
No files were found with the provided path: /__w/_temp/scheduler.log
/home/runner/work/_temp/executor.log. No artifacts will be uploaded.
```

The least common ancestor of `/__w/...` and `/home/runner/...` is `/`, which is the line quoted in the issue, and it is why the one file that did upload in run 29749959848 sat under `__w/_temp/` inside the artifact.

The two-line `path` becomes one glob, `${{ runner.temp }}/*.log`, in all three workflows, with a two-line comment saying why it is a glob so it does not get tidied back into two lines. A single line is rewritten whole, and with one search path the artifact root is `/__w/_temp`, so both files land at the top level of the artifact.

`{scheduler,executor}.log` was not an option: `@actions/glob` constructs minimatch with `nobrace: true`.

Within this repo's own workflow and `ci/` code, the only writer of `.log` files into `runner.temp` is `ci/scripts/ballista_cluster.sh` (`scheduler.log`, `executor.log`, via `CLUSTER_DIR`).

# Are these changes tested?

Yes, with a throwaway workflow on my fork that runs in the same `amd64/rust` container, writes both files to `$CLUSTER_DIR`, and uploads them twice: once with the current two-line `path`, once with the glob. In https://github.com/amitvijapur/datafusion-ballista/actions/runs/35105812662 the two-line step logs `there will be 1 file uploaded` and its artifact holds only `__w/_temp/scheduler.log`; the glob step logs `there will be 2 files uploaded` and its artifact holds `scheduler.log` and `executor.log` at the root. The step's own output also shows the env translation at work: the step header prints `CLUSTER_DIR: /home/runner/work/_temp` while the shell inside the container sees `CLUSTER_DIR=/__w/_temp`.

No. CI only. Failure artifacts from these three jobs now contain `executor.log`, and the files sit at the artifact root instead of under `__w/_temp/`.

Written with AI assistance (Claude Code). The mechanism was checked against the runner and toolkit sources and the run logs quoted above, and the repro was run on my fork.


andygrove (approved):
Thanks @amitvijapur

PR #2462, opened 2026-09-17
feat(scheduler): record scheduler id in job starts

### LINKED ISSUE #2413: history server event log should capture scheduler id

at the moment event logs are written in local directory, so associating logs with scheduler is not hard task, if we ship logs to a centralised location (s3 bucket) being able to associate logs with schedulers would help with diagnostics. current assumption is that shipping event logs to s3 would be done by external process

add additional field to `JobStart` event, like `scheduler_id`

change from

```json
{
  "ev": "JobStart",
  "version": 1,
  "data": {
    "job_id": "0382W17Y0G000",
    "job_name": "Query derived from TPC-H q1",
    "queued_at": 1788250553353,
    "submitted_at": 1788250553364,
```
to

```json
{
  "ev": "JobStart",
  "version": 1,
  "data": {
    "job_id": "0382W17Y0G000",
    "job_name": "Query derived from TPC-H q1",
    "queued_at": 1788250553353,
    "submitted_at": 1788250553364,
    "scheduler_id" : "< random scheduler id >"
```

alternative would be to have a path prefix per scheduler id, which at the moment we cant retrieve from available apis

```bash
curl http://localhost:50050/api/state
{"started":1788384057783,"version":"54.0.0","datafusion_version":"55.0.0","substrait_support":false,"keda_support":false,"prometheus_support":false,"graphviz_support":false,"spark_support":false,"scheduling_policy":"push-staged","enable_embedded_flight_proxy":false}%
```

**Additional context**

we do not have scheduler id associated with the scheduler, so we might need to add it

villebro:
## Summary
History event logs are collected centrally, so `JobStart` records need a scheduler identity that can be used to correlate events with the scheduler that produced them. Until now, the value available on this path was the scheduler callback endpoint (`external_host:bind_port`), which is routing data rather than scheduler identity.

This PR separates identity from routing. Schedulers and executors now both have configurable `--id` values, with UUID-backed instance ids generated by default. Scheduler callback routing keeps using the endpoint built from `external_host` and `bind_port`, now named `scheduler_endpoint`. Executor ids are also validated as unique among active executors registered with a scheduler, which keeps metadata, heartbeats, capacity, and task ownership from aliasing. This also prepares the identity path for future multi-scheduler HA work. Fixes #2413.

## Changes

Identity and CLI behavior:
- Add `SchedulerConfig.scheduler_id` and `ExecutorProcessConfig.executor_id`.
- Add shared `--id` CLI support for scheduler and executor binaries.
- Generate UUID-backed scheduler and executor ids when no id is configured.
- Reject empty configured ids.
- Reject duplicate active executor ids in both push registration and pull-mode metadata refresh, returning gRPC `AlreadyExists`.

Scheduler routing terminology:
- Rename scheduler callback routing from scheduler name/id terminology to `scheduler_endpoint`.
- Rename `SchedulerConfig::scheduler_name()` to `SchedulerConfig::scheduler_endpoint()`.
- Rename `LaunchTaskParams.scheduler_id` and `LaunchMultiTaskParams.scheduler_id` to `scheduler_endpoint` while preserving protobuf field number `2`.

History and API exposure:
- Expose `scheduler_id` from live scheduler `/api/state` and keep history-server `/api/state` aligned.
- Record `scheduler_id` in `JobStart` history events.
- Pass the scheduler id through job state, task management, and execution graphs.
- Remove stale `dead_code` allowances/comments now that graph scheduler id is consumed by event logging.

Docs and tests:
- Document `--id` usage and uniqueness expectations in the scheduler guide, History Server guide, config reference, and 55.0.0 upgrade note.
- Add coverage for configured/default scheduler and executor ids, duplicate executor-id rejection, shared UUID-backed id generation, executor registration metadata, `/api/state`, history-server state compatibility, and serialized `JobStart` output.

## Validation
- `cargo fmt --all --check`
- `npx prettier@2.7.1 --write ...`
- `cargo check -p ballista-core`
- `cargo check -p ballista-scheduler --lib`
- `cargo check -p ballista-executor --lib`
- `cargo check -p ballista-examples --example mtls-cluster --features tls`
- `cargo clippy -p ballista-examples --example mtls-cluster --features tls -- -D warnings`
- `cargo test -p ballista-history`
- `cargo test -p ballista-scheduler --lib`
- `cargo test -p ballista-core ids::tests::instance_ids_are_uuid_backed`
- `cargo test -p ballista-executor --lib executor_metadata`
- `cargo test -p ballista-executor --lib executor_id`
- `cargo test -p ballista-scheduler --lib scheduler_id`
- `cargo test -p ballista-scheduler --lib register_executor`
- `cargo test -p ballista-scheduler --lib save_executor_metadata_rejects_duplicate_id`
- `cargo test -p ballista-scheduler --lib test_pull_work`
- `cargo test -p ballista-scheduler --lib duplicate_executor_registration_maps_to_already_exists`

Local smoke test: started a scheduler with `--event-log-dir` and `--id local-scheduler`, started an executor, ran `select 1`, verified the generated `JobStart` line contained `"scheduler_id":"local-scheduler"`, then started `ballista-history-server` and verified `/api/state`, `/api/jobs`, and `/api/job/<job_id>/stages` returned the replayed history data.


andygrove (approved):
LGTM. Thanks @villebro

milenkovicm (commented):
Thanks @villebro two questions
- Can we create scheduler id on scheduler start and print info log, we do similar thing with executor
- can we expose scheduler id as part of the rest api, there already is endpoint which can host this info,

villebro:
@milenkovicm good point. Looking closer, I think there is an existing naming mismatch here. Executor identity and executor routing are separate today: executors autogenerate an opaque `executor_id` at process start, but the scheduler calls them using `ExecutorMetadata { host, grpc_port, ... }`. For schedulers, the value currently called `scheduler_id` is not equivalent to `executor_id`. It is produced by `SchedulerConfig::scheduler_name()` as `external_host:bind_port`, and executors use it directly as the callback endpoint. I assume the reason for this has been to keep scheduler id more tangible on logs etc.

To align with executor behavior, I think we should split these, so it would be
- `scheduler_id` is an opaque id, like it is for executors.
- introduce `scheduler_host` + `scheduler_port`, or `scheduler_grpc_endpoint` as a routable callback address used by executors.

..or something similar. That would make history entries a bit less immediately tangible because they would carry an opaque scheduler id instead of `host:port`, but I think it is probably the cleaner long-term model: it matches executor semantics and avoids exposing internal routing details in history/API state. Operators could still correlate via the scheduler startup log and `/api/state`, which would expose the generated `scheduler_id`.

 Thoughts?

milenkovicm:
the idea behind scheduler_id is to be able to link events with schedulers. for example if logs are shipped to a S3 bucket we can use scheduler id as part of the path. I don't believe scheduler_id needs to have some meaningful name. 

we can provide a scheduler configuration option to assign id to the executor, or if scheduler id is not configured a uuid can be generated (and printed to logs). this would cover both cases handling schedulers as pets or cattle 

scheduler id can also be part of `/api/state` so we can poll it from external process should we need scheduler id 

does it make any sense, wdyt ?

villebro:
@milenkovicm sure - if we're ok with making the scheduler id opaque, similar to how it's done in executor, I can bake those changes in this PR. Or alternatively do a follow-up.

milenkovicm:
if we can cover them here i it would be great

villebro:
@milenkovicm ok, I think I finally got the diff in a decent state, basically splitting `scheduler_id` into `scheduler_id` and `scheduler_endpoint`, and was able to validate locally that it works as expected. LMKWYT.

Edit: I got slightly carried away with changes here, but I feel all of the changes were related and was too lazy to break them up into multiple PRs :D

andygrove:
@milenkovicm are you ok with merging this one?

milenkovicm:
Apologies @villebro & @andygrove will have a look tomorrow morning. Please @villebro ping me next time for follow up review if I'm unresponsive

villebro:
No worries @milenkovicm - feel free to request downscoping the changes if this feels difficult to review, happy to break this up into a few follow-ups.

milenkovicm (approved):
thanks @villebro

PR #2463, opened 2026-09-17
ci: upload cluster logs with one glob

### LINKED ISSUE #2147: CI: TPC-H SF10 failure artifact is missing executor.log

villebro:
## Summary

The failed TPC-H runs from #2147 were missing `executor.log` from the artifact even though the cleanup step had just tailed that file, so the file existed at upload time. The upload step was passing two explicit absolute paths under `${{ runner.temp }}` in a container job. That is `upload-artifact`'s documented multiple-path mode: its README says that when multiple paths are provided, the least common ancestor of all search paths is used as the artifact root ([docs](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions)). The action source also logs the exact `Multiple search paths detected...` / `least common ancestor` messages when `globber.getSearchPaths()` returns more than one path ([source](https://github.com/actions/upload-artifact/blob/main/src/shared/search.ts#L113-L127)).

In the failed run, that documented multi-path handling appears to have interacted badly with the container path mapping: the action computed `/` as the LCA and uploaded only one resolved file. Using a single `*.log` glob keeps artifact discovery rooted in one temp directory, avoiding that multi-path resolution path while still collecting both `scheduler.log` and `executor.log`. H2O and TPC-DS use the same cluster harness and had the same upload pattern, so this PR updates those workflows too.

Fixes #2147.

## Changes

- Upload benchmark cluster logs with a single `${{ runner.temp }}/*.log` glob.
- Apply the same hardening to the TPC-H, TPC-DS, and H2O workflows.
- Warn when no log files are found instead of silently ignoring the miss.


andygrove:
@villebro this may overlap with #2460?

villebro:
Ah indeed, I had authored this some days ago before this landed. Sorry about the noise!

